PR #2667, opened 2015-12-20
Legion dev - Shadow Priest implementation

kaesebrezen:
Since Twintop asked me to do a pull request, here it is.

Voidform is mostly based around Metamorphosis, but still would be good if someone could look over it. 

Talents and stuff that needs to be checked later on is marked with a fixme.


scamille:
Going to just merge this stuff in, thanks for the pull request. But I added you to the commit team as well, welcome to the team ;)

I'll try to take a look at the voidform implementation as well, seems to be the most complex part we'll have this expansion.

PR #2673, opened 2016-01-09
Mr/blood dk profile update

mrdmnd:
@mserrano - updates T18 profiles for blood dk.

I don't know if it's correct to strip out the default gear, or if it will break the sims. I was under the impression that importing your character sets all the gear anyways. I don't really want to go through and build the BIS lists for each difficulty.

Minor changes to the T18M_BoS APL, where chains of ice is removed from the "Active Breath of Sindragosa" priority list, as the glyph to generate RP from CoI has been removed in T18.

### DIFF
--- a/profiles/Tier18H/Death_Knight_Blood_T18H.simc
+++ b/profiles/Tier18H/Death_Knight_Blood_T18H.simc
@@ -1,0 +1,117 @@
+deathknight="Death_Knight_Blood_T17H"
+level=100
+race=tauren
+role=tank
+position=front
+talents=2013102
+spec=blood
+
+# This default action priority list is automatically created based on your character.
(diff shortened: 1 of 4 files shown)


profiles/Tier18H/Death_Knight_Blood_T18H.simc
```diff
@@ -0,0 +1,87 @@
+deathknight="Death_Knight_Blood_T17H"
```
mrdmnd: Oops, this needs to be updated to T18H. Same for all the other files.

mrdmnd:
Okay, after a short discussion on IRC, this is the wrong place to handle this. Will try updating the APL gen code in the main engine.

PR #2674, opened 2016-01-09
Mr/blood dk profile update

mrdmnd:
Updates to DK Blood APL to clean up some cruft. It doesn't look like anyone's maintaining the live WoD APL stuff; I'm happy to do more work on this, but here's a first pass.

Things changed: 
 - Always use Strength flask, always use Multistrike food.
 - Don't spend runic power on abilities during Breath of Sindragosa; it's going to be a DPS loss. 
    - In particular, it's better to wait or do no action than casting death coil. 
    - Additionally, only cast outbreak if it won't cost runic power (== if it's unglyphed).
   - Remove chains of ice from the priority list; this was changed to not produce as much RP, and is no longer worth casting.
 - Added some default gear sets to the profile generation .simc file. Not sure if these are BIS, but they're probably close.
 - Regenerated default APLs.

@mserrano please take a look when it suits you.


mrdmnd:
I see about a 1% DPS increase on my character with this new APL. I wish there were more (read: any) tests to run to ensure I'm not introducing regressions.

mrdmnd:
I'm going to update this branch with a new commit that ekes out about 5% increased DPS over the current APL.

mrdmnd:
I did a pseudo-merge of the BT priority and the Breath of Sindragosa CD priority, which earned most of the improvment. I also tweaked some constants a bit for runic power thresholds, to reduce the likelihood that breath falls off early.

Here's the result, on a 450 +/- 20% patchwork fight, with my ilvl 722 blood dk
Before (current master branch): http://pastebin.com/Z8mzQiLP (48,341 +/- 30 dps)
After (this branch): http://pastebin.com/vQWaKaqL (51,740 +/- 27 dps)

This is a 7% DPS gain.

mrdmnd:
@Collisionc - @mserrano  suggested I tag you.

mrdmnd:
Also @scamille - looks like you've touched the codebase recently too.

Collisionc:
Thanks for the update, sorry we've all been a bit lazy over the last few months, that's just kind of what happens at the end of an expansion. :)

mrdmnd:
Absolutely. Hope I didn't miss anything or leave any DPS obviously on the table.

Issue #2654, opened 2015-11-23
TMI


serrufs:
why in the "TMI Standard boss" is not present "T18M / H / N / L"?

For http://www.wowhead.com/item=133598 it ignores the demon boss or humanoid.

scamille:
T18 TMI seems to be implemented in the CLI, just not hooked up to the GUI.

serrufs:
And add it to the GUI?

mrdmnd:
I've done this in the PR here: https://github.com/simulationcraft/simc/pull/2678

PR #2678, opened 2016-01-24
Adds TMI Tier 18 bosses to GUI

mrdmnd:
Solves https://github.com/simulationcraft/simc/issues/2654.


serrufs:
Excellent! And when it gets in Windows patskage [http://downloads.simulationcraft.org/?C=M;O=D](url) ?

mrdmnd:
Yes, I don't do any of the packaging, but I'd imagine you can find it there when they do a new SIMC build.

PR #2682, opened 2016-01-29
Hotfix to stone-of-the-elements. Need to verify if this is for real or not.

mrdmnd:
@Valkryst, please update your issue with logs :)


Valkryst:
You might have meant to notify someone else. I've never worked on this repo as far as I can recall.

mrdmnd:
Hah, yes. Sorry! #OpenSource.

scamille:
As discusses on IRC (not sure if that was with you mrdmnd), the .inc files should not be edited manually. They are auto-generated from the WoW Client files.

Instead, void unique_gear::register_hotfixes(); https://github.com/simulationcraft/simc/blob/c215c66f5a45b54ba5f63d223ce0e4ee6d1483ec/engine/player/sc_unique_gear.cpp#L4407 should be adjusted.

I'm therefore closing this pull request. Even if there ICD really might be 40s, it needs to be fixed differently.

PR #2685, opened 2016-02-05
Rppm fix 201602

kenthklui:
@navv1234 @Collisionc @Twintop @Frostedmages 

The ending of the RPPM modelling nightmare, hopefully?

Please review https://github.com/simulationcraft/simc/commit/145b96075f0eed135add37c8dd85f760653ac3b5 for RPPM changes.

### DIFF
--- a/engine/simulationcraft.hpp
+++ b/engine/simulationcraft.hpp
@@ -6260,6 +6260,7 @@ private:
   double       freq;
   double       modifier;
   double       rppm;
+  bool         first_proc_occurred;
   timespan_t   last_trigger_attempt;
   timespan_t   last_successful_trigger;
   timespan_t   initial_precombat_time;
@@ -6303,9 +6304,10 @@ public:
 
   real_ppm_t() :
     player( nullptr ), freq( 0 ), modifier( 0 ), rppm( 0 ),
+    first_proc_occurred( false ),
     last_trigger_attempt( timespan_t::from_seconds( -10.0 ) ),
-    last_successful_trigger( timespan_t::from_seconds( -180.0 ) ),
-    initial_precombat_time( timespan_t::from_seconds( -180.0 ) ),
+    last_successful_trigger( timespan_t::from_seconds( -120.0 ) ),
+    initial_precombat_time( timespan_t::from_seconds( -120.0 ) ),
     scales_with( RPPM_NONE )
   { }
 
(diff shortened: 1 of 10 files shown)


engine/simulationcraft.hpp
```diff
@@ -6314,10 +6316,10 @@ struct real_ppm_t
     freq( frequency ),
     modifier( mod ),
     rppm( freq * mod ),
+    first_proc_occurred( false ),
     last_trigger_attempt( timespan_t::from_seconds( -10.0 ) ),
-    last_successful_trigger( timespan_t::from_seconds( -180.0 ) ), // Blizz done lied to us, or changed it without telling. After going through a lot of logs,
-                                                                   // it seems that it's actually 3 minutes for the precombat timer.
-    initial_precombat_time( timespan_t::from_seconds( -180.0 ) ),
```
mrdmnd: Is there reason to believe this comment is wrong, or the behaviour has changed? I don't know either way, but this seems to contradict the old comment.

engine/simulationcraft.hpp
```diff
@@ -6360,12 +6363,19 @@ struct real_ppm_t
...
+    double chance = proc_chance( player, rppm, last_trigger_attempt,
+                                 last_successful_trigger, scales_with );
+    bool success = player -> rng().roll( chance );
 
     if ( success )
+    {
+      first_proc_occurred = true;
       last_successful_trigger = player -> sim -> current_time();
+    }
+    // Due to a bug, last attempt to trigger is fixed until first proc
```
mrdmnd: What bug? Is this a bug that can be filed against SimC as an issue?

engine/simulationcraft.hpp
kenthklui: Yes. The old model gave an absurdly high first-hit proc chance, more than 2.5x what we observe in game, while the second/third attacks and onward had very low proc rates that were significantly below what we saw in game. New testing shows this model to be much more accurate.

engine/simulationcraft.hpp
kenthklui: It is an in game bug, not a SimC bug. We are merely showing in-game behavior.

engine/simulationcraft.hpp
mrdmnd: Cool. Just checking =)

engine/simulationcraft.hpp
mrdmnd: Is this something that's likely to be fixed in a patch, ever? If we codify in - game bugs, is there mechanism in place to remove our implementation of the bug when the real bug is fixed?

kenthklui: It will most likely be fixed in the upcoming biggest patch - 7.0. There is no mechanism needed because all we need to do is avoid merging this change into `legion-dev`.

mrdmnd: Seems reasonable.

mrdmnd:
:+1: LGTM

PR #2715, opened 2016-06-08
[Rogue] Make Agonizing Poison proc Kingsbane

Zuliane:
Not yet tested ingame, though it should be the case: http://us.battle.net/wow/en/forum/topic/20743504316?page=15#282
Tooltip was recently changed from Wound/Deadly Poison to "Lethal Poison".


navv1234:
Already done in aaeed50.

PR #2720, opened 2016-07-04
[DK] fixed Runic Power generation Bug

jokre33:
[Problem](http://puu.sh/pQnGW/d5599b5bb0.png)
[Full Report](http://files.jokre.eu/simc_report.html)

`this -> energize_amount += std::fabs( this -> base_costs[ RESOURCE_RUNIC_POWER ] );`

somehow sets `energize_amount` to a very large negative number while

`this -> energize_amount = std::fabs( this -> base_costs[ RESOURCE_RUNIC_POWER ] );`

doesn't.

[Expected Result](http://files.jokre.eu/simc_report_0.html) and also result with the change applied

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -1772,7 +1772,7 @@ struct death_knight_action_t : public Base
     {
       this -> energize_type = ENERGIZE_ON_CAST;
       this -> energize_resource = RESOURCE_RUNIC_POWER;
-      this -> energize_amount += std::fabs( this -> base_costs[ RESOURCE_RUNIC_POWER ] );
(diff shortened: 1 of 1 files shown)


navv1234:
Where are you seeing an overgeneration of Runic Power?

jokre33:
[Here](http://puu.sh/pQnGW/d5599b5bb0.png) (technically you'd probably call it undergeneration...)

[Full Report](http://files.jokre.eu/simc_report.html)

E: Changing the += to = results in the expected output on my side [report](http://files.jokre.eu/simc_report_0.html)

navv1234:
Let's fix the real reason instead. Can you try now? energize_amount is not initialized which is probably not a good idea. I pushed a fix.

jokre33:
it seems to work now! closing the request

PR #2731, opened 2016-07-18
[Warrior] Prot and Precombat APL Update

adamhl8:
The Prot APL has been updated for all new talents and abilites.
Updated the precombat APL with the new food/flasks/potions.


Collisionc:
Will add once I get protection up and running.

PR #2738, opened 2016-07-23
Update Rogue Assa APL

skasch:
Reorganizing APL + more optimized (+4% patchwerk) + more talents support


scamille:
those APL changes will be lost once profiles are regenerated. They need to be implemented into https://github.com/simulationcraft/simc/blob/3ed646054021f6c3dce8c791fc4da61314b748c4/engine/class_modules/sc_rogue.cpp#L5673

PR #2783, opened 2016-08-25
[Hunter] SV Pet Melee Fix

dashbrown:
Pet auto attacks no longer trigger Mastery: Hunting Companion. Potentially an in-game bug as they used to, but it seems to have changed since the mastery buff on August 16.


Collisionc:
Thanks

PR #2794, opened 2016-08-29
FrostDK - try to fix "Frozen Pulse"

gixxpunk:
I'm not 100% sure, but I think this talent should not be allowed to crit? If yes, this should fix it.


navv1234:
Is it not supposed to crit?

gixxpunk:
Sorry. Seems it can crit. Re-checked ingame and Recount shows it does crit.

PR #2802, opened 2016-09-06
[Frost DK] Move Remorseless Winter ahead of Frostscythe

wintertouch:
In situations where Frostscythe will be used, Remorseless Winter will never be called due to the catch-all manner of the Frostscythe line.

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -6619,9 +6619,9 @@ void death_knight_t::default_apl_frost()
   // Core rotation
   core -> add_talent( this, "Glacial Advance" );
   core -> add_action( this, "Frost Strike", "if=buff.obliteration.up&!buff.killing_machine.react" );
+  core -> add_action( this, "Remorseless Winter", "if=spell_targets.remorseless_winter>=2" );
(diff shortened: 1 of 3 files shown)


wintertouch:
Did not realize this would update the pull request. Thought it would snapshot a commit. Learning is fun.

navv1234:
Can you adjust engine/class_modules/sc_death_knight.cpp for the APL change (it would otherwise be overwritten when we regenerate profiles), and adjust profiles/Tier19P/generate_Death_Knight_T19P.simc for the relic changes (same thing here).

navv1234:
Are the relics intended to increase specific ranks? If so, the increases should be reflected in the artifact= string. The other option is to provide relic item ids 0 in the artifact string, and give the proper relic data in the main hand weapon.

wintertouch:
"hardcoding trait levels" is my misunderstanding of the artifact string. The increases to trait levels are still included in the "artifact=" string. This change only adds the relic IDs to the artifact string so that the relics show up on the linked Wowhead page.

From testing, the "gem_id" does not appear to be required on the "main_hand" line. I assume this is policed by convention rather than the engine.

navv1234:
Gem_id is only required on the main hand weapon if you are actually using the relic data there. If the main hand weapon has an ilevel= override, and the artifact string contains relics, the gem_id stuff does nothing, yup.

PR #2896, opened 2016-09-26
T19H Typo

reighnman:


reighnman:
Not sure why it failed going to recommit

mrdmnd:
Hi! Thanks for the contribution! The correct way to make apl changes to simcraft is to modify the default code generator  in the engine - any changes to generated profiles here will be blown away once anyone regenerates profiles, which happens daily. Take a look at engine/class_modules/sc_warrior. cpp and search for the AOL generation code - it'll be near the end. 

Thanks, and best wishes!

reighnman:
@mrdmnd I'm at looking engine/class_modules/sc_warrior and I don't see anything for gear or traits.  I'm assuming that is just contained in the profiles/TierX/generate files which is later combined with class modules to generate the actual APL file?

PR #2897, opened 2016-09-26
Fury T19P race change and initial T19H

reighnman:
T19P race change to troll and T19H using 26 point artifact and 865 gear set


reighnman:
resubmitting with generate profile

PR #2900, opened 2016-09-26
[Warrior_Fury] T19P/T19H generate update

reighnman:
Update race in T19P and initial T19H creation


reighnman:
@mrdmnd I think this should do it

mrdmnd:
Aha - yes, my mistake earlier. I didn't see that you were updating gear/traits, which is definitely done the way you've done it here - the default APL generation code isn't what you were modifying; that's my bad.

@xMarok for review?

reighnman:
or @Collisionc

PR #2948, opened 2016-10-05
updated artifact and talent choices as per H2P

seanpeters86:
The artifact path took Sphere over Sinister thoughts, where sphere is a minimal DPS increase. Also to help get your third VoiT PI is the better option over MB for almost all cases.


scamille:
Thank you for the pull request!

Next time please adjust the generate_Priest_T19H.simc file, as I did in a00014fac721638e9d4132bd6d3239a08ed1c5d4 . And any changes in the Action Priority List in https://github.com/simulationcraft/simc/blob/legion-dev/engine/class_modules/sc_priest.cpp#L5353

The final profile files are auto-generated.

PR #3006, opened 2016-10-21
Add interrupt to actions

Hemario:
Ovale addon needs it in order to generate the actions to interrupt.


Hinalover:
I needed to actually add Spear Hand Strike to SimC (never got around to it). Should be fixed with https://github.com/simulationcraft/simc/commit/e1663e274429ae1b85339190ca8ee5831a79a3bb

Hemario:
Ovale thanks you @Hinalover  ;)

PR #3014, opened 2016-10-23
[Death Knight] Frost - Add 7.1 Frozen Pulse changes

gixxpunk:
"Frozen Pulse now triggers if you have fewer than 2 full Runes (was 1)."

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -2984,7 +2984,7 @@ struct melee_t : public death_knight_melee_attack_t
         p() -> buffs.killing_machine -> trigger();
       }
 
-      if ( frozen_pulse && p() -> _runes.runes_full() == 0 )
+      if ( frozen_pulse && p() -> _runes.runes_full() < 2 )
       {
         frozen_pulse -> target = s -> target;
         frozen_pulse -> schedule_execute();


gixxpunk:
Imho Frozen pulse implementation is a bit borked. Without that patch it already does ~14% of our DPS in simc, but Recount shows just 7-8% ingame. Maybe this should be adressed with a different patch than this one?

navv1234:
Was just about to fix this, so closing. About the inflated damage, the only thing that I can easily see that would be "wrong" is whether it procs from both autoattacks, or just main hand. Coincidentally, cutting it to half would match the numbers pretty well too.

If it's not that, then it might be something to do with multipliers and such.

gixxpunk:
@navv1234 
Your guess was right! Frozen Pulse only procs from main hand. I checked on PTR.

See my Recount screenshots:

Only off hand equipped: http://abload.de/img/off_handvuzxl.jpg
Only main hand equipped: http://abload.de/img/main_hand3ablt.jpg

As on PTR Frozen Pulse uptime has now doubled for OB+RA playstyle, numbers match pretty well. Would be nice if you could change it to only proc of main hand. Thanks in advance!

navv1234:
Yeah, Main-hand only now in c7b860cf990759112f444c0b784fb944866a3fd6.

PR #3044, opened 2016-10-27
[Death Knight] Frozen Pulse triggers from main-hand and off-hand auto…

gixxpunk:
…attacks

More detailed analysis showed that Recount was wrong and FP triggers from mh and oh autoattacks.


gixxpunk:
Already changed

Issue #3050, opened 2016-10-27
Shadow Priest, tweaked apl and s2m.check


N1gh7h4wk:
http://pastebin.com/XVJRbCNT

since i can't seem to find a formatting for the apl, i put it in a pastebin
also too tired to make a pull request 

This improved APL works fine for single target, shadow priest comparison and raid sim (24 actors, mixed execute and no execute) s2m, especially tweaked with the "standard" s2m talent setup used in raids.

lgkern:
Created a pull request with N1gh7h4wk changes: https://github.com/simulationcraft/simc/pull/3053

scamille:
Thanks for all the work. Unfortunately I cant use that pull request as is.

The default APL changes need to go into https://github.com/simulationcraft/simc/blob/legion-dev/engine/class_modules/sc_priest.cpp#L5279 , which are then used to create the Priest_Shadow_T19P.simc and Priest_Shadow_T19P_S2M.simc from the generate_Priest.simc file. (and default APL for user sims when they import their character from armory)

So if APL changes go directly to the priest .cpp file, gear/talent/artifact changes to the generate_Priest.simc file. The final profiles will just be overwriten when the next person auto-generates them.

Long story short: I might find time to adjust your changes late next week, but I'm out of town over the long weekend.

scamille:
I played a bit with the shadow APL just now, and just changing the surrender madness start to if=target.health.pct<43 gives such nice and good dps with the H profile.

Of course that would kill Sims for lower-haste priests. But could we still have some simpler formula based on target health.pct and shadow haste? What do you think?

N1gh7h4wk:
The main problem is having multiple actors imo, since your execute phase has to start sooner or later depending on how many actors /execute actors are in the sim.
The change I proposed reduces the haste impact on the s2m length (it's observed in usual gameplay tests) but still accounts for multi actors sims /high differences in low and high haste sims. 
Did you use the other changes for the APL?
We improved the s2m performance with that, before changing the s2m trigger.

scamille:
I simmed with Raid_T19H.simc, without your APL changes.

I had the idea the other day, that we could use raid simc to generate boss hp%-time profiles, which then could be fed into single actor sim to mimic that boss hp-timeline.

PR #3053, opened 2016-10-28
Updated S2M APL

lgkern:
Updates the S2M APL with a more accurate TTD prediction and tweaks to general CD usage during it, providing better results for overall raid sims and single target comparisons between classes.

Thanks N1gh7H4wk for the work he put on those.

Results from the changes can be seen at https://www.dropbox.com/s/scrb3rvv4smez3k/simc_standard_fix.zip


scamille:
See #3050 for why I cant use this directly.

PR #3197, opened 2016-11-23
[Shadow Priest] Mangazas Madness APL tweak

N1gh7h4wk:
Since the last commit got lost somehow.


scamille:
http://simulationcraft.org/reports/Raid_T19H_Legendary.html pretty small difference now between Mangazas and Sharaz

PR #3201, opened 2016-11-25
[Shadow Priest] Interrupt Tweak

N1gh7h4wk:
Stopped the actor from locking himself out in multi target situations during high s2m.
Thanks to Penkek for killing me with that. :)


Strahbehry:
toxic simcrafter right here

PR #3217, opened 2016-12-03
[Mage] Fire APL changes

Twinkeli:
Ran all talent combinations against default and everything was higher or within margin of error


Twinkeli:
that priest thing wasnt supposed to be there, dont know how to remove it though...

Issue #3219, opened 2016-12-04
[Guardian Druid] Adaptive Fur increases incoming magic damage


Symondus:
When simulating my guardian druid i noticed quite inflated incoming magic damage. I could track the issue to the artefact trait adaptive fur that resulted in about 4 times as high magic damage as without the trait.

Symondus:
I could track the error further to 

the target_mitigation function in sc_druid.cpp 

in the line
s -> result_amount *= 1.0 + buff.adaptive_fur -> value();

the value() function returns the spell shool and not the percentage change -0.10.

Collisionc:
Thanks.

Issue #3108, opened 2016-11-04
[Guardian] Adaptive Fur dtps bug


videnfost:
Selecting the Adaptive Fur artifact trait as a Guardian Druid causes a spike in damage taken by 52.5%. There is simply no possibility that said artifact trait would actually cause a damage taken increase.

Collisionc:
Thanks.

PR #3223, opened 2016-12-06
[Guardian Druid] Fix adaptive fur

Symondus:
Fixes Issues #3219 and #3108
Adaptive fur now tracked per basic school as one buff can be active for fire, frost, holy, shadow, arcane and nature at any time (so up to 6 buffs at once).


Collisionc:
Got it, thanks!

PR #3241, opened 2016-12-12
[Priest] Implementation of Current PTR Changes

lgkern:
This commit implements all current changes to Priests on PTR:
- PTR flat % damage buff;
- Mind Flay dealing Area Damage (Mind Sear) if the target has Shadow Word: Pain;
- Mind Spike new incarnation;
- Void Bolt changes, now extending instead of refreshing
- Legendary Shoulders' new effect;
- New Lingering Insanity, now a talent instead of a passive effect;


scamille:
Great work... just one question: Does this preserve live priest mechanics, or is it directly imposing PTR behaviour? Eg. for lingering insanity.

In 29cbaab16ff8384e3774d39364cbf89c3f6d2645 : Please dont use find_spell inside a composite_ function. Create some const spell_data_t* on priest level, initialize somewhere and then use that.

Otherwise this looks good to me. I'll probably adjust some minor stuff once its merged, without changing any functionality.

lgkern:
Oh ok, when doing https://github.com/simulationcraft/simc/commit/29cbaab16ff8384e3774d39364cbf89c3f6d2645 I had no idea how to do it, so I ended up going for that implementation, but I can see that having it initialized only once can be better.

It does not preserve live priest mechanics. In 7.1.5 Shadow is receiving a huge overhaul, so doing an implementation that would work on both live and PTR would add a lot of depth and complexity to it. 

Edit: One thing that can be reviewed is Mind Sear (Now area damage around Mind Flay target). For some reason, while doing the correct damage, it is still doing a bunch of ticks for 0 damage. A way to see that is to do a pure single target sim, it will still show Mind Sear there, with ticks doing 0 damage.

PR #3262, opened 2016-12-19
[Hunter] Auto Shot & Windburst

nuoHep:
The first 2 are strictly bugfixes.
<br />
<br />The last one is a refactor prompted by the Auto Shot <-> Bullseye interaction bugfix. I'm not wedded to it, so if you don't like it I can easily refresh the branch without it.


Collisionc:
Looks good. I like the refactor.

PR #3266, opened 2016-12-20
[Hunter] First batch of 7.1.5 changes

nuoHep:
The first 2 are bugfixes for Live.
Everything else is strictly 7.1.5 stuff. Mostly the "easy" bits.


nuoHep:
Updated the branch as I found a small copy-n-paste bug in
  	[Hunter] Vulnerable is checked on Aimed Shot cast

nuoHep:
Eh, I guess that's not needed anymore...

PR #3292, opened 2016-12-29
Fixing more feral issues

Xanzara:
Thrash changed to be able to proc Ashamanes Bite and Primal Fury, verified behaviour on PTR as of 28-Dec-2016
Added the damage aura added to the Feral Druid passive.
Made sure that Shadow Thrash is also affected by a Moment of Clarity buffed clearcasting, verified behaviour on PTR as of 28-Dec-2016

### DIFF
--- a/engine/class_modules/sc_druid.cpp
+++ b/engine/class_modules/sc_druid.cpp
@@ -353,11 +353,11 @@ public:
     // Guardian
     // adaptive fur for each basic magic school
     buff_t* adaptive_fur_holy;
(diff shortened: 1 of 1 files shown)


gixxpunk:
Could you please redo your changes without the indentation mess and then re-send the pull request?

Xanzara:
Right, changed this commit to remove the indention mess but kept the indention changes in a few places where they actually made sense.

gixxpunk:
Thanks!

PR #3337, opened 2017-01-08
[Hunter] Stomp still refuses to cooperate

Jayezi:


nuoHep:
It's actually a problem with ordering between pet summon & buff gain (which I somewhat broken in the initial fixup commit). My latest PR should fix it for good.
Though there is still a problem of the first Dire Beast melee attack not benefitting from the buff (as it's execute()d higher in the virtual call chain).

PR #3379, opened 2017-01-14
[Rogue] Improved Sub Gear Set

samlongden:
+~10k over previous.


aethys256:
Thanks, I'll put it in the generate script later.
(In fact, you should modifiy the generate file and not the one used to sim)

aethys256:
More complete one here : https://github.com/simulationcraft/simc/pull/3388

PR #3380, opened 2017-01-14
Add interrupts and some spell name fixes

Sidoine:
Needed by Ovale Spell Priority. It fixes some talent name and add interrupt abilities to every script that lack them.


aethys256:
This must be done in the modules and not profiles directly.

PR #3385, opened 2017-01-14
Use https for decorated tooltips

elvador:
Since the WOWDB tooltip js is served over HTTPS there is no reason to use HTTP.

This fixes Chromes Mixed Content errors if the report is served over HTTPS.


Collisionc:
Interesting, thanks.

PR #3400, opened 2017-01-15
[Demon Hunter] Refactor metamorphosis_buff_t

EvanMichaels:
* Merge Havoc and Vengeance to use the same buff creation process
* Add tracking for Demonic triggers to prefer re-triggers by consecutive non-Metamorphosis Eye Beams
* Add buff.metamorphosis.triggered_by_demonic expression to build APL logic based on above functionality
* Call adjust_recharge_multiplier on cooldown.blade_dance when triggering Metamorphosis to ensure Death Sweep benefits from Metamorphosis haste if already on cooldown
* Add 7.1.5 armor buff for Vengeance to Metamorphosis

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -1816,7 +1816,7 @@ struct eye_beam_t : public demon_hunter_spell_t
       }
       else
       {
-        // Trigger before the execute so that the duration is affected by Meta haste
+        // Trigger Meta before the execute so that the duration is affected by Meta haste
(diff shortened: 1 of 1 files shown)


EvanMichaels:
For @Zarania to review, I had to resolve some conflicts from the submit last night.

EvanMichaels:
@Zarania 
Tweaked some logic in one place since Demonic was changed in today's hotfix not extend even in Meta after the first. Also changed the expression to extended_by_demonic instead of triggered_by_demonic due to this. We'll probably want to add APL logic for this to avoid using Eye Beam when it won't extend..

PR #3404, opened 2017-01-16
[Hunter] Reimplement Barrage tick behavior

Jayezi:
Double the damage since the spell data is no longer based off hitting the primary target for double damage and also put the chance to hit logic back in so Barrage works properly with Bullseye.


nuoHep:
I'm not sure about the base multiplier change (will have to test if stuff matches in game) and it's not going to work anyway as tick_actions don't use their own multipliers. So if you can separate out the ::impact() "revert" I'll happily merge that.

Collisionc:
Does the spell really just randomly miss targets? That's weird.

nuoHep:
Yeah, can't find the gif atm, but it basically always had a chance of not hitting secondary targets (which we never modeled correctly) and with 7.1.5 it got changed to a chance of not hitting any targets whatsoever (along with not requiring a target and doing the same damage to everything).

PR #3485, opened 2017-01-30
[Mage] Add variables to action lists to track legendary items

cgdangelo:
I think it would be helpful for the reader to know at a glance what a line is doing without having to know an item id. To that end I've added `variable` actions for any legendary items that were being tracked in each spec's APL; any legendaries that weren't already present were left out, but we can easily add the rest.

There were also some inconsistencies with actions checking for legendary buffs/effects, e.g. Arcane would check for the bracers buff being up as well as the item being equipped, whereas Fire would only check for the buff. I've added the appropriate variable checks to those actions so there's a clearer connection between item and buff.


Collisionc:
You can use equipped.draught_of_souls instead of equipped.number

cgdangelo:
...shit.

I'll use that instead.

Issue #3064, opened 2016-10-29
March of the legion set does not procc


SooiDP:
When I try to sim with the march of the legion ring/neck set it does not proc at all. I set the boss as a demon, so it should proc.

navv1234:
We do not have the proc implemented currently.

PR #3502, opened 2017-02-04
Implement March of the Legion Set

AlexanderKenny:
Not very familiar with the c++ side of SimC so it's not pretty but it seems to work. Test outputs at: https://www.dropbox.com/sh/182rkeem8ul51be/AABYjqAPMj_zBI6dnNBTyNsma?dl=0


Collisionc:
fixed https://github.com/simulationcraft/simc/issues/3064

PR #3525, opened 2017-02-09
[SV Hunter] slightly improved APL

Meatwookie:


engine/class_modules/sc_hunter.cpp
```diff
@@ -6441,60 +6441,62 @@ void hunter_t::apl_surv()
...
 
-  precombat -> add_action( "explosive_trap,if=!talent.steel_trap.enabled" );
+  precombat -> add_action( "explosive_trap" );
   precombat -> add_action( "steel_trap" );
   precombat -> add_action( "dragonsfire_grenade" );
   precombat -> add_action( "harpoon" );
 
   default_list -> add_action( "auto_attack" );
 
-  add_racial_actions( default_list );
```
Zuliane, line 6451: we want to keep this one I suppose

nuoHep: no, both blood fury & berserking are added a couple lines below
if we leave this here the newly added ones will basically have no effect as the "generic" ones are less strict

engine/class_modules/sc_hunter.cpp
```diff
@@ -6441,60 +6441,62 @@ void hunter_t::apl_surv()
...
   add_item_actions( default_list );
-
-  default_list -> add_action( "potion,name=prolonged_power,if=buff.spitting_cobra.up" );
+  
+  default_list -> add_action("berserking,if=(buff.spitting_cobra.up&buff.mongoose_fury.stack>2&buff.aspect_of_the_eagle.up)|(!talent.spitting_cobra.enabled&buff.aspect_of_the_eagle.up)");
+  default_list -> add_action("blood_fury,if=(buff.spitting_cobra.up&buff.mongoose_fury.stack>2&buff.aspect_of_the_eagle.up)|(!talent.spitting_cobra.enabled&buff.aspect_of_the_eagle.up)");
+  default_list -> add_action( "potion,name=prolonged_power,if=(talent.spitting_cobra.enabled&buff.spitting_cobra.remains)|(!talent.spitting_cobra.enabled&buff.aspect_of_the_eagle.remains" );
```
nuoHep: missing arcane torrent

engine/class_modules/sc_hunter.cpp
```diff
@@ -6441,60 +6441,62 @@ void hunter_t::apl_surv()
...
 
   default_list -> add_action( "auto_attack" );
 
-  add_racial_actions( default_list );
   add_item_actions( default_list );
-
-  default_list -> add_action( "potion,name=prolonged_power,if=buff.spitting_cobra.up" );
+  
+  default_list -> add_action("berserking,if=(buff.spitting_cobra.up&buff.mongoose_fury.stack>2&buff.aspect_of_the_eagle.up)|(!talent.spitting_cobra.enabled&buff.aspect_of_the_eagle.up)");
+  default_list -> add_action("blood_fury,if=(buff.spitting_cobra.up&buff.mongoose_fury.stack>2&buff.aspect_of_the_eagle.up)|(!talent.spitting_cobra.enabled&buff.aspect_of_the_eagle.up)");
```
nuoHep, line 6454: missing arcane torrent

engine/class_modules/sc_hunter.cpp
nuoHep, line 6455: missing a closing brace

engine/class_modules/sc_hunter.cpp
Zuliane: then we should remove https://github.com/simulationcraft/simc/blob/legion-dev/engine/class_modules/sc_hunter.cpp#L6290-L6295 else it's pointless, either keep the racial list and edit the changes there or erase the unneeded lines and write it below

nuoHep: I was actually going to drop add_racial_actions() after merging this pr as every spec wants different conditions on them anyway so there is no point in the "shared" function.

PR #3545, opened 2017-02-14
[Mage] Increase when running RoP to just use obelisk on CD since it e…

AlexanderKenny:
…nds up getting held, conditioning on RoP being up when RoP is not being run is also bad. Both cases seems best to just use it (assuming it will be back up for Combust). Simple conditional might not play well with kindling (but no one is taking it anyway/shouldn't be any worse than the old conditional in that regard). Basic tests here: https://www.dropbox.com/s/q7zlpde6fb0vkac/Obelisk.html?dl=0


AlexanderKenny:
On the note of special trinkets KJs was not being used for arc/frost. Added a commit for those.

PR #3547, opened 2017-02-14
[Mage] Adjust Flurry travel speed to match in-game data

vituscze:
Going through combat logs of Flurry cast at various distances and putting the resulting travel times through linear regression gives roughly 110 ms delay and 45 yd/s travel time.

Flurry travel speed is set manually via hotfix so that users can mess with it via `override.spell_data`.


vituscze:
Used data: https://docs.google.com/spreadsheets/d/1QvTPsK8Rs67cUF_L8WHljZqgTIkH-pezPBFwrBlP7Jo/edit?usp=sharing

PR #3596, opened 2017-02-27
Extend APL 'variable' operations to include a 'setif' operation.

AlexanderKenny:
Sets variable to 'value' if 'condition' returns true. Otherwise sets variable to 'value_else'. Currently this if/else assignment on variables can only be done inelegantly with two sequential set operations with an if= operator on the action. This improves readability in the APL.

### DIFF
--- a/engine/player/sc_player.cpp
+++ b/engine/player/sc_player.cpp
@@ -6271,12 +6271,14 @@ struct variable_t : public action_t
 {
   action_var_e operation;
   action_variable_t* var;
-  std::string value_str, var_name_str;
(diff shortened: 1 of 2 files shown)


AlexanderKenny:
While not strictly necessary (as I said you can work around it) I feel like it would be a strong addition to the variable functionality as I found it annoying to have to do so (and it only adds confusion to the APL for an unfamiliar reader). Additionally the variable functionality should likely be documented on the wiki but that is separate from this pull request.

Collisionc:
I like this change. Documentation is of course... on the list of things to do at some point. :)

PR #3597, opened 2017-02-27
[Hunter] Mok'Nathal Survival optimizations

jcadduono:
A little bit of dps increase, could probably be improved somewhat more.


nuoHep:
Thanks.

PR #3605, opened 2017-03-02
[Hunter] Better Warbelt usage

Jayezi:
Based on a tip from Huntermangun on Discord where forcing a 10 stack Aimed on 2 targets was better. Modified to roughly scale as a multiple of the target count.


nuoHep:
Thanks

PR #3612, opened 2017-03-05
[Hunter] Use accurate costs and cast times for MM apl

Jayezi:
Hardcoding values was not a good habit for 4pc. Small bump for dps.


nuoHep:
Thanks.
I actually did basically the same some time ago but was hesitating pushing it (I believe I actually pushed a branch with it in my fork).
If you have time you can also try to play with targetdie apl as I believe it's somewhat suboptimal (dropping it for our t19_m profile resulsted in a sligh bump in dps of around 2-3k).

PR #3626, opened 2017-03-16
Remove player interrupt from the casting raid event.

vituscze:


scamille:
What was the reason for this change? Just hiding bugs in the class modules?

So the only thing casting raid event now does is provide some interrupt possibilities for casters?

vituscze:
You don't usually stop casting and interrupt as soon as the boss starts casting. You try to time the interrupt in a way that doesn't disrupt your rotation, if possible.

Also, the interrupt part of this event was already available as a separate event (`interrupt_event_t`).

Frostedmages:
Additionally, this better models what the game actually does.

When a boss begins casting, the event itself does not interrupt the player. The player chooses to interrupt (either via a stopcasting macro as in the old days, or by casting their interrupt which normally has the stopcasting functionality built into it now).

PR #3664, opened 2017-04-01
[Rogue] Loaded Dice Implementation

Mystler:
This adds the implementation for the new Outlaw Rogue 7.2 gold trait "Loaded Dice" in both combinations (with Roll the Bones and Slice and Dice).


aethys256 (approved):
Looks ok as we discussed :)

PR #3674, opened 2017-04-04
[Hunter] Fix Vulnerable stats secondary target procs

Jayezi:
Was simming Piercing Shot and thought the APL was broken at first.

### DIFF
--- a/engine/class_modules/sc_hunter.cpp
+++ b/engine/class_modules/sc_hunter.cpp
@@ -730,9 +730,9 @@ struct vulnerability_stats_t
     }
   }
 
-  void update( hunter_t* p, const action_state_t* state )
+  void update( hunter_t* p, const action_t* a )
   {
-    if ( ! p -> get_target_data( state -> target ) -> debuffs.vulnerable -> check() )
+    if ( ! p -> get_target_data( a -> target ) -> debuffs.vulnerable -> check() )
     {
       no_vuln -> occur();
     }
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_hunter.cpp
```diff
@@ -730,11 +730,9 @@ struct vulnerability_stats_t
     }
   }
 
-  void update( hunter_t* p, const action_t* a )
+  void update( hunter_t* p, const action_state_t* state )
```
nuoHep: You could name it `s` and avoid most of the churn below.

engine/class_modules/sc_hunter.cpp
```diff
@@ -3531,10 +3529,21 @@ struct piercing_shot_t: public hunter_ranged_attack_t
...
+    if ( execute_state -> target != p() -> target )
+    {
+      original_target = execute_state -> target;
+      execute_state -> target = p() -> target;
+    }
+
+    vulnerability_stats.update( p(), execute_state );
+
+    if ( original_target )
+      execute_state -> target = original_target;
```
nuoHep: This one I'm not sure about. I always though that the target from execute state is the actual target of the action, so modifying it to point to the current player target is actually "backwards". Maybe I'm simply wrong or missing smth.

nuoHep: Yeah... looks like I was wrong this whole time, at least for aoe actions.
But, using player target is also wrong, and I believe we should be using action -> target here.

nuoHep:
2 comments, the second one I'm not sure about.
One thing worth checking is if using `s -> action -> target` inside the update method (and dropping the whole hunk inside pis execute) will give the same results (I believe it should).
Or simply using `a -> target` and leaving update signature the same.

nuoHep:
Thanks

navv1234:
I'm unsure what all this is, but generally speaking, action_state_t::target is always the safe bet for the actual target of the ability.

action_t::target is not the correct solution in all situations.

Jayezi:
For piercing shot the state's target had been looped to a secondary target by the time execute returned but I wanted to track specific info on the primary target (it does more damage to primary/primary is what we're trying to maximize the patient sniper stacks on) and other info on the secondary targets. Based on the apl the primary target is always the player target but yea potentially the action target could be different, but at that point something else is going on or something happened to the apl since all the vulnerable window stuff it uses is based off the current target and done before it gets to actions.

navv1234:
Yeah we went through this with @nuoHep in discord, it's all good.

PR #3703, opened 2017-04-16
Adding gear output to the JSON reporting

pontifechs:
The HTML report already has this, so it'd be nice if the JSON had it as well.

### DIFF
--- a/engine/report/sc_report_json.cpp
+++ b/engine/report/sc_report_json.cpp
@@ -58,6 +58,7 @@ public:
   JsonOutput& operator=( T v )
   { v_ = v; return *this; }
 
+  // Assign a string to the current value (v_), and have RapidJSON do the copy.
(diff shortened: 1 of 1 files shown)


navv1234:
Change the "thing" to "options" or something. Also if you want to do this properly, it might actually make sense to change the "thing" to an "options" object, and put the field information from item_t::parsed_input_t there.

pontifechs:
Yeah, I couldn't quite come up with a good name for it, so I just put in thing and moved on. Forgot to come back and think about the name. I think just "encoded_item" works pretty well. 

As far as splitting it out into an object, that sort of defeats the purpose I had in mind. The original thought was to be able to recreate that item exactly in another sim execution. If I split it out into the individual components, now the consumer has to be able to reassemble them to be able to send it back to simc.

PR #3719, opened 2017-04-22
Subtlety Rogue 7.2.5 PTR Changes

Mystler:
This should cover all of the current PTR changes for Subtlety Rogues and introduces an additional APL that is run when ptr is set. The PTR APL only contains minor changes for now, though.


aethys256 (approved):
Everything looks fine from my first look, good job!

PR #3726, opened 2017-04-27
[Mage] Changing APL to use Prolonged Power over Deadly Grace for Fire

ForgyGaming:
Changed the fire mage potion override from Deadly Grace + Prolonged power (that had no conditions assigned) to just Prolonged Power since it seems to perform better. Results of sim of my gear and default NH gear attached. Normal talent setup and ones that should favor deadly grace over prolonged, yet prolonged always won

http://www.files.com/set/590153ba5fa1f


Frostedmages:
I can do this without merging a few different commits.

ForgyGaming:
If the end result is the same its all good :) Thanks

PR #3729, opened 2017-04-27
[Mage] Arcane Explosion over Arcane Blast

ShinLei:
Im not sure if im on the right section for this, sry if its wrong.

Is it possible to write an Profile for Arcane Explosion over Arcane blast?

The reason for it is this post.

https://www.altered-time.com/forum/viewtopic.php?f=3&t=4315


Frostedmages:
No, we are not going to be merging Mage-Legion-Dev into Legion-Dev right now. If you want to make a clean pull request for legion-dev, I can look at that. I'll take a look at that thread too.

PR #3733, opened 2017-04-28
[Mage] Arcane/Frost proc tracking

cgdangelo:
Fire has several procs tracking Heating Up/Hot Streak events, but none exist for Arcane/Frost.

I've added procs for both specs: Arcane Missiles for Arcane, Fingers of Frost/Brain Freeze for Frost. For BF, I made a distinction between _generated_ (no current BF proc) and _regenerated_ (BF proc applied after delay).

There's admittedly a lot of copy/paste here so it might be worthwhile to refactor a bit and minimize code duplication. If everything else seems alright otherwise then I can look into that as well.

### DIFF
--- a/engine/class_modules/sc_mage.cpp
+++ b/engine/class_modules/sc_mage.cpp
@@ -387,6 +387,7 @@ public:
           * arcane_missiles_wasted; // Additional AM generated at max stacks
 
     proc_t* brain_freeze_generated, // BF proc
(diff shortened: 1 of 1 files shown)


vituscze:
As discussed on Discord, default buff tracking does most of this and the extra stuff it doesn't (tracking expired stacks, BF generate/regenerate) is not all that useful.

PR #3740, opened 2017-05-04
[Hunter] Implement Aspect of the Wild GCD reduction from 7.2.5 PTR

overmind2:
Following the implementation of the Adrenaline Rush GCD modification in sc_rogue.cpp.

Also remove two spurious lines in sc_action.cpp: P_{CRIT,COOLDOWN}
are valid values for misc_value1() but not for subtype().


overmind2:
Closing this as 80eafff already implemented this in a more correct way.

PR #3746, opened 2017-05-09
[Hunter] Make FotE and its effect on Mongoose Fury work accurately

Jayezi:
The sv apl dude was in discord trying to mess with interrupting the channel and it turned out this hadn't been updated in a year.


nuoHep:
I'd drop `virtual` from the `last_tick` override, but meh. Not worth respinning it due to this. Merged as is.
Thanks.

PR #3768, opened 2017-05-21
[Blood] fixing marrowrend + bone shield stacks + adding basic apl

seanpeters86:
Marrowrend now properly adds stacks based on artifact abilities, and whether dancing rune weapon is active. 

Bloodworms is also a WIP addition as well.

First pass at a rotation has been added, did not add food buff as currently there is no way to manually set suramar feast to 500 STR (default to 500 Stamina for tanks, however most tanks chose to eat the food buff in DPS spec to get the strength instead). Able to add ```enchant_strength=500``` to combat this, but not sure if this can easily be added in the apl.

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -7363,8 +7363,42 @@ void death_knight_t::default_apl_dps_precombat()
 
 void death_knight_t::default_apl_blood()
 {
-    // TODO: mrdmnd - implement
-  default_apl_frost();
+  action_priority_list_t* def = get_action_priority_list( "default" );
+  action_priority_list_t* st  = get_action_priority_list( "st" );
+
+  // Setup precombat APL for DPS spec
+  default_apl_dps_precombat();
+
+  // Racials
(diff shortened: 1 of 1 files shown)


AlexanderKenny:
For the basic APL probably want to setup default food for blood instead of 'disabled' and you may wish to just use 'potion' instead of 'potion,name=prolonged_power' for the in-combat potion (same as the pre-combat one currently is). This would allow a user to use the potion override to edit the in-combat potion without having to generate the APL and then edit it manually.

seanpeters86:
@AlexanderKenny just changed the potion line, good call there. I added the explanation of the food buff problem above, not sure how to handle it. 

```
First pass at a rotation has been added, did not add food buff as currently there is no way to manually set suramar feast to 500 STR (default to 500 Stamina for tanks, however most tanks chose to eat the food buff in DPS spec to get the strength instead). Able to add enchant_strength=500 to combat this, but not sure if this can easily be added in the apl.
```

AlexanderKenny:
Fair point on the feast missed that on your initial comment. Probably still worth putting something there in the meantime (whether lemon_herb_filet or one of the 375 stat foods).

PR #3780, opened 2017-05-29
Fix arcane burn duration value to return previous burn length

AlexanderKenny:
if outside burn phase (or current burn duration if inside one). last_disable and last_enable default to timespan_t::zero() anyway so in the scenario that neither has been set and we are outside burn (I.E. before the first burn takes place) they will return 0 duration as previous anyway.


vituscze:
It looks like current Arcane APLs do not rely on this behavior, so I'm just going to adjust the wiki and leave it as-is.

Issue #3769, opened 2017-05-23
Frost DK APL & Default Profile update


Melekus:
I've been working on an APL update for frost Death Knight which is a slight improvement for all talent builds currently in use and adds custom support for a few trinkets (Horn of Valor and Draught of Souls).

https://pastebin.com/V0qRM44P

I've also made a new default profile for frost DK : it still uses Nighthold-only gear but has a better stat distribution for a ~1% dps increase.

https://pastebin.com/eR1tzyZv

PR #3786, opened 2017-05-31
Frost Death Knight APL and default profiles update

Melekus:
I modified the class module and the various default frost DK profiles to reflect the updated APL and profile I submitted in https://github.com/simulationcraft/simc/issues/3769 a few days ago.


navv1234:
You should not manually update the sample profiles. Instead, update ```generate_Death_Knight_XX.simc``` with the correct talent, artifact, and gear setup in the corresponding tier sample profile directory, an optionally run "generate_profiles.sh (or .bat on windows). The default APL comes from the c++ side.

Reason for this is, the sample profiles are automatically generated from those "generate" .simc files, and any changes you make to the generated ones are automatically lost when someone re-generates the sample profiles.

Melekus:
Changed the generate files as well

PR #3818, opened 2017-06-12
Remove _NH from Raid_T20M profile

wfarr:
All 3 hunter profiles were using this suffix still causing this profile to be invalid


aethys256:
Oups, ty :P

PR #3830, opened 2017-06-15
[Priest] Adding AS + Sephuz

seanpeters86:
Swapping to AS instead of SL, and replacing the Cloak legendary with Sephuz


seanpeters86:
2.5% increase for the Mythic profile, and a 3.1% increase for the Heroic Profile.

PR #3840, opened 2017-06-20
Add pet and base spell modifiers for Elemental Shaman spec aura

wfarr:
I'm not 100% sure this covers everything appropriately just yet, but should be a good starting point to implement support for the spec aura for Elemental Shaman, which recently got put to real use in the latest round of hotfixes.

cc @Bloodmallet for 👀

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -382,6 +382,7 @@ public:
     const spell_data_t* chain_lightning_2; // 7.1 Chain Lightning additional 2 targets passive
     const spell_data_t* elemental_focus;
     const spell_data_t* elemental_fury;
+    const spell_data_t* elemental_shaman;
     const spell_data_t* flame_shock_2; // 7.1 Flame Shock duration extension passive
     const spell_data_t* lava_burst_2; // 7.1 Lava Burst autocrit with FS passive
     const spell_data_t* lava_surge;
@@ -1050,6 +1051,12 @@ public:
 
       maelstrom_gain = effect.resource( RESOURCE_MAELSTROM );
       ab::energize_type = ENERGIZE_NONE; // disable resource generation from spell data.
+    }
+
+    if ( ab::data().affected_by( player -> spec.elemental_shaman -> effectN( 5 ) ) )
+    {
+      ab::base_multiplier *= 1.0 + player -> spec.elemental_shaman -> effectN( 5 ).percent();
+      ab::base_multiplier *= 1.0 + player -> spec.elemental_shaman -> effectN( 7 ).percent();
     }
 
     if ( ab::data().affected_by( player -> spec.enhancement_shaman -> effectN( 1 ) ) )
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_shaman.cpp
```diff
@@ -1052,6 +1053,12 @@ struct shaman_action_t : public Base
       ab::energize_type = ENERGIZE_NONE; // disable resource generation from spell data.
     }
 
+    if ( ab::data().affected_by( player -> spec.elemental_shaman -> effectN( 5 ) ) )
+    {
+      ab::base_multiplier *= 1.0 + player -> spec.elemental_shaman -> effectN( 5 ).percent();
+      ab::base_multiplier *= 1.0 + player -> spec.elemental_shaman -> effectN( 7 ).percent();
```
wfarr: This should be the pet aura. I'm not sure if this is the correct way to implement this or not.

engine/class_modules/sc_shaman.cpp
```diff
@@ -1052,6 +1053,12 @@ struct shaman_action_t : public Base
       ab::energize_type = ENERGIZE_NONE; // disable resource generation from spell data.
     }
 
+    if ( ab::data().affected_by( player -> spec.elemental_shaman -> effectN( 5 ) ) )
+    {
+      ab::base_multiplier *= 1.0 + player -> spec.elemental_shaman -> effectN( 5 ).percent();
```
wfarr, line 1058: This should be the effect covering all non-pet spells. Effect 6 appears identical to it, so I just chose one of the two of them to use.

wfarr:
Here are the relevant spell effects: https://github.com/simulationcraft/simc/blob/b951d34880b968cc641e4fa431c3fec3598607af/SpellDataDump/shaman.txt#L2816-L2826

engine/class_modules/sc_shaman.cpp
wfarr: SpellDataDump confirms effect 5 is what's in actual use.

engine/class_modules/sc_shaman.cpp
```diff
@@ -1850,6 +1856,11 @@ struct pet_action_t : public T_ACTION
 
     this -> special = true;
     this -> may_crit = true;
+
+    if ( this->data().affected_by( p() -> o() -> spec.elemental_shaman -> effectN( 7 ) ) )
+    {
+      this->base_multiplier *= 1.0 + p() -> o() -> spec.elemental_shaman -> effectN( 7 ).percent();
+    }
```
wfarr: I'm unsure if this is working as I intend just yet.

Bloodmallet:
I'm adding @navv1234 for the pet stuff. the other part looks good for my untrained eye.

Navv please have a look at the pet multiplier. Wfarr wrote that somehow the dmg of the Elementals wasn't changed.

wfarr:
> I'm adding @navv1234 for the pet stuff. the other part looks good for my untrained eye.
> 
> Navv please have a look at the pet multiplier. Wfarr wrote that somehow the dmg of the Elementals wasn't changed.

I ended up removing the conditional to see if the player was affected by the aura's pet effect and always applying it for now, and the numbers all make sense.

If you know of a better conditional to wrap it in rather than always applying it, I'm happy to fix up.

navv1234:
You can put the pet damage effect to player_t::composite_player_pet_damage_multiplier() (where it should honestly be)

wfarr:
> You can put the pet damage effect to player_t::composite_player_pet_damage_multiplier() (where it should honestly be)

4522c7f0e should address this

PR #3846, opened 2017-06-22
[Hunter] Small apl improvements

Jayezi:
Add checks for better BM cooldown usage, adds ~5k dps.
Handle MM T20 2p better, adds a decent ~20k dps to the 2p value.


engine/class_modules/sc_hunter.cpp
```diff
@@ -6219,15 +6219,15 @@ void hunter_t::add_item_actions( action_priority_list_t* list )
...
         else if ( item.name_str == "tome_of_unraveling_sanity" )
         {
-          list -> add_action( "use_item,name=" + item.name_str + ",if=((buff.trueshot.up|cooldown.trueshot.remains<13|cooldown.trueshot.remains>30)&(target.time_to_die>cooldown+duration*2))|target.time_to_die<26|buff.bullseye.react=30" );
+          list -> add_action( "use_item,name=" + item.name_str + ",if=((cooldown.trueshot.remains<13|cooldown.trueshot.remains>30)&(target.time_to_die>cooldown+duration*2))|target.time_to_die<26|buff.bullseye.react=30" );
         }
         else if ( item.name_str == "kiljaedens_burning_wish" )
         {
-          list -> add_action( "use_item,name=" + item.name_str + ",if=buff.trueshot.up|cooldown.trueshot.remains>20" );
+          list -> add_action( "use_item,name=" + item.name_str + ",if=cooldown.trueshot.remains>20" );
         }
```
nuoHep, line 6231: Are you sure about at least this one? Looks pretty broken to me as it basically ensures it will never get used under Trueshot. Other use_item changes also look kinda suspicious.

Jayezi: It was just something I didn't think about right the first time when I added them, buff.trueshot.up is always covered under cooldown.trueshot.remains since it's on cooldown if it was just used.

nuoHep: Ah, right.

nuoHep:
Thanks

PR #3894, opened 2017-07-07
[Demon Hunter] Adjust for new CB duration

talst:


EvanMichaels:
Hi Talst, that change looks good but would also need to make the change in the sc_demon_hunter.cpp 

cd->add_talent(this, "Chaos Blades", "if=buff.metamorphosis.up|cooldown.metamorphosis.adjusted_remains>60|target.time_to_die<=12");

I can make this change if you want. Let me know if you want to change your branch or have me do it. Thanks for noticing this.

EvanMichaels:
Fixed in https://github.com/simulationcraft/simc/commit/36e2ed61f2982afa8bcb5d1443a4230696a067e6

PR #3925, opened 2017-07-24
[Hunter] Update Warbelt logic to account for changes since f24b97a

Jayezi:
warbelt
Aimed is stronger in general due to UA being added since,
it also results in forcing more T20 2pc buffs.
There's no longer any visible cutoff where Aimed falls behind like before.
Dumping the buff when the next multishot would waste about half of the stacks
seems to be the most consistent.

From BM discord members, Qa'pla condition apparently just barely prefers 1 over 3.


nuoHep:
Thanks

PR #3926, opened 2017-07-24
Persist copied target list between calls

tomasolander:
In `action_t::ready()` and while cycling targets, recreating the `std::vector` for the copied target list calls is a rather expensive operation, and the function is called very frequently during a simulation. Declaring it `static` persists it between calls and results in a minor to modest performance increase depending on the sim.

I did some performance analysis testing, and this seemed like low-hanging fruit. As far as I've been able to tell the function is only called from the main thread, so a `static` declaration should be a safe and "easy" enough fix, but might be a bit hacky. Perhaps it should be implemented in some other way, though. Let me know what you think!

To benchmark, I repeatedly ran [this SimC script](https://github.com/simulationcraft/simc/files/1169681/t20.txt) which includes most T20 sample profiles, running via the SimC CLI built locally with VS 2015. It might require more extensive testing that I'm not aware of.

**Multi-threaded:**

```
deterministic=1
threads=4
iterations=10000
```

Before:
```
Baseline Performance:
  RNG Engine    = sse2-sfmt (deterministic)
  Iterations    = 10000 (2500, 2500, 2500, 2500)
  TotalEvents   = 1189955637
  MaxEventQueue = 1269
  TargetHealth  = 11433676292
  SimSeconds    = 3008477
  CpuSeconds    = 2460.969
  WallSeconds   = 627.227
  SpeedUp       = 1222
```

After:
```
Baseline Performance:
  RNG Engine    = sse2-sfmt (deterministic)
  Iterations    = 10000 (2500, 2500, 2500, 2500)
  TotalEvents   = 1189787064
  MaxEventQueue = 1282
  TargetHealth  = 11429384357
  SimSeconds    = 3007994
  CpuSeconds    = 2333.859
  WallSeconds   = 588.722
  SpeedUp       = 1289
```

**Single-threaded:**

```
deterministic=1
threads=1
iterations=10000
```

Before:
```
Baseline Performance:
  RNG Engine    = sse2-sfmt (deterministic)
  Iterations    = 10000
  TotalEvents   = 1189806900
  MaxEventQueue = 1298
  TargetHealth  = 11430916356
  SimSeconds    = 3007278
  CpuSeconds    = 1560.375
  WallSeconds   = 1563.768
  SpeedUp       = 1927
```

After:
```
Baseline Performance:
  RNG Engine    = sse2-sfmt (deterministic)
  Iterations    = 10000
  TotalEvents   = 1189806900
  MaxEventQueue = 1298
  TargetHealth  = 11430916356
  SimSeconds    = 3007278
  CpuSeconds    = 1449.297
  WallSeconds   = 1451.828
  SpeedUp       = 2075
```

### DIFF
--- a/engine/action/sc_action.cpp
+++ b/engine/action/sc_action.cpp
@@ -1943,7 +1943,11 @@ bool action_t::ready()
     // Note, need to take a copy of the original target list here, instead of a reference. Otherwise
     // if spell_targets (or any expression that uses the target list) modifies it, the loop below
     // may break, since the number of elements on the vector is not the same as it originally was
-    std::vector< player_t* > ctl = target_list();
+
+    // Since this is a hot function, declare this vector static as an optimization to persist the
+    // copied vector instance between calls, since constructing it from scratch is expensive.
+    static std::vector< player_t* > ctl;
+    ctl = target_list();
     size_t num_targets = ctl.size();
 
     if ( ( option.max_cycle_targets > 0 ) && ( ( size_t ) option.max_cycle_targets < num_targets ) )


navv1234:
action_t::ready() is called from all threads, so static here is not a good idea. The other problem is that expressions and execution currently shares the same target cache, so if an expression modifies it (possible, for example spell_targets expression), then things will break horribly.

One way to fix this (I have not had time, since it's maybe somewhat extensive) would be to separate target caches for different purposes. Then grabbing a reference there would work.

tomasolander:
Oh, I see - my mistake. Apologies!

I'm afraid my understanding of the simulator isn't good enough yet to have poked at the target cache, so please let me know if I'm just way down a false path here, or one that isn't worthwhile. Would using TLS to make it local to the thread be an option, if that'd make it "safe"? From what I can tell, *constructing* the vector from scratch every time this path is taken ends up costing a lot, which is what I'm trying to take a crack at making more performant.

navv1234:
I'm unsure of what the effects of making the vector thread local would have, but it would likely make things thread safe in this context.

As a side note, if your interest is mainly single target simulations (as above), a significantly better optimization would be to ensure that cycle_targets options are "no-op" in single target simulations. Something like changing the ```if ( cycle_targets )``` to ```if ( sim -> target_list.size() > 1 && cycle_targets )``` would probably yield quite a bit of benefit too. Note that same treatment should be applied to the target_if_mode line above (```target_if``` option is the slightly saner way of doing "cycle targets").

If you want to test the effects of different solutions, please use a multi-target simulation with dynamic enemies (HecticAddCleave or BeastLord fight styles).

tomasolander:
Duh, that makes sense. Thank you very much for the pointers!

I'll take a crack at things and hopefully return with a new PR. :)

PR #4039, opened 2017-09-20
[Death Knight] Blood support update

Melekus:
Improved DRW : Consumption and legendary shoulders' stacks are now properly mirrored and mouth of hell now creates two pets.
Fixed Consumption damage.
Added the spec aura buff as well as the 7.2 trait that increases damage by 10%

The damage is still low compared to live values. But at least it's much better than before.


Melekus:
Travis still working, will merge after sleep if everything is fine

PR #4088, opened 2017-10-16
[Hunter] Track Critical Aimed benefit for Aimed Shot, clean redundant…

Jayezi:
… procs

Declutter the procs section since some of the same information is in the buff table.


nuoHep:
Thanks. Been meaning to clean this up some day too.

PR #4132, opened 2017-11-13
Extra tweaks to Enhancement T21 profile

Gengirawr:
Food buff, and a talent change were missing


Bloodmallet:
Dismissing this because it wasn't in the generator file. Fixed in the next pull request.

PR #4137, opened 2017-11-14
optimized crucible for enhancement

Gengirawr:
Enhancement crucible is optimized when having 3 windstrikes traits and it was changed for 2 only (which is the same as elemental so maybe that was the confusion, or maybe something it was something that I'm not aware) in any case, proposing this change to go back to the 3 windstrike traits


vituscze:
The profile currently has 3 extra ranks in that trait, two from crucible and one from a relic. Having another one from crucible is thus impossible. Closing.

N1gh7h4wk:
IF you want to change the crucible to windstrike you need to change the relic (if there is a better one instead of the windstrike relic).

Gengirawr:
We are not using a windstrike relic though, we are using a windsurge, gathering storms and gathering of the maelstrom
https://gyazo.com/c3852f83296bad69b4ed7ab9f5d4ca29 (the crucible you propose)
https://gyazo.com/e769446eed10bf1adde3a4073b307320 (the one I propose)
as you see windstrikes is 4+0+3
where is this discrepancy coming from?

vituscze:
I think someone else changed the relics on that profile. We had to change it because simc would refuse to generate the profile.

Gengirawr:
Regarding relics, we found optimal to use all Argus trinkets due to the higher ilvl:
fire: flickering ember of rage
iron: Mote of the Forgemaster
storm: Conch of the Thunderer
none of them wind strikes so the crucible should reflect that since it is our best trait.

Not only that, but it is very weird even the 2 second trinket is changed to a less optimized one 
https://gyazo.com/5b03d926d44e1af0f353ea558e978ebe
using scorgewing instead of gorshalach (and the stack present right now is using gorshalach so not sure how that change from one build to another) so it should be changed back

sorry for causing so much trouble, initiating my days of learning and trying to support this

vituscze:
Oh, one minor thing. You want to do the changes in the T21_Generate_Shaman.simc file,  not the generated profile.

Kutikuti:
I made the changes. Current setting is what i found is best. Show me better sims with a different settings and then we can change.

Gengirawr:
@Kutikuti somehow copying the gear you propose has some issue with stats even though it says it's 960 ilvl gear it looks like stats of 930
https://www.raidbots.com/simbot/report/dc3s9Eni4rnJbVzaMy844t (example on my tests I brought it up for 50% haste for the extra damage bump, and yet your gear (which is the same bar the trinket) I get only 38% both in raidbots and lastest nightly

Gengirawr:
correction, since I changed to hailstorm, haste is only 43%, but there is still a discrepancy

Gengirawr:
Found the problem, off hand is not there

Gengirawr:
Ok so here is the issue:
2 windsurge relics at 970 ilvl when there is only one: https://gyazo.com/e2c13783e30d1c097d8e4152c0575a80
https://gyazo.com/0cfc6993c68edd6d7f6bdc313a2cf988
This combination does not exist baseline, I was under the impression this was discouraged

Gengirawr:
@Kutikuti 
https://www.raidbots.com/simbot/report/cLve8J8GxYHSA7rgkoKPRT
baseline relics or not, Gorsha still seems to be the best trinket, your thoughts?

N1gh7h4wk:
Seems like a problem with the bonus id on Kutikutis part

Kutikuti:
yes, didn't know relic was different than other gear. And now legacy is better, yes

vituscze:
Should be fixed in 55a8a9acd3463674156b34645710800c89a9b115 and ab07768315f6fa6728b6eeaad4ca8c9f88fccfc9

PR #4138, opened 2017-11-14
Update T21_Death_Knight_Unholy.simc

Maxweii:
Optimizations to profile.


vituscze:
You want to do the changes here: https://github.com/simulationcraft/simc/blob/legion-dev/profiles/generators/Tier21/T21_Generate_Death_Knight.simc

Maxweii:
Put it in the right area now

PR #4148, opened 2017-11-17
Update T21_Druid_Balance.simc

Gebuz:
crucible: shadowbind -> torment the weak, empowerment -> dark side of the moon
gear: changed boots, neck and ring.


vituscze:
You want to modify this file: https://github.com/simulationcraft/simc/blob/legion-dev/profiles/generators/Tier21/T21_Generate_Druid.simc

The one you're modifying is the generated profile and the changes would be lost the next time someone ran the generate script.

Gebuz:
thanks, I'll do that

PR #4159, opened 2017-11-24
[Balance] Updated APL and fixed a small bug

Jundarer:


Bloodmallet:
Gebuz in Discord wrote that this PR came from them.

PR #4184, opened 2017-12-05
[Hunter] Reduce the BM T21 4pc effect when wearing Convergence

Jayezi:


Mystler:
Looks like Blizzard changed their mind again. https://twitter.com/Wowhead/status/938074929232097280

PR #4213, opened 2017-12-29
Update T21_Demon_Hunter_Havoc.simc

Petra105:
Swapping out the 960 Sliver of Corruption relics for 970 Condensed Blight Orb relics is a dps increase.


Jundarer:
Please change T21_Generate_Demon_Hunter.simc which is used to generate this profile.

PR #4229, opened 2018-01-09
Hectic Add Cleave - Beastlord  profiles Update

CrystalGGG:
Added Timers for Add spawns when talented in Whirling Dragon Punch talent - 10% dps increase on Hectic add cleave - 5% beastlord


vituscze:
Changes to default APL need to be made in the class module, not in the generated simc profiles.

lgkern:
Those changes need to be made on class_modules/sc_monk.simc and not on the profiles.

PR #4289, opened 2018-03-03
Add proc_sephuz for warlocks

sharpobject:


sharpobject:
Heard on warlock discord that fake actions aren't getting into the warlock module. So I should come back with spell_lock and singe_magic. Thanks!

PR #4302, opened 2018-03-27
Add consistent stats reporting to JSON

seriallos:
* Add crit/haste/mast/vers percents to actor
collected_data.buffed_stats. Add consistently named stats to JSON
* Get `profileset_output_data` working again
* Add "stats" arg to `profileset_output_data` which will add
`overrides.stats` to profileset JSON output.

I think this is mostly fine but there are probably some questionable choices I made. I'll highlight them in comments to see if anyone has a strong feeling about fixing/changing them.

### DIFF
--- a/engine/sim/sc_profileset.cpp
+++ b/engine/sim/sc_profileset.cpp
@@ -1231,9 +1231,12 @@ void save_output_data( profile_set_t& profileset, const player_t* parent_player,
 
     // tertiary stats
 
-    profileset.output_data().avoidance( buffed_stats.avoidance );
-    profileset.output_data().leech( buffed_stats.leech );
-    profileset.output_data().speed( buffed_stats.run_speed );
+    profileset.output_data().avoidance_rating( player -> composite_avoidance_rating() );
+    profileset.output_data().avoidance_pct( buffed_stats.avoidance );
+    profileset.output_data().leech_rating( player -> composite_leech_rating() );
+    profileset.output_data().leech_pct( buffed_stats.leech );
+    profileset.output_data().speed_rating( player -> composite_spell_haste_rating() );
+    profileset.output_data().speed_pct( buffed_stats.run_speed );
   }
 }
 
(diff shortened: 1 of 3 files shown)


engine/sim/sc_profileset.hpp
```diff
@@ -188,11 +188,27 @@ class profile_output_data_item_t
...
-  std::string                             m_artifact;
-  std::string                             m_crucible;
-  std::vector<profile_output_data_item_t> m_gear;
+  race_e                                       m_race;
+  std::vector<talent_data_t*>                  m_talents;
+  std::string                                  m_artifact;
+  std::string                                  m_crucible;
+  std::vector<profile_output_data_item_t>      m_gear;
+
+  double    m_crit_rating,
```
seriallos, line 197: I played around with having these stored in a different class but got in over my head with implicitly deleted copy constructors and the like.

These will be allocated if a user has any non-empty `profileset_output_data` option but it shouldn't affect the default profilesets without that option.

engine/sim/sc_profileset.cpp
```diff
@@ -1115,7 +1123,7 @@ statistical_data_t metric_data( const player_t* player, scale_metric_e metric )
   }
 }
 
-void save_output_data( std::unique_ptr<profile_set_t>& profileset, const player_t* parent_player, const player_t* player, std::string option )
+void save_output_data( profile_set_t* profileset, const player_t* parent_player, const player_t* player, std::string option )
```
navv1234: Would rather make sense to pass it as reference here, and above in the `save_output_call` pass it as a reference too (the `set` parameter).

engine/sim/sc_profileset.hpp
```diff
@@ -418,7 +524,7 @@ void create_options( sim_t* sim );
 
 statistical_data_t collect( const extended_sample_data_t& c );
 statistical_data_t metric_data( const player_t* player, scale_metric_e metric );
-void save_output_data( std::unique_ptr<profile_set_t>& profileset, const player_t* parent_player, const player_t* player, std::string option );
+void save_output_data( profile_set_t* profileset, const player_t* parent_player, const player_t* player, std::string option );
```
scamille: I would change the signature to 
void save_output_data( profile_set_t& profileset, const player_t* parent_player, const player_t* player, std::string option );

engine/sim/sc_profileset.hpp
```diff
@@ -227,6 +243,96 @@ class profile_output_data_t
 
   profile_output_data_t& gear( const std::vector<profile_output_data_item_t>& v )
   { m_gear = v; return *this; }
+
+  double crit_rating() const
```
scamille, line 250: Do you really want all those accessors? Why not just use public class members?

Of course this is the proper way to do it, so if you are comfortable with the maintenance costs and since you already added it, why not.

engine/sim/sc_profileset.hpp
scamille: Edit: Should have re-read what navv wrote, the exact same thing ;)

engine/sim/sc_profileset.hpp
seriallos: Mainly because I was copy/pasting and wanted to stay consistent with the original code

PR #4308, opened 2018-04-21
Misc elemental updates

wfarr:
First pass at updating Elemental to account for alpha talent changes over the past few months.

* Update Aftershock to reflect new implementation
* Remove Molten Fury
* Remove Elemental Attunement
* Add Exposed Elements

cc @Bloodmallet


Bloodmallet:
From what I tested: works

PR #4311, opened 2018-04-28
[Arcane] Base BfA APL

ronoudgenoeg:


vituscze:
It's a bit early to push an experimental APL as the default one. Especially given that Mana Adept might be changing (hopefully sooner than later) and there are multiple Arcane APLs that are being worked on.

Keep iterating on it and once we get closer to a state where the game is more finalized, we can think about what APL should be in simc by default.

PR #4330, opened 2018-06-16
[survival] harpoon fix

KuroTheCrazy:
Fixed harpoon to be able to be used more than once per fight by changing a bool set.

### DIFF
--- a/engine/class_modules/sc_hunter.cpp
+++ b/engine/class_modules/sc_hunter.cpp
@@ -3381,7 +3381,7 @@ struct harpoon_t: public hunter_melee_attack_t
       terms_of_engagement -> execute();
     }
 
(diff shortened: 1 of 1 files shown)


nuoHep:
You may as well remove it (the bool) completely at this point.

But, using Harpoon rotationally 'properly' requires more work tbh:
* Movement simulation (going out of melee range). While its effect is pretty small it still should be modeled somehow
* Proper handling of Helbrine, as it's now going to get applied on each harpoon (though tbh the current implementation of it is kinda shit anyway, as it expects there being only one harpoon line in the apl). While moot in bfa it's pretty important for prepatch.

nuoHep:
I've pushed some stuff which should resolve this a bit better.

PR #4331, opened 2018-06-18
Add `target_error_role` option for target_error behavior

seriallos:
Allows user choice of which role to use for target_error. Default has
been changed to DPS rather than decide based on the player role. I
believe this matches the vast majority of SimC usage.

It may be worth looking into allowing choice over the specific metric
rather than just role but that would have required deeper work in the
target metric code which I wasn't comfortable with (and, again, isn't
used very much today)

### DIFF
--- a/engine/player/sc_player.cpp
+++ b/engine/player/sc_player.cpp
@@ -11745,7 +11745,11 @@ void player_collected_data_t::collect_data( const player_t& p )
   {
     double metric=0;
 
-    switch( p.primary_role() )
+    // ROLE is used here primarily to stay in-line with the previous version of the code.
+    // An ideal implementation is probably to rewrite this to allow specification of a scale_metric_e
+    // to make it more flexible. That was beyond my capability/available time and it would also likely be
+    // very, very low use (as of legion/bfa, almost all tanks are simming DPS, not survival).
+    switch( p.sim -> target_error_role )
     {
     case ROLE_ATTACK:
     case ROLE_SPELL:
--- a/engine/player/sc_player.cpp
+++ b/engine/player/sc_player.cpp
@@ -11745,11 +11745,17 @@ void player_collected_data_t::collect_data( const player_t& p )
   {
     double metric=0;
 
+    role_e target_error_role = p.sim -> target_error_role;
+    // use player's role if sim didn't provide an override
+    if (target_error_role == ROLE_NONE) {
+      target_error_role = p.primary_role();
+    }
+
     // ROLE is used here primarily to stay in-line with the previous version of the code.
     // An ideal implementation is probably to rewrite this to allow specification of a scale_metric_e
     // to make it more flexible. That was beyond my capability/available time and it would also likely be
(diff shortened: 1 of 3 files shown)


seriallos:
Two core objectives:

1. User control over target error metric
2. Sensible defaults

I imagine 1. is pretty unobjectionable. Right now I'm not aware of any sane method to sim tanks using DPS at the metric for target error (all the workarounds I know about are incomplete/non-optimal (changing the role and position all have weird side effects)). This provides a reasonably simple and clear way to change that.

Sensible defaults is not as clear cut. As far as I'm aware, most folks simming tanks are doing DPS sims especially since survival simming is basically unmaintained right now. There may be a "seasonality" to this (early xpac is focused on survival, mid/late xpac is focused on DPS) but I don't have enough context to really know how true this is. I also know basically nothing about the state of healing - my understanding is that it's basically non-existent right now and the only APLs being worked on for healing specs are also DPS focused.

It really comes down to what we think the user expectation is - if someone sims a tank or healer profile, what do they expect? At the moment for me, it definitely feels like SimC is almost entirely DPS focused right now and, as far as I'm aware, there are no realistic dev efforts on getting survival/healing sims up and running.

Should probably circulate this around some more to determine the default setting. It's certainly easy enough to default `target_error_role` to ROLE_NONE and decide the metric based on `role`. That approach strikes me as more idealistically correct but doesn't reflect current usage.

seriallos:
More specifically for Raidbots, I'm likely going to be relying on `target_error` more heavily moving forward which is why I'm looking at it right now. Being able to specify the metric is the main thing I need for that since tank DPS sims are effectively broken using target_error right now. The default value in SimC comes down more to expected/intended usage for everyone.

scamille:
As you say, in practice right now DPS is probably good enough for all sims done. But that could change again, and if we can avoid it we shouldn't hinder healing defaults in the core.

My idea with ROLE_NONE was like this, in sc_player.cpp:
```
role_e role_for_target_error = p.primary_role();
if ( p.sim -> target_error_role != ROLE_NONE )
 role_for_target_error = p.sim -> target_error_role;
``` 

And additionally just set metric to DPS for ROLE_TANK, because that just makes more sense.

Or merge your pull request and I try to change target_error_role to target_error_metric?

seriallos:
Modified to go back to the previous behavior for defaults. If we want healers to default to a heal metric even though it's not fully implemented, we should likely do the same for tanks to keep things consistent.

Melekus:
Most people that sim tanking specs at the moment do it to optimize their dps output, not their survival. 

Accurate survival simulations in general also require custom boss scripts and custom Action Priority Lists to best play around those boss. That kind of user is in my opinion much more likely to tweak settings (such as the target_error_role) if needed than the kind of user who just paste their /simc input and hit Simulate. Thus 
 the default target error setting for tanks should be DPS.

As for healers, to my knowledge there is currently no active support for any healing simulation, so I don't see any harm in using DPS as the default metric until that changes.

Issue #4341, opened 2018-06-29
Simc WebAssembly


athei:
I am currently experimenting with creating a web assembly module from the simulationcraft codebase by compiling it with emscripten.  In the long run I want something that can be used to start a side like raidbots.com but without the need to have some costly server to run the simulations on. I also do not want the users to have some client to install and connect it to a side. Running simulationcraft as part of the side is the best of both worlds. What I basically want to create is an easy way to compile simulationcraft to an wasm file that can be included into a website and called into from javascript like this:
```
var simc = Simc()
simc.simulate("profile")
```

I managed to come up with some proof of concept where I have a simc running inside the browser and doing successful simulations. There is still a lot of work to do. I am interested in mainstreaming it.

**Would that be something you are interested in merging even though that would mean to merge some larger patches that touch multiple and restructure the build system?**

The largest part that need to change the build system  that needs to be more flexible and needs to allow for enabling and disabling features. This is necessary because not all parts of simc are useful for the wasm target and we need a way to easily switch them on and off. Right now I removed all dependencies on phtreads because the threading support in the browser is highly unstable and only possible with Firefox Nightly.

I think the easiest thing would be to apply the needed changes only to the cmake system and make it to a first class citizen.
Therefore I would start with merging some patches where I add some missing functionality to the cmake scripts and and add some more `#defines` that are switchable through cmake options. For example:
* **SC_MULTITHREADING**: Threading support is not mature in browsers. Should be disabled for the wasm target until browsers get better.

Other things can be removed too for wasm because we need fewer things and want to shrink the size of the binary since it must be transferred via network on page load.

I also would also remove the file globbing from the cmake script and replace it with explicitly defined *.cpp files to compile. In the process I might change the directory structure a bit.  The reason is that I need to select different files based on selected options and wether I compile for wasm or for your computer. For example I implemented a `mock_concurrency.cpp `that does basically nothing and removed the `sc_profileset.cpp` files from the build since they use `std::thread `directly without the wrapper from `concurrency.hpp`.

There are some other issues that make this hard. For example I am not sure from which branch I should start my wasm branch. There is the default `legion-dev` branch and the `bfa-dev` feature branch. Problem is that non-bfa code changes (like improvements to cmake scripts) are merged to the feature branch instead of to the main branch.

**Will `bfa-dev`and `legion-dev` be merged into a new master on bfa release and should I wait for that before I start a pull request? And for right now should I branch off `bfa-dev` to have fewer merge conflicts later on?**

[...]

PR #4347, opened 2018-07-02
cmake improvements (integrate into synchronize.py)

athei:
Like discussed in #4341 I made some improvements to the cmake build system and integrated it into the `synchronize.py` architecture. This pull request only touches cmake files and the python thing.

This is the preparation pull request for the one that adds an option to cmake to disable threading.

The normal build file still builds. This is no wonder as I did not touched any files of them. However, the qmake build does not build for me:
`make: *** No rule to make target profiles/Tier22, needed by SimulationCraft.app/Contents/Resources/profiles/Tier22.  Stop.`

But this is also the case for `bfa-dev` and was not broken by my changes.


scamille:
Looks good. I haven't tested it, but that can be done after the merge ;)

athei:
Glad to hear. We are waiting for nav to look into it? I am just asking because I have follow up patches in the pipeline :)

scamille:
add_compile_options(-fcolor-diagnostics) does not work for me. (Ubuntu 18.04 on Windows, using G++8.1.0):

c++: error: unrecognized command line option ‘-fcolor-diagnostics’

athei:
> c++: error: unrecognized command line option ‘-fcolor-diagnostics’

Ooops. Thought it was the same command line option for clang and gcc. Postet a fix.

PR #4351, opened 2018-07-05
Add option to disable features

athei:
As a preparation for the wasm target I added two options to the cmake:
```
# disable various features that may be anvailable or unneeded
option(SC_NO_THREADING "Disable all dependencies on pthreads" OFF)
option(SC_NO_NETWORKING "Disable all networking related stuff." OFF)
```

which are disabled by default. The good old Makefile is untouched by this as it does not set this flag and builds with all features like it always did.


navv1234:
Instead of peppering all this profileset-related preprocessing code all around the core, wouldn't it make more sense to disable profilesets in the profileset files if no threading is done? I havent looked at how much that'd require, but it should not be that much.

That way all the profileset calls could be there, but they'd essentially just no-op.

athei:
Yes, that might be a better solution. Depends on how large the API surface is. I will look into it.

My prefered solution would be to make profilesets use the `concurrency.hpp `wrappers like the rest of the threading code. These are already mocked when `SC_NO_THREADING` is set. However, I have no idea how much work that would be. The blocker there might be that it depends in inheriting the thread class which seems to be a legacy design and was the reason not use it for profilesets.

navv1234:
So from my perspective, the "concurrency" interface/wrapper we had was relevant when Simulationcraft was still in the C++03 era. We had no STL-defined threading API that was cross-platform, so there had to be a simple wrapper to do this in a platform-independent way.

As we moved to C++11, the need for such a wrapper (imo) went away. Profilesets are a new construct, and when I wrote the code, I saw no need to go back to the concurrency wrapper.

However, if there's now an actual need to disable threading from the simulator, it might make sense to convert Profilesets to use the concurrency stuff. My only issue with that is as you said, I don't personally like the way we do threading there (through inheritance), but it probably would not be a show-stopper.

athei:
Most people would agree with you that the inheritance design is not how you would do that today. It just couples it to tightly. 

Looking at this we should not port code to the old design. I look into mocking profilesets as I did for `concurrency.cpp` as you initially proposed.

Maybe further down the road we can think about a refactoring of the old threading code and a thinner wrapper around std `threading`, which then obsoletes the profilesets mock.

But basically all of this is a temporary solution until browsers reactivate support for multi threading when the spectre situation has cooled down. simcraft without multithreading does not make sense in the long run.

scamille:
Well with C++11 and std::thread also came the options to offer a thread interface in concurrency.hpp similar to that of std::thread - using some template magic with variadic templates and perfect forwarding.

So profileset could use the concurrency.hpp wrapper without any code changes and athei gets to disable threading in a single place.

scamille:
Oh btw: have you actually checked that std::thread implementations in wasm do multi-threading at all/are problematic?

We no longer do any threading stuff ourself (like creating pthread/windows threads), just delegate to std::thread, std::mutex and so on. And since those are implemented somewhat platform-specific,maybe you don't even need to disable anything?

athei:
Under the hood the `std::thread` does use pthreads and that it is what emscripten implements in terms of APIs that are available in the browser.

A `std::thread` and therefore a pthread maps to webworkers inside the browser. The shared memory of pthreads  is implemented using `SharedArrayBuffer`. But this last bit is disabled in all browsers because it opens up a security hole together with spectre. 

I succesfully used simc multithreading with an development build of Firefox. It works. It is just disabled for now in release builds. And even if it is enabled it is still very new and experimental.

athei:
I updated the commit. Now I do not touch all files to deactivate the profileset. I only change `sc_profileset.{hpp,cpp}` just like I removed the other functionality.

athei:
Do you like that way better? Changes necessary before merging that?

scamille:
Looks good. Profileset is a bit invasive, navv should probably comment if its ok or too burdensome for maintenance/extension.

navv1234:
Yeah better this way.

PR #4364, opened 2018-07-18
T21_Feral profile as requested by Bloodmallet

Tomslack:


Bloodmallet:
Thank you. I'll update the generator files accordingly.

PR #4366, opened 2018-07-19
Correcting a talent on the T21 Feral profile

Tomslack:
Swapped Incarnation for JW. Note to self: Talents have changed location.

### DIFF
--- a/profiles/Tier21/T21_Druid_Feral.simc
+++ b/profiles/Tier21/T21_Druid_Feral.simc
@@ -5,7 +5,7 @@ race=night_elf
 timeofday=day
 role=attack
 position=back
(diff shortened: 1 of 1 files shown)


N1gh7h4wk:
The important part to change is not the profile itself, but rather the generator file, since that would generate the old profile again [https://github.com/simulationcraft/simc/tree/bfa-dev/profiles/generators/Tier21](url)

scamille:
@aggixx  Can you manually make this adjustment?

Or if it is not sensible, close it.

seanpeters86:
@Tomslack can you make the changes made above and provide data on your change being an increase?

PR #4369, opened 2018-07-20
[Druid] Fix Brutal Slash not awarding combo points

ketilovre:
Effect no. 1 for Brutal Slash is the AP mod, so it wasn't awarding any combo points when used.


Collisionc:
Checked spelldata, yep, that fixes it.

PR #4370, opened 2018-07-20
[GUI] Improve render engine selection for Linux

tigew:
Since depending on your flavor of Linux you may or may not have webengine, do an additional check for webengine specifically.


scamille:
Thanks.

While you are at it, feel free to give cmake GUI build as well.

Works on my windows for Linux subsystem Ubuntu 18.04 bash, but probably hasn't been widely used yet.

PR #4376, opened 2018-07-22
Change some variable names to more accurately reflect their usage, and keep them consistent with the names blizzard uses internally

xeropresence:
During the BFA beta blizzards internal names used for DB2 files were leaked and can be seen here: https://github.com/Arctium/ClientDB-Structures/

I've spent a good amount of time reverse engineering the wow binary to get accurate spelltooltips. Comparing my work against simcraft was a big help but was made difficult by the sometimes obtuse names used for variables.

This PR focuses on changing two variables that I think are the best candidates for being renamed.

The first is the "_m_avg" variable. 
Blizzard internally refers to this as "Coefficient" and throughout the code that is how the variable is used.  By changing the name from _m_avg its use is made clear that it is not the average its self but is used in conjunction with the different forms of scaling to get the actual average.

The second is "item_effect"  which is referred to as DamageReplaceStat in both the SpellScaling.txt gametable and the RandPropPoint.db2 

I've had these changes made locally for a while now and keeping the naming of the variables consistent was a big help when tabbing between the reversed code and simcraft.

### DIFF
--- a/dbc_extract3/dbc/generator.py
+++ b/dbc_extract3/dbc/generator.py
@@ -3977,7 +3977,7 @@ class RandomPropertyPointsGenerator(DataGenerator):
         for id in ids + [ 0 ]:
             rpp = self._randproppoints_db[id]
 
-            fields = rpp.field('id', 'item_effect')
+            fields = rpp.field('id', 'damage_replace_stat')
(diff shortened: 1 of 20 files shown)


navv1234:
If the conflicts get fixed we could merge this. I don't really care about the naming, at all.

xeropresence:
I've rebased my changes onto the latest commits

scamille:
I think you somehow need to update your pull request?

xeropresence:
New commit didn't get pushed, re did it and should be good now

PR #4382, opened 2018-07-24
Updated Hunter APL

dj-riff:
Updated Hunter APL

Now includes use cases for Aspect of the Eagle based on target.distance >= 6. Results in a massive increase in DPS when boss is constantly out of range and a minor gain in Heavy Movement sims.


engine/class_modules/sc_hunter.cpp
```diff
@@ -1,6031 +1,6034 @@
...
-// Shared attacks
-//==============================
-
-// Barrage ==================================================================
-
-struct barrage_t: public hunter_spell_t
-{
-  struct barrage_damage_t: public hunter_ranged_attack_t
-  {
-    barrage_damage_t( const std::string& n, hunte
```
nuoHep: You don't need `buff.aspect_of_the_eagle.up` checks as the actions themselves check that in `ready()`.

Duh, line-endings broke github line resolution :/

nuoHep (changes requested):
Fix-up line-endings please (we use LF only).

nuoHep (approved):

PR #4388, opened 2018-07-29
Remove unnecessary null pointer checks

### LINKED ISSUE #4385: Remove unnecessary null pointer checks
[An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") like the following.
- [reforge_plot_t::analyze_stats](https://github.com/simulationcraft/simc/blob/78b35c0bdcef84a2050dd6ef8d210f9be8ecab52/engine/sim/sc_reforge_plot.cpp#L217 "reforge_plot_t::analyze_stats function")
- [SC_QueueListView::initListView](https://github.com/simulationcraft/simc/blob/78b35c0bdcef84a2050dd6ef8d210f9be8ecab52/qt/simulationcraftqt.hpp#L357 "SC_QueueListView::initListView function")
- [SC_TabBar::initContextMenu](https://github.com/simulationcraft/simc/blob/78b35c0bdcef84a2050dd6ef8d210f9be8ecab52/qt/util/sc_tabbar.cpp#L51 "SC_TabBar::initContextMenu function")

cameronjphillips:
Resolves #4385

### DIFF
--- a/engine/sim/sc_reforge_plot.cpp
+++ b/engine/sim/sc_reforge_plot.cpp
@@ -214,11 +214,8 @@ void reforge_plot_t::analyze_stats()
       }
     }
 
-    if ( current_reforge_sim )
-    {
-      delete current_reforge_sim;
-      current_reforge_sim = nullptr;
-    }
+    delete current_reforge_sim;
+    current_reforge_sim = nullptr;
   }
 }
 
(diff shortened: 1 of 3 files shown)


scamille:
Thanks.

PR #4397, opened 2018-07-31
Implement bfa.feast_as_dps, copied from the legion version

OrcinusDrake:
Use legion.feast_as_dps, and copy paste with name changes to bfa. Affects bountiful_captains_feast


engine/player/sc_unique_gear_x7.cpp
```diff
@@ -74,11 +74,28 @@ void consumables::galley_banquet( special_effect_t& effect )
...
+      effect.trigger_spell_id = 259454;
+      break;
+    case STAT_INTELLECT:
+      effect.trigger_spell_id = 259455;
+      break;
+    default:
+      break;
+  }
+
+  if (effect.player -> role == ROLE_TANK && !effect.player->sim->bfa_opts.bountiful_feast_as_dps )
```
nuoHep: You could've simply added this check to `init_feast()`, so it works for both of them.

nuoHep, line 92: Why not do a generic `feast_as_dps` and stuff the check inside `init_feast()`?

nuoHep: Especially considering the options is already named this way...

nuoHep: I'm also not sure there is any point in it being expansion-scoped.

OrcinusDrake:
new version here: https://github.com/simulationcraft/simc/pull/4417 cause I don't understand how updating works

PR #4398, opened 2018-08-01
[Priest] Small RE addition to DV

seanpeters86:


lgkern:
Is that time frame enough? you could use DV at 10s before adds, adds spawn, last < 10 and you won't be able to DV them.

scamille:
Feel free to push directly to bfa-dev now :-)

PR #4412, opened 2018-08-04
Updated Blood DK APL

cyfirgit:
-DS only dumps RP if the next rune spender will overcap RP
-Prioritize DnD over HS at 3+ targets


Melekus:
Changes added and optimised in https://github.com/simulationcraft/simc/commit/baac220cf732900f398aa1e83c5b7690d6105b15

PR #4413, opened 2018-08-04
+2.5% to Pre-Raid Feral Sim (Thanks Robo)

RobosaurusFeral:
See https://roboferal.wordpress.com/2018/08/04/paging-expert-feral-druid-tc-for-consideration-3-august-2018
for all change discussions.

### DIFF
--- a/profiles/PreRaids/PR_Druid_Feral.simc
+++ b/profiles/PreRaids/PR_Druid_Feral.simc
@@ -6,6 +6,8 @@ timeofday=day
 role=attack
 position=back
 talents=3000212
+#ST profile iH8Shred, #Please also consider unban Robo from Dreamgrove.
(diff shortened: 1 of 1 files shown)


RobosaurusFeral:
how does this even work

RobosaurusFeral:
Ahhh might have to change some formatting accidentally added white space.

seanpeters86:
To change the preraid profile you actually need to edit the generator file here: https://github.com/simulationcraft/simc/blob/bfa-dev/profiles/generators/PreRaids/PR_Generate_Druid.simc

To edit the APL you would change that in the module itself: https://github.com/simulationcraft/simc/blob/bfa-dev/engine/class_modules/sc_druid.cpp#L7453

PR #4415, opened 2018-08-05
T22_Death_Knight_Blood.simc

squee666:
Requires gear summary to be adjusted and maybe other changes?


squee666 (commented):
removed an X(i  had left) and left gear summary blank  since needs filled

PR #4433, opened 2018-08-12
[Arcane] Remove Time Warp line for Exodar

ronoudgenoeg:


seanpeters86:
Can you give some more detail here? Looking at the commit it isn't obvious why you are making this change, and you also removed arcane torrent for some reason. @ronoudgenoeg

ronoudgenoeg:
@seanpeters86 Casting arcane torrent is a DPS loss at 120. The time warp line was removed because legendaries are removed from the mage module, and it only existed to support the Shard of Exodar legendary.

Issue #4434, opened 2018-08-13
Balance druid ST APL Starlord tweaks


alexwyler:
I was tweaking the ilvl 340 ST APL provided in the most recent raid report profile for balance druid (http://www.simulationcraft.org/reports/PR_Raid.html).  By playing around Starlord a bit more, I am able to consistently sim higher than the profile in that report.

Here's a copy of the new profile I was using to sim: https://www.dropbox.com/s/os0fp8dxtcfpdk3/balance.simc?dl=0

Here's a visual representations of the changes made: 
https://www.dropbox.com/s/g1xtqasgil553u4/Screenshot%202018-08-12%2018.49.03.png?dl=0

I'm not sure the process for verifying APL changes, but here are the results of sims I ran showing the consistent improvement:

Base profile:
```
alexwyler@Alexs-MacBook-Pro simcraft $ simc balance-base.simc | grep PR_Druid_Balance
  10391  100.0%  PR_Druid_Balance
Player: PR_Druid_Balance troll druid balance 120
alexwyler@Alexs-MacBook-Pro simcraft $ simc balance-base.simc | grep PR_Druid_Balance
  10391  100.0%  PR_Druid_Balance
Player: PR_Druid_Balance troll druid balance 120
alexwyler@Alexs-MacBook-Pro simcraft $ simc balance-base.simc | grep PR_Druid_Balance
  10392  100.0%  PR_Druid_Balance
Player: PR_Druid_Balance troll druid balance 120
alexwyler@Alexs-MacBook-Pro simcraft $ simc balance-base.simc | grep PR_Druid_Balance
  10391  100.0%  PR_Druid_Balance
Player: PR_Druid_Balance troll druid balance 120
```

New profile:
```
alexwyler@Alexs-MacBook-Pro simcraft $ simc balance.simc | grep PR_Druid_Balance
  10405  100.0%  PR_Druid_Balance
Player: PR_Druid_Balance troll druid balance 120
alexwyler@Alexs-MacBook-Pro simcraft $ simc balance.simc | grep PR_Druid_Balance
  10404  100.0%  PR_Druid_Balance
Player: PR_Druid_Balance troll druid balance 120
alexwyler@Alexs-MacBook-Pro simcraft $ simc balance.simc | grep PR_Druid_Balance
  10403  100.0%  PR_Druid_Balance
Player: PR_Druid_Balance troll druid balance 120
[...]

PR #4435, opened 2018-08-13
Balance APL - Pool AP before triggering Starlord

alexwyler:
...and spend AP immediately when starlod stacks are < 3

This is the strategy being used by boomkins with the set bonus at the end of legion.
See: https://github.com/simulationcraft/simc/issues/4434

It's a 1-2 basis point increase in DPS in single target at ilvl 340, which seems like it could be under the threshold to add complexity.


Jundarer:
Now irrelevant due to new APL.

PR #4445, opened 2018-08-15
[trinket] first pass at Leyshock's Grand Compilation

wfarr:
Does not currently work. Modeled after Harlan's implementation.

I spent some time testing this on Beta, and it appears that all
casts and DOT ticks are capable of proc'ing this, and the proc
rate is fairly high. It behaves similarly to Acrid Catalyst
Injector, if ACI lacked the "all" buff.

This implementation doesn't work quite yet, but I figured I'd throw
this stab up for review, as someone more familiar with the process of
implementing custom item behavior will surely have better ideas on
how to do this properly.

cc @Bloodmallet

### DIFF
--- a/engine/player/sc_unique_gear_x7.cpp
+++ b/engine/player/sc_unique_gear_x7.cpp
@@ -370,6 +370,8 @@ void items::snowpelt_mangler( special_effect_t& effect )
 
 void items::leyshocks_grand_compilation( special_effect_t& effect )
 {
+  effect.proc_flags_ = PF_NONE_SPELL | PF_NONE_HEAL | PF_MAGIC_SPELL | PF_MAGIC_HEAL | PF_PERIODIC;
+
   auto crit = create_buff<stat_buff_t>( effect.player, "precision_module", effect.player->find_spell( 281791 ), effect.item );
   auto haste = create_buff<stat_buff_t>( effect.player, "iteration_capacitor", effect.player->find_spell( 281792 ), effect.item );
   auto versatility = create_buff<stat_buff_t>( effect.player, "adaptive_circuit", effect.player->find_spell( 281793 ), effect.item );
--- a/engine/player/sc_unique_gear_x7.cpp
+++ b/engine/player/sc_unique_gear_x7.cpp
@@ -374,7 +374,7 @@ void items::leyshocks_grand_compilation( special_effect_t& effect )
 
   auto crit = create_buff<stat_buff_t>( effect.player, "precision_module", effect.player->find_spell( 281791 ), effect.item );
   auto haste = create_buff<stat_buff_t>( effect.player, "iteration_capacitor", effect.player->find_spell( 281792 ), effect.item );
-  auto versatility = create_buff<stat_buff_t>( effect.player, "adaptive_circuit", effect.player->find_spell( 281793 ), effect.item );
+  auto versatility = create_buff<stat_buff_t>( effect.player, "adaptive_circuit", effect.player->find_spell( 281795 ), effect.item );
   auto mastery = create_buff<stat_buff_t>( effect.player, "efficiency_widget", effect.player->find_spell( 281794 ), effect.item );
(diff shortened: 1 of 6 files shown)


navv1234:
Some comments

1. The individual buffs, except for the first one (crit) are not linked to anything in the spell data so our exporter cannot find them. It needs to be told to explicitly include them. you can add them in `dbc_extract3/dbc/generator.py` file to the `_spell_id_list` in `SpellDataGenerator`. Should be self explanatory, there are already entries for some of the 8.0 trinket stuff we have implemented. Once that's done, someone has to regenerate the export data.
2. The description of the spell says "based on the spell", what is the categorization logic here? It will need to go into the proc callback system, instead of the uniform random selection.
3. The driver of the spell (i.e., in game the spell that procs your stat buff) currently holds no proc flags, potentially because it's so special (see 2). Simc will need to be told what kinds of spells can proc it. You can do this by overriding the proc flags of the special effect, something like `effect.proc_flags_ = PF_NONE_SPELL | PF_NONE_HEAL | PF_MAGIC_SPELL | PF_MAGIC_HEAL | PF_PERIODIC`. That will make it so that simc will consider any kind of "spell" and periodic damage to proc it.

edit: "Damaging and healing" here may actually mean that one should not add `PF_NONE_SPELL` or `PF_NONE_HEAL` in to the list.

wfarr:
> The individual buffs, except for the first one (crit) are not linked to anything in the spell data so our exporter cannot find them. It needs to be told to explicitly include them. you can add them in dbc_extract3/dbc/generator.py file to the _spell_id_list in SpellDataGenerator. Should be self explanatory, there are already entries for some of the 8.0 trinket stuff we have implemented. Once that's done, someone has to regenerate the export data.

All right I added the 4 buff spells to the generator and re-generated the spell dump (small patch bump, and includes hotfixes).

> The description of the spell says "based on the spell", what is the categorization logic here? It will need to go into the proc callback system, instead of the uniform random selection.

I haven't been able to make rhyme nor reason for the procs on Live with a version of this trinket. It seems generally a bit bugged though, as it sometimes procs even without combat or spells being cast in-game.

In general in combat in-game, I'm seeing between 80-90% uptime for each of the buffs on my Affliction Warlock (does not account for stacks, just presence of any).

> The driver of the spell (i.e., in game the spell that procs your stat buff) currently holds no proc flags, potentially because it's so special (see 2). Simc will need to be told what kinds of spells can proc it. You can do this by overriding the proc flags of the special effect, something like effect.proc_flags_ = PF_NONE_SPELL | PF_NONE_HEAL | PF_MAGIC_SPELL | PF_MAGIC_HEAL | PF_PERIODIC. That will make it so that simc will consider any kind of "spell" and periodic damage to proc it.

Sounds good. I'll start with the above list and if it turns out to be incorrect or inaccurate (especially once the bugs in-game are fixed around this trinket), we can correct it later.

---

This appears to be in working order now, albeit with a guesstimated rng roll driven proc chance until we have more info.

engine/player/sc_unique_gear_x7.cpp
```diff
@@ -365,6 +366,44 @@ void items::snowpelt_mangler( special_effect_t& effect )
...
+    leyshocks_cb_t( const special_effect_t& effect, const std::vector<buff_t*>& b ) :
+      dbc_proc_callback_t( effect.item, effect), buffs( b )
+    { }
+
+    void execute( action_t*, action_state_t* ) override
+    {
+      // We don't know true proc rate or spell list just yet, so this is roughly educated guesstimate from in-game dummies
+      if ( rng().roll(0.15) ) {
+        // which stat procs is entirely random
+        unsigned idx = ::floor(rng().range(0, 3));
```
navv1234: `auto idx = static_cast<size_t>( rng().range( 0, as<double>( buffs.size() ) );`

seanpeters86:
Closing due to already implemented

PR #4478, opened 2018-08-24
[Arcane] Increase DPS on 2 targets

ronoudgenoeg:


seanpeters86:
@ronoudgenoeg can you provide some examples of what you tested for this APL change?

PR #4485, opened 2018-08-28
Profile update

squee666:
more damage orientated


squee666 (commented):
forgot shoulder tag

squee666 (commented):
fixed names and alents i was half asleep okay im done

PR #4507, opened 2018-09-02
[Priest-Holy] profile improvements

sprd-tovo:
* add low priority instant casts for movement
* delay AE spells for adds
* add missing racial abilities
* add and adjust target thresholds
* improve Holy Fire handling
* optimize gear


Juliacare:
As discussed on discord, APL improvements into the module itself and profile changes into the generator file and regenerate after.

PR #4517, opened 2018-09-07
Fix Shredding Fury double decrement

### LINKED ISSUE #4516: [Feral][Azerite] Shredding Fury buff is not deincremented correctly
### Describe the bug
In viewing the feral Sample Sequence Table when Shredding Fury is equipped (Azerite ID 209) one can see that when using shred with the buff up the buff is reduced from 4 to 2, then from 2 to 0.

This can also be seen in the buff details, where it only shows uptime at 4 stacks and 2 stacks, never 3 or 1.

### Expected behavior
The Shredding Fury buff should apply 4 damage increases for Shred, not only 2.

### To Reproduce
Simply view the Sample Sequence Table with Shredding Fury equiped.

### Additional information
Using version 801-02

https://pastebin.com/zRbMadnj

kyogoi:
closes: #4516

### DIFF
--- a/engine/class_modules/sc_druid.cpp
+++ b/engine/class_modules/sc_druid.cpp
@@ -3862,6 +3862,14 @@ struct shred_t : public cat_attack_t
       td( s -> target ) -> debuff.bloodletting -> trigger(); // Druid module debuff
   }
 
+  virtual void execute() override
+  {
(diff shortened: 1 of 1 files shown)


kyogoi:
Forgot to call base execute ``cat_attack_t::execute()``

PR #4550, opened 2018-09-20
[Hunter] update SV potions

Pewtro:
https://www.raidbots.com/simbot/report/81amJFFaD1AiCqpKF2Qtc5
https://www.raidbots.com/simbot/report/cBqTozRqe2NGz2fnnf5gAy

### DIFF
--- a/engine/class_modules/sc_hunter.cpp
+++ b/engine/class_modules/sc_hunter.cpp
@@ -5151,10 +5151,7 @@ void hunter_t::init_assessors()
 
 std::string hunter_t::default_potion() const
 {
-  std::string lvl120_potion =
-    specialization() == HUNTER_SURVIVAL ? "bursting_blood" : "battle_potion_of_agility";
-
-  return ( true_level >  110 ) ? lvl120_potion :
+  return ( true_level >  110 ) ? "battle_potion_of_agility" :
          ( true_level >= 100 ) ? "prolonged_power" :
          ( true_level >= 90  ) ? "draenic_agility" :
          ( true_level >= 85  ) ? "virmens_bite":
--- a/profiles/PreRaids/PR_Hunter_Survival.simc
+++ b/profiles/PreRaids/PR_Hunter_Survival.simc
@@ -8,7 +8,7 @@ position=back
 talents=3101022
 
 # Default consumables
-potion=bursting_blood
+potion=battle_potion_of_agility
 flask=currents
 food=bountiful_captains_feast
 augmentation=battle_scarred
--- a/profiles/Tier22/T22_Hunter_Survival.simc
+++ b/profiles/Tier22/T22_Hunter_Survival.simc
@@ -8,7 +8,7 @@ position=back
 talents=3101022
 
 # Default consumables
-potion=bursting_blood
+potion=battle_potion_of_agility
 flask=currents
 food=bountiful_captains_feast
 augmentation=battle_scarred


leeterry:
Was also unable to find a place to change the hunter pet. They have the current pet set as a cat which is not horrible but not the best either. I would change that to Nether Ray or Bat for there ability to remove buffs since it still has the Primal rage ability. I also did not see the bm hunter use multi shot at all which is important when there are multiple targets (obviously). Not only because it lets you shoot them all but it also makes your pet attacks hit multiple mobs at a time (that dps adds up). I am usually top dps in any raids/dungeons I have done on my bm hunter but I have never gone below 3rd on total dps for the raid/dung. This is because I use multishot when fighting multiple mobs and I have really good azerite powers on my gear. Icy Veins has a list of the bis azerite traits you can get and I use those whenever I get gear to find the best traits I can get with it and compare it to my current gear/traits before deciding if its worth swapping.

Azortharion:
Neither of those things are relevant for SimulationCraft, since it doesn't
use dispel, and Bloodlust is supplied to all sims regardless of Priaml Rage
status.

The BM Hunter module in SimCraft uses Multi-Shot just fine.

Den fre. 21. sep. 2018 kl. 10.13 skrev leeterry <notifications@github.com>:

leeterry:
I just looked again (http://simulationcraft.org/reports/T22_Raid.html) under damage results, for bm hunter on the raid sim it does not show any damage from multishot (just as it didn't before) and that is simply not possible if it is being used. Thats a problem because everyone who plays a raid knows the bosses often spawn adds which in turn makes the results misleading. Was able to run a sim myself and had to set it up as multiple mobs to get the multishot to fire off without switching off of patchwork and patchwork was suposed to be a mix of all of them from what I understood. In reality at least 1 of the bosses is more of a hectic add cleve fight. Zek'voz spwans adds often plus has allot of things that you need to move to avoid. MOTHER also has a decent amount of movement involved (unless you want to die) and a time where you will be fighting ads but not nearly as bad as the other when it comes to the ads. The simulationcraft.org site does not show that and seeing as bm hunters can attack while moving and deal decent aoe dps that makes them have higher DPS in the particular fights. Would be better to run diff sims (1 for each boss) then set the options on those sims to accurately reflect the type of fight for each boss and add it all up in the end to post the end results (you could even post the results of each sim for those interested). I also noticed you said nothing about good azerite powers comment and thats setting off alarms too. Do you not have those coded in properly or something. Update: After running a sim on my own toon it seems the dps is lower then what I get while I am playing the toon myself (but I always pickup the orbs of power from my azerite trait) And in the section that shows all the damage I did it does not show anything from tital surge or heed my call (2 azerite traits I use that do damage). Furthermore I see no buffs from the orbs that I get becasue of secrets of the deep. You are saying good azerite powers are not relevant but I disagree 100% on that. Lastly I use the Darkmoon deck trinket (Fathoms) and it's not showing any dps from that in the sim logs either.

Azortharion:
The chart on the front page is a single-target comparison. BM Hunters do not use Multi-Shot on a single target. 

The results are only misleading insofar as the person reading them decides to attribute importance to them. They are not meant to be a ranking of the best DPS specs. It is a ranking of the highest single-target DPS potential that we can currently squeeze out in the simulation tools.

Patchwerk doesn't mix all of them. Patchwerk is just a blank target dummy boss.

The SimulationCraft.org page is not meant to be a DPS ranking. Why would anyone use such a thing when you can simply look at the real-world statistics on WarcraftLogs?

You're commenting on a miniscule potion update commit that is part of a project with thousands upon thousands of contributions. If you have all of these concerns (many of which are misguided) then maybe you should make an Issue about it here: https://github.com/simulationcraft/simc/issues

But I can tell you now that most of the concerns you have raised are misguided (although well-intentioned as far as I can tell), and they have been brought up hundreds of times before by other people who are misguided and do not quite understand what SimulationCraft is, what it does, or what the front page sim attempts to do.

I am literally the author of the Icy-Veins Hunter Guides, I know exactly what they recommend since I am the one making those pages and the charts they present. They are coded in properly and the Icy-Veins chart literally uses SimulationCraft to generate it.

I very much doubt that you are beating your SimDPS unless you are simply simulating incorrectly. Given the lack of understanding you show in your posts this is not surprising at all. You are very confrontational about things that you do not seem to understand the basics of.

The buff from the orbs of the Secrets of the Deep is accounted for in the sim.

I feel like you're simming with an old Legion-based version of SimulationCraft and that is the source of all of these things. Tried using Raidbots, or just an updated version of SimulationCraft from here?
http://downloads.simulationcraft.org/?C=M;O=D

PR #4555, opened 2018-09-23
[Warrior] Fury no longer uses Arcane Torrent

kyogoi:


seanpeters86:
Can you provide the comparison sim that shows the increase?

PR #4564, opened 2018-09-30
Implement Waycrests Legacy Set (and heal portion of Music Box)

AlexanderKenny:
So that can implement set bonus damage component (waycrest_legacy)


AlexanderKenny:
It works but I strongly suspect this is not a good way of doing this if one of the power users can give it a look over @navv1234 @vituscze @aethys256  etc.

PR #4591, opened 2018-11-01
Ret APL Fix

kfoster5:
Divine Storm can only be used for AoE with 3+ targets.  The current default APL uses it over Templar's Verdict when there are 2 targets


seanpeters86:
@kfoster5 to change an APL that is generated you'll want to change the engine file here: https://github.com/simulationcraft/simc/blob/bfa-dev/engine/class_modules/paladin/sc_paladin_retribution.cpp#L890

Then after new profiles are generated (and all sims using the module) it will take your apl changes. Also when requesting APL modifications it would help to give us a before/after of Sim DPS difference and various scenarios you tested it with.

seanpeters86 (changes requested):
These changes should be made in the module, not the profile itself. See comment above.

PR #4605, opened 2018-11-29
[Priest] implement Thought Harvester for PTR

seanpeters86:
- Implement Thought Harvester in Mind Sear and Vampiric Touch
- Added TODO to fix `chain=1` behavior
- Added prelim APL for harvested thoughts


lgkern (approved):
Looks good besides not having the Mind Sear buff usable right now.

PR #4638, opened 2018-12-16
Update buff_t in order to take into account Bone Shield's ICD on gains

### LINKED ISSUE #4637: Marrowrend's bone_shield generation is affected by the ICD in spellData
### Describe the bug
Any death knight profile will find itself unable to generate bone shield charges from more than one marrowrend in a 2.5s window. This is visible on the sample runs in simcraft, where the APL goes as follows (visible here: https://www.raidbots.com/simbot/report/ivBHT2jbT4Uf5QWHMJ85uN/simc )
1. Blooddrinker
2. DRW
3. Marrowrend
4. Blood boil
5. Marrowrend
6. Marrowrend

### Expected behavior
The default APL should not cast marrowrend more than twice on the opener, and should gain an appropriate number of bone shield charges.

### To Reproduce
Run the default T22 profile and check the output report: `simc ../profiles/Tier22/T22_Death_Knight_Blood.simc html=report.html`

Under the APL section, the cast sequence should ideally not cast marrowrend more than twice in the first 12s of the fight.

### Additional information
* Simcraft versions affected: nightly, 8.1, and probably earlier

Do note that this bug may also affect other buffs working in a similar fashion, where the `internal_cooldown` only applies to charges lost, not gained.

srenauld:
A newcomer on Acherus spotted the default APL marrowrending 3 times during DRW in quick succession. Turns out the issue is that
the Bone Shield ICD is copied to the cooldown value on buff_t, which then prevents stacks from being granted.

As I could see no way to fix this without inadvertently breaking other things, and since simc does not seem
to be thoroughly covered by automatic tests on this regard, I opted to fix it in a slightly kludgey way -
if the stacks are positive and the buff is bone_shield, the cooldown check is considered void.

In the future, it may be worth separating cooldown and internal_cooldown to avoid such issues.

fixes #4637

### DIFF
--- a/engine/buff/sc_buff.cpp
+++ b/engine/buff/sc_buff.cpp
@@ -1168,8 +1168,10 @@ bool buff_t::trigger( int stacks, double value, double chance, timespan_t durati
   if ( _max_stack == 0 || chance == 0 )
     return false;
 
-  if ( cooldown->down() )
-    return false;
+  if ( cooldown->down() ) {
+    if (name_str != "bone_shield" || stacks <= 0)
+      return false;
+  }
 
   if ( player && player->is_sleeping() )
     return false;


Melekus:
Handling a buff's internal cooldown for one specific class in the core isn't the right solution, I'll make a fix in sc_death_knight.cpp

Melekus:
Fixed the issue by manually setting bone shield's buff cooldown to zero in https://github.com/simulationcraft/simc/commit/8213c552eb6946eeccb6c5e698c5ef82e9d44739

PR #4665, opened 2018-12-31
[Priest] update cleave APL and remove AOE APL

seanpeters86:
- Adjust precombat actions
- Remove AOE APL
- Adjust priority of Vampiric Touch with Thought Harvester on main
target
- Adjust Dark Void action to keep SWP uptime on DungeonSlice
- Simplify Mind Sear expressions
- Adjust TTD calculations for VT and SWP


lgkern (changes requested):
Overall the changes look good, however, they execute a lot of algebra in runtime that can be moved to the pre-combat section and calculated only once per simulation, instead of every GCD.

lgkern (approved):

Issue #4676, opened 2019-01-09
[Warlock] Havoc sets spells to new target permanently.


Azevara:
### Describe the bug
When cycling targets to apply Havoc to a secondary target, casting a spell that duplicates during Havoc will set the target of that spell to permanently be the second target.

This appears to be the case for the baseline Havoc copying behavior in the Destruction module, and will most likely occur for any spell that can be copied using Havoc.

### Expected behavior
Since the target is never cycled or set in the action in question, it should always hit the primary target.

### To Reproduce
You can reproduce the problem concisely using a two line APL for Destruction in a 2 target sim:
```
actions+=/havoc,cycle_targets=1,if=!(target=sim.target)
actions+=/incinerate
```

After the first Incinerate, all subsequent Incinerates will be targeting the second enemy.

### Additional information
* SimulationCraft Version (required)
SimulationCraft 810-02 for World of Warcraft 8.1.0 Live (wow build 28768)

vituscze:
Should be fixed now, closing.

PR #4677, opened 2019-01-10
[Warlock] Fix/Rewrite Destruction Havoc Behavior

Azevara:
A fix for Issue #4676 and related bugs discovered while investigating it. Havoc no longer triggers a second copy of the spell with target reassignment, but instead uses aoe and generates a two-target list using `available_targets()`.

Generation of Soul Shards is now done with ENERGIZE_PER_HIT behavior to properly account for this change. Affected spells are Immolate, Conflagrate, Incinerate (non-fnb), Soul Fire, and Shadowburn.

Additionally, the change to Havoc functionality corrects a major bug where Chaos Bolt would consume double the appropriate number of Soul Shards when a second bolt was fired due to Havoc. 

Since execution of spells are no longer done twice for Havoc situations, execute vs hit counting in reporting should now be accurate.

Unused Legion/WoD effect code was also removed from the Destruction module.


Azevara:
Going to fix some issues and create a new PR.

PR #4679, opened 2019-01-11
[Prot Warrior] Enable in simc, updated APL and consumables

tlexii:
Hi, I'm new to simc codebase so all I've done is basically enabled prot and updated the APL, primarily for Unstoppable Force and old food/flask/potions.

Happy for any advice about what is missing from prot that needs fixing, otherwise I will just investigate what I can when I can.


scamille:
Thanks for the work. Someone actually contributing something instead of just asking about prot warrior :-)

Some things you can do:
* Compare stats of your ingame prot warrior to what simc html report shows if you sim him.
* Check all the other things like ability damage and %, buffs, etc.. At least if they are plausible.

tlexii:
Made a spreadsheet comparing a few minutes on a target dummy vs 0.01% error simc

https://docs.google.com/spreadsheets/d/1nh9suiLUZzvYMEYy2U6lQjtqfWFaLtrBo0W-4pMrUKc/edit?usp=sharing

Can you suggest what sort of discrepancies are normal?
The variation in Thunder Clap damages seems large to me, are there any recommended strategies to determine the source?

Jundarer:
You have to make your sheet public if you want to share it. "Normal" discrepancies are 0 if everything is being accounted for. What you want to do is calculate for yourself what the damage should be, check if your calculation is correct in game against the correct training dummy (raid dummy in most cases) and then see if every modifier is being accounted for in simc.

Best way to check what it does in simc is by enabling logging and then going through the log. Target error in simc has no impact on the damage calculation.

tlexii:
Yep, failed to set sharing to global, it was my domain only, I've updated the link.

Melekus:
The best way to compare damage from abilities one on one is to run simc with a very simple APL (typically a single spell per simulation) without raid buffs and with minimal procs on your character. That way you will get the minimal amount of damage of each ability and can compare it with how much damage you deal with this ability on a dummy.

Something to keep in mind since warrior deals physical damage is to make sure that the simc and ingame dummies have the same level for the purpose of armor calculations.

Once every ability's minimal damage is correct, you can use log=1 or debug=1 (more verbose) in simc to make sure damage buffs interact properly with abilities, and that other spells' properties (cooldowns, resource usage, etc.) are working as expected.

tlexii:
[simc_debug_extract.txt](https://github.com/simulationcraft/simc/files/2751480/simc_debug_extract.txt)
Right, and having done that can you suggest how to locate the source of mult=1.15 and tgt_da_mul=1.15 on fluffy_pillow from the very first auto attack at the start of the sim?  (I'm guessing the mult is for the attack and is the product of all relevant multpliers of which there is only one that isnt 1.0... tgt_da_mul)

I dont see any mention in the debug log of a matching buff/debuff.

Melekus:
What I meant with log/debug is just to check that the damage is correctly modified after an effect is applied. I'm not sure what the mult corresponds to in the case of auto attacks. 
Debug is more verbose than log=1 and I would advise using it only if you find a discrepancy between simulated and ingame damage and want to investigate.

On the topic of auto attacks, simc will usually not apply the variance in damage due to weapon damage rolls, so you can only compare the mean you see ingame with the simc results. You can disable that by adding average_range=0 in your input if you want to thoroughly check the auto attack's damage range

tlexii:
I started simple, auto attacks, no armour, 1 damage weapon, very low level shield, simc is 15% higher damage than live. Thunder Clap is the same 15% too much damage, 
1.15 multiplier comes from somewhere but I'm not seeing it in the debug log and some time looking at the code yielded nothing. 
Cant see any point checking buffs if something as simple as this isn't right.

Melekus:
Could you link the test profile you're working with? 

I'm currently using a profile without talents or azerite with gear copied from the T22 Protection paladin default profile and the only thing I'm seeing in mult= for thunder clap is the versatility multiplier.
(I also made sure to disable every raid buff and debuff)

tlexii:
I started fresh and got past that problem, verified many damage amounts but Avatar is causing me problems so I'm learning about spell selection and effects implementation a.t.m.

PR #4689, opened 2019-01-16
Fix NoNetwork Compile

adamhl8:
I was trying to build SimC GUI on Windows (VS2017) using the NoNetwork config, but the build kept failing due to it trying to use curl (which I didn't have installed/set up).

After some digging it seems like `SC_NO_NETWORKING` was never added as a Preprocessing Definition on the NoNetwork config, so I added that in.

Additionally, there were a couple of functions that were being called that were only defined if `SC_NO_NETWORKING` was *not* added, so I added a check for that as well.


navv1234:
Hm, did i forget to update that file. Weird.

PR #4692, opened 2019-01-17
[Warrior] protection initial azerite trait Bastion of Might

tlexii:
Implements Azerite trait Bastion of Might... still learning so hopefully not a total hack job.
Happy for feedback/tips.

### DIFF
--- a/engine/class_modules/sc_warrior.cpp
+++ b/engine/class_modules/sc_warrior.cpp
@@ -3941,7 +3941,6 @@ struct avatar_t : public warrior_spell_t
     {
       p()->buff.bastion_of_might->trigger();
 
-      action_t* bastion_of_might_ip_active = p()->active.bastion_of_might_ip;
       bastion_of_might_ip_active->execute( );
     }
   }
--- a/engine/class_modules/sc_warrior.cpp
+++ b/engine/class_modules/sc_warrior.cpp
@@ -3941,7 +3941,7 @@ struct avatar_t : public warrior_spell_t
     {
       p()->buff.bastion_of_might->trigger();
 
-      bastion_of_might_ip_active->execute( );
+      p() -> active.bastion_of_might_ip -> execute( );
     }
   }
 
--- a/engine/class_modules/sc_warrior.cpp
+++ b/engine/class_modules/sc_warrior.cpp
@@ -247,7 +247,6 @@ public:
   {
     const spell_data_t* spell_reflection;
     const spell_data_t* recklessness;
-    const spell_data_t* bastion_of_might;
     const spell_data_t* bladestorm;
     const spell_data_t* bloodthirst;
     const spell_data_t* colossus_smash;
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_warrior.cpp
```diff
@@ -245,6 +247,7 @@ struct warrior_t : public player_t
   {
     const spell_data_t* spell_reflection;
     const spell_data_t* recklessness;
+    const spell_data_t* bastion_of_might;
```
Melekus:
```suggestion
```
You don't need to set spec.bastion_of_might for two reasons:
1) It doesn't belong in the spec. namespace because it's for spells associated with a spec's core components, as in spells that can be called with find_specialization_spell()
2) You're not using it anywhere

engine/class_modules/sc_warrior.cpp
```diff
@@ -3934,6 +3937,14 @@ struct avatar_t : public warrior_spell_t
     warrior_spell_t::execute();
 
     p()->buff.avatar->trigger();
+
+    if ( p()->azerite.bastion_of_might.enabled() )
+    {
+      p()->buff.bastion_of_might->trigger();
+
+      action_t* bastion_of_might_ip_active = p()->active.bastion_of_might_ip;
```
Melekus:
```suggestion
```

engine/class_modules/sc_warrior.cpp
```diff
@@ -3934,6 +3937,14 @@ struct avatar_t : public warrior_spell_t
...
+      bastion_of_might_ip_active->execute( );
```
Melekus:
```suggestion
      p() -> active.bastion_of_might_ip -> execute( );
```
You can just execute it right away without assigning it to a new bastion_of_might_ip variable

engine/class_modules/sc_warrior.cpp
```diff
@@ -4313,6 +4324,24 @@ struct ignore_pain_t : public warrior_spell_t
   }
 };
 
+// A free Ignore Pain from Bastion of Might when Avatar is cast
+struct ignore_pain_bom_t : public ignore_pain_t
```
Melekus, line 4326: Did you check how it interacts with spells that interact with normal Ignore Pains like bloodsport, vengeance, or never surrender? If it is different from a regular Ignore Pain, you will need to override execute too.

engine/class_modules/sc_warrior.cpp
```diff
@@ -4557,6 +4586,7 @@ void warrior_t::init_spells()
   mastery.unshackled_fury  = find_mastery_spell( WARRIOR_FURY );
 
   // Spec Passives
+  spec.bastion_of_might     = find_spell( 287378 );
```
Melekus:
```suggestion
```
see line 250

Melekus (commented):
There are a few things that can be improved but it's looking good 👍

engine/class_modules/sc_warrior.cpp
tlexii: No, not extensively, will try and find time

tlexii: interestingly, doesn't get affected by bloodsport and doesn't consume vengeance buff, I will update the spell

PR #4708, opened 2019-01-23
[Warrior] Protection added T22/T23 profiles, tweaks to APL

tlexii:
Gear for the profiles may not be the absolute best and is still aimed at DPS sims primarily.
T23 sim on raidbots: https://www.raidbots.com/simbot/report/wLSoYxEhnHCGxzLptvigr6/simc


Melekus:
The T22 gearset has 340 ilevel shoulders so there's probably a mistake here. (Even if it's intentional, they should be 345)

Melekus:
I just noticed something else about the T22 gearset, the shoulders in question are from M+ and thus not allowed at ilvl 390, so you'll have to put it back to 340 or replace it with an item from Uldir.

This is not the case for the T23 set which allows azerite gear from M+ that you can get from the residuum vendor.

The full rulesets can be found here: 
- T22: https://github.com/simulationcraft/simc/blob/bfa-dev/profiles/generators/Tier22/T22_Generate.simc
- T23: https://github.com/simulationcraft/simc/blob/bfa-dev/profiles/generators/Tier23/T23_Generate.simc

Melekus:
It's not like T22 matters much now though, so i'll be merging this PR anyway 👍

PR #4714, opened 2019-01-25
[Warlock] Affliction Fixes

Azevara:
Several fixes and clean up for the Warlock modules, particularly regarding Drain Life and Seed of Corruption. Changes are as follows:

~~Split Drain Life into each spec separately, to handle Affliction specific behavior (Seed of Corruption)~~ (Replaced in later commit with general handling that includes pet damage and procs).
~~Drain Life now affected by Mastery for Destruction.~~ This is not necessary.
Inevitable Demise now expires if Drain Life is recast mid channel.
Simplified Chaos Shards expression.
Removed Sephuz code.
Fixed edge cases regarding Seed of Corruption refreshing.

### DIFF
--- a/engine/class_modules/warlock/sc_warlock.cpp
+++ b/engine/class_modules/warlock/sc_warlock.cpp
@@ -144,7 +144,7 @@ namespace warlock
 
     if ( debuffs_shadowburn->check() )
     {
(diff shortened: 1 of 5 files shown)


Azevara:
The function for calculating the remaining number of ticks left (full + partial) for Deathbolt should now function correctly for all cases.

Azevara:
Nightfall was being consumed on cast and benefiting the casted Shadowbolt, instead of affecting the next.

PR #4722, opened 2019-01-31
[Warrior] Protection - implement contributed APL/T23

tlexii:
as supplied by Melekus


Melekus:
There should be a more updated version coming (hopefully) soon, i'll push it myself once I have it 👍

PR #4739, opened 2019-02-16
Fix wrong decimal point for European LC_NUMERIC

kwuerl:
There is an issue with the generation of doubles in json in the German (DE_DE) LC_NUMERIC locale.
Due to the indirect use of _vprint_  in the custom _sc_json_writer_t_ doubles are printed as 1,22 instead of 1.22 resulting in this (While generously ignoring setlocale in the main function):

`... {"y":6,00,"name":"10848 to 10892"},{"y":16,00,"name":"10892 to 10936"},{"y":22,00,"name":"10936 to 10980"},{"y":49,00,"name":"10980 to 11024"},{"y":73,00,"name":"11024 to 11068"},{"y":155,00,"name":"11068 to 11112"},{"y":220,00,"name":"11112 to 11156"},{"y":350,00,"name":"11156 to 11200"},{"y":558,00,"name":"11200 to 11244"},{"y":850,00,"name":"11244 to 11288"},{"y":1171,00,"name":"11288 to 11332"},{"y":1692,00,"name":"11332 to 11376"} ...`

Honestly i did not understand why doubles need to be limited to precision in machine-read json files so i removed the custom _sc_json_writer_t_ and defaulted to rapidjson's really good dtoa implementation.

If you find you need the fixed point doubles in the json output another possibillity ist to simply replace the comma (but i find this pretty ugly...):

**engine/report/sc_highchart.cpp:513ff**
```
for ( unsigned i = 0; i < buffer.size(); ++i )
  {
    if(buffer.data()[ i ] == ',')
         this->os_->Put( '.' );
    else
         this->os_->Put(` buffer.data()[ i ] );
  }
```

### DIFF
--- a/engine/report/sc_highchart.cpp
+++ b/engine/report/sc_highchart.cpp
@@ -146,7 +146,7 @@ std::string chart_t::to_target_div() const
 std::string chart_t::to_data() const
 {
   rapidjson::StringBuffer b;
-  sc_json_writer_t<rapidjson::StringBuffer> writer( b, sim_ );
+  rapidjson::Writer<rapidjson::StringBuffer> writer( b );
 
   js_.Accept( writer );
   std::string str_ = "{ \"target\": \"" + id_str_ + "\", \"data\": ";
@@ -159,7 +159,7 @@ std::string chart_t::to_data() const
 std::string chart_t::to_aggregate_string( bool on_click ) const
 {
   rapidjson::StringBuffer b;
-  sc_json_writer_t<rapidjson::StringBuffer> writer( b, sim_ );
+  rapidjson::Writer<rapidjson::StringBuffer> writer( b );
 
   js_.Accept( writer );
   std::string javascript = b.GetString();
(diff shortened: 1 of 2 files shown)


kwuerl:
This PR is connected to #4740. Both PRs fix the Problem on their own, but each of them a different cause.

navv1234:
At a guess, the precision override was there because we tend to use the "report precision" to control the decimals in output. Does not really matter for the json report though, as you mention, and the use is inconsistent in any case.

scamille:
The inconsistent application really confused me because I started looking at the json report, even though this only touches the highcharts :)
Html does not look in any way different in browser, the file is just slightly larger.

The original problem probably no longer applies, since it should have been inadvertently fixed by me upgrading to fmt 6.0 (https://github.com/fmtlib/fmt/releases/tag/6.0.0 Floating-point formatting is now locale-independent by default).

@navv1234 @seriallos how much do we care about html file size? The T24_Raid.simc I used changed from 21.6MB to 24.8MB.

scamille:
Closing this, since it should have been fixed by PR #4740 and fmt upgrade.

PR #4745, opened 2019-02-23
[Warlock] Add expression support for tracking incoming wild imps

Azevara:
Added two new expressions for tracking Wild Imp spawning events.

`incoming_imps` will return the number of events currently pending.

`time_to_imps.X.remains` returns the time until X imps will spawn. If X is greater than the number of incoming imps or if X is <= 0, it will return the time to the last imp waiting to be spawned. Additionally, `time_to_imps.all.remains` can be used in place of a number to achieve the same effect

### DIFF
--- a/engine/class_modules/warlock/sc_warlock.cpp
+++ b/engine/class_modules/warlock/sc_warlock.cpp
@@ -797,6 +797,56 @@ void warlock_t::create_options()
   add_option( opt_string( "default_pet", default_pet ) );
 }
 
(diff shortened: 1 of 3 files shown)


Hekili:
Is there a reason to build out `time_to_imps.X` to `time_to_imps.X.remains`?  Are there other potential properties under `time_to_imps.X` that may need to be supported later?

If not, why not just `time_to_imps.X` and `time_to_imps.all`?

PR #4779, opened 2019-03-18
[Priest] APL updates

Urcra:
There are two main additions to the APL. ShadowFiend delay and Shadow Word: Void banking.

The current APL just uses ShadowFiend on CD instead of when we are in Void Form, causing the ShadowFiend to deal slightly less damage than it could and causing us to waste some insanity. Instead using it later on in VF increases the damage of the ShadowFiend and makes us have longer Void Forms. Since the CD of ShadowFiend is so long delaying it will not result in a dps loss.

Shadow Word: Void banking. In the current APL SW:V is used on CD, which causes us to overflow some of the insanity generated by it(even more apparent with WOTD). We can hold on to the charges for some time to decrease the amount of insanity overflow we get without losing out on any casts.

The APL changes the Insanity overflow from SW:V from 9.1% to 5.6%, and the ShadowFiend overflow from 2.29% to 0.32%. And results in an overall increase of 1.3% extra dps

[Raidbots report](https://www.raidbots.com/simbot/report/qAWrTedHQnoRi55frEuyjB)

I tested these APL changes on my main character, and verified them on the T23 profile(with 3111111 talents) and a T23 profile where I changed the chest to use WOTD. And saw an increase for both profiles using the new APL

T23 Profile dps comparisons:
Capture
Capture2
Capture3

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -1951,13 +1951,18 @@ void priest_t::generate_apl_shadow()
                       "cooldown.shadow_word_death.remains<gcd.max)" );
   single->add_talent( this, "Surrender to Madness", "if=buff.voidform.stack>10+(10*buff.bloodlust.up)" );
   single->add_talent( this, "Dark Void", "if=raid_event.adds.in>10" );
-  single->add_talent( this, "Mindbender" );
+  single->add_talent( this, "Mindbender", "if=talent.mindbender.enabled|(buff.voidform.stack>18|target.time_to_die<15)" );
   single->add_talent( this, "Shadow Word: Death",
                       "if=!buff.voidform.up|"
                       "(cooldown.shadow_word_death.charges=2&"
                       "buff.voidform.stack<15)" );
   single->add_talent( this, "Shadow Crash", "if=raid_event.adds.in>5&raid_event.adds.duration<20" );
-  single->add_action( this, "Mind Blast", "if=variable.dots_up" );
+  // Bank the Shadow Word: Void charges for a bit to try and avoid overcapping on Insanity.
+  single->add_action( this, "Mind Blast",
+                      "if=variable.dots_up&"
+                      "(!talent.shadow_word_void.enabled|buff.voidform.down|"
+                      "buff.voidform.stack>14&(insanity<70|charges_fractional>1.33)|"
+                      "buff.voidform.stack<=14&(insanity<60|charges_fractional>1.33))" );
   single->add_talent( this, "Void Torrent",
                       "if=dot.shadow_word_pain.remains>4&"
                       "dot.vampiric_touch.remains>4&buff.voidform.up" );
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -1960,7 +1960,8 @@ void priest_t::generate_apl_shadow()
   // Bank the Shadow Word: Void charges for a bit to try and avoid overcapping on Insanity.
   single->add_action( this, "Mind Blast",
                       "if=variable.dots_up&"
-                      "(!talent.shadow_word_void.enabled|buff.voidform.down|"
+                      "((raid_event.movement.in>cast_time+0.5&raid_event.movement.in<4)|"
+                      "!talent.shadow_word_void.enabled|buff.voidform.down|"
                       "buff.voidform.stack>14&(insanity<70|charges_fractional>1.33)|"
                       "buff.voidform.stack<=14&(insanity<60|charges_fractional>1.33))" );
   single->add_talent( this, "Void Torrent",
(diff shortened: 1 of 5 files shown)


seanpeters86:
Did a few tests here:

### Base Composite Profile: 

#### Battle for Dazar'alor Composite
| Actor | DPS | Increase |
|---|:---:|:---:|
|Mindbender|25042|1.05%|
|Both|24981|0.80%|
|Base|24782|0.00%|
|Mind_Blast|24730|-0.21%|

#### Single Target
| Actor | DPS | Increase |
|---|:---:|:---:|
|Mindbender|23546|1.23%|
|Both|23493|1.00%|
|Base|23261|0.00%|
|Mind_Blast|23219|-0.18%|

There's some issues with the mind blast apl on movement situations, not comfortable committing that until it is tweaked for light movement and number of targets. This was without WotD using `talents=3111111`.


### If I add in one whispers: 

#### Battle for Dazar'alor Composite
| Actor | DPS | Increase |
|---|:---:|:---:|
|Both|25015|1.29%|
|Mindbender|24978|1.14%|
|Mind_Blast|24729|0.13%|
|Base|24697|0.00%|

#### Single Target
| Actor | DPS | Increase |
|---|:---:|:---:|
|Both|23925|1.58%|
|Mindbender|23856|1.29%|
|Mind_Blast|23605|0.22%|
|Base|23553|0.00%|

The increase here looks much better, need to add some sort of changes to accommodate NOT having whispers so we don't overvalue this trait.

### Leaving whispers on testing with FotM:

#### Battle for Dazar'alor Composite
| Actor | DPS | Increase |
|---|:---:|:---:|
|Mindbender|24881|1.08%|
|Both|24881|1.08%|
|Base|24614|0.00%|
|Mind_Blast|24602|-0.05%|

#### Single Target
| Actor | DPS | Increase |
|---|:---:|:---:|
|Mindbender|23872|1.23%|
|Both|23869|1.22%|
|Base|23582|0.00%|
|Mind_Blast|23568|-0.06%|

Looks like this change even with 1 whispers isn't worth for FotM, so we would want to exclude or change this.

LMK if you need help or have questions with any of these changes, thanks for contributing!

## Misc comments
If you aren't familiar with composite, it's just a weighted sim that varies [Patchwerk, Light Movement, Heavy Movement] X [1 Target, 2 Targets] X [No Adds, Small Adds, Big Add]. Weights for BoD can be found [here](https://github.com/WarcraftPriests/bfa-shadow-priest/blob/master/weights.py#L22-L41).

Another change I would love to see is to use `insanity_drain_stacks` or something similar to what we did back [in legion](https://github.com/simulationcraft/simc/blob/legion-dev/engine/class_modules/sc_priest.cpp#L5194), this is a smaller point but definitely helps us handle better varieties of situations with a single APL.

CC: @lgkern

seanpeters86 (changes requested):
see above ^

Urcra:
Yeah I will fix it so we don't end up overvaluing whispers, and see if I can optimize around movement. But I am not quite sure why the FoTM is simming lower since I did notice that this would only work for SW:V so I added in a check for it

"/mind_blast,if=variable.dots_up&**(!talent.shadow_word_void.enabled**|buff.voidform.down|buff.voidform.stack>14&(insanity<70|charges_fractional>1.33)|buff.voidform.stack<=14&(insanity<60|charges_fractional>1.33))"

Ill try to look into the drain stacks aswel

Urcra:
Changed the Mind blast to only bank the charges if there are no movement events upcoming. With these changes we no longer lose any dps with these changes compared to the old.

3111111 no whisper:

pc: https://www.raidbots.com/simbot/report/iT1zrvW7VuovU2DnR5auFJ
lm: https://www.raidbots.com/simbot/report/bSZDfvL6A47mZ3uAwDYy5X
hm: https://www.raidbots.com/simbot/report/c3viNG1DRfNRkjHPMSFj63


FOTM:

pc: https://www.raidbots.com/simbot/report/xtNBR1zjKv3MVBXHY9Fgdz
lm: https://www.raidbots.com/simbot/report/tpWccnx7LR6dYxWYLr5Rrf
hm: https://www.raidbots.com/simbot/report/vqJTuQ4spFmFofdzqpNtan


3111111 WOTD:

pc: https://www.raidbots.com/simbot/report/9pmstSRKG6bsM1jLVS9r3c
lm: https://www.raidbots.com/simbot/report/hQ3qWh81mpbRc9adE4LVAy
hm: https://www.raidbots.com/simbot/report/xfrXrrwEdfMN39xXsnRwb

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1951,13 +1951,18 @@ void priest_t::generate_apl_shadow()
...
-  single->add_talent( this, "Mindbender" );
+  single->add_talent( this, "Mindbender", "if=talent.mindbender.enabled|(buff.voidform.stack>18|target.time_to_die<15)" );
   single->add_talent( this, "Shadow Word: Death",
                       "if=!buff.voidform.up|"
                       "(cooldown.shadow_word_death.charges=2&"
                       "buff.voidform.stack<15)" );
   single->add_talent( this, "Shadow Crash", "if=raid_event.adds.in>5&raid_event.adds.duration<20" );
-  single->add_action( this, "Mind Blast", "if=variable.dots_up" );
+  // Bank the Shadow Word: Void charges for a bit to try and avoid overcapping on Insanity.
+  single->add_action( this, "Mind Blast",
```
seanpeters86, line 1961: Can you add the same `movement.in` block to the module? This needs to be here so we generate the proper APL.

seanpeters86 (changes requested):

seanpeters86 (approved):
Looks great!

# Battle for Dazar'alor Composite
| Actor | DPS | Increase |
|---|:---:|:---:|
|Both|25058|1.48%|
|Mindbender|24973|1.13%|
|Mind_Blast|24775|0.33%|
|Base|24694|0.00%|

# Single Target
| Actor | DPS | Increase |
|---|:---:|:---:|
|Both|23969|1.78%|
|Mindbender|23848|1.27%|
|Mind_Blast|23657|0.46%|
|Base|23549|0.00%|

PR #4785, opened 2019-03-28
changed to some mythic gear/azerite

squee666:
because default only uses raid gear

also bonestorm enabled because its dungeon splice should be

### DIFF
--- a/profiles/generators/DungeonSlice/DS_Generate_Death_Knight.simc
+++ b/profiles/generators/DungeonSlice/DS_Generate_Death_Knight.simc
@@ -4,23 +4,23 @@ race=undead
 spec=blood
 role=tank
 position=front
-talents=2220022
+talents=2220023
 
-head=glaciercrest_helm,id=165823,bonus_id=4824/1537,azerite_powers=348/561/31/13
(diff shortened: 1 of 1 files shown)


Melekus:
Could you provide a raidbots/simc report comparing the current profile and your proposed improvements?

On top of that, engineering and other BoP crafted gear are not authorized in default profiles, make sure you read the rules in https://github.com/simulationcraft/simc/blob/bfa-dev/profiles/generators/DungeonSlice/DS_Generate.simc

squee666 (commented):
changed helmet. donig sims

also rules dont say no crafted azerite which is normal considering the helmet was allowed before in uldir

squee666:
https://www.raidbots.com/simbot/report/7h5YkzeUZXrTjurm9KipH1

squee666:
fixed sim i hope https://www.raidbots.com/simbot/report/j9HtFicfsXsDJhFTSZxDLx

Melekus:
Looks like the profile "T23_Blood_Death_Knight updated" in your link isn't the same as the one in your PR. At the very least the helm is different.

Also no, crafted engineering helmets were not allowed in Uldir either

Melekus:
The author is inactive, i'll close the pull request and see if there's anything to salvage from the changes

PR #4796, opened 2019-05-03
[Shaman] Restoration DPS implementation

niseko:
- Updates some older code and implements a crude resto APL. All of the spells and azerite already work seamlessly through the Elemental implementations with only minor changes.
- Removes resto spells that haven't existed for quite a few expansions at this point.
- Adds and generates the T23 profile


engine/class_modules/sc_shaman.cpp
```diff
@@ -644,7 +643,10 @@ struct shaman_t : public player_t
     // Azerite Effects
     lightning_conduit = nullptr;
 
-    regen_type = REGEN_DISABLED;
+    if ( specialization() == SHAMAN_ELEMENTAL || specialization() == SHAMAN_ENHANCEMENT )
+      regen_type = REGEN_DISABLED;
+    else
+      regen_type = REGEN_STATIC;
```
Bloodmallet: After some discussion in SimCMinMax, there should be an option for `REGEN_DYNAMIC`. Try that out. Static has fixed resource ticks. Which is no longer the case ingame. Dynamic checks for the current new resource level everytime the actor could do something.

engine/class_modules/sc_shaman.cpp
```diff
@@ -5303,6 +5291,13 @@ struct spiritwalkers_grace_t : public shaman_spell_t
                       options_str )
   {
     may_miss = may_crit = harmful = callbacks = false;
+    // Reduces the cooldown of Spiritwalker's Grace by ${$m1/-1000} sec
+    if ( p()->talent.graceful_spirit->ok() )
+    {
+      timespan_t reduction = timespan_t::from_seconds( p()->talent.graceful_spirit->effectN( 1 ).base_value() / -1000 );
```
Bloodmallet: `timespan_t reduction = p()->talent.graceful_spirit->effectN( 1 ).time_value();`

engine/class_modules/sc_shaman.cpp
```diff
@@ -5303,6 +5291,13 @@ struct spiritwalkers_grace_t : public shaman_spell_t
...
+      cooldown->duration -= reduction;
```
Bloodmallet: Or make this a one liner:
`cooldown->duration += p()->talent.graceful_spirit->effectN( 1 ).time_value();`

"+" there because the time value is negative.

engine/class_modules/sc_shaman.cpp
```diff
@@ -5303,6 +5291,13 @@ struct spiritwalkers_grace_t : public shaman_spell_t
...
+      p()->buff.spiritwalkers_grace->cooldown->duration -= reduction;
```
Bloodmallet: Why does the buff have a cooldown here? Only the action should have the CD.

engine/class_modules/sc_shaman.cpp
```diff
@@ -5673,6 +5668,10 @@ struct flame_shock_t : public shaman_spell_t
     {
       proc_chance = p()->azerite.igneous_potential.spell_ref().effectN( 3 ).percent();
     }
+    if ( p()->spec.restoration_shaman->ok() )
+    {
+      proc_chance += p()->spec.restoration_shaman->effectN( 8 ).percent();
+    }
```
Bloodmallet, line 5667: While this results in the correct value, I'd add the actual `Igneous Potential` effect 4 for restoration, too. 

```cpp
    proc_chance = p()->spec.lava_surge->effectN( 1 ).percent();
    
    if ( p()->spec.restoration_shaman->ok() )
    {
      proc_chance += p()->spec.restoration_shaman->effectN( 8 ).percent();
    }

    if ( p()->azerite.igneous_potential.ok() )
    {
      if ( p()->spec.restoration_shaman->ok() ) {
        proc_chance = p()->azerite.igneous_potential.spell_ref().effectN( 4 ).percent();
      } else if ( p()->spec.elemental_shaman->ok() {
        proc_chance = p()->azerite.igneous_potential.spell_ref().effectN( 3 ).percent();
      }
    }
```

engine/class_modules/sc_shaman.cpp
```diff
@@ -7024,6 +7022,12 @@ void shaman_t::init_base_stats()
   if ( specialization() == SHAMAN_ELEMENTAL || specialization() == SHAMAN_ENHANCEMENT )
     resources.base[ RESOURCE_MAELSTROM ] = 100;
 
+  if ( specialization() == SHAMAN_RESTORATION )
+  {
+    resources.base[ RESOURCE_MANA ]                  = 20000;
+    resources.initial_multiplier[ RESOURCE_MANA ]    = 5;
```
Bloodmallet: `spec.restorarion_shaman->effect(5)` has the multiplier. I don't know where one can find the base mana value.

Bloodmallet (changes requested):
I added comments to some code lines. Would be great if you could respond or update the PR.

Overall: good.

engine/class_modules/sc_shaman.cpp
niseko: i've removed the cooldown from the buff itself now which should be cleaner

engine/class_modules/sc_shaman.cpp
niseko: could just divide the base it has by default (100k, not sure where it gets that from) by that multiplier as well; just need the base mana to be 20k since the spells calculate their cost off of that. that way it would at least scale

Bloodmallet (approved):

PR #4896, opened 2019-07-23
Catch for some on uses with razor coral

Tomslack:
Font of Power and Cyclo as most common
Added font to precombat for feral as well
Will need additional reviews for more trinkets going forwards


Tomslack:
fixed some stuff because i'm an idiot

PR #4901, opened 2019-07-28
Brewmaster bugfixes & APL tweaks

emallson:
This PR has two fixes:

1. RJW no longer sometimes stops doing damage (depending on your exact haste value). The underlying issue is the chaining of `tick_t` events in `sc_buff.cpp`. Under certain conditions, the last tick can fire (and set the `tick_event` pointer to null) before the buff actually expires. When this happens, refreshing the buff didn't set up a new `tick_t` event. 95% of the fix is doing that (in `sc_monk.cpp` via a new public member function `schedule_tick()` because `tick_t` is local to the anonymous namespace in `sc_buff.cpp`). The remaining 5% is accomodating RJW's unique behavior regarding tick hasting: tick period is determined on cast, and is *not* dynamically updated with haste levels (but is updated on refresh).
2. Expel Harm now (approximately) deals damage. Technically, WoW heals you with each orb (which can crit) and then deals damage equal to 10% of the amount of healing done (including overhealing and absorbed healing). My implementation just calculates the EV of that given the number of orbs you have. This is slightly higher variance since you cut the number of events by 3-5x depending on APL, but has the same overall EV and gives a good estimate of damage done.

It also updates the APL slightly to cast expel harm and concentrated flame (effectively our only active DPS major). At some point, I'd like to get it updated to cast whatever essence you have equipped but that is a project for another day.


vituscze:
We're currently discussing changing the default buff tick behavior to handle some of these corner cases. If you want to fix the other stuff unrelated to ticking buffs (expel harm, APL) before we change the core buff code, you might want to split the PR.

emallson:
@vituscze how soon is the buff stuff changing? the RJW bug has been screwing up people's top gear / stat weight sims for a while now and even if the fix only goes in temporarily it'd be nice to not have to constantly address that question

navv1234:
See 2155010aaf8b056f0037714fb4b5bd3de9bf1f80

emallson:
Since it appears the new buff code is in (?) I just went ahead and updated my RJW fix to use that instead of splitting it into something separate entirely. If things are still up in the air, I can cut it out and hold off on the RJW fix til later.

navv1234:
As far as I am concerned, things are not up in the air with the buff tick stuff (and partial ticking) support with regards to the feature.

PR #4926, opened 2019-08-10
[Unholy DK]Fix vision of perfection and unholy frenzy APL + Optimization

nyterage:
added a modifier into the APL to look for vision of perfection and cast unholy frenzy on cooldown rather than waiting until Apocalypse has <2 seconds left with the major or minor active. 

before this change
https://www.raidbots.com/simbot/report/huPuGMYenjhJarFqFHGqgW/simc

after this change
https://www.raidbots.com/simbot/report/rSTqUaSCFozJyKXsm1chQB/simc

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -7841,6 +7841,7 @@ void death_knight_t::default_apl_unholy()
   cooldowns -> add_action( this, "Apocalypse", "if=debuff.festering_wound.stack>=4" );
   cooldowns -> add_action( this, "Dark Transformation", "if=!raid_event.adds.exists|raid_event.adds.in>15" );
   cooldowns -> add_talent( this, "Summon Gargoyle", "if=runic_power.deficit<14" );
+  cooldowns -> add_talent( this, "Unholy Frenzy", "if=essence.vision_of_perfection.enabled" );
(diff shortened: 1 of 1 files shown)


seanpeters86 (approved):
Howdy, not a DK maintainer, but this looks good to me. The only thing I’d ask is why are there so many different entries for Unholy Frenzy? You could condense these lines to a single entry with an | (OR) in between each condition around parenthesis. 

I’ll approve for now though and let a DK merge.

nyterage:
so currently, Unholy Frenzy has 4 different scenarios it has to look for;
Vision of Perfection
Magus of the Dead Enabled
Magus of the Dead Disabled
AoE
due to the different scenarios im not sure it can be combined into a single line, i will see what i can do though. 

Edit: Looks like it works combining all but AoE into a single line, committed the change, will need to be approved.

nyterage:
added some further APL optimization for Condensed Life Force, 

Before, 
https://www.raidbots.com/simbot/report/aQarUTwXGMmB8JcysXyzxc

After,
https://www.raidbots.com/simbot/report/i4qGg8JHXT4wK45pmm1e5a

Bicepspump:
This all looks good to me.

nyterage:
added yet another optimization to the APL for ashvanes razor coral, 
before
https://www.raidbots.com/simbot/report/gY7V25e64at4UkFrqwxLDv
after
https://www.raidbots.com/simbot/report/xdkcADrzt1wA7QgfoYoUX2

PR #4927, opened 2019-08-12
[DK Unholy] Razor Coral APL Interaction

nyterage:
Change made allows razor coral to work properly without Unholy Frenzy or Condensed life force selected, was not interacting properly before. 
Before
https://www.raidbots.com/simbot/report/6f4tctf3dTigq2dvYcqunJ
After
https://www.raidbots.com/simbot/report/bRdvzGytaTwYT84mjtxcqq

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -7810,7 +7810,7 @@ void death_knight_t::default_apl_unholy()
   def -> add_action( "berserking,if=buff.unholy_frenzy.up|pet.gargoyle.active|!talent.summon_gargoyle.enabled" );
   def -> add_action( "use_items,if=time>20|!equipped.ramping_amplitude_gigavolt_engine|!equipped.vision_of_demise", "Custom trinkets usage" );
   def -> add_action( "use_item,name=ashvanes_razor_coral,if=debuff.razor_coral_debuff.stack<1" );
-  def -> add_action( "use_item,name=ashvanes_razor_coral,if=(pet.guardian_of_azeroth.active&buff.unholy_frenzy.up)|(cooldown.apocalypse.remains<2&!essence.condensed_lifeforce.enabled&!talent.unholy_frenzy.enabled)|(target.1.time_to_die<cooldown.apocalypse.remains)|(cooldown.apocalypse.remains<2&target.1.time_to_die<cooldown.condensed_lifeforce.remains)" );
+  def -> add_action( "use_item,name=ashvanes_razor_coral,if=(pet.guardian_of_azeroth.active&buff.unholy_frenzy.up)|(cooldown.apocalypse.remains<2&!essence.condensed_lifeforce.enabled&!talent.unholy_frenzy.enabled)|(target.1.time_to_die<cooldown.apocalypse.remains)|(cooldown.apocalypse.remains<2&target.1.time_to_die<cooldown.condensed_lifeforce.remains+20)" );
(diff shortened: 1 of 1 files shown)


Bicepspump:
Would you mind explaining what you've done here exactly? Also, it's generally not a good idea to use army cooldown timer to set time conditions. Simply use time>120 for instance.

Chewsdaydk:
I would triple check how this interacts with vision of perfection and sims that do not have Magus of the Dead - from my memory from changes i made back in January i think this line will fail when Magus isn't equipped or with Vision of Perfection Minor as it's checking for both CLF and Frenzy and Unholy Frenzy is only held under certain conditions. I would advise changing it so that Unholy Frenzy is held for Guardian if that is so

nyterage:
okay so, what ive done here ill go through step by step, the first part, 

"(pet.guardian_of_azeroth&buff.unholy_frenzy.up)"
this checks for condensed life force and unholy frenzy together, something that will happen during the 3rd Apocalypse cooldown of any given simulation, i added buff.unholy_frenzy.up to the line due to a later line "guardian_of_azeroth,if=cooldown.apocalypse.remains<6" with this line, the pet guardian of azeroth can/will be summoned a few seconds before apocalypse for haste stacking prior to ghouls/magus coming out (this was about a 200dps gain to condensed life force vs the previous line that was "guardian_of_azeroth,if=cooldown.apocalypse.ready"). 

next up is 
"(buff.unholy_frenzy.up&!essence.condensed_lifeforce.enabled)"
this will look for unholy frenzy being up, but only if you're not using condensed life force as a essence, allowing it to operate properly with unholy frenzy selected as a talent, while not using condensed life force, as in that situation using any other major essence its no different to save coral vs using it for every apocalypse. could easily be changed to "(cooldown.apocalypse.remains<2&!essence.condensed_lifeforce.enabled)" with a small loss to dps in simulations without condensed life force, but with unholy frenzy enabled. 

next we have 
"(buff.unholy_frenzy.up&cooldown.army_of_the_dead.remains<360)"
this one was added specifically for simulation lengths of less than 6 and a half minutes witch are fairly standard, this allows razor coral to be used on the 4th or later cast of apocalypse. due to it being a condition only being met after 4 minutes with razor coral and, condensed life force with the unholy frenzy talent selected. 

and finally the fix for when you have neither unholy frenzy or condensed life force selected, 
"(cooldown.apocalypse.remains<2&!talent.unholy_frenzy.enabled)"
this just looks for apocalypse cooldown and uses razor coral stacks about 2 seconds prior to it coming off of cooldown for situations where unholy frenzy is not selected, and condensed life force is not selected. as all of the other parts apply specifically to when the DK being simmed is using UF, CLF, or both in conjunction, i felt this line was necessary for people who aren't following the "BiS" setup to not be hindered in their Razor coral simulations.

Bicepspump:
I'm concerned that unholy frenzy won't always be up for condensed life-force. Might be a good idea to add a CLF conditional to UF saving for Apoc.

Using Coral with unholy frenzy doesn't sound ideal. Isn't it better to attach it to apocalypse?

Third condition should have a (cooldown.condensed_life_force.remains + 20<target.time_to_die) instead of the army condition. What we are really looking for is CLF usage before fight finish.

Think you could bunch up last and second condition to both follow apoc.

nyterage:
ooh i didnt think of time to die as a condition, thats much better than the solution i had thought of. yes ill modify UF usage to save for CLF when using both in conjunction as well to ensure that it doesnt break things. 

As for using coral with apoc, in my testing it was about a 50 dps difference between the two in advantage to using it with UF rather than apoc, so i went that route.

PR #4931, opened 2019-08-15
[DK Unholy] Razor Coral APL Interaction

nyterage:
Modified Razor Coral's APL in order for it to work properly without Unholy Frenzy, Condensed Lifeforce or Magus of the Dead, as well as optimize slightly.


Melekus:
The line for Ashvane's Razor Coral doesn't seem to cover all the cases, what if you have Unholy Frenzy talented but aren't using Condensed Life Force's major? It will only trigger from the time to die conditions.

nyterage:
ahh good point, i entirely missed that variable. should be fixed now. sorry bout that melekus!

PR #4955, opened 2019-09-19
Fix expression buff.potion to properly match only potions

mbattersby:
This is a fix for issue 4952.

Modifies how buff.potion is calculated so that it doesn't match nil_item_data nor things that don't trigger the shared potion cooldown, like augment runes.

As an aside, it's possible that potion_filter_t in engine/dbc/sc_item_data.cpp potion_filter_t is no longer required.

### DIFF
--- a/engine/buff/sc_buff.cpp
+++ b/engine/buff/sc_buff.cpp
@@ -1927,7 +1927,9 @@ struct potion_spell_filter
 
   bool operator()( const item_data_t* item ) const
   {
-    return range::contains_value(item->id_spell, spell_id);
+    // Augment runes and other things look like potions. Only match things
+    // that trigger the potion shared cooldown.
+    return item->cooldown_group[0] == 4 && range::contains_value(item->id_spell, spell_id);
   }
 };
 }  // namespace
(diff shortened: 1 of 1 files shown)


navv1234:
I need to find some time to look at why the filtering by subclass does not work. The fact that the accessor does not work isn't really the issue here, but rather has something changed in the client data so that we can no longer accurately match on potions.

EDIT: This can probably be merged though, you should probably hardcode the cooldown group as an enum somewhere, see `ITEM_TRINKET_BURST_CATEGORY` for example in sc_enum.hpp

mbattersby:
I don't know how this ever worked though, because item lookups (via find(), find_gem() and find_consumable()) return nil_item_data when not found, not 0.  If ( item ) is then by definition always true.

mbattersby:
I'll re-submit with and enum for ITEM_COOLDOWN_GROUP_POTION.

I looked through the DBC item data and there's nothing but 0 for none and 4 for potions.

navv1234:
Broke during the optimization pass it looks like to me, a long long time ago.

PR #4960, opened 2019-09-23
[Unholy DK] Beserking APL update

nyterage:
this one is a bit odd, i believe prior to this change it was utilizing beserking a few seconds before unholy frenzy, even though it was specified to utilize it with unholy frenzy.

i feel the `!talent.summon_gargoyle.enabled` portion was overriding the unholy frenzy bit for some reason. 

with specifics already in place for both unholy frenzy, and gargoyle i felt changing it to specify army of the damned would make more sense, guess it results in 200 extra dps too. 

heres a sim with the change
https://www.raidbots.com/simbot/report/sYXk8EFBVqcXsPjfTXH9n7


Melekus:
Wouldn't it make more sense to sync Berserking with Apocalypse if neither Gargoyle nor Unholy Frenzy are talented? Rather than using it on cooldown.

nyterage:
yeah, it does make more sense. honestly didn't think of that in my confusion over the odd behavior it was exhibiting. 

https://www.raidbots.com/simbot/report/8b5Sydjd3PPBUaXJuwkzPE

ill update to reflect that change.

PR #4964, opened 2019-10-04
Prot warrior food & minor essence change

Mwahi:
Updated old feast with new food and changed minor Aegis to C&S


Melekus:
This won't work to change the default food, you should update sc_warrior.cpp instead

PR #4965, opened 2019-10-04
Actually cache the results of timespan_t::{zero,min,max}.

Mazorlion:
Simc creates the timespans in timespan_t::{zero,min,max} repeatedly.

`constexpr` on functions is only honored when used in a `constexpr`; many callers of the above functions do not have that property.  This effect is more noticeable in debug mode because the operators on timespan_t have assertions in them that call these functions. In debug mode this change was a ~5-10% CPU savings, in NDEBUG it was closer to ~1-1.5%. 

[Approximate Profile Data, simming a DH with defaults.](https://docs.google.com/spreadsheets/d/e/2PACX-1vRS09q4NirSxYKB_kcf1Tfs0PizbWnxBjZ9AEIxmzo7EcFN1359ir022qt5ylnRV1-Bv3el85TCsuDx/pubhtml)

### DIFF
--- a/engine/sc_timespan.hpp
+++ b/engine/sc_timespan.hpp
@@ -250,18 +250,22 @@ namespace timespan_adl_barrier
       return timespan_t(t);
     }
 
-    static constexpr timespan_t zero()
+    static timespan_t zero()
     {
-      return timespan_t();
+      static constexpr timespan_t cached_zero = timespan_t();
+      return cached_zero;
     }
-    static constexpr timespan_t max()
+    static timespan_t max()
     {
-      return timespan_t(std::numeric_limits<time_t>::max());
+      static constexpr timespan_t cached_max = timespan_t( std::numeric_limits<time_t>::max() );
+      return cached_max;
     }
-    static constexpr timespan_t min()
+    static timespan_t min()
     {
-      return std::is_floating_point<time_t>::value ? timespan_t(-std::numeric_limits<time_t>::max()) :
-          timespan_t(std::numeric_limits<time_t>::min());
+      static constexpr timespan_t cached_min = std::is_floating_point<time_t>::value
+                                                   ? timespan_t( -std::numeric_limits<time_t>::max() )
+                                                   : timespan_t( std::numeric_limits<time_t>::min() );
+      return cached_min;
     }
   };
 


scamille:
Hi that does look nice.

Can you explain to me why this does make a difference? I know that simc uses timespan_t 99% of the time in a non-constexpr context. The keyword was mostly tucked on it since I could, and maybe help in those rare places where we do some constexpr-enabled calculations by hand.

One would assume there is something to gain from sharing those commonly used values in a static function-local variable, but then we also store most of our timespan value based anyway. Or does the saving come from the frequent comparisons against 0 of non-constexpr timespan objects?

Mazorlion:
> Hi that does look nice.
> 
> Can you explain to me why this does make a difference? I know that simc uses timespan_t 99% of the time in a non-constexpr context. The keyword was mostly tucked on it since I could, and maybe help in those rare places where we do some constexpr-enabled calculations by hand.
> One would assume there is something to gain from sharing those commonly used values in a static function-local variable, but then we also store most of our timespan value based anyway. Or does the saving come from the frequent comparisons against 0 of non-constexpr timespan objects?

Previously when timespan_t::zero was called in a runtime context it would execute the content of the function each time. The result is that every non-constexpr call to zero(), basically all of them, actually constructed a new timespan_t object and this was essentially doubling the expense of calling zero(). This was particularly bad in debug builds (where assertions are executed) because timespan_t::operator+=, timespan_t::operator-=, timespan_t::operator+, timespan_t::operator- all have assertions that call these functions multiple times per invocation:
```
    timespan_t& operator+=(timespan_t right)
    {
      assert(right < zero() || *this < zero() || *this <= max() - right);
      assert(right > zero() || *this >= min() - right);
      time += right.time;
      return *this;
    }
    timespan_t& operator-=(timespan_t right)
    {
      assert(right < zero() || *this >= right + min());
      assert(right > zero() || *this <= right + max());
      *this = *this - right;
      return *this;
    }
```
To clarify this from the profile: See CPU usage below from the previous version (numbers are % of total CPU). Notice how timespan_t::zero costs 13.78 but only 6.24 of that come from the function itself, the rest comes from the constructor (timespan_t::timespan_t) being called in the function.

Previous version:

| Function                                     | Function + Children | Function Only |
|----------------------------------------------|---------------------|---------------|
| timespan_adl_barrier::timespan_t::timespan_t | 7.53                | 7.53          |
| timespan_adl_barrier::timespan_t::zero       | 13.78               | 6.24          |


Optimized version:

| Function                                     | Function + Children | Function Only |
|----------------------------------------------|---------------------|---------------|
| timespan_adl_barrier::timespan_t::timespan_t | 0.03               | 0.03          |
| timespan_adl_barrier::timespan_t::zero       | 8.16             | 8.16          |

After this change timespan_t::zero costs ~roughly the same (total CPU is slightly lower so it's a slightly higher fraction, also there is going to be some variance) but we're calling the no-arg constructor a tiny fraction of the number of times we were calling it before. The same effect can be seen for min/max.

Also, thanks for the quick review! I should hopefully have additional improvements soon.

scamille:
Ok so even though the min/max/zero functions and also the ctor was constexpr, since they get called in a non-constexpr context nearly everywhere, the object construction overhead was so massive that this made a difference. Thanks.

One last question:
Why did you change the function signature to non-constexpr?

Mazorlion:
See https://en.cppreference.com/w/cpp/language/constexpr

A constexpr function must satisfy the following requirements:
* the function body must not contain:
  * a definition of a variable of static or thread storage duration

PR #4971, opened 2019-10-27
[Monk] Fix lucid dreams

Juko8:
Should only be affecting energy


Juko8:
@Hinalover

Hinalover:
Considering Lucid Dream works with mana, this only fixes part of the issue. The issue before was that Lucid Dream was refunding Chi when it is not supposed to.

PR #4978, opened 2019-11-26
Update T24_Hunter_Beast_Mastery.simc

Nakauri:
Lucid Dreams is pointless but even more-so to overlap with Aspect of the Wild.


Nakauri:
This isn't to be merged which is why it's on the profile and not the engine. There are additional apl changes since the last hunter merge (beyond this). It's just one line and I didn't want to have a convo with anyone to get azerite traits looked at for further refining 👍

Pewtro:
Irrelevant now as this was T24 and only on the profile

PR #4993, opened 2020-01-05
Add moonkin form to druid restoration apl

dengelgit:
Currently, the Druid Restoration apl uses Starsurge and Lunar Strike. 
However, these spells can only be used if the player uses the Moonkin form before.
Therefore we have to add the Moonkin form as an action.


Jundarer:
This is handled by the precombat apl.

PR #5043, opened 2020-02-07
Add T25 generator for Ret Paladins

Prejudice182:


Melekus:
Feel free to add more corruption effects to the profile, the maximum allowed for T25 ruleset is 39 net corruption (your current value is -40).

Full ruleset: https://github.com/simulationcraft/simc/blob/bfa-dev/profiles/generators/Tier25/T25_Generate.simc

You can find a list of corruption effects' bonus IDs at https://www.raidbots.com/developers

PR #5046, opened 2020-02-11
[Paladin]Avenging Wrath may be activated if another Avenging Wrath is…

LordSkeletor:
… active (such as from Vision of Perfection)

-Also removed Crusade check
-Will verify behavior Tuesday morning to make sure it extends duration


Melekus:
Implemented along with the special crusade behavior in https://github.com/simulationcraft/simc/commit/934390cabc6372ca9997db0130e52f5803a0834d

PR #5064, opened 2020-02-28
[Unholy DK] Minor APL Fixes

nyterage:
Fix some uncommon use cases with Vision of Perfection and 0 Magus of the dead traits

Wait to use Blood of the Enemy until after DnD


Bloodmallet:
This change might've reduced the power of `Festermight` (Azerite trait). Was this intended?

See this auto-data generation as an example. https://github.com/Bloodmallet/bloodmallet.github.io/commit/59609a1f9829a363665234d129aba9625c237de5#diff-32f14ef6d75f0644acee5fe28fec9cd3L283

Chewsdaydk:
I had a look at this a couple of times, I couldn't spot anything obvious and it seemingly had no impact on individuals actual character simulations. The fact it seemingly reduced its power in your sims makes me a bit uncomfortable but not really sure what's going on.

While looking for what this might be I discovered that the simulation gets stuck when no traits are taken, there's a PR from @nyterage to fix this. Perhaps that will solve the above issue. If not then I would probably undo the Festering Strike/Scourge Strike lines and try and solve the original problem another way.

PR #5068, opened 2020-03-01
[Havoc] Optimize BotE/Fel Barrage usage

kyogoi:
Implements better synergy between Fel Barrage and BotE, as well as optimizing Fel Barrage usage within meta windows.

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -5171,7 +5171,7 @@ void demon_hunter_t::apl_havoc()
   apl_normal->add_action( this, "Throw Glaive", "if=talent.demon_blades.enabled" );
 
   action_priority_list_t* apl_demonic = get_action_priority_list( "demonic" );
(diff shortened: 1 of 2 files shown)


EvanMichaels:
@wisnoi thanks for the PR. However this seemed to go through a lot of iteration. You may need to regenerate the profiles again because I don't think the output matches the code. I'm happy to review otherwise. Do you have Raidbots comparison for the difference?

kyogoi:
@EvanMichaels Sorry! I've been a bit busy these past few days, what with a global pandemic going on :p  
I don't really know how to generate the profiles o: I was under the impression it was done manually, and by looking at previous commits I thought I was doing it right ><  
  
As for comparisons:  
My changes: https://www.raidbots.com/simbot/report/mwhDrM6GQwQdjTDRCCo5MP
Current: https://www.raidbots.com/simbot/report/iPwuC93SqCD9hWJ9wjDG8u  
  
It went through a lot of iterations because I initially only set out to improve performance when both BotE and FB were used together, but did not anticipate that the changes I made would also serve the purpose of buffing sim output for both independently.

EvanMichaels:
I had some additional sync logic that was a major gain for AoE sims. I have included some of your changes that I verified in https://github.com/simulationcraft/simc/commit/a0f230c34609be29ff35cd7f291c9a83c745600a and this should resolve all the remaining issues. Thanks!

PR #5070, opened 2020-03-04
Update sc_shaman.cpp

Lavindar:
Removed code for the old version of the Ancestral Resonance azerite trait.


Bloodmallet:
What would be the benefit?

Bloodmallet:
Stale. Closing

PR #5074, opened 2020-03-23
[DK] More APL Fixes

nyterage:
Fix another random issue with Vision of Perfection and 0 magus of the dead traits.


nyterage:
Reverted previous changes that caused the weirdness with Festermight on bloodmallet, and fixed by holding unholy frenzy for apocalypse no matter if you have magus or not in ST with vision of perfection.

PR #5096, opened 2020-05-02
Split out real_ppm.cpp

nuoHep:
Move all of the inlines of real_ppm_t out of `simulationcraft.hpp` into their
own source file. They should've not been inline in the first place (maybe apart
from the constructor), as they are too big for that (imo).

Fixes link errors in `sc_buff.cpp` with clang+lld as they are bit more picky
regarding odr violations (and it effectively is, as it [sc_buff.cpp] includes
`real_ppm.hpp` without including `simulationcraft.hpp` and so does not see the
inlines).

----
TODO: add `real_ppm.cpp` to all of the projects' source lists. I don't really know how the diff project files are managed (apart from the ones touched here) so I'd appreciate fixups on top.


navv1234:
There's a python3 script that can handle it in source_files. You add your new file(s) to the QT_engine.pri and run the sync script in the directory, it will automatically update all versions.

navv1234:
(Serge just now pushed a fix for that, presumably)

nuoHep:
> There's a python3 script that can handle it in source_files. You add your new file(s) to the QT_engine.pri and run the sync script in the directory, it will automatically update all versions.

Thanks. Will try to remember next time I touch smth :)

PR #5104, opened 2020-05-11
Touch up range:: stuff a bit

nuoHep:
Clean up some stuff.
Add range versions of binary search related routines.
Add projection-enabled `range::find()` overload.

Mostly to tickle ci (vs).

### DIFF
--- a/engine/class_modules/sc_hunter.cpp
+++ b/engine/class_modules/sc_hunter.cpp
@@ -137,7 +137,7 @@ struct player_data_t
 
   action_data_t* get( const action_t* a )
   {
-    auto it = range::find_if( data_, [ a ] ( const record_t& r ) { return a -> name_str == r.first; } );
+    auto it = range::find( data_, a -> name_str, &record_t::first );
     if ( it != data_.cend() )
       return it -> second.get();
 
@@ -646,7 +646,7 @@ public:
   template <typename T, typename... Ts>
   T* get_background_action( const std::string& n, Ts&&... args )
   {
-    auto it = range::find_if( background_actions, [ &n ]( action_t* a ) { return a -> name_str == n; } );
+    auto it = range::find( background_actions, n, &action_t::name_str );
     if ( it != background_actions.cend() )
       return dynamic_cast<T*>( *it );
 
--- a/engine/dbc/azerite.cpp
+++ b/engine/dbc/azerite.cpp
@@ -1,4 +1,3 @@
-#include <algorithm>
 #include <array>
 
 #include "config.hpp"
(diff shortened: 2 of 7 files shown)


scamille:
Is there any benefit in having those projection-enabled overloads, compared to just forcing the caller to provide a lambda?

nuoHep:
I'm not sure what kind of a lambda do you have in mind exactly. But generally speaking projections are strictly more powerful than a lambda, as a lambda *can* be a projection.

Projections are anything that is `INVOKE`able (`invokable` concept in C++20, anything you can `std::invoke` in C++17) with a single argument. It can be a pointer to a member, a lambda, a regular function. The only difference with a projection over a lambda, implementation-wise, is a requirement to `invoke` it, instead of simply calling the lambda.

I can only speak from experience and this is all subjective. We've been using a range library internally for around half a year now. Projections were one of the things that was kinda slow to adopt at first, but is now considered (in the team) one of the most powerful things in ranges (for some algorithms*). They make the code much more concise, clear and reviewable. Granted, this is mostly because lambdas in C++ are pretty verbose, but alas.

[*] I've never seen a projection used for `find_if` for example (and I'd push back during review to remove it tbh). But for things like `contains`, `sort`, `find`, binary search stuff they are used constantly and even encouraged over writing custom ad-hoc comparators (or worse, slapping an `operator<` on the type).

nuoHep:
I guess code speaks better than my mostly incoherent prose, so I'll just drop some examples:
```cpp
struct foo {
  // [...]
  int bar();
};
int bar = 42;
```
```cpp
std::vector<foo> foos;
ranges::contains(foos, bar, &foo::bar); // pointer to member projection
ranges::contains(foos, bar, [](auto&& foo){ return foo.bar(); }); // lambda projection
ranges::any_of(foos, [bar]{auto&& foo){ return foo.bar() == bar; }; // lambda comparator
```
```cpp
// the nice thing is that pointer to member projections are
// "polymorphic" over values/pointers (including smart pointers)
std::vector<std::unique_ptr<foo>> foos;
ranges::contains(foos, bar, &foo::bar); // pointer to member projection
ranges::contains(foos, bar, [](auto&& foo){ return foo->bar(); }); // lambda projection
// notice how ptr to member stayed the same while the lambda is slightly different
```
---
Pointers to members are somewhat uncommon, I'll give you that. At first code like this does seem kinda foreign to people unfamiliar with ranges/projections or pointers to members. So I'm not even going to try pushing for their usage, especially in the class modules.

`lower_bound`/`equal_range` changes were contained to "core" stuff so I thought it ok to convert. But if you don't like I'm fine reverting/"de-projectionifying"/whatever.

scamille:
Thanks for helping me understand the benefits.

I have nothing against using them at all, I was just curious about the rationale/motivation for adding the extra invoke code.

Adoption of this in SimC will likely be slow, but it is always good to have various tools available.

I think the whole range:: helpers fit the SimC codebase nicely anyway, and with our slow adoption of newer C++ standards, it is great to have  the range::invoke as well.

PR #5106, opened 2020-05-12
Add SC_DBC_GET_DATA() helper

nuoHep:
Returns a 'dynamic' extent span to a corresponding data array depending on the
ptr flag. Hides all of the boilerplate associated with it:
* supression of -Wunused for the ptr flag in non-ptr builds
* possible error in a ternary with mismatching span extents

Aside from being more concise it also helps preventing bugs like in 76450d67


navv1234:
Looks good and significantly less error prone than what we have now.

PR #5110, opened 2020-05-15
PVS Studio support and some fixes

krahen94:
+ Added PVS comment headers so PVS Studio analyses the files
* Fixed multiple reported critical issues - extra asserts and null checks added and fixed a memory leak
* Applied the included clang_format

Basically just added support for more analysis, since the new PVS license can be used with open source with no cost. Since I don't really have domain knowledge in the code, only fixed the most obvious reported critical issues.
https://www.viva64.com/en/b/0457/

### DIFF
--- a/engine/action/absorb.cpp
+++ b/engine/action/absorb.cpp
@@ -1,3 +1,5 @@
+// This is an open source non-commercial project. Dear PVS-Studio, please check it.
+// PVS-Studio Static Code Analyzer for C, C++, C#, and Java: http://www.viva64.com
 // ==========================================================================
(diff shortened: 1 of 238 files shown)


scamille:
Thanks for the effort, but there are a couple reasons why we don't want to accept this pull request:
- Adding the PVS comments to the top of every file is a absolute no-go. Please design your software/solution so that it can be run without being this intrusive.
- Your commits mix a lot of formatting with actual changes. This way it is nearly impossible to pick out the the actual changes in between all the noise of reformatting and adding comments.

PR #5113, opened 2020-05-17
Refactor SC_HIGH_PRECISION_STOPWATCH handling

nuoHep:
All of the stopwatch types can be represented 1-to-1 with clockid_t clocks with
recent enough (really, quite old) kernel + libc. Simplify stuff by using
clock_gettime() for everything.

Change CLOCK_REALTIME to CLOCK_MONOTONIC to align it with the Windows
implementation (which uses QPC). CLOCK_REALTIME is bad for stuff like this
anyway.

While at it make it unconditional on Linux. While the comment said we need to
link with -lrt, it got changed with glibc 2.17. That is old enough to be relied
upon. We require C++14 and all of the compilers supporting that were released
after that anyway. Distro wise both Debian 8 (Jessie) and Ubuntu 14.04 have
glibc 2.19.

### DIFF
--- a/engine/config.hpp
+++ b/engine/config.hpp
@@ -32,6 +32,10 @@
 #  define SC_SIGACTION
 #endif
 
+#if defined(__linux) || defined(__linux__) || defined(linux)
+#  define SC_LINUX
+#endif
+
 // ==========================================================================
 // Compiler Definitions
 // ==========================================================================
--- a/engine/util/stopwatch.cpp
+++ b/engine/util/stopwatch.cpp
@@ -22,13 +22,6 @@
 #include <sys/time.h>
 #endif
 
-// If you turn this on, you will need to add -lrt to LINK_LIBS in Makefile
-// #define SC_HIGH_PRECISION_STOPWATCH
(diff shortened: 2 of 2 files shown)


nuoHep:
Tickling CI. Will merge once it's ok.

scamille:
Is there a reason not to use std::chrono for all this?

And if the clocks in std::chrono are not good enough for our purposes, do we want to add our custom high_resolution clock conforming to std::chrono clock interface?

nuoHep:
We actually have 3 clocks here. And while `STOPWATCH_WALL` maps perfectly to `high_resolution`/`steady` `_clock`, there is nothing in `<chrono>` for the other 2.

But yeah, I guess we could change it to `std::chrono::hrc` + custom clocks with `std::chrono` interface. Will write it up later today.

scamille:
Just an idea I had from a while ago when last looking at stopwatch, not saying we really need it.

std::Chrono is a bit template heavy as far as I remember, but I feel quite strongly that having type-safe durations vs. timepoint distinction is a big advantage for any software, even if it sometimes brings slight inconveniences to the programmer.

PR #5116, opened 2020-05-19
Stopwatch & chrono stuff

nuoHep:
The first commit is:

Move clock related stuff into a new header under a top-level `chrono`
namespace.

Add 2 clocks meeting `Clock` requirements (`std::chrono` ones) for process cpu
time & thread cpu time. As wall clock time can be handled through an
`std::chrono` clock, add an alias for that to make it less to type and be
consistent with our clocks. `steady_clock` is chosen instead of
`high_resolution_clock` as the latter has a history of being "broken" on
different platforms (a lot of the time it's and alias to `system_clock`).

Add a `chrono::elapsed_fp_seconds()` helper as a replacement for
`util::duration_fp_seconds()` which also takes into account the clock type
(instead of always using hrc).

Leave `stopwatch_t` a simple type templated on a clock type. The latter must meet
requirements for `Clock` (the C++ named requirements one, basically must be
modeled after `std::chrono` clocks).

---

The latter commits refactor some places that actually do time related stuff to these new things.


scamille:
Great work.

The chrono definitions are simpler than I would have thought, and the implementation of the custom clocks is very tidy now.

PR #5118, opened 2020-05-20
Refactor random number generators.

scamille:
Remove Mersenne Twister, <random> ad other rng generators.

Add xoshiro256+ as a new default rng genreator.

Add SplitMix64 as a seed generator which itself accepts 0 as a initial seed value.

Refactor rng test / benchmark code.

### DIFF
--- a/engine/Makefile
+++ b/engine/Makefile
@@ -261,7 +261,7 @@ sc_http$(MODULE_EXT): interfaces$(PATHSEP)sc_http.cpp util$(PATHSEP)sc_io.cpp sc
 	-@echo [$@] Linking
 	$(CXX) $(CPP_FLAGS) -std=c++0x -DUNIT_TEST $(OPTS_INTERNAL) $(OPTS) $(LINK_FLAGS) $^ $(LINK_LIBS) -o $@
 
-rng$(MODULE_EXT): util$(PATHSEP)rng.cpp util$(PATHSEP)fmt$(PATHSEP)format.cpp
+rng$(MODULE_EXT): util$(PATHSEP)rng.cpp lib$(PATHSEP)fmt$(PATHSEP)format.cpp util$(PATHSEP)chrono.cpp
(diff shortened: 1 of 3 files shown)


nuoHep:
I'd go even further and completely remove support for choosing a prng. Just pick one and stick with it.

I never actually saw any point in it [prng] being selectable in the first place tbh. The default is already supposed to be the "best" one (for some definition of), so we basically let the user choose an inferior one for... what reason exactly?

On the plus side we could drop all of this virtual inheritance gunk, etc. Which will even slightly speed up the sim in the process (as we could store the prng by value, remove the virtual method indirection, let the compiler see through, etc.).

navv1234:
The point is that some people like to try out things and people are free to implement something that interests them in the simulator.

To that end though, I agree, we could just use one and speed it up a tiny bit.

scamille:
I played with that a long time ago, and the impact on performance was not measurable.

Back then it was voted in favor of keeping the runtime rng selection.

Removing it would certainly allow even more code removal, but that's only the virtual modifier and the factory. Changing the public API is too much work in any case.

nuoHep:
Fair enough.

I'll prototype it up in a pr if you don't mind after this lands. Just to see how invasive it is.

scamille:
The main goal with removing MT is that while the implementation was hardened and stable over the last years, the SSE implementation did cause quite some trouble in the past. There might even be some remaining alignment hacks on some sim_t or GUI objects somewhere, which were necessary on MinGW to not cause crashes in multi-threaded sims on that platform.

scamille:
Sure, feel free to do any further changes after this.

I personally don't see much benefit in having the runtime selection as well, but don't feel strongly about it.

A lot of stuff in SimC might have an impact on optimization and especially cache/memory access which you can't test in isolation, but that would require doing some correct, rigorous benchmarks on full simulations :)

PR #5122, opened 2020-05-22
Report level

scamille:
Initial attempt at a report version / report configuration system.

The idea is to parse all json= or jsonN= options, put the parsed level into a configuration object and pass to the report code. In the report, this level can then be used to tailor the report to the given version.
Currently this is only fully implemented for the json report.

The motivation for all this is to have some system which allows to easily add new report versions and deprecate ( or keep) older ones.

@navv1234 @nuoHep Please take a look if the architecture looks ok, or if there are any fundamental things which should be changed before adding this. I might refine it some more anyway, but wanted to get some initial feedback.


nuoHep:
One quick high level comment: do we actually need it for html & text?

I completely agree we need something like this for json as it's structured and requires versioning in some way (we'd need it for xml if we had it for the same reason).

But I don't see much point extending this to html & text reporting. HTML is "whatever", no one parses it (or at least I hope so), so as long as the information is there in some form for the user we can change it as we please (and I still hope we'll eventually get https://github.com/simulationcraft/simc/projects/2 so we don't have to dump raw html). And text report is basically immutable.

scamille:
I agree with you on the html report. It might still make sense for eg. migrating to a json-based web page, which some people planned a while ago.

For the text report it could make a lot of sense, since that is parsed quite often as well. The problem is that it is implicitly generated for each simulation.

In any case, the intention was to have a generic system which in theory could support any report types, even if it does not really benefit each of them. Right now the main "extra" feature would be supporting multiple outputs (which is not really that useful for most cases, except where you want to have different versions), but I can also imagine tying more report configuration options to this if we want to.

scamille:
The simpler option would be to just let json2= and json3= hardcoded options also set some int json_report_level, and pass that to the json report.

nuoHep:
I'm not against this. I simply don't think we need to make it generic from the start. Right now only json needs this. So I'd do it for json only now and then, when (if) we need this for another report type we could easily factor it out and make it generic.

scamille:
One of the things I did not want to do, is to just pass an int to the report function, and instead at least some object encapsulating it.

But yeah I think I'll redo a bit of the code and just hardwire json2 and json3 options. Thanks for the feedback.

scamille:
I discussed some more json report ideas with Aethys, and his desire to have a JSON-Schema for them to auto-generate stuff from it, bot for hero damage and an eventual html report.

After some discussion about how to version it, and me only wanting to use integers, I thought I would give this cpp-semver library a try. After a bit of fiddling, I got it working.

JSON report will now have semver versions, which allows us to more precisely indicate what changed on a new report version (maybe there will be a changelog for it).

It also allows to use these version ranges when specifying the json report option.
things like
json=foo,version=">=2.0.0"
json=foo,version="3.*"

scamille:
Closing this and pushing to simc repository as branch json-report-version.

PR #5123, opened 2020-05-23
[NOT-FOR-MERGE] Add util::static_map/set

nuoHep:
Tickling CI. Not really expecting it to compile on vs2017.


nuoHep:
Huh, that's... unexpected. vs2017 has better c++14 support than gcc5...

PR #5124, opened 2020-05-25
A bunch of ubsan fixes + Github Actions CI

nuoHep:
The main thing is the ubsan fixes to make the build somewhat more clean.

---

The CI is my experiment at setting up Github Actions. I simply wanted a clang-10 build with ubsan & asan that runs a bunch of different stuff with output (json, html etc). Not really married to it and can easily drop it. The nice thing though is that it's fast (compared to travis & appveyor) - all jobs finish in under 10 minutes (and gcc-9 smoke run is 5-6 minutes).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,98 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [ bfa-dev ]
+  push:
+    branches: [ bfa-dev ]
+
+env:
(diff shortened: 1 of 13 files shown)


navv1234:
If its more reliable and easier to add new test stuff than travis i'm all for it.

scamille:
If we can just run this in parallel to existing CI builds, I am all for it to just activate it.

We can then disable travis later if we are happy with it.

PR #5130, opened 2020-05-30
A bunch of misc cleanups

nuoHep:


scamille (commented):
Edit: This comment addresses 8cf6ca4 
Oh that is a good catch, copying the sample data container probably did cost quite a bit both in terms of memory and performance.

Is the size ==0 change not (potentially) less performant for non-random access iterators?

What would best match the intent is probably std::empty. Maybe we want to add that to range:: ?

nuoHep:
It should be either the same or faster actually. `range::size()` directly calls the containers `.size()` member function. It is constant time (since c++11) for all std containers that have it*.

[*] `std::forward_list<>` does not have `.size()`, but I don't see much point catering to it considering its perf characterstics for the usecase at hand (abysmal)

PR #5132, opened 2020-05-31
Include ostream required by VS2019

Eliviel:
Problem in VS2019:
> error C2027: use of undefined type 'std::basic_ostream<char,std::char_traits<char>>'

More info in answer:
https://developercommunity.visualstudio.com/content/problem/503648/simple-code-no-longer-compiles-when-using-and-stdo.html

Mainly:
> This behavior change is by design. The previous arrangement of <xstring> created a number of problems:
> * Programs that only wanted to use std::string facilities were including all of the iostreams machinery. The consequence of this was that a program that used only <string> took ~40% longer to compile than it should have.

### DIFF
--- a/engine/player/player_talent_points.hpp
+++ b/engine/player/player_talent_points.hpp
@@ -12,6 +12,7 @@
 #include <vector>
 #include <functional>
 #include <string>
+#include <ostream>
 
 struct spell_data_t;
 


scamille:
We have been changing our includes from one big header with all std includes, to separate specific includes for each translation unit. That is why this problem likely occured, and because different platforms have different implicit std includes.

What I find strange is that our automated CI build with VS2019 is not broken without this.

scamille:
Thanks a lot for your pull request.

I wanted to have a fix which does not bring back <iostream> to a header file. Since the iostream function seems to be unused anyway, I just removed it, and changed the to_string() implementation to use fmt formatting.

PR #5143, opened 2020-06-03
RFC: hook up max spell targets from dbc info

nuoHep:
Not sure if worth. Also, I suck at naming things, not even sure if `max_targets` is how this should be called.

@navv1234 dunno where you are at with your dbc export refactorings, just fyi that we may need this info

### DIFF
--- a/dbc_extract3/dbc/data.py
+++ b/dbc_extract3/dbc/data.py
@@ -86,6 +86,14 @@ class RawDBCRecord:
         db = dbc.db.datastore().get(child_db)
         return db.records_for_parent(self._id)
 
+    def child(self, child_db):
+        if not dbc.db.datastore():
+            raise ValueError
+
+        db = dbc.db.datastore().get(child_db)
+        records = db.records_for_parent(self._id)
+        return len(records) and records[0] or db.default()
+
     def child_refs(self, child_db):
         if not dbc.db.datastore():
             return []
--- a/dbc_extract3/dbc/generator.py
+++ b/dbc_extract3/dbc/generator.py
@@ -2856,6 +2856,11 @@ class SpellDataGenerator(DataGenerator):
             fields += category.field('dmg_class')
             f, hfd = category.get_hotfix_info(('dmg_class', 47))
 
+            # 48
+            targeting = spell.child('SpellTargetRestrictions')
+            fields += targeting.field('max_affected_targets')
+            f, hfd = targeting.get_hotfix_info(('max_affected_targets', 48))
+
             # Pad struct with empty pointers for direct access to spell effect data
             # 47, 48, 49, 50, 51
             fields += [ u'0', u'0', u'0', u'0', u'0', ]
--- a/engine/dbc/sc_spell_info.cpp
+++ b/engine/dbc/sc_spell_info.cpp
@@ -1435,6 +1435,15 @@ std::string spell_info::to_str( const dbc_t& dbc, const spell_data_t* spell, int
     s << ( int ) spell -> max_range() << " yards" << std::endl;
   }
 
+  if ( spell -> max_targets() != 0 )
+  {
+    fmt::print( s, "Max Targets      : {}{}{}\n",
+      spell -> max_targets() == -1 ? "Unlimited(" :
+        spell -> max_targets() < 0 ? "Unknown(" : "",
+      spell -> max_targets(),
+      spell -> max_targets() < 0 ? ")" : "" );
+  }
+
   if ( spell -> _cast_min > 0 || spell -> _cast_max > 0 )
   {
     s << "Cast Time        : ";
--- a/SpellDataDump/deathknight.txt
+++ b/SpellDataDump/deathknight.txt
@@ -7338,6 +7338,7 @@ Class            : Death Knight
 School           : Shadow
 Spell Type       : None
 Resource         : 30 Runic Power (6) (id=11019)
+Max Targets      : 20
 GCD              : 1.5 seconds
 Labels           : 16: Agonizing Backlash (320035 effect#3)
 Affecting spells : Sudden Doom (81340 effect#1)
@@ -7686,6 +7687,7 @@ Class            : Death Knight
 School           : Shadow
 Spell Type       : Magic
 Range            : 100 yards
+Max Targets      : 7
 Labels           : 16: Agonizing Backlash (320035 effect#3)
 Affecting spells : Mastery: Dreadblade (77515 effect#1), Sudden Doom (81340 effect#1), Frost Death Knight (137006 effect#1), Unholy Death Knight (137007 effect#1), Blood Death Knight (137008 effect#1), Master of Ghouls (246995 effect#1), 
 Family Flags     : 43, 78, 121
@@ -12792,6 +12794,7 @@ Class            : Death Knight
 School           : Shadow
 Spell Type       : Magic
 Range            : 20 yards
+Max Targets      : 1
 Labels           : 16: Agonizing Backlash (320035 effect#3)
 Affecting spells : Frost Death Knight (137006 effect#1), Unholy Death Knight (137007 effect#1), Blood Death Knight (137008 effect#1), Master of Ghouls (246995 effect#1), 
 Triggered by     : Abomination Limb (315443)
--- a/SpellDataDump/demonhunter.txt
+++ b/SpellDataDump/demonhunter.txt
@@ -3533,6 +3533,7 @@ School           : Shadow
 Spell Type       : Magic
 Spell Level      : 34
 Range            : 20 yards
+Max Targets      : 1
 GCD              : 1.5 seconds
 Duration         : 60 seconds
 Cooldown         : 45 seconds
--- a/SpellDataDump/druid.txt
+++ b/SpellDataDump/druid.txt
@@ -1382,6 +1382,7 @@ Spell Type       : Magic
 Resource         : 8% Base Mana (0) (id=1447)
 Spell Level      : 20
 Range            : 40 yards
+Max Targets      : 1
 GCD              : 1.5 seconds
 Duration         : 15 seconds
 Labels           : 16: Agonizing Backlash (320035 effect#3)
@@ -3671,6 +3672,7 @@ School           : Nature
 Spell Type       : Magic
 Spell Level      : 1
 Range            : 40 yards
+Max Targets      : 1
 Labels           : 16: Agonizing Backlash (320035 effect#3)
 Attributes       : ........ ........ ........ ........   ........ ..x..... ........ ........ 
                  : ........ ........ ........ .....x..   ........ ........ ........ ......x. 
@@ -3710,6 +3712,7 @@ School           : Physical
 Spell Type       : None
 Spell Level      : 19
 Range            : 40 yards
+Max Targets      : 1
 Velocity         : 20 yards/sec
 Duration         : 3600 seconds
 Cooldown         : 30 seconds
@@ -4581,6 +4584,7 @@ Class            : Druid
 School           : Nature
 Spell Type       : Magic
 Range            : 40 yards
+Max Targets      : 60
 Duration         : 8 seconds
 Labels           : 16: Agonizing Backlash (320035 effect#3)
 Stacks           : 5 maximum
@@ -11493,6 +11497,7 @@ School           : Arcane
 Spell Type       : None
 Resource         : 2% Base Mana (0) (id=192730)
 Range            : 40 yards
+Max Targets      : 1
 Cast Time        : 2.5 seconds
 GCD              : 1.5 seconds
 Duration         : 3600 seconds
--- a/SpellDataDump/hunter.txt
+++ b/SpellDataDump/hunter.txt
@@ -316,6 +316,7 @@ Spell Type       : Ranged
 Resource         : 40 Focus (2) (id=11195)
 Spell Level      : 29
 Range            : 40 yards
+Max Targets      : 5
 GCD              : 1.5 seconds
 Velocity         : 50 yards/sec
 Requires weapon  : Bow, Gun, Crossbow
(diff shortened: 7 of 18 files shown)


navv1234:
Yeah, that's fine and useful info for us in certain situations.

My intent with the refactoring is to split everything into their own files instead of growing our primary structs even more (spell, item). For that to happen we need to figure out a way to do something like what you had (the pseudo-runtime linking of data X with our "spell_data_t" using spans or something similar).

That change will also mean that the whole interface on how we access that data could change, as there might be multiple entries in a linked db2 file for a single spell (might not be true for player-related spells though, this is more on the enemy side).

dbc_extract3/dbc/generator.py
```diff
@@ -2856,6 +2863,11 @@ def generate(self, ids = None):
             fields += category.field('dmg_class')
             f, hfd = category.get_hotfix_info(('dmg_class', 47))
 
+            # 48
+            targeting = spell.get_link('targeting')
+            fields += targeting.field('max_affected_targets')
+            f, hfd = targeting.get_hotfix_info(('max_affected_targets', 48))
+
```
navv1234, line 2863: You should be able to do something like:
```
targeting = spell.children('SpellTargetRestrictions')
if len(targeting):
    fields += targeting[0].field('max_affected_targets')
    f, hfd = targeting[0].get_hotfix_info(('max_affected_targets', 48))
else:
    fields += self.db('SpellTargetRestrictions').default().field('max_affected_targets')
```

This is kinda clunky though, so I'm also slightly wondering if that children() interface should change to return a default entry if there are no actual children available for the parent. Note also that children() will return a list of entries, because technically, a single parent record can have multiple child records in a db2 file.

dbc_extract3/dbc/data.py
```diff
@@ -643,6 +643,9 @@ def initialize_data_model(options):
         if options.build >= dbc.WowVersion(8, 2, 0, 30080):
             SpellName.link('azerite_essence', AzeriteEssencePower)
 
+        if options.build >= dbc.WowVersion(9, 0, 0, 0):
+            SpellName.link('targeting', SpellTargetRestrictions)
+
```
navv1234: If you go by the new data model stuff (see comment below where you add the field to spell_data_t record output), linking is completely unnecessary, if the current format file defines this linkage (i.e., the `parent` value for `SpellTargetRestrictions` is set to `SpellName` in the format file).

dbc_extract3/dbc/generator.py
```diff
@@ -1993,6 +1993,9 @@ def __init__(self, options, data_store):
         if self._options.build >= dbc.WowVersion(8, 3, 0, 0):
             self._dbc.append('ItemBonus')
 
+        if self._options.build >= dbc.WowVersion(9, 0, 0, 0):
+            self._dbc.append('SpellTargetRestrictions')
+
```
navv1234: If you go by the new data model stuff (see comment below where you add the field to spell_data_t record output), explicitly instructing the generator to load db2 files is unnecessary.

dbc_extract3/dbc/generator.py
```diff
@@ -2029,6 +2032,10 @@ def initialize(self):
         self._data_store.link('ItemSetSpell', 'id_item_set', 'ItemSet', 'bonus')
 
         self._data_store.link('ItemEffect', self._options.build < 25600 and 'id_item' or 'id_parent', self._options.build < 23436 and 'Item-sparse' or 'ItemSparse', 'spells')
+
+        if self._options.build >= dbc.WowVersion(9, 0, 0, 0):
+            self._data_store.link('SpellTargetRestrictions', 'id_parent', 'SpellName', 'targeting')
+
```
navv1234: If you go by the new data model stuff (see comment below where you add the field to spell_data_t record output), explicitly linking stuff into a parent db2 file is unnecessary, **if** the current format file defines this linkage (i.e., the `parent` value for `SpellTargetRestrictions` is set to `SpellName` in the format file).

engine/dbc/sc_spell_info.cpp
```diff
@@ -1435,6 +1435,14 @@ std::string spell_info::to_str( const dbc_t& dbc, const spell_data_t* spell, int
     s << ( int ) spell -> max_range() << " yards" << std::endl;
   }
 
+  if ( spell -> max_targets() != 0 )
+  {
+    fmt::print( s, "Max Targets      : {}{}{}\n",
+      spell -> max_targets() < 0 ? "Unlimited(" : "",
```
navv1234: Might make sense to call this ´Unknown´ instead of Unlimited, from the spell query dump we can see some values of at least -1 and -6, meaning Blizzard might be using negative values to denote some thing server-sidey.

dbc_extract3/dbc/generator.py
nuoHep: That's much nicer, yeah. I decided to go a bit further and introduced a `.child()` accessor. Not sure... but for stuff that we either know there is only 1 thing (or we simply don't care) it seems it's much easier.

engine/dbc/sc_spell_info.cpp
nuoHep: Yeah, so `-1` is def unlimited from what I've seen. So I decided to do:
* `Unlimited(-1)` if n == -1
* `Unknown(n)` otherwise if n < 0

nuoHep:
> My intent with the refactoring is to split everything into their own files instead of growing our primary structs even more (spell, item). For that to happen we need to figure out a way to do something like what you had (the pseudo-runtime linking of data X with our "spell_data_t" using spans or something similar).
> 
> That change will also mean that the whole interface on how we access that data could change, as there might be multiple entries in a linked db2 file for a single spell (might not be true for player-related spells though, this is more on the enemy side).

I'm not sure about splitting `spell_data_t` tbh. While it's natural for labels, powers and (especially) effects (it's always 1-to-n and the linked data is somewhat big) I'd be very cautious about splitting off other stuff. Mostly perf wise. No matter how you implement that it's going to be an indirection.

While you could say that we don't do runtime spell data access that much that's... not really true. Or at least my testing shows that it isn't.
I was looking at reducing simc binary size recently. This basically amounts to reducing the size of the different dbc structs. One thing that I already knew (and is a pet peeve of mine[*]) is that we use doubles *everywhere*, even for client originated data (which is 32b floats). A simple change from double to float for all dbc exported structs and I got a 4-5% slowdown depending on the profile... Not nice (but it reduced the binary size by ~5MB /shrug).

I did't really analyze the results (no `perf`) and I didn't really look how spell data is split up between the different db2 files. So maybe it wouldn't be as bad, but we must be wary of possible perf regressions with such a split up.

[*] So yeah, doubles. This is a huge pet peeve of mine. I think I remember asking Nate a long time ago why simc used doubles everywhere. I don't remember the exact answer, but I do remember it making sense then - smth along the lines of: we had no spell data back then, so it was all manual constants and using doubles was just easier (no requirement to sprinkle `f`s around).
But right now we mostly use client spell data, which is in floats. As I come from game dev myself I would be *very* surprised if there are actually any computations done with doubles in game (we build with `-Werror=double-promotion` for example). So we don't need the extended precision (I wouldn't even be surprised if some of the inconsistencies we get from the sim vs in-game numbers are due to us using doubles). But alas...

engine/dbc/sc_spell_info.cpp
navv1234: Yep, that's good.

dbc_extract3/dbc/generator.py
navv1234: Yeah that's fine, although it does hide problems if blizzard makes an unconventional spell data change (adds another child record and the ordering is such that the old one is no longer the first entry).

navv1234:
> > My intent with the refactoring is to split everything into their own files instead of growing our primary structs even more (spell, item). For that to happen we need to figure out a way to do something like what you had (the pseudo-runtime linking of data X with our "spell_data_t" using spans or something similar).
> > That change will also mean that the whole interface on how we access that data could change, as there might be multiple entries in a linked db2 file for a single spell (might not be true for player-related spells though, this is more on the enemy side).
> 
> I'm not sure about splitting `spell_data_t` tbh. While it's natural for labels, powers and (especially) effects (it's always 1-to-n and the linked data is somewhat big) I'd be very cautious about splitting off other stuff. Mostly perf wise. No matter how you implement that it's going to be an indirection.

Yes this is the problem. I want to do it because in my opinion the spell data generator is too complicated right now. I want to simplify it. I also want to make changes more visible, figuring out what changed in our massive spell data from diffs is quite difficult in some cases. However, I am not willing to sacrifice performance too much there, if the indirection is a significant hit, then it's no go .. but there's no way of knowing that without doing some tests, and splitting up the huge struct is a pain.
 
> While you could say that we don't do runtime spell data access that much that's... not really true. Or at least my testing shows that it isn't.
> I was looking at reducing simc binary size recently. This basically amounts to reducing the size of the different dbc structs. One thing that I already knew (and is a pet peeve of mine[*]) is that we use doubles _everywhere_, even for client originated data (which is 32b floats). A simple change from double to float for all dbc exported structs and I got a 4-5% slowdown depending on the profile... Not nice (but it reduced the binary size by ~5MB /shrug).
> 
> I did't really analyze the results (no `perf`) and I didn't really look how spell data is split up between the different db2 files. So maybe it wouldn't be as bad, but we must be wary of possible perf regressions with such a split up.

Are we talking about spell data or effect access? I would expect spell data during iteration to be minimal (almost all modules cache pointers), however effect data is another thing entirely, that happens very often (`->effectN( x ).something()`). But that stuff is already behind an indirection.

Re: doubles, yes, I want to change client data export from double to float. It should have been like this from the start, but I suspect my original motivation when implementing the system was "simc uses doubles, so client data export uses doubles". And yes, client data export came later, if memory servers right around WotLK. If it's a noticeable performance regression though, I'm not sure we should do it. I'd rather take larger binary.

> 
> [*] So yeah, doubles. This is a huge pet peeve of mine. I think I remember asking Nate a long time ago why simc used doubles everywhere. I don't remember the exact answer, but I do remember it making sense then - smth along the lines of: we had no spell data back then, so it was all manual constants and using doubles was just easier (no requirement to sprinkle `f`s around).
> But right now we mostly use client spell data, which is in floats. As I come from game dev myself I would be _very_ surprised if there are actually any computations done with doubles in game (we build with `-Werror=double-promotion` for example). So we don't need the extended precision (I wouldn't even be surprised if some of the inconsistencies we get from the sim vs in-game numbers are due to us using doubles). But alas...

dbc_extract3/dbc/generator.py
nuoHep: I don't think we would be able to handle such a change for stuff we expect a 1-to-1 mapping even with an explicit `children()[0]` access. It would also hide the problem from us.

What I think we can do for data we expect a 1-to-1 link is to slap a warning inside `child()` in case the number of linked records is more than one. It would actually better protect us from changes in spell data.

navv1234: No, of course not, better way would be to analyze the data before generation and warn about multiple entries.

EDIT: And yes, can add the warning to child() also.

nuoHep:
> [ ... ]
> 
> Are we talking about spell data or effect access? I would expect spell data during iteration to be minimal (almost all modules cache pointers), however effect data is another thing entirely, that happens very often (`->effectN( x ).something()`). But that stuff is already behind an indirection.
> 

Spell data. What I mean is: say we split off `SpellAuraOptions.db2` data from `spell_data_t` into its own struct and slap a pointer into spell data to "link" it. While the `spell_data_t` pointers are cached, accessing any field from aura options now is behind another pointer. Aura options contains `proc_chance` field which, from a quick grep, is accessed quite a lot from `action_t::execute()`s in different modules.
It obviously has to be profiled.

And yeah, effects right now are behind a triple indirection (ptr to vector -> ptr to buffer -> ptr to effect). Reducing it was one of the motivations for my "spanification". It removed 1 indirection. And I think we can actually do even better (at the cost of slightly more complexity for runtime data overrides machinery).

> Re: doubles, yes, I want to change client data export from double to float. It should have been like this from the start, but I suspect my original motivation when implementing the system was "simc uses doubles, so client data export uses doubles". And yes, client data export came later, if memory servers right around WotLK. If it's a noticeable performance regression though, I'm not sure we should do it. I'd rather take larger binary.
> 
> [ ... ]

Thinking about it a bit more, what I think we can safely do is change all item-related exports to floats and leave spell/effect data as doubles (at least as a first step). While spell data gets a lot of runtime accesses (where an extra `cvtss2sd` actually slows down things) item data is accessed only on init afaik.

navv1234:
> > [ ... ]
> > Are we talking about spell data or effect access? I would expect spell data during iteration to be minimal (almost all modules cache pointers), however effect data is another thing entirely, that happens very often (`->effectN( x ).something()`). But that stuff is already behind an indirection.
> 
> Spell data. What I mean is: say we split off `SpellAuraOptions.db2` data from `spell_data_t` into its own struct and slap a pointer into spell data to "link" it. While the `spell_data_t` pointers are cached, accessing any field from aura options now is behind another pointer. Aura options contains `proc_chance` field which, from a quick grep, is accessed quite a lot from `action_t::execute()`s in different modules.
> It obviously has to be profiled.

Right, yeah splitting off things would have implications on that for fields that are accessed during runtime.

> And yeah, effects right now are behind a triple indirection (ptr to vector -> ptr to buffer -> ptr to effect). Reducing it was one of the motivations for my "spanification". It removed 1 indirection. And I think we can actually do even better (at the cost of slightly more complexity for runtime data overrides machinery).
> 
> > Re: doubles, yes, I want to change client data export from double to float. It should have been like this from the start, but I suspect my original motivation when implementing the system was "simc uses doubles, so client data export uses doubles". And yes, client data export came later, if memory servers right around WotLK. If it's a noticeable performance regression though, I'm not sure we should do it. I'd rather take larger binary.
> > [ ... ]
> 
> Thinking about it a bit more, what I think we can safely do is change all item-related exports to floats and leave spell/effect data as doubles (at least as a first step). While spell data gets a lot of runtime accesses (where an extra `cvtss2sd` actually slows down things) item data is accessed only on init afaik.

Item data is only accessed during initialization, yes. I also want to split the item struct a bit (carve off item effects; we already have item effect export for the new item bonus system, so we should be able to do this for item effects easy enough.

nuoHep:
Closing. Rebased and resubmitted as #5180

PR #5154, opened 2020-06-19
Bunch of dbc refactoring

nuoHep:
The main thing is the removal of runtime hotfix linking. There are also some other cleanups.

PR to check msvc compatibility for some of the template stuff.


scamille:
Is it correct that the actual hotfix application to a spell/effect/etc. is done directly in dbc_extract, and all SimC uses the hotfix data is for reporting about it?

Otherwise I am not yet fully understanding your commit :-)

nuoHep:
Yes, hotfix data is only for spell query output. Actual hotfixing is done by the extractor so the exported spell/effect/etc arays already have the proper hotfixed values.

PR #5163, opened 2020-06-27
[DBC] separate out item stats from item data

nuoHep:
Modern wow items typically have no more than 4 stats. We currently "store" 10,
just like dbcs do. By storing them separately from the main item data we can
greatly reduce item data storage requirements by effectively not storing more
stats than strictly required.

Separate stats storage also lets us "deduplicate" it - a lot of items have same
stat allocation. For the current build it reduces ~250k stat records to ~46k,
which is even less than the number of currently exported items (~68k). The only
downside is that it makes item data export diffs potentially illegible.

One complication is that `item_data_t` is currently used both as the client
exported data type (immutable data) and for storing player items which are
mutable data. Handle this by separating it into 2 distinct types:
`dbc_item_data_t` for the former and `parsed_item_data_t` for the latter.

Additional steps taken to further reduce storage requirements:
 * stat "arrays" are stored as a pointer + size pair (instead of an
   `util::span`) - lets us use a smaller type for size
 * stat type & amount are stored as int16_t (instead of ints); in the client
   data type is a int8 so we are fine; amount is an int32 so we may have
   problems, but we will catch it at compile time as brace-enclosed init lists
   forbid narrowing conversions
 * `dbc_item_data_t` fields are shuffled a bit to reduce holes due to alignment

An explicit goal was to minimize the amount of changes to non-item related
code. There are possible cleanups on top of that (like changing the way stats
are stored in `parsed_item_data_t`).

For a clang-10 `-O3 -flto-thin -ffast-math` build reduces binary size by ~7MB:
```
    FILE SIZE        VM SIZE
  [NEW]  +363Ki  [NEW]  +363Ki    __item_stats_data
  [NEW]    +760  [NEW]    +687    parsed_item_data_t::init()
   +64%    +617   +70%    +617    (anonymous namespace)::bfa::init_punchcard()
   +23%    +610   +24%    +610    enchant::initialize_relic()
  +0.0%    +125  +0.0%      +6    [775 Others]
 -32.6% -7.26Mi -32.6% -7.26Mi    __item_data
 -13.8% -6.90Mi -14.5% -6.91Mi    TOTAL
```


nuoHep:
Added stats sorting & "full" deduplication. Compared to initial commit shaved off another ~80KB (almost 10k stat records less).

PR #5167, opened 2020-06-30
[DBC] separate out item effects from item data

nuoHep:
This is a bit more complicated than stats as the effects are accessed both
through the `dbc_item_data_t` (which is immutable) and through
`parsed_item_data_t` (where effects can come from wowhead/bcp_api/item
bonuses).

To minimize potential problems with effects getting out of sync (in case a
`parsed_item_data_t` is accessed through a `dbc_item_data_t` pointer) we use an
inner pointer for effects and forbid copying of `dbc_item_data_t`. Pretty sure
the problem is a bit theoretical and maybe there is a more elegant solution,
comments welcome.

We also can't deduplicate as much - the effects have indexes (so we can't sort
them) and are not always "contiguous" (there are "holes"). We *seem* to care
about that. But we still reduce `6439` total effects to 4171 unique "effect
records". The comparison is a bit off as the latter also includes 470 'nil'
effects where client data has "holes".

Reasons for not using `item_effect_t`:
 * exporting the effects as part of ItemEffectGenerator is much more
   complicated - we would have to store per item effect id arrays and then do
   some kind of runtime linking in init (which would require memory
   allocations and making item data non-const); this seems pretty suboptimal
   to me

 * exporting `item_effect_t`s as part of ItemDataGenerator simply wastes more
   space as we can't deduplicate the effects - `item_effect_t` contains an
   (id, index) tuple which is pretty much unique (for comparison, 6439
   `item_effect_t`s is ~176KiB + we would need either 470 "nils" (~12.8KiB) or
   6909 pointers (~54KiB))

In the end it's both more simple and takes less space to use a "custom" struct
that has only what we need (and uses smaller types for some fields) storing it
in a flat array that is easier to deduplicate.

```
    FILE SIZE        VM SIZE
  [NEW] +65.2Ki  [NEW] +65.2Ki    __item_effects_data
  +0.0% +1.39Ki  +0.0% +1.12Ki    [753 Others]
  [NEW]    +613  [NEW]    +472    parsed_item_data_t::add_effect()
 -19.1%    -626 -19.6%    -626    enchant::initialize_relic()
 -40.0%    -633 -42.2%    -633    (anonymous namespace)::bfa::init_punchcard()
 -41.4% -6.22Mi -41.4% -6.22Mi    __item_data
 -14.3% -6.16Mi -15.1% -6.16Mi    TOTAL```

### DIFF
--- a/dbc_extract3/dbc/generator.py
+++ b/dbc_extract3/dbc/generator.py
@@ -906,7 +906,7 @@ class ItemDataGenerator(DataGenerator):
 
             def items(self):
                 return self._list
-            
+
             def _find_pos(self, data):
                 for pos in self._dict[data[0]]:
                     if self._list[pos:pos + len(data)] == data:
@@ -914,9 +914,11 @@ class ItemDataGenerator(DataGenerator):
                 return None
 
         items_stats_index = Index()
+        items_effects_index = Index()
 
         for id in ids:
             item = self._itemsparse_db[id]
+
             stats = []
             for i in range(1, 11):
                 stat_type = getattr(item, 'stat_type_{}'.format(i))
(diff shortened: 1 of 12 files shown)


nuoHep:
One clarification/observation:

>[...] We also can't deduplicate as much - the effects have indexes (so we can't sort
them) and are not always "contiguous" (there are "holes"). We seem to care
about that.

I'm not sure if we do it properly now nor if we should care tbh. At first I tried full dedup filtering out "nil" effects (not caring about holes). This changed deterministic `T25_Priest_Shadow.simc` results. The culprit turned out to be `Psyche Shredder (id=174060)` - in item data it has a hole at index 0 (the only inherent effect is at index 1) so when it gets an effect through `bonus_id=6560` current code places it into this hole. The "interesting" part is that the corresponding `item_effect_t` record contains an index of **2**, so if I get it right (and if effects granted through bonus ids respect this field) it should be placed *after* the inherent effect (and then there would be no change to deterministic results).
In the end I decided to fully preserve current behavior.

navv1234:
The indices in item_effect_t did not always seem to correspond with empty slots on the base item, and thus I scrapped the plan with using them (based on the fact that I clearly did not understand what the use of the field was).

It might be worth it for shadowlands to implement inserting on indices, and then adding a warning if it is trying to overwrite something. In that case probably best to pick the next empty index.

EDIT: That being said, I'm not sure simc care about the ordering at all. There are some cases where an item can have multiple effects that reference each other, but I think that's pretty rare.

nuoHep:
`item_effect_t`-based export is turned out to be better overall
Closing in favor of #5168

PR #5168, opened 2020-07-01
[DBC] separate out item effects from item data

nuoHep:
An "alternatative" to https://github.com/simulationcraft/simc/pull/5167 that uses `item_effect_t`.

~~Compared to the other impl we don't save as much (but it's still 6MiB less) and it's a bit more complicated for bcp/wowhead items. We also have to make item data non-const as we now do runtime linking).~~
~~Not merge ready as item effect filtering is a hack (we instantiate a local `ItemDataGenerator` to filter items; somewhat surprisingly it's actually pretty fast).~~

Tuned out to be overall better:
* saves more space (by shrinking `dbc_item_data_t` more)
* *less* complicated on the C++ side (no inner pointers and related headaches)
* should be more maintainable in the generator
* vs2017 does not choke on it

```
    FILE SIZE        VM SIZE
 +31e2%  +150Ki +32e2%  +150Ki    __item_effect_data
  [NEW] +13.0Ki  [NEW] +12.9Ki    __item_effect_id_index
  +1.9% +1.14Ki  +1.9% +1.15Ki    (anonymous namespace)::print_html_player_()
  [NEW]    +717  [NEW]    +576    parsed_item_data_t::add_effect()
  +0.0%    +525  +0.0%    +197    [773 Others]
 -44.8% -6.77Mi -44.8% -6.77Mi    __item_data
 -15.3% -6.61Mi -16.2% -6.61Mi    TOTAL
```


nuoHep:
* backed off runtime linking - turned out to be very clean
* reworked `item_effect_t` storage a bit - effects are now sorted by `(item_id, index, id)` tuple and there is a separate small index "map" to speedup searches by effect id
* effects in `parsed_item_data_t` are stored by value, which simplifies quite some stuff
* inherited `ItemDataGenerator` instead of using a local instance

nuoHep:
Did some mild perf testing (all caveats about concrete setup obviously apply).

Both binaries compiled by clang10 with `-O3 -ffast-math -flto-thin`. 500 runs of `CI.simc deterministic=1 iterations=2000` for both. Times are obtained by grepping simc output (`wall` is `WallSeconds`, `init` is `InitSeconds`).

Init time does not seem to regress:
```
nuohep@r2d2:~/dev/simc> ministat -q times-init-8f4e022afa.csv times-init-HEAD.csv
x times-init-8f4e022afa.csv
+ times-init-HEAD.csv
x 500     0.4294766     0.7700738     0.5748266    0.57533303   0.060453633
+ 500     0.4127324     0.7947556      0.567055    0.57294091   0.064620165
No difference proven at 95.0% confidence
```

Wall time is actually lower
```
nuohep@r2d2:~/dev/simc> ministat -q times-wall-8f4e022afa.csv times-wall-HEAD.csv
x times-wall-8f4e022afa.csv
+ times-wall-HEAD.csv
x 500     11.247789     15.091807     11.540288     11.551576    0.25980009
+ 500      11.12745     15.612369     11.446557     11.452805    0.22267765
        -0.0987706 +/- 0.0299927
        -0.85504% +/- 0.259641%
        (Student's t, pooled s = 0.241952)
```
(it's less than 1% so meh)

PR #5170, opened 2020-07-04
[DBC] pseudo-runtime spell data linking

nuoHep:
Generate per-spell spell power/label/effect/driver "arrays" during spell data export generation.

Specifically we generate array of pointers to relevant struct sorted by (parent_spell_id, index/id) tuple and store the record counts in each exported spell. Linking then is a simple matter of "slicing" the pointer arrays using the counts and a running index.

While it somewhat complicates SpellDataGenerator we get several benefits on the c++ side:
  * one less indirection on access to the records through a spell
  * no dynamic memory allocations in init
  * in the end we get to slap const on a bunch of pointers
  * no need to "deinit" spell data

While this increases binary size a bit we should have less memory consumption overall (not like this is a major source of memory allocations in the sim, so meh).
```
nuohep@r2d2:~/dev/simc> bloaty -d symbols -n 5 build-optimized/simc.HEAD -- build-optimized/simc.b4326944e6
    FILE SIZE        VM SIZE
  [NEW]  +179Ki  [NEW]  +179Ki    __spelleffect_index_data
  +2.3%  +123Ki  +2.3%  +123Ki    __spell_data
  [NEW] +31.7Ki  [NEW] +31.6Ki    __spelllabel_index_data
  [NEW] +31.6Ki  [NEW] +31.5Ki    __spelldriver_index_data
  [NEW] +4.75Ki  [NEW] +4.67Ki    __spellpower_index_data
  -0.2% -16.3Ki  -0.2% -15.5Ki    [3307 Others]
  +0.9%  +354Ki  +1.0%  +354Ki    TOTAL
```

Runtime-wise this is a *very* slight speedup (at least for my setup here with all the usual caveats). `profiles/CI.simc deterministic=1 iterations=2000` compiled by clang-10 `-O3 -flto-thin -ffast-math` comparing `WallSeconds`.
```
nuohep@r2d2:~/dev/simc> ministat -q wall-b4326944e6.csv wall-HEAD.csv
x wall-b4326944e6.csv
+ wall-HEAD.csv
x 500     11.412602     14.173276     11.574831     11.590959    0.14507114
+ 500     11.355447     13.196773     11.478599     11.493844     0.1127418
        -0.0971155 +/- 0.0161046
        -0.837855% +/- 0.138941%
        (Student's t, pooled s = 0.129916)
```

### DIFF
--- a/SpellDataDump/deathknight.txt
+++ b/SpellDataDump/deathknight.txt
@@ -1,4 +1,4 @@
-SimulationCraft 901-01 for World of Warcraft 9.0.1.34972 shadowlands-BETA (git build shadowlands 2fa6bde648)
+SimulationCraft 901-01 for World of Warcraft 9.0.1.34972 shadowlands-BETA (git build dbc_spells 1877d36fa6)
 
 Name             : Dual Wield (rank=Passive) (id=674) [Passive] 
 Class            : Frost Warrior, Hunter, Rogue, Death Knight, Monk, Demon Hunter
@@ -8512,7 +8512,7 @@ Class            : Warrior, Paladin, Hunter, Rogue, Priest, Death Knight, Shaman
 School           : Physical
 Spell Type       : None
 Azerite Power Id : 21
-Triggered by     : Elemental Whirl (268953), Elemental Whirl (268956), Elemental Whirl (268955), Elemental Whirl (268954)
+Triggered by     : Elemental Whirl (268953), Elemental Whirl (268954), Elemental Whirl (268955), Elemental Whirl (268956)
 Attributes       : ......x. ........ ........ ........   ........ ........ ........ ........ 
                  : ........ ........ ........ ........   ........ ........ ........ ........ 
                  : ........ ........ ........ ........   ........ ........ ........ ........ 
@@ -8616,7 +8616,7 @@ Class            : Warrior, Paladin, Hunter, Rogue, Priest, Death Knight, Shaman
 School           : Physical
 Spell Type       : None
 Azerite Power Id : 31
-Triggered by     : Gutripper (270668), Gutripper (269031)
+Triggered by     : Gutripper (269031), Gutripper (270668)
 Attributes       : ......x. ........ ........ ........   ........ ........ ........ ........ 
                  : ........ ........ ........ ........   ........ ........ ........ ........ 
                  : ........ ........ ........ ........   ........ ........ ........ ........ 
@@ -9006,7 +9006,7 @@ Class            : Warrior, Paladin, Hunter, Rogue, Priest, Death Knight, Shaman
 School           : Physical
 Spell Type       : None
 Azerite Power Id : 98
-Triggered by     : Crystalline Carapace (271539), Crystalline Carapace (271538)
+Triggered by     : Crystalline Carapace (271538), Crystalline Carapace (271539)
 Attributes       : ......x. ........ ........ ........   ........ ........ ........ ........ 
                  : ........ ........ ........ ........   ........ ........ ........ ........ 
                  : ........ ........ ........ ........   ........ ........ ........ ........ 
--- a/SpellDataDump/demonhunter.txt
+++ b/SpellDataDump/demonhunter.txt
@@ -1,4 +1,4 @@
-SimulationCraft 901-01 for World of Warcraft 9.0.1.34972 shadowlands-BETA (git build shadowlands 2fa6bde648)
+SimulationCraft 901-01 for World of Warcraft 9.0.1.34972 shadowlands-BETA (git build dbc_spells 1877d36fa6)
 
 Name             : Dual Wield (rank=Passive) (id=674) [Passive] 
 Class            : Frost Warrior, Hunter, Rogue, Death Knight, Monk, Demon Hunter
@@ -3619,7 +3619,7 @@ School           : Physical
 Spell Type       : None
 Duration         : 15 seconds
 Labels           : 16: Agonizing Backlash (320035 effect#3)
-Triggered by     : Soul Fragment (204255), Soul Fragment (203795)
+Triggered by     : Soul Fragment (203795), Soul Fragment (204255)
 Attributes       : .......x ........ ........ x..x....   .....x.. ........ ........ ........ 
                  : ........ ......x. ........ ........   ........ ........ ........ ........ 
                  : ........ ........ ........ ........   ...x.... ........ .xx..... ........ 
@@ -4630,7 +4630,7 @@ Class            : Warrior, Paladin, Hunter, Rogue, Priest, Death Knight, Shaman
 School           : Physical
 Spell Type       : None
 Azerite Power Id : 21
-Triggered by     : Elemental Whirl (268953), Elemental Whirl (268956), Elemental Whirl (268955), Elemental Whirl (268954)
+Triggered by     : Elemental Whirl (268953), Elemental Whirl (268954), Elemental Whirl (268955), Elemental Whirl (268956)
 Attributes       : ......x. ........ ........ ........   ........ ........ ........ ........ 
                  : ........ ........ ........ ........   ........ ........ ........ ........ 
                  : ........ ........ ........ ........   ........ ........ ........ ........ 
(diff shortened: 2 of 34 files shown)


nuoHep:
I explicitly wanted to minimize the api changes of this.

As a follow up I'm considering trying to be a it more invasive and remove another layer of pointer indirection:
- We don't really need `spelllabel_data_t` at this point - we don't need anything apart from the label value anywhere. We can store the labels as a flat `uint16_t` array sorted by (spell_id, label_id) and slice into that.
- Storing `spellpower_data_t` sorted by (spell_id, power_id) is also easy. I don't even think we would need an index for optimizing access by id - the only place doing that is the runtimes overrides machinery and we could live with a linear scan for it imo (considering power count is comparatively small).
- In theory we could also sort `spelleffect_data_t` export by (spell_id, effect_index) plugging the holes by copies of `spelleffect_data_t::nil`. We would def require an index for optimizing "id" searches and it would depend on vs2017 not choking on it and compilers doing the copy at compile time.

scamille:
I don't understand your last bullet point: how are you doing the pseudo runtime linking for spell effects if you are not sorting them by spell_id, effect_index?

I don't think we necessarily need any optimization for looking things like effect up by id. Is there anything else besides spell query doing that?

nuoHep:
> I don't understand your last bullet point: how are you doing the pseudo runtime linking for spell effects if you are not sorting them by spell_id, effect_index?

Currently I generate an array of pointers which *is* sorted by `(spell_id, index)` and during linking we slice it up for spells. What I meant in my last bullet point is that I want to try sorting the "main" spell effect data array and remove the indirection through the pointer array.

navv1234:
I wouldn't remove the spell label information, it may be that Blizzard thinks they want to apply spell labels through the bonus id system at some point.

I do think some of this could be separated to a completely different set of info, much like item effects did. We honestly only need spell effects to be directly linked to spells, powers and labels both could just as well be a very simple separate compilation unit that we access when we need the information during init. That would simplify the whole thing at the cost of some initialization overhead, I suppose.

scamille:
Ah so you still have the indirection through the new linking array to the actual effects. But since before we had a heap allocated vector doing the linking, this new version is still one less indirection.

nuoHep:
Separating powers & labels completely is complicated by the fact that spells don't carry any context of where they "came from" and almost nothing in the sim really cares about that. For both powers and labels the "context" includes ptr/non-ptr (same as for item effects). Powers also can come from the runtime override systems. This is further complicated by the fact that there are "indirect" accessors to powers & labels right on `spell_data_t` (`::cost()`, `::affected_by_label()` etc).

So while we could separately export this stuff we can't really do the same as with item effects. We'd still have to keep the pointers in `spell_data_t` and link it all up during init with searches. Which, admittely, should be fine.

At this point we could actually export even effects separately.

Re labels: I just don't like that we waste 10 bytes to store 2. Granted, the overall size of label data has historically been pretty small (disregarding the current blowup because of label `16`). But I don't care enough to push for that.

nuoHep:
> Ah so you still have the indirection through the new linking array to the actual effects. But since before we had a heap allocated vector doing the linking, this new version is still one less indirection.

Yes, exactly

navv1234:
I don't really understand why we would need to keep the pointers in spell_data_t. If you apply the requirement that spell_data_t needs to have a "powers()" method to access them, then of course this is true, but I don't think we need that. We need to basically get the powers for a given spell once (during action construction), and in some cases class modules need to get them for some purpose. In my opinion, all of that can be just as easily handled by a couple of accessor methods in dbc_t.

nuoHep:
It can, it's just a bit of churn.

navv1234:
From a quick scan through the code, I see 5ish places where powers() is called, I don't think that's too bad. Override code may also become a bit simpler, since there would not be any need to update spell data pointers.

Personally speaking, I find the code a lot easier to work with and a lot easier to maintain if the data access is separated to it's own thing. There are some places where I think it's just not worth it though (spell effects are accessed too much during runtime), but almost everything else can be separated imo.

That being said, if you want to keep it all together I'm ok with that too, just think it'd overall simplify the code / make it easier to maintain.

nuoHep:
I'm mostly "worried" about calls like
https://github.com/simulationcraft/simc/blob/shadowlands/engine/class_modules/sc_druid.cpp#L2868
rn it's pretty easy to grasp and use for class devs. Separating powers out would require changing it to smth like `data().cost( player, POWER_COMBO_POINT )` (we need the player for overrides and the dbc pointer). Which is... not too bad as it kinda makes sense. Dunno. For calls to `powers()` the requirement to pass in a player pointer would be somewhat non-obvious.

Moving the accessors off of `spell_data_t` is a complete no-go from my pov (as a hunter module dev).

Anyway, my current plan:
1. try sorting the underlying data arrays
2. merge it
3. look at splitting spell data as a follow-up

Is it ok with you?

navv1234:
Sure, if you want to keep the cost() and such stuff in spell_data_t, no point in trying to split anything honestly at least in terms of powers.

scamille:
Is the spell effect usage something we want to fully support with / strive for perfect runtime performance since it is so widely used in class modules, or is it something we want to gradually adjust class modules to some different usage system?

And assuming we want the first option, is there really anything using spell effects separate from spell data?

Anyway, I am just trying to follow along, trying to get some better understanding about the dbc system with it. I don't have any strong opinions for either way.

Regarding this PR, I would say the new runtime linking using spans is desirable compared to the existing one, independent from any further separation or complete removal of eg. power to spell data linking.

navv1234:
In my opinion the spell effect linking is definitely wanted, both in terms of the interface, but also in terms of performance.

nuoHep:
Sorted the underlying arrays. A bit of churn for the users but it's better this way.

We get an expected bigger speedup:
```
nuohep@r2d2:~/dev/simc> ministat -q wall-b4326944e6.csv wall-HEAD.csv
x wall-b4326944e6.csv
+ wall-HEAD.csv
x 500     11.412602     14.173276     11.574831     11.590959    0.14507114
+ 500     10.967504     20.825408     11.186813       11.2536    0.61028788
        -0.337359 +/- 0.0549847
        -2.91054% +/- 0.474376%
        (Student's t, pooled s = 0.443563)
```
and it's also smaller
```
nuohep@r2d2:~/dev/simc> bloaty -d symbols -n 5 build-optimized/simc -- build-optimized/simc.b4326944e6
    FILE SIZE        VM SIZE
  +2.3%  +123Ki  +2.3%  +123Ki    __spell_data
  +1.3% +56.1Ki  +1.3% +56.1Ki    __spelleffect_data
  [NEW] +44.3Ki  [NEW] +44.3Ki    __spelleffect_id_index
  [NEW] +31.6Ki  [NEW] +31.5Ki    __spelldriver_index_data
   +34% +4.41Ki   +36% +4.41Ki    paladin::paladin_action_t<>::paladin_action_t()
  -0.4% -24.5Ki  -0.4% -23.8Ki    [3353 Others]
  +0.6%  +234Ki  +0.7%  +235Ki    TOTAL
``` 
(`paladin_action_t` increase is because the compiler decided to inline a lambda)

As a nice effect of this labels, powers & effects are now mostly decoupled from `spell_data_t` on the c++ side. I think we can now do linking only in the sim with `dbc::find_many()`. Will look at it later.

scamille:
Is doing linear search instead of creating a separate index that much slower? Do we care for those cases where we access labels/powers/effects by id?

Curious how direct linking spell effects will affect performance. I assume it has the biggest impact because of the many use cases.

nuoHep:
Ughh, vs2017, why so slow?.. Will try to fix it up later (pretty sure it's the same case of nested pointers etc as with items).

> Is doing linear search instead of creating a separate index that much slower? Do we care for those cases where we access labels/powers/effects by id?

We definitely need it for effects - there are a lot a places doing by id searches and I'm not sure they are all init-time only. I don't do it for labels. And for spellpowers, while we don't do much searches, it's just easy to do and doesn't cost us much (as the index is less than 1KiB).

> Curious how direct linking spell effects will affect performance. I assume it has the biggest impact because of the many use cases.

The first output (`ministat -q`) in my last comment is comparing wall times of 500 runs of `profiles/CI.simc iterations=2000 deterministic=1` between HEAD of this pr and b4326944e6. Binary compiled with clang10 `-O3 -flto=thin -ffast-math` and run on i5-4670. So there is an almost 3% speedup.

nuoHep:
Pushing it now as-is so it does not rot. Will come back to a more fine grained split as a follow-up later.

PR #5176, opened 2020-07-23
Drastically (5-10%) improve performance by eliding to_string call when debug logging is disabled.

Mazorlion:
I had started doing some performance analysis on simc last year, but got sidetracked and left it on the ground (besides one small PR). Now that I have some more time and interest, the number one low hanging fruit on my list is debug logging.

The common pattern for debug logging in simc is:

```c++
sim->print_debug( "my object {} with field {} and field {} and field {}", name(), other_function, other_field, ... );
```

And in various places the pattern of first checking if `sim->debug()` is true has been adopted (as in this CL)
```c++
  if ( sim->debug )
  {
    sim->print_debug( "my object {} with field {} and field {} and field {}", name(), other_function, other_field, ... );
  }
```

The issue with the former pattern is that all the functions that feed into the log message are evaluated regardless of whether we're actually going to log or not. This might not seem like a big deal, but some of these functions are called thousands or millions of times per run and there are numerous functions that call `print_debug`.

It's unfortunately a bit hard to quantify the total CPU that would be saved by moving all calls to the latter pattern because the wasted calls are not trivially associated with the print_debug statement in most perf tools (I'm working on this).  That being said, the following change addresses the most egregious of these calls to print_debug where in some of my sample sims it was consuming >10% of the total CPU.

I have plans to propose larger scale changes to how logging is done to improve on this point, but the savings in this instance is so large I just wanted to get it in. 

FWIW the problematic print_debug was the second one, but I figured I'd change both while I'm changing this function.

Note: CPU includes the function and all child functions.
(Resolution is a bit funky, just click and enlarge the images)
Windows (NDEBUG Release) - VS Profiler Warlock (11.18% total CPU)
to_string_cpu_usage_warlock
Windows (NDEBUG Release) DH - VS Profiler (4.28% total CPU)
to_string_cpu_usage
Linux (gcc, optimized LTO) warlock patchwerk sim - Valgrind (callgrind) (9.33% total CPU)
to_string

Fixed Windows (NDEBUG Release) Warlock - VS Profiler (0.00% total CPU)
fixed_to_string
Fixed Linux (gcc, optimized LTO) warlock patchwerk - Valgrind (callgrind) (0.21% total CPU)
fixed_player_t_to_string_cpu

### DIFF
--- a/engine/player/sc_player.cpp
+++ b/engine/player/sc_player.cpp
@@ -5069,7 +5069,10 @@ void player_t::reset()
   // Restore default target
   target = default_target;
 
-  sim->print_debug( "{} resets current stats ( reset to initial ): {}", *this, current.to_string() );
+  if ( sim->debug )
+  {
+    sim->print_debug( "{} resets current stats ( reset to initial ): {}", *this, current.to_string() );
+  }
 
   for ( auto& buff : buff_list )
     buff->reset();


nuoHep:
As a general point, as you somewhat alluded to yourself in
>The issue with the former pattern is that all the functions that feed into the log message are evaluated regardless of whether we're actually going to log or not. 

the issue **are** the function calls performed in `print_debug()` calls and not the unguarded `print_debug()` calls themselves.

While this could be "fixed" by guarding all `print_log()`/`print_debug()` calls (basically reverting to before their introduction) another alternative (which I like a bit more) is to defer "stringification" by adding proper formatting support to types and eliminating the function calls. This basically amounts to removing most of `to_string()` methods and adding a specialization of `fmt::formatter` (the former can always be done through the latter anyway).

This has a nice property of streamlining how we format types, reducing our usage of streams and eliminating possible stray unwanted calls.

So looking at this PR:
- the first hunk seems unrelated and shouldn't result in any perf difference (or even a change is assembly)
- `gear_stats_t` should drop `to_string()` and get smth like:
```cpp
namespace fmt {
template <>
struct formatter<gear_stats_t> {
  constexpr auto parse( format_parse_context& ctx ) { return ctx.begin(); }
  format_context::iterator format( const gear_stats_t& stats, format_context& ctx );
};
} // namespace fmt
```
```cpp

fmt::format_context::iterator
fmt::formatter<gear_stats_t>::format( const gear_stats_t& stats, format_context& ctx )
{
  auto out = ctx.out();
  for ( stat_e i = STAT_STRENGTH; i < STAT_MAX; i++ )
  {
    if ( i > 0 )
      out = fmt::format_to( out, " " );
    out = fmt::format_to( out, "{}={}", util::stat_type_abbrev( i ), stats.get_stat( i ) );
  }
  return out;
}
```

NB. Catching stuff like this through profiling is going to be pretty hard as compilers can actually optimize it "out" depending on the inlining decisions and how complex the called functions are. I *think* your best bet is to slap an `__attribute__((noinline))` on `print_log()`/`print_debug()` and go from there (it *will* skew the data though).

nuoHep:
Basically smth like the top 2 commits in https://github.com/nuoHep/simc/commits/print_debug-opt (only compile tested).

scamille:
Using sim->print_debug directly is a new format, in the past nearly everything was guarded by manual checks. The print_debug function hopefully gets inlined in most cases so that it does not really make a difference.

Doing performance intensive stuff when creating the parameters is something which I think should be fixed on a case by case basis. I really like the structured logging approach we can get by adding fmt formatters, and in the few cases where that is not possible we should just add back a manual sim->debug check.

In any case, good catch, and we definitively should do something so that we don't pay extra for logging in non-debug runs.

@nuoHep shouldn't it already be possible to defer the stringification with fmt as long as the type has a std::ostream<< operator? I know that we want to replace that with fmt::formatter eventually, but in the meantime it might be good enough. So just removing the to_string() call might already solve it for gear_stats_t.

nuoHep:
> @nuoHep shouldn't it already be possible to defer the stringification with fmt as long as the type has a std::ostream<< operator? I know that we want to replace that with fmt::formatter eventually, but in the meantime it might be good enough.

Yeah, `operator<<(osteam)` works too.
> So just removing the to_string() call might already solve it for gear_stats_t. 

The problem is... it does not have an `operator<<(ostream)` :)

Mazorlion:
I like the formatter approach in this case, but I think it's unfortunately not broadly applicable to most print_debug calls as they don't just stringify an entire object (usually it's just a few fields to create a specific message).  Relying on most of these being inlined seems like kind of a weak approach and given how often some of these functions are called and the number of print_debug calls, even a slight amount of CPU can result in a big overhead.

What are your opinions on a using a macro instead of a function call? i.e. something like SIM_PRINT_DEBUG(sim, "{} thing thing {}", varargs) that does the check outside of the function call, but simplifies how it appears everywhere? I think the same pattern could be broadly applied to sim->print_log, which already has callsites that pre-check sim->log as well.

You could even go a step further and use the macro to hide the function, to protect against non-macro calls, but I'm not sure that's necessary.

I think the macro approach has the benefit of not punishing devs for wanting to ever log a message, protects against either calling functions they shouldn't or against a forgotten pre-check or formatter. All it takes is one new print_debug call to slip in (like the one in this change) for performance to take a hit.

Specific comments:
>the first hunk seems unrelated and shouldn't result in any perf difference (or even a change is assembly)

I noted this and would be happy to drop it.

And I agree `operator<<(ostream)` or the formatter approach is likely the right thing to do in this case anyways.

nuoHep:
Fwiw:
* I'm against macros for this
>I like the formatter approach in this case, but I think it's unfortunately not broadly applicable to most print_debug calls as they don't just stringify an entire object (usually it's just a few fields to create a specific message). Relying on most of these being inlined seems like kind of a weak approach 

* it does not matter if we print a full object or some of its fields, what matters is that we don't call any functions while doing so

If we are worried about missed inlining (at which point I'd question the compiler used tbh, for these functions at least) I'd advocate slapping `__forceinline`/`__attribute__((always_inline))` on them.

Mazorlion:
Apologies, I'm abusing this PR a little bit to talk about debug logging in general. I think it's worth separating this discussion: In this case I totally agree with the formatter or ostream change and I'll make that change to this PR, regardless of the rest of this discussion.

Do you have any additional thoughts about why you don't like macros in this case? It's not hard to come up with reasons why macros are ideally avoided, and I'm not particularly partial to them. However, I think there may be strong enough reasoning to fight that urge in this case. (Also I want to hear your thoughts to maybe learn something)

>it does not matter if we print a full object or some of its fields, what matters is that we don't call any functions while doing so

I agree in principle but in practice this is asking for trouble.
- There are a whole bunch of people working on simc
- There is no automated enforcement of these rules
- There are other concerns like [doing comparisons](https://github.com/simulationcraft/simc/blob/86b75080e269b0b19e13a30a0ec134098e6c6699/engine/player/azerite_data.cpp#L3595) or even [dereferencing a pointer](https://github.com/simulationcraft/simc/blob/a1534d87efe82ca561abf9eed45b81bf04ba6c17/engine/buff/sc_buff.cpp#L143)

I'd also say that the majority of these calls currently have some flavor of function call or other funkiness that has non zero cost, and even if they didn't, the current approach does nothing to prevent misuse. Given that these additional costs also apply to print_log without a precheck, the number of log calls, and the number of times the functions containing them are called... I think it's not a stretch to say there are non-trivial performance gains to be had here. 

I was trying a bit to avoid starting this conversation without first motivating it with performance numbers but as we agreed before it's not easy to get exact numbers. Even without them I think it's probably worth continuing this discussion.

scamille:
You make some very good points, and it is definitively hard, on average, to avoid any costly function call to slip in somewhere.

But I would vote against any new macros being introduced. I would rather go back to doing manual checks before doing that. Even if that potentially allows for some performance regessions.

scamille:
Oh and with all that discussion going on: I think your PR should be merged like this, since it does fix the problem just fine.

For shadowlands we probably will switch to fmt formatters at some point anyway instead of std::ostream<<, and we can then also add that for gear_stats_t.

nuoHep:
>Do you have any additional thoughts about why you don't like macros in this case? It's not hard to come up with reasons why macros are ideally avoided, and I'm not particularly partial to them. However, I think there may be strong enough reasoning to fight that urge in this case. (Also I want to hear your thoughts to maybe learn something)

Because they obfuscate and because (from my pov as mostly a class module dev) they are somewhat hostile to potential class contributors that are not always C++ developers in the first place (by potentially generating illegible compiler error messages in case of misuse). So the 'whole bunch of people' (in my view at least) is actually a reason to **not** employ macros here.

I agree that enforcing the "no-calls" policy is hard to impossible and there is no automated way to do this. But I think we should strive to make it easier to print stuff "properly" in the first place, by adding the formatting support for the types people would want to print. So that devs can write the "obvious" `print_debug("{} {} {}", val0, val1, val2)` and it would just work(tm) (instead of the current status quo where they have to find how they should print stuff on a type by type basis, for non-builtin types at least).

---
As for other concerns:
* [comparisons](https://godbolt.org/z/Kc78re)
* [dereferencing a pointer](https://godbolt.org/z/v7v938)

Modern optimizing compilers are [smart](https://godbolt.org/z/4EYoxq) :) non-trivial function calls are really the only things that force evaluation at the call site in this case. 

---
And yeah, I'm also ok merging this (apart from the first hunk) as is.

Mazorlion:
Removed unnecessary check, ready to merge if you wish.

--------------

Godbolt is awesome, had not seen it before. I played with some [other](https://godbolt.org/z/o6cv5z) [cases](https://godbolt.org/z/9fK33r) out of curiosity and on recentish gcc releases (anything sub 10 in that case) it doesn't optimize everything to after the `debug` check.

---------------
Fair point about the log messages, I hadn't really considered their impact on a project like this.

I do feel better about the current state of things given your observations (and a bit of research). However part of these calls being efficient relies on the compiler, and that seems kind of brittle given how widely the usage of the function varies and that people may be using any number of compilers. That being said, I expect even imperfect optimization wouldn't result in much of a cost for most callsites. 

What do you think about a middle of the road approach like requiring a `sim->debug` check before via an `assert`?  It would go towards the goal of preventing bad logging, while avoiding macros. The cost would be some increased boilerplate for debug logging, but there are already a decent number of callsites that precheck for both `debug` and `log`. It might also be worthwhile just standardizing it (I recognize that simc intended to move away from that approach with this function). 

Realistically, this is almost certain to happen again at some point in the future without some kind of change; while the cases might not be numerous, all it takes is one to have an impact on performance. Also without some form of monitoring on performance for simc (is there already something? - I expect it could be tricky given that different sims hit totally different paths), it could go unnoticed for a long time.

-----------------------------------
(I like your use of horizontal dividers in comments)

>I think we should strive to make it easier to print stuff "properly" in the first place, by adding the formatting support for the types people would want to print

One comment about this is that I think it goes both ways: Making it easier to properly log things without worrying about the consequences would also go towards avoiding inefficiency. Though your point about printing objects is taken.

-------------------------------------

Thanks for the discussion! I've rebuilt many of the thoughts I had around this.

scamille:
>What do you think about a middle of the road approach like requiring a sim->debug check before via an assert? It would go towards the goal of preventing bad logging, while avoiding macros. The cost would be some increased boilerplate for debug logging, but there are already a decent number of callsites that precheck for both debug and log. It might also be worthwhile just standardizing it (I recognize that simc intended to move away from that approach with this function).

Where exactly would you want to place the check, assuming no macro is introduced? sim->print_debug already has its own check in it, so a assert would do nothing there. And the whole idea of sim->print_debug having the sim->debug check inside is that it gets easier for the caller to do logging.

Old sim->debug_str.printf() call (or whatever the exact syntax was) always require manual checks, and it would be very quickly visible if a check got forgotten there, since you would see it on any normal sim.

I don't know of any way to check that no "expensive" computation is done on the caller site.

Mazorlion:
I think I was unclear; I was indeed suggesting going back to the old method of requiring callers to manually check `sim->debug` or `sim->log`. I was suggesting this as a middle ground between the code today, and using something like macros of lambdas. However, one minor potential addition to prevent stray logs could be an `assert(debug)` in the function to enforce that callers have checked it beforehand. 

The major cost here is a minor inconvenience to callers, but I'd bet that for many users it will result in some level of immediate performance improvement (relative to how good their compiler is, with potentially big gains for some users) and it confers long term protection against mistakes.

FWIW, I would be happy to go and make the changes if there is agreement that it's worthwhile.

scamille:
I could certainly live with manual checks at the caller site - that was the de facto standard for the last 10 years in SimC.

But the whole idea of the new sim->print_debug function (besides using fmt formatting) was that it would make the call sites look prettier, not requiring extra if statements and indentation. It reduces small logging statements from 4 lines of code to 1, which is quite nice.

So all in all, I personally would prefer the way it currently works, even if it does open up some potential performance pitfalls.

nuoHep:
Right... adding an assert like this would just be bad. As you would effectively replace a potential performance issue with a "correctness" one. Even worse, it would be a "hidden" api contract which is super fragile. You *really* want compiler enforcement for stuff like this.

Further, as an assert would manifest *only* in "no -NDEBUG" builds, you would still have to leave the inner check anyway for release builds (just in case a stray unguarded `print_debug()` slips through), at which point - why bother?
We also don't test all permutations in debug builds. Our travis bot tests all talent combinations, that's pretty much as far as we go atm.
So stuff can easily slip through, accumulate and then blow up on some random dev that simply had to debug something completely unrelated that *happened* to also trigger this assert.

---
For the code examples - sure, you can come up with cases where the compiler does not fully optimize it. In both of these cases gcc is just doing a bad job, especially in the 2nd case (like... it emits the `cmp` first, but then proceeds to emit a bunch of `mov`s before the `jne`, wtf?). Fyi clang does optimize both of them to the same assembly.
But here comes the "fun" part of low-level optimizations and modern x86 (or arm for that matter) wide-issue out-of-order cpus - sometimes it does not matter how bad your compiler is as long as it does a semi decent job.
The key to this is data dependencies - the flag check (`cmp`+`jne` pair in the assembly) does not depend on any of the arguments (basically everything in the function prior to this, or any `mov`s between them). This allows the cpu to freely reorder instructions, effectively moving all `mov`s to after `jne` (jump) which **is** dependent on `cmp`. That's instruction reordering.
Then there is the fact that cpus are pipelined and `mov`s are cheap (they have low issue latency and there are usually several load/store ports) - so even if the cpu does issue all of the `mov`s, as soon as it hits the `cmp`+`jne` pair it will quickly abandon all of them (if the check failed). There are obviously limits to all of this, but they are (increasingly) very high.

Remember, cpus are optimized to execute real world code. Which includes large amounts of virtual machine code that's not always jitted. Which by its nature presents the cpu with a lot of "dead" `mov`s and generally suboptimal instruction sequences.

In closing, I can almost guarantee you the difference in both of these code examples will be in the noise. I will still assert that only function calls really matter here. And in the case of this PR further compounded by the fact that std streams are **really** bad.

---
Now you can rightfully say that that's further relying on stuff out of our control. That's somewhat true. But I think we *can* rely on most of it, just as we rely on the compiler inlining stuff properly (otherwise modern C++ would be slow as hell). And ease of use here trumps possible gains from micro-optimizations *as long as* we remove the main source of potential performance pitfalls - function calls.

Mazorlion:
Sounds like I'll just shelve this discussion for now. I'll likely do a benchmark just for myself at some point to see what happens if I did change all the log statements and then try it on a few compilers (just to learn some things, and get a sense of what difference this makes - if any at all). Thanks for your feedback!

------
>  you can come up with cases where the compiler does not fully optimize it.

To be fair, all of these were pulled straight from simc, but your point about processors being smart still stands.

> you would still have to leave the inner check anyway for release builds

I realized this but figured that while yucky, it still fixes the issue at hand without being particularly costly (in that at worst it just checks twice only when `debug` is set). That being said, your assessment is fair.

PR #5183, opened 2020-07-26
Switch more stuff to fmt formatting

nuoHep:
* Add `util/format.h` - small helper to reduce boilerplate necessary to add fmtlib formatting support to a type. With this one can add a `fmt::format_context::iterator format_to( T, fmt::format_context& )` overload to make the type formattable.
* Add fmtlib formatting support to a bunch of enums that already have `util::xx_string()` functions. Makes it possible to format them directly (so `fmt::format( "{}", enum_value )` just works).
* Switch a bunch of types that had only `.to_string()` member functions to be formattable (through the `format_to()` helper)
* Switch `special_effect_t` & `item_t` to `format_to()` from `operator<<(ostream)`
* Some misc conversions to python-style formatting and formatting enums directly.

Somewhat of a draft PR to see if this makes sense.

One caveat about fully switching to fmtlib is that it produces different output for floating point compared to streams because streams seem to apply some kind of "clever" rounding. There are 2 main differences I noticed:
* streams skip the decimal part when it's zero, fmtlib doesn't; effectively we get a bunch of `.0` added everywhere
* fmtlib does "round-trip", for most stuff it does not really matter but there are some at first glance odd cases where a `0.7` gets changed to `0.7000000000000001`

Now, the way rounding works in streams is platform-specific in the first place. And I don't think we care about that here. So dunno.

### DIFF
--- a/engine/action/dbc_proc_callback.cpp
+++ b/engine/action/dbc_proc_callback.cpp
@@ -84,8 +84,8 @@ void dbc_proc_callback_t::trigger( action_t* a, void* call_data )
 
   bool triggered = roll( a );
   if ( listener->sim->debug )
-    listener->sim->out_debug.printf( "%s attempts to proc %s on %s: %d", listener->name(), effect.to_string().c_str(),
-                                     a->name(), triggered );
+    listener->sim->print_debug( "{} attempts to proc {} on {}: {:d}", listener->name(), effect, a->name(), triggered );
+
   if ( triggered )
   {
     // Detach proc execution from proc triggering
--- a/engine/action/sc_action.cpp
+++ b/engine/action/sc_action.cpp
@@ -4695,7 +4695,7 @@ void action_t::apply_affecting_effect( const spelleffect_data_t& effect )
           case P_RESOURCE_COST:
             base_costs[ resource_current ] += effect.base_value();
             sim->print_debug( "{} base resource cost for resource {} modified by {}", *this,
-                              util::resource_type_string( resource_current ), effect.base_value() );
+                              resource_current, effect.base_value() );
             break;
 
           case P_TARGET:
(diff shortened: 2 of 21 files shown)


nuoHep:
@scamille thoughts?

scamille:
Looks nice, and adding new formatters seems to be rather easy with this system in place.

The functions are friends of the class, and not freestanding. I assume that ties them more to the type that you try to format?

Formatted enums are a great thing, but we should strive to really get all/most formattable. Otherwise it will become a burden to always check if something is formattable or not. Same thing goes for all types, really.
Or does fmt already cause a compile-time error when it cannot format the type?

nuoHep:
> The functions are friends of the class, and not freestanding. I assume that ties them more to the type that you try to format?

Ah, I did them as hidden friends out of habit (we do this internally to disallow calling them directly and give access to the internals). They can easily be freestanding.

The latter actually makes this thing possible to also use for stuff like enums or types from third-party code.

> Formatted enums are a great thing, but we should strive to really get all/most formattable. Otherwise it will become a burden to always check if something is formattable or not. Same thing goes for all types, really.

fmtlib will format enums (even enum classes) as integers by default. So that's ok - if we miss some enum it will still "work".

> Or does fmt already cause a compile-time error when it cannot format the type?

Compile time error - it hits `static_assert` here: https://github.com/simulationcraft/simc/blob/shadowlands/engine/lib/fmt/core.h#L1015

scamille:
Ok great that we get a static assert on types without a formatter. That does make it a lot safer to use.

I am all for going down the fmt formatter route, and this PR certainly is a good start for that.

The query about using friend vs freestanding was mainly curiosity, but maybe we also want to have a consistent style for it.

Same probably goes for how to format a type/class in general. I added the class name for action_t and player_t std::ostream<< formatter, something like player '{}' and action '{}'. But I am not 100% happy about it, and it does differ quite a bit from some of the older debug-log statements we had.
We should at least get some rough understanding of what we want and then apply that for consistency. Maybe there are some better guidelines on how to best stringify types (for logging/debug purposes).

PR #5189, opened 2020-07-31
String split with string view

scamille:
Change util::string_split to return vector<string_view>

The motivation behind this change is that most of the times, we only inspect the split entries directly at the call site, without actually storing them or otherwise requiring them to be full std::string's.

After stumbling upon a few cases where we actually do want to store all of the split parts as string, added a new util::string_split_as_string function which has the old behaviour, so that it makes the call sites easier, just as before.

also add some util::to_int/to_unsigned/to_double functions/overloads accepting string_view. While this for new just convert back to a std::string, they might at some point offer the chance to replace this with a string view based number parsing.

@nuoHep Please take a view and leave some critique, or ideas if this can be improved in some form.


nuoHep:
Sorry, I couldn't get back to you sooner on this. Anyway, overall I'm not a huge fan tbh. At least given the direction taken here.

The 2 problems I see with it:

* Proliferation of `auto`
While I typically don't care either way, here I believe explicitly stating the types is much preferred. We do care if it's an `std::string` vs an `util::string_view` (ditto for vectors of them) because they have very different ownership semantics. And it's easy to trip over something subtle if you are not careful enough.

* Converting `string_split()` to return an `std::vector<util::string_view>`.
This is simply making a safe-to-call by default function to an unsafe one. Made worse by the fact you changed the return type of an existing function, instead of introducing a new one and carefully converting to it.

While the first can be argued as stylistic. The 2nd one is major imo.
So prior to 389631c this call here: https://github.com/simulationcraft/simc/pull/5189/commits/4c9a3f0e263d0f8d0989fbccd316674a07111ab6#diff-6af2102e2d6bdb5068d34a9503ba9258R742 was UB (either use after free or stack stomping, depending on the size of the temporary substring).
All of the calls to `util::string_split()` here: https://github.com/simulationcraft/simc/pull/5189/commits/4c9a3f0e263d0f8d0989fbccd316674a07111ab6#diff-645fc59abb86985a1d8501c01675b52f are broken for the same reason.

I could also argue that optimizing this is kinda moot - we are still doing the allocation for the vector, at that point it does not matter that much if we are also allocating strings (which we may not considering a lot of the splits are in expression parsing where most parts should fit into `std::string` sso anyway).

So if I were doing this:
* First, I'd not do autos (at least as a first step and def not for `std::string`/`std::string_view`).
* Second, I would leave `util::string_split()` returning `std::string`s by default and introduce some other way to split a string into `util::string_view`s.

---
The latter can be accomplished through different means.
The simplest would be to lift the implementation template introduced in this PR into the header and make it the primary function, defaulting `StringType` to `std::string`. So the signature would be:
```cpp
template <typename StringType = std::string>
std::vector<StringType> string_split( util::string_view s, util::string_view delim );
```
That makes a generic call like `util::string_view( blah, "." );` safe by default (if not "optimal"). That also means that all existing calls are also "safe" and can't subtly invoke UB. After that you could start auditing the calls and changing them to `util::string_split<util::string_view>( blah, "." )` where it's safe and makes sense.

The other alternative is to make `string_split()` a lazy range which is convertible to a vector but can also be iterated in a completely non-allocating fashion. It could also be made safe by construction for rvalue `std::strings`. But that requires a bunch of template mumbo-jumbo and maybe an overkill for our case.

---
Re `to_unsigned`/`to_int`/`to_double`. There are some optimizations possible and I'm not particularly fond of the `_ignore_error` naming (can't come up with something better though, so /shrug) but makes sense overall.

scamille:
Thanks for the feedback.

I mostly focused on the returned splits, and checked how they are used further down the line to make sure I do not introduce any problems there. I see that the input to the string split function needs to be checked much more carefully as well.

Overall, the occurences where string_split isn't used with a l-value are quite rare though, but I will check them.

Merging the two string split versions by making the template version public is certainly an option. Having the std::string version as the default version can help, but at the end of the day it just changes the naming of things. It doesn't really prevent me from messing up the lifetime of my source string when using the string_view version, if I choose to.

I might have gone a bit overboard on the auto usage for the actual string/string_view objects. I can see how to it makes the ownership semantic clearer. Auto let the appropriate type flow freely through the function, that's why I chose to do so.


--------------

about the str to number functions: I am hoping that we at some point can optimize this for string views, that is why I wanted to switch more stuff to using those. But as far as I understand, even now not all compilers have to_chars stdlib implementation and we are not on c++17 anyway.

I couldn't come up with a better naming for the _ignore_error functions as well, feel free to change it if you find something. The main reason I changed it like this is that I wanted to make it explicit when someone uses the 0 as a magic value to represent an invalid input.

nuoHep:
> I mostly focused on the returned splits, and checked how they are used further down the line to make sure I do not introduce any problems there. I see that the input to the string split function needs to be checked much more carefully as well.

Well, `string_view` has by design an explicitly different api compared to an `std::string` where it could introduce problems. E.g. it does not have a `.c_str()` as it can't return a nul-terminated string in the general case. Stuff like that.
So, unless there was some questionable api use to begin with, any problems should've resulted in compile-time errors. 
Accidentally introducing UB by consuming an r-value `std::string` is much more likely though and can't be caught by a compiler.

> Merging the two string split versions by making the template version public is certainly an option. Having the std::string version as the default version can help, but at the end of the day it just changes the naming of things.

But naming matters. What experience taught me is that apis should be easy to use hard to misuse. What this usually means for naming is, unless you can make a single api (e.g. function) always safe, you want to reserve the short & concise "go to" name to the obviously safest thing (if not the most optimal). Any other variant that requires some thought before use should have a more verbose name.

> It doesn't really prevent me from messing up the lifetime of my source string when using the string_view version, if I choose to.

Right, it doesn't. What it does is that it makes the "default" thing (`util::string_split()`) always safe, and you have to *explicitly* opt-in into the potentially unsafe variant by using the more verbose `util::string_split<util::string_view>()`.

You also basically followed this "rule" here by introducing "unchecked" "noexcept" variants of `to_xxx` functions under different, more verbose names.

> about the str to number functions: I am hoping that we at some point can optimize this for string views, that is why I wanted to switch more stuff to using those. But as far as I understand, even now not all compilers have to_chars stdlib implementation and we are not on c++17 anyway.

Right. `<charconv>` is a bit of a failure tbh, availability wise.

scamille:
I guess the main open question/task is to resolve the string split API now.

Of course you do make good points, and it makes sense to have a default safe string_split. The only argument against that is that people will likely use that 100% of the time instead of the string_view variant, even if there is no reason to do so. But the argument about hard to misuse default API is certainly stronger.

So I guess I will go with the public templated string split, default to std::string, rename all util::string_split to util::string_split<<util::string_view>> and rename util::string_split_as_string back to util::string_split.

Ok?

nuoHep:
Yeah, seems good to me. I may look at it a bit later to make it more resilient to misuse.

scamille:
About UB string_split behaviour:
Of course the https://github.com/simulationcraft/simc/commit/4c9a3f0e263d0f8d0989fbccd316674a07111ab6#diff-645fc59abb86985a1d8501c01675b52f usages are UB, since they use string view into a temporary that goes out of scope.

You also linked https://github.com/simulationcraft/simc/commit/4c9a3f0e263d0f8d0989fbccd316674a07111ab6#diff-6af2102e2d6bdb5068d34a9503ba9258R742 . But wouldn't that be ok, since a const std::string& is valid even for a temporary during the duration of a function call? There was some special rule about reference to temporary lifetimes.

nuoHep:
> You also linked [4c9a3f0#diff-6af2102e2d6bdb5068d34a9503ba9258R742](https://github.com/simulationcraft/simc/commit/4c9a3f0e263d0f8d0989fbccd316674a07111ab6#diff-6af2102e2d6bdb5068d34a9503ba9258R742) . But wouldn't that be ok, since a const std::string& is valid even for a temporary during the duration of a function call? There was some special rule about reference to temporary lifetimes.

`std::string::substr()` returns a temporary (prvalue).

As to reference lifetime extension - it's a **very** subtle thing that works in basically one specific case; if you are interested - [this](https://abseil.io/tips/107) is a nice writeup.

PR #5194, opened 2020-08-03
Drop the action name tokenization. It is unlikely that we get any unt…

scamille:
Drop the action name tokenization. It is unlikely that we get any untokenized action names in the APL since it does not support whitespace anyway without escaping. (eg. writing action+=/"Shadow Bolt" )
At least for CI.simc, there is no problem with any action name.

Also change html action priority list to use the actions signature_str, which includes the final input used to create the action, including any modifications by modify_action. Again, not sure if anyone is using modify_action= at all, but if they do you can now see the effective action string in the html report.

### DIFF
--- a/engine/player/sc_player.cpp
+++ b/engine/player/sc_player.cpp
@@ -2692,19 +2692,20 @@ void player_t::create_actions()
 
   init_action_list();  // virtual function which creates the action list string
 
-  std::string modify_action_options;
+  util::string_view modify_action_name = modify_action;
+  util::string_view modify_action_options;
 
   if ( !modify_action.empty() )
   {
-    std::string::size_type cut_pt = modify_action.find( ',' );
+    util::string_view modify_action_view = modify_action;
+    util::string_view::size_type cut_pt = modify_action_view.find( ',' );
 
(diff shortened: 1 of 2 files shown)


nuoHep:
The code looks sane but I wonder why do we need/want it? I guess to minimize temporary string allocations, but do we care that much here?

One problem I see with that is that `tokenize()` also does lowering, and all `player_t::create_action()` implementations do a straightforward `==` on the action name. So while we can't have spaces in apls (yet?), it will break stuff like `actions+=/Shadow_Bolt` which I think works now.

scamille:
Yes you are correct, it would reduce the freedom of what you can enter a little bit.

But I haven't seen anyone do that ever, and I think it is reasonable to just expect a fully tokenized name.

Expressions also expects quite a bit of extra matching and does not do case insensitive compares for everything.

Just to emphasize: I wanted to discuss if anybody thinks there is a benefit to tokenizing all names, since I don't see much. The changes are not really that beneficial, I jus wanted to see if I can use string view through the whole input call stack. If we want to keep it, I will just scrap this PR.

scamille:
I'll discard this.

PR #5202, opened 2020-08-10
Checking what broke during the refactor, to prepare updating the refactor branch

Bloodmallet:


Bloodmallet:
Looks like only Ember Elemental and Spark Elemental somehow create merge conflicts.

scamille:
Is it possible for you to rebase the shaman changes on top of shadowlands instead of merging it, or does that cause too much trouble?

Bloodmallet:
For me this was just a way to see what might need changes. Your question needs the attention of @Rusah

scamille:
If shadowlands-shaman is to be considered a public branch where more than one people are working on, and you do not want to alter the history anyway, then merging is just fine.

I thought of it as more of a private branch, where rebasing might be an option and would give a cleaner change history when it eventually gets merged back to shadowlands. Both options have their pro and cons.

PR #5203, opened 2020-08-13
Merge Shadow Priest Changes

Saeldur:
Fix @seanpeters86 initial work on mind blast charges with voidform
Implement an initial working version of the Dark Thoughts passive including some tracking
Allow mindblast to be instant cast during the dark thoughts passive, but only during mind flay or sear. Thank you @vituscze for the assistance.
Slight APL adjustments to accommodate the passive

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -79,8 +79,6 @@ public:
                                    ->effectN( 1 )
                                    .resource( RESOURCE_INSANITY ) )
   {
-	bool cwc = false;
-	add_option(opt_bool("cwc", cwc));
     parse_options( options_str );
 
     energize_amount *= 1 + priest().talents.fortress_of_the_mind->effectN( 2 ).percent();
@@ -93,12 +91,7 @@ public:
     }
 
     cooldown->hasted = true;
-	if (cwc)
-		usable_while_casting = true;
-	/*usable_while_casting = true;
-	use_while_casting = true;
-	cooldown->add_execute_type(execute_type::CAST_WHILE_CASTING);
-	internal_cooldown->add_execute_type(execute_type::CAST_WHILE_CASTING);*/
+    usable_while_casting = use_while_casting;
 
     // Reduces CD of Mind Blast but not SW:V
     apply_affecting_aura( player.find_rank_spell( "Mind Blast", "Rank 2", PRIEST_SHADOW ) );
@@ -106,11 +99,11 @@ public:
 
   bool usable_during_current_cast() const override
   {
-	  if (!priest_spell_t::usable_during_current_cast()) return false;
-	  if (!priest().buffs.dark_thoughts->check()) return false;
-	  if (player->channeling == nullptr) return false;
-	  if (player->channeling->name() == "mind_flay" || player->channeling->name() == "mind_sear") return true;
-	  return false;
+    if (!priest_spell_t::usable_during_current_cast()) return false;
+    if (!priest().buffs.dark_thoughts->check()) return false;
+    if (player->channeling == nullptr) return false;
+    if (player->channeling->name_str == "mind_flay" || player->channeling->name_str == "mind_sear") return true;
+    return false;
   }
 
   void schedule_execute( action_state_t* s ) override
@@ -216,10 +209,10 @@ public:
     //   );
     // }
 
-	if (priest().buffs.dark_thoughts->check()) 
-		priest().buffs.dark_thoughts->decrement();
-	else 
-		priest_spell_t::update_ready(cd_duration);
+  if (priest().buffs.dark_thoughts->check()) 
+    priest().buffs.dark_thoughts->decrement();
+  else 
+    priest_spell_t::update_ready(cd_duration);
   }
 };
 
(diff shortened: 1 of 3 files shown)


scamille:
Does dark thoughts have 2% independent proc chance for each dot up, or are they cumulative as in your code?

Otherwise looks good. Would be nice if you could run clang-format on those changes, but not strictly necessary.

The usable while casting expression can be a bit more optimized for performance later, but that's not that important now.

Saeldur:
It's currently unknown precisely how dark thoughts works, it would logically be cumulative for each dot up per target but there hasn't been enough verification on that. This is kind of just an initial implementation to see how it works/if it works properly to be easier to change later after more information is known. That's why I threw in a bunch of TODOs around some of the proc.
The usable while casting definitely needs to be optimized so that it can return false early as possible but this was just an initial implementation.
Sorry for the messed up formatting, my vstudio had issues with having clang-format run automatically so I had to disable it (Most of the messed up format is from vstudio trying to automatically format stuff for me).

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -316,6 +348,24 @@ struct mind_flay_t final : public priest_spell_t
...
+    dot_t* dp       = td.dots.devouring_plague;
+
+    int dots = swp->is_ticking() + vt->is_ticking() + dp->is_ticking();
+
+    /// TODO: Confirm is procrate is matching ingame
+    /// TODO: Confirm if this needs to be adjusted to be on its own PRNG system or has an ICD
+
+    if ( rng().roll( priest().specs.dark_thoughts->effectN( 1 ).percent() * dots ) )
+    {
+      // sim->print_log("{} rolled DT Proc success from flay with {} chance with {} dots", *player,
```
seanpeters86: can you swap this to just be a debug log? that way we can leave it in uncommented and still see it when debug is on

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1553,6 +1614,38 @@ struct vampiric_embrace_t final : public priest_buff_t<buff_t>
   }
 };
 
+struct dark_thoughts_t final : public priest_buff_t<buff_t>
+{
+  dark_thoughts_t( priest_t& p ) : base_t( p, "dark_thoughts", p.find_spell( 341205 ) )
+  {
+    this->set_max_stack( 5 );
```
seanpeters86, line 1629: is this not in spelldata anywhere?

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1553,6 +1614,38 @@ struct vampiric_embrace_t final : public priest_buff_t<buff_t>
...
+    this->set_duration( timespan_t::from_seconds( 6 ) );
```
seanpeters86, line 1630: same as above, this should either be in spell data or duration should be pulled automagically

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -2163,27 +2262,46 @@ void priest_t::generate_apl_shadow()
...
-                      "TODO see if this is worth even without Painbreaker Psalm" );
+                    "if=(!runeforge.painbreaker_psalm.equipped|(runeforge.painbreaker_psalm.equipped&variable.dots_up))"
+                    "&target.health.pct>30",
+                    "TODO see if this is worth even without Painbreaker Psalm" );
   main->add_talent( this, "Shadow Crash", "if=raid_event.adds.in>5&raid_event.adds.duration<20",
-                      "Use Shadow Crash on CD unless there are adds incoming." );
+                    "Use Shadow Crash on CD unless there are adds incoming." );
+  main->add_action(
+      this, "Mind Sear",
+      "target_if=spell_targets.mind_sear>1&buff.dark_thoughts.up,chain=1,interrupt_immediate=1,interrupt_if=ticks>=2" );
```
seanpeters86: Can we add comments to these more complex lines to the APL?

```suggestion
      "target_if=spell_targets.mind_sear>1&buff.dark_thoughts.up,chain=1,interrupt_immediate=1,interrupt_if=ticks>=2", "Use Mind Sear to consume any Dark Thought procs on AoE" );
```

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -2163,27 +2262,46 @@ void priest_t::generate_apl_shadow()
...
+  main->add_action( this, "Mind Flay",
+                    "if=buff.dark_thoughts.up&variable.dots_up,chain=1,interrupt_immediate=1,interrupt_if=ticks>=2&"
+                    "cooldown.void_bolt.up" );
```
seanpeters86: same as above

seanpeters86 (changes requested):

engine/class_modules/priest/sc_priest_shadow.cpp
Saeldur: Testing without manually setting these does not appear to be getting it from the spelldata, it appears to not currently be in there or I'm missing something.

Updated the p.find_spell to instead get by name however in https://github.com/simulationcraft/simc/pull/5203/commits/cddecfc83431b3ddd487cec96d21750b5d466fed

engine/class_modules/priest/sc_priest_shadow.cpp
Saeldur: Added to https://github.com/simulationcraft/simc/pull/5203/commits/cddecfc83431b3ddd487cec96d21750b5d466fed alongside of one for mind sear.

engine/class_modules/priest/sc_priest_shadow.cpp
Saeldur: Added with a TODO statement in https://github.com/simulationcraft/simc/pull/5203/commits/cddecfc83431b3ddd487cec96d21750b5d466fed

engine/class_modules/priest/sc_priest_shadow.cpp
Saeldur: Added with a TODO statement in https://github.com/simulationcraft/simc/pull/5203/commits/cddecfc83431b3ddd487cec96d21750b5d466fed

Saeldur:
Updated with all of the suggestions

engine/class_modules/priest/sc_priest_shadow.cpp
Saeldur: Answered in next conversation

seanpeters86 (approved):

PR #5212, opened 2020-08-25
[Frost DK] Rank 2 Spells

Khazakdk:
This includes all the ranks listed in the spreadsheet pinned in Discord, and does include rank 2 of the newly added Might of the Frozen Wastes.  This is my first PR so I expect its not perfect.  I've left comments in places I'm not sure about.


engine/class_modules/sc_death_knight.cpp
```diff
@@ -5297,6 +5311,11 @@ struct obliterate_strike_t : public death_knight_melee_attack_t
     {
       p() -> buffs.icy_citadel_builder -> trigger();
     }
+
+    if ( p()->buffs.killing_machine->up() )
+    {
+      school = SCHOOL_FROST;
+    }
```
Khazakdk: `apply_affecting_aura` seems to be the new standard for rank 2 passives based on what I saw in other class files but that didn't seem to do the trick here.  The spell data for rank 2 seems to be more or less empty outside of the name so this made sense to me and I resorted to manually setting the school.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5503,6 +5522,18 @@ struct pillar_of_frost_bonus_buff_t : public buff_t
...
+{
+  pillar_of_frost_buff_rank_2_t(death_knight_t* p) :
+    buff_t(p, "pillar_of_frost_rank_2")
+  {
+    set_duration( p -> spec.pillar_of_frost -> duration() );
+    set_default_value( p -> spec.pillar_of_frost_rank_2 -> effectN( 1 ).percent() );
+
+    add_invalidate( CACHE_STRENGTH );
+  }
+};
```
Khazakdk: Another case where `apply_affecting_aura` seemed to not work.  Doing it this way also offers the benefit of explicitly listing the rank 2 in the buffs section of the output, which makes it clear to the user that rank 2 is present as the base Pillar of Frost has the rank 1 tooltip.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -7498,14 +7538,14 @@ void death_knight_t::default_apl_dps_precombat()
 std::string death_knight_t::default_potion() const
 {
   std::string frost_potion = ( true_level > 110 ) ? "potion_of_unbridled_fury" :
-	                           ( true_level > 100 ) ? "prolonged_power" :
+                             ( true_level > 100 ) ? "prolonged_power" :
```
Khazakdk: Set VS from Mixed to 2 Spaces

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5117,6 +5130,7 @@ struct howling_blast_t : public death_knight_spell_t
     if ( p() -> buffs.rime -> up() )
     {
       m *= 1.0 + p() -> buffs.rime -> data().effectN( 2 ).percent();
+      m *= 1.0 + p() -> spec.rime_rank_2 -> effectN( 1 ).percent();
```
Khazakdk: Had to manually implement the rank as `apply_affecting_aura` didn't seem to buff Howling Blast like it should.  My guess at this point was that `apply_affecting_aura` is meant for direct spells and that I should manually implement rank 2s that affect buffs.

engine/class_modules/sc_death_knight.cpp
scamille: You are setting the school to frost without setting it back to what it was before. And I am not sure how well supported adjusting school during combat is anyway.

It might also be just a different spell or additional damage as frost. But I am no DK and don't know how this exactly works in-game.

Khazakdk: Fixed it not setting back.  The way in game works right now is it just changes the school for the hit to frost.  Looking at logs, in game, and poking about spell data there's no separate spell involved for the frost damage other than [Killing Machine Rank 2](https://shadowlands.wowhead.com/spell=317214/killing-machine).  Its the largely the same SPELL_DAMAGE event in the raw log, just with a different school.  Confirmed that adjusting the school works by disabling all the other rank 2s and got the expected dps increase.

engine/class_modules/sc_death_knight.cpp
Melekus: According to spelldata, pillar of frost rank 2 only increases the amount of strength on the buff.
You shouldn't need a buff for that, just a spelldata object and a modifier on the default value of pillar of frost's buff.
I don't think adding yet another pillar of frost buff helps visibility, especially when the user can check the html report for pillar of frost's buff value, which should be either 0.10 or 0.20, depending on spell ranks.

I'd rather keep apply_affecting_aura only for wide range whitelisted effects. I think it's better to implement rank 2 spells manually to keep the code clear.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -7275,14 +7312,21 @@ void death_knight_t::init_spells()
...
+  spec.rime_rank_2         = find_specialization_spell( 316838 );
   spec.runic_empowerment   = find_specialization_spell( "Runic Empowerment" );
   spec.killing_machine     = find_specialization_spell( "Killing Machine" );
+  spec.killing_machine_rank_2     = find_specialization_spell( 317214 );
   spec.empower_rune_weapon = find_specialization_spell( "Empower Rune Weapon" );
+  spec.empower_rune_weapon_rank_2 = find_specialization_spell( 317230 );
   spec.pillar_of_frost     = find_specialization_spell( "Pillar of Frost" );
+  spec.pillar_of_frost_rank_2    = find_specialization_spell( 316849 );
   spec.remorseless_winter  = find_specialization_spell( "Remorseless Winter" );
+  spec.remorseless_winter_rank_2 = find_specialization_spell( 316794 );
```
Melekus, line 7559: You should be able to use find specialization_spell( name, token) to get spell ranks without looking for specific IDs.

example used in unholy: ``  spec.raise_dead_2        = find_specialization_spell( "Raise Dead", "Rank 2" );``

engine/class_modules/sc_death_knight.cpp
Khazakdk: Oh that would be a lot more straight forward.  To do that I would just modify the call to `set_default_value` in the pillar of frost buff to also add the rank 2 value?

engine/class_modules/sc_death_knight.cpp
Melekus: This part is missing a check for Killing Machine's rank 2 on the player, and with this order of operation, only the obliterate after the one on which you spend KIlling Machine will be affected by the school change.
You should change the school to frost before the ``death_knight_melee_attack_t::execute();`` call, and set it back to physical after it.

engine/class_modules/sc_death_knight.cpp
Melekus: Rime's ranks 1 and 2 are additive, you should put both modifiers on the same line to avoid multiplying them.

engine/class_modules/sc_death_knight.cpp
Khazakdk: Fixed and confirmed in the debug log that the frost damage is on the right cast.  Not 100% sure I got the KM Rank 2 existing on the player check right so that should be looked at.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4450,6 +4457,7 @@ struct empower_rune_weapon_t : public death_knight_spell_t
     dot_duration = base_tick_time = 0_ms;
 
     cooldown -> duration *= 1.0 + p -> vision_of_perfection_minor_cdr;
+    cooldown -> adjust( timespan_t::from_seconds( abs( p -> spec.empower_rune_weapon_rank_2->effectN(1).base_value() / 1000 ) ) );
```
Melekus: adjust() dynamically adjusts the remaining duration on a cooldown object, the only thing you need here is to change cooldown -> duration like the line above.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4450,6 +4457,7 @@ struct empower_rune_weapon_t : public death_knight_spell_t
     dot_duration = base_tick_time = 0_ms;
 
     cooldown -> duration *= 1.0 + p -> vision_of_perfection_minor_cdr;
+    cooldown -> duration -= timespan_t::from_seconds( abs( p -> spec.empower_rune_weapon_rank_2->effectN(1).base_value() / 1000 ) );
```
Melekus: You can save a few operations by replacing ``.base_value() / 1000`` with ``.time_value()`` and simply adding that value to cooldown -> duration, instead of substracting the absolute value.

Khazakdk: Oh that's much more straightforward

PR #5213, opened 2020-08-26
Wrathful faerie

seanpeters86:
Implements the Wrathful Faerie part of the Night Fae Priest Covenant ability: [Fae Guardians](https://shadowlands.wowhead.com/spell=327661/fae-guardians).

```
Wrathful Faerie: Direct attacks against the target restore 0.5% Mana or 3 Insanity. Follows your Shadow Word: Pain.
```

So what I do here is the following:
1. If Fae Guardians is active when casting SW:P we apply `wrathful_faerie` to our target. Before this we make sure to call `remove_wrathful_faerie()` to clear out any existing debuffs. We do this because this can only be active on a single target
2. By overriding the default `impact()` in `priest_spell_t` we can check for if the actor is Shadow spec, if the impact action is a `DMG_DIRECT`, and that it hit for more than 0. We do this to avoid hard-coding which spells can proc this and which ones can't, since DoT damage/Mind Flay doesn't proc it since it doesn't have direct damage. If all of these conditions are true we `trigger_wrathful_faerie` to call our `wrathful_faerie_t` spell
3. The `wrathful_faerie_t` spell is just here to give the actor insanity, and make sure the ICD of 1.5s is respected
4. Once Fae Guardians expires we call `remove_wrathful_faerie()` again to eliminate the debuff

Unfortunately all of these spell IDs need to be hard coded since they are only in the tooltip string for Fae Guardians.

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -1654,6 +1654,8 @@ void priest_t::init_background_actions()
 {
   action.ascended_eruption = new actions::spells::ascended_eruption_t( *this );
 
+  active_spells.wrathful_faerie = new actions::spells::wrathful_faerie_t( *this );
+
   init_background_actions_shadow();
 }
 
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -364,6 +364,17 @@ struct wrathful_faerie_t final : public priest_spell_t
     energize_type     = action_energize::ON_HIT;
     energize_resource = RESOURCE_INSANITY;
     energize_amount   = insanity_gain;
+
+    cooldown->duration = data().internal_cooldown();
+  }
+
+  void trigger()
+  {
+    if ( priest().cooldowns.wrathful_faerie->is_ready() )
+    {
+      execute();
+      priest().cooldowns.wrathful_faerie->start();
+    }
   }
 };
 
@@ -1113,6 +1124,7 @@ priest_t::priest_t( sim_t* sim, util::string_view name, race_e r )
 /** Construct priest cooldowns */
 void priest_t::create_cooldowns()
 {
+  cooldowns.wrathful_faerie    = get_cooldown( "wrathful_faerie" );
   cooldowns.holy_fire          = get_cooldown( "holy_fire" );
   cooldowns.holy_word_serenity = get_cooldown( "holy_word_serenity" );
   cooldowns.void_bolt          = get_cooldown( "void_bolt" );
@@ -1509,7 +1521,7 @@ void priest_t::trigger_lucid_dreams( double cost )
 
 void priest_t::trigger_wrathful_faerie()
 {
-  active_spells.wrathful_faerie->execute();
+  active_spells.wrathful_faerie->trigger();
 }
 
 void priest_t::init_base_stats()
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -1507,6 +1507,11 @@ void priest_t::trigger_lucid_dreams( double cost )
   }
 }
 
+void priest_t::trigger_wrathful_faerie()
+{
+  active_spells.wrathful_faerie->execute();
+}
+
 void priest_t::init_base_stats()
 {
   base_t::init_base_stats();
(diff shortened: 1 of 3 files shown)


engine/class_modules/priest/sc_priest.hpp
```diff
@@ -1322,6 +1327,15 @@ struct priest_spell_t : public priest_action_t<spell_t>
       {
         priest().buffs.twist_of_fate->trigger();
       }
+
+      if ( priest().specialization() == PRIEST_SHADOW && s->result_type == result_amount_type::DMG_DIRECT && s->result_amount > 0 )
+      {
+        const priest_td_t* td = find_td( s->target );
+        if ( td && td->buffs.wrathful_faerie->check() )
+        {
+          priest().trigger_wrathful_faerie();
```
seanpeters86, line 1339: this is still causing the sim to crash atm need to figure out why

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -353,6 +353,20 @@ struct fae_guardians_t final : public priest_spell_t
...
+{
+  double insanity_gain;
+
+  wrathful_faerie_t( priest_t& p )
+    : priest_spell_t( "wrathful_faerie", p, p.find_spell( 342132 ) ),
+      insanity_gain( p.find_spell( 327703 )->effectN( 2 ).resource( RESOURCE_INSANITY ) )
+  {
+    energize_type     = action_energize::ON_HIT;
+    energize_resource = RESOURCE_INSANITY;
+    energize_amount   = insanity_gain;
```
seanpeters86, line 366: for some reason this is not respecting the ICD of the action, i suspect its because im calling the action with `active_spells.wrathful_faerie->execute();` which may be circumventing the ICD? but im not sure.

```
Name             : Wrathful Faerie (desc=Night Fae) (id=342132) [Spell Family (6)] 
Class            : Priest
School           : Nature
Spell Type       : None
Range            : 100 yards
Max Targets      : 1
Velocity         : 60 yards/sec
Duration         : 20 seconds
Internal Cooldown: 1.5 seconds
Proc Chance      : 100%
Proc Flags       : ........ .....x.. .x...... ........ 
                 : Generic Hostile Spell Taken, Magic Hostile Spell Taken
```

You can see that even though spelldata has a 1.5s ICD the sim doesn't respect it:

```
9.958 Base gains 3.00 (3.00) insanity from wrathful_faerie (70.00/100.00)
10.924 Base gains 3.00 (3.00) insanity from wrathful_faerie (77.00/100.00)
```

Trying to set it manually also gives the same result:

```
internal_cooldown->duration = timespan_t::from_seconds( 1.5 );
```

seanpeters86: @scamille any ideas here?

scamille: Afaik background actions don't get any cooldown restrictions.
you could provide custom cooldown activation by overriding update_ready.
Or maybe something completely different than a normal action, with hand-crafted cooldown and generate insanity code.

scamille:
About 2: an alternative might be adding it to assess_damage or some of those assessor callbacks. That way it would affect all actions instead of only priest_action_t ones. I dunno if that will ever be the case, but it might be.

What happens in game when SW:D drops off before guardian expires?

lgkern:
Nothing, SWP is just a "pointer" to the fairy.

scamille:
Another question: If you activate Guardian, and never recast SW:P, you won't benefit from wrathful_faerie at all?

seanpeters86:
That's correct, you need to cast SW:P once after activating Fae Guardians to get any benefit

seanpeters86:
> About 2: an alternative might be adding it to assess_damage or some of those assessor callbacks. That way it would affect all actions instead of only priest_action_t ones. I dunno if that will ever be the case, but it might be.

The only reason i didn't do this is i didn't want things like trinkets to proc this, as far as i can tell it should only be priest abilities

engine/class_modules/priest/sc_priest.cpp
seanpeters86: I took a bit of a mix from suggestions from you and Ansh here and came up with this:

Set the cd to the ICD:
```
cooldown->duration = data().internal_cooldown();
```

Create a custom trigger function that we can start/check the CD:
```
void trigger()
  {
    if ( priest().cooldowns.wrathful_faerie->is_ready() )
    {
      execute();
      priest().cooldowns.wrathful_faerie->start();
    }
  }
```

Looks good: (not exactly 1.5s because of the GCD on casting spells)
```
5.131 Base gains 3.00 (3.00) insanity from wrathful_faerie (29.00/100.00)
7.065 Base gains 3.00 (3.00) insanity from wrathful_faerie (40.00/100.00)
```

seanpeters86:
@lgkern @scamille LMK what you guys think of it, if we are good with it i can merge

scamille:
Yeah that looks fine. You could even add a "icd wasted" proc, if that provides any useful information.

PR #5214, opened 2020-08-26
[Frost DK] 2H weapon support for Obliterate and Frost Strike

Armadk:


Melekus:
Looks fine, but these spells will also require adding the new strike actions specific to the two-handed versions of Obliterate and Frost Strike (respectively IDs 325461 and 325464).

Armadk:
> Looks fine, but these spells will also require adding the new strike actions specific to the two-handed versions of Obliterate and Frost Strike (respectively IDs 325461 and 325464).

Sorry I haven't contributed since Wrath, and the core has changed enough that it's almost unrecognizable from then, so I'm learning again from scratch.

I'm not entirely sure what you mean here?  I was under the impression that in the line:

mh = new obliterate_strike_t( p, "obliterate_2h", &( p -> main_hand_weapon ), data().effectN( 4 ).trigger() );

the data.effectN( 4 ).trigger() portion would take care of that?  Is there some other section that I also need to do something with these two id's?  I couldn't find anything that matched the existing Id's like 222024, and 66198 for obliterate for instance.

Melekus:
You are correct, I forgot that they were set in the obliterate_t constructor and read your changes too quickly 👍

PR #5215, opened 2020-08-26
[DBC] try to properly handle hotfixed entries 

nuoHep:
We currently have a problem where we don't filter out duplicate entries
from the parent & reference caches. This can happen at least for hotfixed
entries.

In the DataGenerators this manifests in "duplicate" entries being
returned for `children()` and `child_refs()` (first for the "base" db2 record,
second for the hotfixed record) and `child()` returning the non-hotfixed
record (and complaining that there is more than one).

Fix that up by storing the references in a dict and converting them to a
tuple on first access.

---
As can be seen in the diff we have 2 hotfixed power entries that get "duplicated". This catches a lot more when trying to convert `SpellDataGenerator` off `get_link()`.

---
@navv1234 
Not sure if it's the best way to fix this. What's here is a bit fragile in case we ever add a record after accessing references/children. Currently this is fine as DBCDB seems to be immutable after init, so we are fine. Dunno.

### DIFF
--- a/dbc_extract3/dbc/db.py
+++ b/dbc_extract3/dbc/db.py
@@ -1,4 +1,5 @@
 import os, logging, types, sys
+from collections import defaultdict
 
 import dbc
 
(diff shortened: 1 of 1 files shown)


navv1234:
I at least never had any plans to make DBCDB anything else than a read-only store of run-time parsed records. And the intent definitely was and is that hotfixed records overwrite base-DB2 file record if they exist.

EDIT: I'm not sure what other things could cause duplicate records in the files with the system we have in place.

PR #5216, opened 2020-08-27
Change most remaining ostream operator<< to fmt::format_to

scamille:
Use util/format.hpp for easier implementation.


engine/util/timespan.cpp
```diff
@@ -3,15 +3,13 @@
 // Send questions to natehieter@gmail.com
 // ==========================================================================
 
-#include "sc_timespan.hpp"
-
 #include "fmt/format.h"
-#include "fmt/ostream.h"
+#include "sc_timespan.hpp"
```
nuoHep: Leave the order as-is?

nuoHep (approved):
LGTM overall

engine/util/timespan.cpp
scamille: that was clang-format, because the files have different names. Will fix.

scamille:
I thought it would be a good time to move sc_timespan.hpp :-)

PR #5220, opened 2020-08-31
[Covenant] Initial implementation of soulbind (Grove Invigoration)

gastank:
* Registered & initialized via existing special effect subsystems
* `cast_callback` added to covenant_state_t to handle buffs that trigger when covenant class ability is cast
    * Covenant class ability is determined via __covenant_ability_data dbc table
    * Only simple self-buffs are supported for now
    * Add new callbacks via `covenant::add_covenant_callback_buff( buff_t* )`

* Grove Invigoration (night_fae/niya) initial implementation
    * Buff named 'redirected_anima_stacks' and 'redirected_anima'
    * `buff.redirected_anima_stacks.X` expression universally accessible without soulbind


engine/player/covenant.hpp
```diff
@@ -144,8 +147,33 @@ class covenant_state_t
...
   /// Register player-scope options
   void register_options( player_t* player );
+
+  /// Returns the spell_id of the covenant class ability or the generic ability as determined via the
+  /// __covenant_ability_data dbc table. Will need to be converted into a vector if multiple class abilities are added
+  /// in later.
+  unsigned get_covenant_ability_spell_id( bool generic = false ) const;
+
+  /// List of soulbind spells
+  std::vector<unsigned> soulbind_spells() const { return m_soulbinds; }
```
navv1234: You can return a const reference to the vector here, no need to make a copy.

engine/player/covenant.cpp
```diff
@@ -499,4 +517,123 @@ report::sc_html_stream& covenant_state_t::generate_report( report::sc_html_strea
...
+    if ( !spell->ok() )
+      continue;
+
+    special_effect_t effect { player };
+    effect.type = SPECIAL_EFFECT_EQUIP;
+    effect.source = SPECIAL_EFFECT_SOURCE_SOULBIND;
+
+    unique_gear::initialize_special_effect( effect, soulbind_spell );
+
+    player->special_effects.push_back( new special_effect_t( effect ) );
```
navv1234, line 543: It might be wise here to ensure that the soulbind special effect is a custom one. We have had issues where the "auto infer" special effect system has tried to be clever and completely messed up for example Legion legendary spells.

So, adding a if ( !effect.is_custom() ) { continue; } here  before we push the special effect copy to player would probably be a good idea.

engine/player/covenant.cpp
```diff
@@ -449,28 +451,44 @@ void covenant_state_t::register_options( player_t* player )
           this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3 ) ) );
 }
 
-report::sc_html_stream& covenant_state_t::generate_report( report::sc_html_stream& root ) const
+unsigned covenant_state_t::get_covenant_ability_spell_id( bool generic ) const
```
scamille, line 455: You could just return the found spell data here instead of the id, since you seem to be using spell data later on anyway.

engine/player/covenant.cpp
```diff
@@ -499,4 +517,123 @@ report::sc_html_stream& covenant_state_t::generate_report( report::sc_html_strea
...
+    player->special_effects.push_back( new special_effect_t( effect ) );
+  }
+}
+
+void register_soulbinds()
+{
+  unique_gear::register_special_effect( 322721, soulbinds::grove_invigoration );
+}
+
+struct covenant_ability_cast_cb_t : public dbc_proc_callback_t
```
scamille, line 616: I think the class here needs to somehow have the "buff" part in the name as well, and the player->covenant->cast_callback variable as well.

It could also make sense to have another class between dbc_proc_callback_t and this class which only adds the part of setting up a convenant ability callback and filtering, without the buff_list in it.

That way it is also easy to set up other covenant ability callbacks which do arbitrary stuff, separated from the players single covenant ability buff triggering callback.

gastank: I was thinking since procs are either buffs or actions, if we need action procs in the future, we can add an action_list and also iterate over that in the callback trigger, and just keep everything within a single callback.

Can also set up a base class for the elements of the list vector and have subclasses do a trigger() or schedule_execute() or other as appropriate when the list is iterated on.

scamille: It certainly.simplifies the usage. You could also add both buff and action callbacks into the same callback.

PR #5221, opened 2020-09-01
[Frost DK] Might of the frozen wastes rank 1 + 2

Armadk:
Cherry picked two commits from bfa-dev, as it looks like they are still relevant, at least so far in SL.

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -415,6 +415,7 @@ public:
   // Counters
   double eternal_rune_weapon_counter;
   bool triggered_frozen_tempest;
+  unsigned int km_proc_attempts;
 
   // Enemies affected by festering wounds
   unsigned int festering_wounds_target_count;
@@ -848,6 +849,7 @@ public:
   {
     double lucid_dreams_minor_proc_chance = 0.15;
     bool disable_aotd = false;
+    bool killing_machine_rppm = false;
   } options;
 
   // Runes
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_death_knight.cpp
```diff
@@ -856,6 +859,7 @@ struct death_knight_t : public player_t {
   {
     double lucid_dreams_minor_proc_chance = 0.15;
     bool disable_aotd = false;
+    bool killing_machine_rppm = false;
```
Melekus: I don't think the rppm killing machine options and override are necessary going into shadowlands, it was mostly just a testing tool to compare how off it was compared to the new formula.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -6925,14 +6944,33 @@ unsigned death_knight_t::replenish_rune( unsigned n, gain_t* gain )
 }
 
 // Helper function to trigger Killing Machine, whether it's from a "forced" proc, a % chance, or the regular rppm proc
-void death_knight_t::trigger_killing_machine( double chance, proc_t* proc, proc_t* wasted_proc )
+void death_knight_t::trigger_killing_machine( double chance, proc_t* proc, proc_t* wasted_proc, weapon_e weapon_type )
```
Melekus: You don't need to pass the player's weapon type from the triggering actions to death_knight_t::trigger_killing_machine(), it should be accessible just fine within it.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5394,7 +5404,15 @@ struct obliterate_strike_t : public death_knight_melee_attack_t
     background = special = true;
     may_miss = false;
     weapon = w;
-    base_multiplier *= 1.0 + p -> spec.obliterate_2 -> effectN( 1 ).percent();
+    if ( p -> spec.obliterate_2->ok() )
```
Melekus: This works just fine without the -> ok() check

engine/class_modules/sc_death_knight.cpp
Armadk: Perfect, I'll pull those sections out.

PR #5226, opened 2020-09-02
[shaman] Remove vestiges of Vision of Perfection from elementals

wfarr:
The `essence_proc` argument was still in the function signature
for these abilities, but is no longer necessary.

Also remove commented out guardians tied to Azerite.

cc @Bloodmallet @HawkCorrigan

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -19,7 +19,6 @@
 // - Update Elemental Blast
 //
 // Elemental
-// - Remove traces of VOP and co
 // - Implement Static Discharge
 // - Update Aftershock to only work on Earth Shock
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_shaman.cpp
```diff
@@ -4644,7 +4552,6 @@ struct flame_shock_t : public shaman_spell_t
 {
   flame_shock_spreader_t* spreader;
   const spell_data_t* elemental_resource;
-  const spell_data_t* t20_4pc_bonus;
```
wfarr, line 4647: This got caught looking for unused variables. Didn't feel necessary to split out into its own change given its just the one line.

Bloodmallet (approved):

PR #5227, opened 2020-09-02
Github Actions for Visual Studio

scamille:


scamille:
VS 2017 smoke test is failing, but I can't tell why. Any idea @nuoHep ?

Edit: In other words: Works on my machine, querying return code with ```$LastExitCode``` gives 0. I restarted the job here, and got the same result for the VS-2017 build.

scamille:
Looks like this has been an issue in the past in https://github.com/facebook/jest/issues/9413 https://github.com/actions/virtual-environments/issues/606 and is related to the windows-latest vm.

nuoHep:
Looks fine to me.

Re vs2017 failing the smoke test - I dunno, but if it's not fixable on our end (and it doesn't seem like it is) I'd simply disable the smoke test for it. We were not doing it on AppVeyor, so we are not losing anything. And we do it for the vs2019 which should suffice tbh.

It also may be nice to merge it into a single job with a strategy matrix, but meh, this works.

PR #5231, opened 2020-09-02
[shaman] Add support for Fae Transfusion

wfarr:
* Adds Fae Transfusion
* Update APLs to use it
* Add to default profile for now

~~This is a bit of a work-in-progress as I wasn't 100% sure how to wire this up correctly. Most of it is based on cribbing from Earthquake, since it is another ground-targeted AOE ability. I could use some suggestions on how to make the channeling / ticking behavior work appropriately, as I don't believe it is right now. For example, the base duration should be 3 seconds with base tick rate of 0.5 seconds, but when reviewing the actual simulation behavior it doesn't seem _quite_ right.~~

~~Here is the HTML output of an example running PR_Shaman_Elemental.simc for reference: https://wfarr-static-web.s3.amazonaws.com/fae.html~~

cc @Bloodmallet @HawkCorrigan @navv1234

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -19,7 +19,6 @@
 // - Update Elemental Blast
 // - Covenants
 //   - Kyrian
-//   - Night Fae
 //   - Venthyr
 // - Class Legendaries
 // - Covenant Conduits
@@ -138,16 +137,6 @@ static std::vector<player_t*>& __check_distance_targeting( const action_t* actio
 }
 typedef std::pair<std::string, simple_sample_data_with_min_max_t> data_t;
 typedef std::pair<std::string, simple_sample_data_t> simple_data_t;
-
-template <typename Action, typename Actor, typename... Args>
-action_t* get_action( util::string_view name, Actor* actor, Args&&... args )
-{
-  action_t* a = actor->find_action( name );
-  if ( !a )
-    a = new Action( name, actor, std::forward<Args>( args )... );
-  assert( dynamic_cast<Action*>( a ) && a->name_str == name && a->background );
-  return a;
-}
 
 struct shaman_t;
 
@@ -5366,7 +5355,7 @@ struct fae_transfusion_t : public shaman_spell_t
 
     affected_by_master_of_the_elements = false;
     channeled    = true;
-    tick_action = get_action<fae_transfusion_tick_t>( "fae_transfusion_tick", player );
+    tick_action = new fae_transfusion_tick_t( "fae_transfusion_tick", player );
   }
 };
 
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -139,6 +139,16 @@ static std::vector<player_t*>& __check_distance_targeting( const action_t* actio
 typedef std::pair<std::string, simple_sample_data_with_min_max_t> data_t;
 typedef std::pair<std::string, simple_sample_data_t> simple_data_t;
 
+template <typename Action, typename Actor, typename... Args>
+action_t* get_action( util::string_view name, Actor* actor, Args&&... args )
+{
+  action_t* a = actor->find_action( name );
+  if ( !a )
+    a = new Action( name, actor, std::forward<Args>( args )... );
+  assert( dynamic_cast<Action*>( a ) && a->name_str == name && a->background );
+  return a;
+}
+
 struct shaman_t;
 
 enum totem_e
@@ -384,6 +394,9 @@ public:
   {
     // Necrolord
     const spell_data_t* necrolord; // Primordial Wave
+
+    // Night Fae
+    const spell_data_t* night_fae; // Fae Transfusion
   } covenant;
 
   // Gains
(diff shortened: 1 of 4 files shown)


engine/class_modules/sc_shaman.cpp
```diff
@@ -5272,6 +5280,58 @@ struct thundercharge_t : public shaman_spell_t
...
+struct fae_transfusion_damage_t : public shaman_spell_t
+{
+  fae_transfusion_damage_t( shaman_t* player, const std::string& options_str )
+    : shaman_spell_t( "fae_transfusion_", player, player->find_spell( 328928 ), options_str )
+  {
+    aoe        = 4;
+    ground_aoe = background = true;
+    dynamic_tick_action     = true;
+    base_tick_time          = 500_ms;
+    dot_duration            = 3_s;
```
vituscze: This dot/tick stuff is unnecessary. The spell doesn't have any ticking damage in its data so it wouldn't do ticking damage anyways.

engine/class_modules/sc_shaman.cpp
```diff
@@ -5272,6 +5280,58 @@ struct thundercharge_t : public shaman_spell_t
...
+  }
+
+  void impact( action_state_t* state ) override
+  {
+    shaman_spell_t::impact( state );
```
vituscze: Can be removed.

engine/class_modules/sc_shaman.cpp
```diff
@@ -5272,6 +5280,58 @@ struct thundercharge_t : public shaman_spell_t
...
+  {
+    shaman_spell_t::cancel();
+    event_t::cancel( area_of_effect );
+  }
+
+  void execute() override
+  {
+    shaman_spell_t::execute();
+
+    area_of_effect = make_event<ground_aoe_event_t>(
```
vituscze: This won't work as envisioned. ground_aoe_event_t creates new events after each tick, so the area_of_effect pointer isn't pointing to the current event when you want to cancel it. I'd suggest doing it in the same way as shifting_power_t in the mage module (i.e. channel that has non-nullptr tick_action).

engine/class_modules/sc_shaman.cpp
```diff
@@ -5272,6 +5280,58 @@ struct thundercharge_t : public shaman_spell_t
...
+};
+
+struct fae_transfusion_t : public shaman_spell_t
+{
+  fae_transfusion_damage_t* transfusion;
+  event_t* area_of_effect;
+
+  fae_transfusion_t( shaman_t* player, const std::string& options_str )
+    : shaman_spell_t( "fae_transfusion", player, player->covenant.fae_transfusion, options_str ),
+      transfusion( new fae_transfusion_damage_t( player, options_str ) )
```
vituscze: The secondary damage spell should not be constructed with any options. Options are only ever relevant for spells that go into the APL.

engine/class_modules/sc_shaman.cpp
```diff
@@ -5272,6 +5280,58 @@ struct thundercharge_t : public shaman_spell_t
...
+  void impact( action_state_t* state ) override
+  {
+    shaman_spell_t::impact( state );
+  }
+};
+
+struct fae_transfusion_t : public shaman_spell_t
+{
+  fae_transfusion_damage_t* transfusion;
+  event_t* area_of_effect;
```
scamille: you should set area_of_effect to nullptr in the constructor initializer list.

Bloodmallet:
Thank you for the reviews, I poke wfarr in a different channel about it. 👍

wfarr:
Thanks all! This feedback was helpful. I believe I've addressed everything and in my own local tests Fae Transfusion appears to work correctly now.

engine/class_modules/sc_shaman.cpp
```diff
@@ -139,6 +139,16 @@ static std::vector<player_t*>& __check_distance_targeting( const action_t* actio
...
 
+template <typename Action, typename Actor, typename... Args>
+action_t* get_action( util::string_view name, Actor* actor, Args&&... args )
+{
+  action_t* a = actor->find_action( name );
+  if ( !a )
+    a = new Action( name, actor, std::forward<Args>( args )... );
+  assert( dynamic_cast<Action*>( a ) && a->name_str == name && a->background );
+  return a;
+}
```
vituscze: Doesn't the shaman module have an existing way to get secondary actions?

engine/class_modules/sc_shaman.cpp
```diff
@@ -5322,6 +5335,39 @@ struct thundercharge_t : public shaman_spell_t
...
+    return result_amount_type::DMG_DIRECT;
+  }
+};
+
+struct fae_transfusion_t : public shaman_spell_t
+{
+  fae_transfusion_t( shaman_t* player, const std::string& options_str )
+    : shaman_spell_t( "fae_transfusion", player, player->find_covenant_spell( "Fae Transfusion" ), options_str )
+  {
+    parse_options( options_str );
```
vituscze: Options are currently parsed twice for this spell, only one is necessary.

wfarr: This should be fixed in 3e1c4e2884

engine/class_modules/sc_shaman.cpp
wfarr: I do see `void trigger_secondary_ability()` but I didn't spot a way to get the action itself. That said, I noticed some other spells that set `tick_action` to an instance of some other spell directly, so I've done so in 397c962ae2 and that appears to work fine, so I've removed the `get_action` stuff borrowed from the mage module.

vituscze:
Looks good now.

PR #5232, opened 2020-09-03
[shaman] Add maelstrom back to Elemental

wfarr:
The rework to Fulmination was abandoned and Elemental is once again
using maelstrom as a resource.

Of note in this change:

* The spell ID for maelstrom is now 343725, with the previous
  spell ID of 190493 still hanging around as Fulmination.
* Tooltips for Lava Beam are bugged and show incorrect values
* Tooltips are bugged for Icefury and don't show Frost Shock's
  maelstrom generation value at all, so it is hardcoded based
  on in-game testing on beta.
* Fix Aftershock to only work on Earth Shock

Note that the default APL is currently causing a segfault due to
some conditional hidden within there. I'm going to be sending
along a separate Pull Request shortly to remove all the APL
complexity from Elemental until we're ready to begin optimizing
when things are feature-complete.

cc @Rusah @HawkCorrigan @Bloodmallet

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -1026,7 +1026,12 @@ public:
   {
     ab::execute();
 
-    trigger_maelstrom_gain( ab::execute_state );
+    if (p()->specialization() == SHAMAN_ELEMENTAL)
+    {
+      trigger_maelstrom_gain( ab::execute_state );
+    }
+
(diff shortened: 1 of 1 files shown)


wfarr:
Link to new shadowlands maelstrom spell for reference: https://shadowlands.wowhead.com/spell=343725/maelstrom

Bloodmallet (approved):

PR #5236, opened 2020-09-04
[shaman] Add Primordial Wave (necrolord covenant)

wfarr:
This *only* implements the spell itself and the buff on-cast.

I have a WIP of the duplication of spells on Flame Shocked
targets, but it's not working 100% correctly at the moment.

This at least gets the easy part of the spell implemented
with TODOs for later.

cc @Bloodmallet

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -5,6 +5,7 @@
 
 #include "simulationcraft.hpp"
 
+#include "player/covenant.hpp"
 #include "player/pet_spawner.hpp"
 #include "sc_enums.hpp"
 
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_shaman.cpp
```diff
@@ -372,6 +380,13 @@ struct shaman_t : public player_t
     cooldown_t* strike;  // shared CD of Storm Strike and Windstrike
   } cooldown;
 
+  // Covenant Class Abilities
+  struct
+  {
+    // Necrolord
+    const spell_data_t* necrolord; // Primordial Wave
```
wfarr, line 387: This naming convention is what I saw used in the Druid module so I ran with that.

Bloodmallet (approved):

PR #5237, opened 2020-09-05
[Elemental] Elemental Blast % change

HawkCorrigan:


scamille:
As a general rule, you should use buff->check() in const functions, and up() only once per buff for benefit tracking (if that is possible at all. Usually inside execute() of a single action having something benefiting from the buff would be the right place).

Most const functions in action_t can be called multiple times (through expressions, ready() checks), so that will skew benefit calculations if called in there.

Or as a more generalized rule: Use buff->check() as default mindset, only use buff-up() in specific places where you want to do benefit tracking.


Anyway, not really that important, just wanted to let you know so that you won't draw the wrong conclusions from the benefit% column in the html report buff section.

HawkCorrigan:
lemme change that and read up on it a little more and then reopen

scamille:
As said it isn't really that important, just a minor optimization mainly. No need to close or change this, just something which can be taken into account for future changes.

PR #5238, opened 2020-09-05
[Elemental] Elemental Blast % change

HawkCorrigan:


engine/class_modules/sc_shaman.cpp
```diff
@@ -6415,12 +6417,15 @@ void shaman_t::create_buffs()
...
-  buff.elemental_blast_mastery->set_max_stack( 1 );
+  buff.elemental_blast_crit = make_buff( this, "elemental_blast_critical_strike", find_spell( 118522 ) )
+          ->set_max_stack(1)
+          ->add_invalidate(CACHE_CRIT_CHANCE);
+  buff.elemental_blast_haste = make_buff( this, "elemental_blast_haste", find_spell( 173183 ) )
+          ->set_max_stack(1)
+          ->add_invalidate(CACHE_HASTE);
+  buff.elemental_blast_mastery = make_buff( this, "elemental_blast_mastery", find_spell( 173184 ) )
+          ->set_max_stack(1)
+          ->add_invalidate(CACHE_MASTERY);
```
vituscze, line 6428: Unlike stat buffs, regular buffs make no attempt to extract the crit/haste/etc from the spell data. You need to set the default values here in some way (set_default_value, set_default_value_from_effect, set_default_value_from_effect_type) so that you can use check_value in the later code.

HawkCorrigan:
Makes sense
Some talking to people made it clear that the change in behavior was automatically applied without me needing to change anything. So basically I just changed automatic code into manual one. 🤷‍♂️

vituscze:
No, what you did was definitely on the right track. stat_buff_t can only apply rating buffs automatically, buffs that give raw stats (no ratings) need to be handled in the way you had it in the PR

PR #5244, opened 2020-09-08
[Frost DK] Frost updates for SL

Armadk:
Highlights:
- Frostwyrm made baseline
- Frostscythe target cap
- 2H support for remaining spells
- Bugfix for spec.death_knight lookup
- With the fix applied, frostscythe was double dipping on the crit damage multipler, so it was removed.

At line 2660 I'd like some additional review.  After fixing the spec aura for death knight, I don't think these two are necessary anymore.

Aside from that, I think this should finish off 2h support for frost.

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -923,7 +923,6 @@ public:
     cooldown.empower_rune_weapon = get_cooldown( "empower_rune_weapon" );
     cooldown.icecap_icd          = get_cooldown( "icecap" );
     cooldown.pillar_of_frost     = get_cooldown( "pillar_of_frost" );
-    cooldown.raise_dead          = get_cooldown( "raise_dead" );
     cooldown.vampiric_blood      = get_cooldown( "vampiric_blood" );
 
     resource_regeneration = regen_type::DYNAMIC;
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -4848,7 +4848,7 @@ struct frostscythe_t : public death_knight_melee_attack_t
     parse_options( options_str );
 
     weapon = &( player -> main_hand_weapon );
-    aoe = -1;
+    aoe = as<int>( data().effectN( 5 ).base_value() );
 
     weapon_req = WEAPON_1H;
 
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_death_knight.cpp
```diff
@@ -922,6 +923,7 @@ struct death_knight_t : public player_t {
     cooldown.empower_rune_weapon = get_cooldown( "empower_rune_weapon" );
     cooldown.icecap_icd          = get_cooldown( "icecap" );
     cooldown.pillar_of_frost     = get_cooldown( "pillar_of_frost" );
+    cooldown.raise_dead          = get_cooldown( "raise_dead" );
```
Melekus: You don't need to store a cooldown object for raise_dead if you're not going to adjust it throughout the simulation. One is automatically created on spell construction.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4846,11 +4858,7 @@ struct frostscythe_t : public death_knight_melee_attack_t
     parse_options( options_str );
 
     weapon = &( player -> main_hand_weapon );
-    aoe = -1;
-
-    weapon_req = WEAPON_1H;
-
-    crit_bonus_multiplier *= 1.0 + p -> spec.death_knight -> effectN( 5 ).percent();
```
Melekus, line 4853: Could add a comment mentionning that the crit multiplier is handled in apply_affecting_auras()

engine/class_modules/sc_death_knight.cpp
```diff
@@ -2655,6 +2657,7 @@ struct death_knight_action_t : public Base
       gain = this -> player -> get_gain( util::inverse_tokenize( this -> name_str ) );
     }
 
+    // These both need review to see if the spec aura is applying or not, if it is, this will double dip
```
Melekus: Only a boolean is set here, it shouldn't double dip and the worse case scenario is one superfluous set of test and operation is being run once on spell construction. But i can check later

Melekus:
The frost APL will also require changing add_talent() lines to add_action() lines for Frostwyrm's Fury

engine/class_modules/sc_death_knight.cpp
Armadk: Great idea, added in

Armadk:
> The frost APL will also require changing add_talent() lines to add_action() lines for Frostwyrm's Fury

Great catch on that, thank you, updated.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -497,6 +497,7 @@ struct death_knight_t : public player_t {
   // Cooldowns
   struct cooldowns_t {
     // Shared
+    cooldown_t* raise_dead;
```
Melekus: This should be removed too.

PR #5246, opened 2020-09-09
[shaman] Add Kyrian covenant ability: Vesper Totem

wfarr:
* When summoned, create a hidden aura for the damage pulses it can
  perform
* When the player casts a non-background ability, trigger the
  totem's pulse action
* The totem's pulse action does AOE damage within the defined
  range, and removes a stack of the buff

TODO:

- [ ] Properly wire up a pulse to trigger when a spell / attack
      is successfully cast

This one is a little tricky being unfamiliar with semantics around pets/totems and getting the player to trigger events on them given my lack of experience around some of the internals here. I could use some suggestions on how to proceed with wiring this up properly. 

References:

* base spell https://shadowlands.wowhead.com/spell=324386/vesper-totem
* damage pulse spell https://shadowlands.wowhead.com/spell=324520/vesper-totem
* healing pulse spell https://shadowlands.wowhead.com/spell=324522/vesper-totem

Observed behavior in-game on beta is that the pulse does trigger on spell cast success rather than on spell impact.

cc @Bloodmallet @vituscze

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -326,6 +326,8 @@ public:
 
     // Covenant Class Ability Buffs
     buff_t* primordial_wave;
+    buff_t* vesper_totem_damage;
+    // buff_t* vesper_totem_healing;
 
     // Elemental, Restoration
     buff_t* lava_surge;
@@ -381,6 +383,8 @@ public:
   // Covenant Class Abilities
   struct
   {
+    // Kyrian
+    const spell_data_t* kyrian; // Vesper Totem
     // Necrolord
     const spell_data_t* necrolord; // Primordial Wave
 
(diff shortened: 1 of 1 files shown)


navv1234:
I can push a mockup version of how you could do it, whipped up one to see if we can abuse ground_aoe_event_t for this as Norrinir suggested.

wfarr:
Sure. That’d be rad.

You’re also free to modify any of my changes directly in this PR if that’s easier to iterate on for you.

navv1234:
See 2b920a7d0be19684d0a7cd2238b31f69eb15b9f1

EDIT: Note that the replacement is not implemented right now, as that'll need some thinking on what's the best way to implement it in the APL. I'm guessing default behavior should be "don't replace", but add an option to allow it.

PR #5248, opened 2020-09-09
[shaman] Add Call of Flame conduit

wfarr:
* Increases the duration of Fire/Storm Elemental by N%

Reference spell: https://shadowlands.wowhead.com/spell=338303/call-of-flame

cc @Bloodmallet


scamille (commented):
~~You could add a function void trigger_call_of_flame() to shaman_spell_t and merge the two identical blocks you currently have.~~
Edit: Forget that. I did not fully realize that it affects two separate pets.

Bloodmallet (approved):

PR #5249, opened 2020-09-09
[shaman] Add Venthyr covenant class ability: Chain Harvest

wfarr:
Chain Harvest is a chained spell that deals %SP damage on up to
5 targets, with a secondary effect that reduces the remaining
cooldown on Chain Harvest by 5 seconds for each target that was
critically hit.

cc @Bloodmallet

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -19,7 +19,6 @@
 // - Update Elemental Blast
 // - Covenants
 //   - Kyrian
-//   - Venthyr
 // - Class Legendaries
 // - Covenant Conduits
 //
@@ -386,6 +385,9 @@ public:
 
     // Night Fae
     const spell_data_t* night_fae; // Fae Transfusion
+
+    // Venthyr
+    const spell_data_t* venthyr; // Chain Harvest
   } covenant;
 
   // Gains
@@ -5384,6 +5386,50 @@ struct primordial_wave_t : public shaman_spell_t
 };
 
 // ==========================================================================
+// Chain Harvest - Venthyr Covenant
+// ==========================================================================
+struct chain_harvest_t : public chained_base_t
+{
+  int critical_hits = 0;
+
+  chain_harvest_t( shaman_t* player, const std::string& options_str )
+    : chained_base_t( player, "chain_harvest", player->covenant.venthyr, 0, options_str )
+  {
+    if ( !player->covenant.venthyr->ok() )
+      return;
+
+    aoe = 5;
+    spell_power_mod.direct = player->find_spell( 320752 )->effectN( 1 ).sp_coeff();
+  }
+
+  void impact( action_state_t* s ) override
+  {
+    chained_base_t::impact( s );
+
+    if ( s->result == RESULT_CRIT )
+    {
+      critical_hits++;
+    }
+  }
+
+  void update_ready( timespan_t ) override
+  {
+    auto cd = cooldown_duration();
+    cd -= critical_hits * timespan_t::from_seconds( 5 );
+
+    critical_hits = 0;
+
+    chained_base_t::update_ready( cd );
+  }
+
+  void execute() override
+  {
+    // do not consume stormkeeper, as happens in chained_base_t
+    shaman_spell_t::execute();
+  }
+};
+
+// ==========================================================================
 // Shaman Custom Buff implementation
 // ==========================================================================
 
@@ -5587,6 +5633,8 @@ action_t* shaman_t::create_action( util::string_view name, const std::string& op
     return new primordial_wave_t( this, options_str );
   if ( name == "fae_transfusion" )
     return new fae_transfusion_t( this, options_str );
+  if ( name == "chain_harvest" )
+    return new chain_harvest_t( this, options_str );
 
   // elemental
   if ( name == "chain_lightning" )
@@ -6032,6 +6080,7 @@ void shaman_t::init_spells()
   // Covenants
   covenant.necrolord = find_covenant_spell( "Primordial Wave" );
   covenant.night_fae = find_covenant_spell( "Fae Transfusion" );
+  covenant.venthyr   = find_covenant_spell( "Chain Harvest" );
 
   //
   // Misc spells
@@ -6766,6 +6815,7 @@ void shaman_t::init_action_list_elemental()
   aoe->add_talent( this, "Stormkeeper", "if=talent.stormkeeper.enabled" );
   aoe->add_action( this, "Flame Shock", "target_if=refreshable&spell_targets.chain_lightning<5" );
   aoe->add_action( this, "Earthquake" );
+  aoe->add_action( "chain_harvest" );
   aoe->add_action( "fae_transfusion" );
   aoe->add_action( "lava_beam,if=talent.ascendance.enabled" );
   aoe->add_action( this, "Chain Lightning" );
@@ -6774,6 +6824,7 @@ void shaman_t::init_action_list_elemental()
 
   // Single target APL
   single_target->add_action( this, "Flame Shock", "target_if=!ticking|dot.flame_shock.remains<=gcd&!buff.surge_of_power.up" );
+  single_target->add_action( "chain_harvest" );
   single_target->add_action( "fae_transfusion" );
   single_target->add_action( "primordial_wave" );
   single_target->add_talent( this, "Elemental Blast", "if=talent.elemental_blast.enabled" );


engine/class_modules/sc_shaman.cpp
```diff
@@ -5383,6 +5387,39 @@ struct primordial_wave_t : public shaman_spell_t
...
+
+  void impact( action_state_t* s ) override
+  {
+    chained_base_t::impact( s );
+
+    if ( result_is_hit( s->result ) && s->result == RESULT_CRIT )
+    {
+      // This value appears to hardcoded in the spell
+      p()->cooldown.chain_harvest->adjust( -timespan_t::from_seconds(5) );
+    }
```
wfarr: I'm not sure this is behaving as expected. As best I can tell in the output, the execution interval remains approx 90 seconds when I would expect it to go down a bit from there.

I'm wondering if the spell isn't considered to be on cool down when impact() executes, though I would expect it is based on in-game behavior.

Looking at similar effects, like Rogue's Sepsis, setting up the `cooldown_t*` and calling `adjust( timespan_t* )` on it appears to be the preferred way.

scamille:
The cooldown reduction code does look ok. Have you checked with debug=1?

You could also hook up a proc object to track the number of occurrences.

wfarr:
> The cooldown reduction code does look ok. Have you checked with debug=1?

Yeah, I've tried a couple approaches and I've seen two behaviors here:

* The interval of execution basically doesn't move at all
* The interval goes from 90s to 3s, which is clearly very wrong

When debugging, I'm seeing output that indicates the spell hasn't started its cooldown yet like so (for the sake of testing I've set desired targets to 5 and critical strike to be 100% for the character, which should yield the maximum effect of a 25s cooldown reduction):

```
3.301 PR_Shaman_Elemental recharge cooldown chain_harvest adjustment=-5.000, remains=0.000
3.301 PR_Shaman_Elemental recharge cooldown chain_harvest adjustment=-5.000, remains=0.000
3.301 PR_Shaman_Elemental recharge cooldown chain_harvest adjustment=-5.000, remains=0.000
3.301 PR_Shaman_Elemental recharge cooldown chain_harvest adjustment=-5.000, remains=0.000
3.301 PR_Shaman_Elemental recharge cooldown chain_harvest adjustment=-5.000, remains=0.000
3.301 Player 'PR_Shaman_Elemental' starts cooldown for Action chain_harvest (Cooldown chain_harvest, 1/1). Duration=-9223372036854776.000 Delay=0.000. Will be ready at 93.301
```

scamille:
Can you paste a full profile to easily replicate this?

Edit: Now that I come to think of it, this is probably a ordering issue in execute/impact.
Assuming the spell has no travel time, the cooldown reduction comes before action_t::update_ready which starts the cooldown.

Would it be ok to complicate redesign this so that instead of starting with 90s and then reduction the cooldown through adjust, to just directly start it with 90s-5s*num_targets hit?

wfarr:
> Can you paste a full profile to easily replicate this?

Sure, here's the dummy profile I'm using: https://gist.github.com/wfarr/f3daad8f32592c6ecdf9f73573cbd0c6

> Edit: Now that I come to think of it, this is probably a ordering issue in execute/impact.
> Assuming the spell has no travel time, the cooldown reduction comes before action_t::update_ready which starts the cooldown.

This sounds accurate the the behavior I'm seeing. I need to do a quick test in-game to see if it _should_ have a travel time or not.

> Would it be ok to complicate redesign this so that instead of starting with 90s and then reduction the cooldown through adjust, to just directly start it with 90s-5s*num_targets hit?

The tricky thing here is the CDR only occurs if the hit is a critical strike. I could be incorrect, but my understanding is we needed impact in order to evaluate the result of the hit.

scamille:
You could try the following: In chain_harvest_t add:
```
void update_ready(timespan_t) override
{
 auto cd = cooldown_duration();
 cd -= num_targets_hit * timespan_t::from_seconds( 5 );

  chain_base_t::update_ready( cd );
}
```
(Remove your current cd -> adjust calls)

Edit: I forgot about the crit part :/

wfarr:
Checked on beta and can confirm there is no travel time.

RE: overloading update_ready, maybe I could have `impact` update an internal variable for tracking, and then utilize that when update_ready applies the cooldown reduction?

wfarr:
OK, this appears to work as expected on initial testing:

```c++
struct chain_harvest_t : public chained_base_t
{
  int critical_hits = 0;

  chain_harvest_t( shaman_t* player, const std::string& options_str )
    : chained_base_t( player, "chain_harvest", player->covenant.venthyr, 0, options_str )
  {
    if ( !player->covenant.venthyr->ok() )
      return;

    aoe = 5;
    spell_power_mod.direct = player->find_spell( 320752 )->effectN( 1 ).sp_coeff();
  }

  {
    chained_base_t::impact( s );

    if ( s->result == RESULT_CRIT )
    {
      critical_hits++;
    }
  }

  void update_ready( timespan_t ) override
  {
    auto cd = cooldown_duration();
    cd -= critical_hits * timespan_t::from_seconds( 5 );

    critical_hits = 0;

    chained_base_t::update_ready( cd );
  }

  void execute() override
  {
    // do not consume stormkeeper, as happens in chained_base_t
    shaman_spell_t::execute();
  }
};
```

scamille:
Your solution looks fine.

I asked in Discord channel for ideas on how to solve this. Another option would be to force an event inbetween the action execution and the cooldown adjustment by either forcing a travel event with 0ms (override schedule_travel),
or having a custom event in impact for the cooldown adjustment like this:
```
void impact( ... )
{
  base::impact( ... );

  if ( is_crit )
    make_event( *sim, 0_ms, [ this ] { cooldown->adjust( -5.0_s ); } );
}
```

Then your original cooldown->adjustment calls would work.

wfarr:
Word. I took a quick stab at the above options but they still weren't quite working. If there are no major concerns about the current implementation, I think I'd be keen to get it merged and revisit later if another implementation is preferred.

scamille:
Yes this is fine, especially if you have been able to verify it is working as it should.

wfarr:
Yep, here's output from a sample with 5 targets and 100% crit:

```
3.301 Player 'PR_Shaman_Elemental' starts cooldown for Action chain_harvest (Cooldown chain_harvest, 1/1). Duration=65.000 Delay=0.000. Will be ready at 68.301
70.921 PR_Shaman_Elemental chain_harvest cooldown waste tracking waste=2.620 exec_time=2.407
70.921 Player 'PR_Shaman_Elemental' starts cooldown for Action chain_harvest (Cooldown chain_harvest, 1/1). Duration=65.000 Delay=0.000. Will be ready at 135.921
139.176 PR_Shaman_Elemental chain_harvest cooldown waste tracking waste=3.255 exec_time=2.407
139.176 Player 'PR_Shaman_Elemental' starts cooldown for Action chain_harvest (Cooldown chain_harvest, 1/1). Duration=65.000 Delay=0.000. Will be ready at 204.176
206.763 PR_Shaman_Elemental chain_harvest cooldown waste tracking waste=2.587 exec_time=2.360
206.763 Player 'PR_Shaman_Elemental' starts cooldown for Action chain_harvest (Cooldown chain_harvest, 1/1). Duration=65.000 Delay=0.000. Will be ready at 271.763
274.123 PR_Shaman_Elemental chain_harvest cooldown waste tracking waste=2.360 exec_time=2.360
274.123 Player 'PR_Shaman_Elemental' starts cooldown for Action chain_harvest (Cooldown chain_harvest, 1/1). Duration=65.000 Delay=0.000. Will be ready at 339.123
```

And the same again with 50% crit:

```
3.301 Player 'PR_Shaman_Elemental' starts cooldown for Action chain_harvest (Cooldown chain_harvest, 1/1). Duration=80.000 Delay=0.000. Will be ready at 83.301
85.661 PR_Shaman_Elemental chain_harvest cooldown waste tracking waste=2.360 exec_time=2.360
85.661 Player 'PR_Shaman_Elemental' starts cooldown for Action chain_harvest (Cooldown chain_harvest, 1/1). Duration=70.000 Delay=0.000. Will be ready at 155.661
159.631 PR_Shaman_Elemental chain_harvest cooldown waste tracking waste=3.970 exec_time=2.407
159.631 Player 'PR_Shaman_Elemental' starts cooldown for Action chain_harvest (Cooldown chain_harvest, 1/1). Duration=75.000 Delay=0.000. Will be ready at 234.631
236.945 PR_Shaman_Elemental chain_harvest cooldown waste tracking waste=2.314 exec_time=2.314
236.945 Player 'PR_Shaman_Elemental' starts cooldown for Action chain_harvest (Cooldown chain_harvest, 1/1). Duration=75.000 Delay=0.000. Will be ready at 311.945
```

Bloodmallet (approved):

PR #5250, opened 2020-09-09
[Frost DK] Added Eradicating Blow Conduit

Khazakdk:
Also changed Biting Cold to use benefit tracking after talking to Melekus


scamille:
Those buff->up() calls in frost atrik should go into an execute() override.
The const functions should use check().
That way you ensure having one correct benefit tracking call per execution.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5037,6 +5053,11 @@ struct frost_strike_t : public death_knight_melee_attack_t
         p() -> replenish_rune( 1, p() -> gains.obliteration );
       }
     }
+
+    if ( p() -> buffs.eradicating_blow -> up() )
```
Melekus: As @scamille pointed out, this should be in an action's execute() override.
Either in frost_strike_strike_t::execute() if it loses two stacks per cast when dual wielding, or in frost_strike_t::execute() if it triggers only once regardless of DW/2H.

Adding it right after the call to death_knight_melee_t::execute() will ensure that the damage bonus is properly accounted for.

Khazakdk: Verified the behavior in the latest beta build, it changed sometime recently to consume both stacks per cast regardless of dw/2h so I updated the code accordingly.  Will need to keep an eye on it in case in changes in the future.

PR #5256, opened 2020-09-11
[shaman] Add support for Tumbling Waves conduit

wfarr:
Primordial Wave has an N% chance to not incur its cooldown (though
technically I am resetting the cooldown immediately here instead).

cc @Bloodmallet

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -5436,7 +5436,7 @@ struct primordial_wave_t : public shaman_spell_t
 
     if ( p()->conduit.tumbling_waves->ok() && rng().roll( p()->conduit.tumbling_waves.percent() ) )
     {
-      cooldown->reset( false );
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_shaman.cpp
```diff
@@ -5428,7 +5431,13 @@ struct primordial_wave_t : public shaman_spell_t
 
   void execute() override {
     shaman_spell_t::execute();
+
     p()->buff.primordial_wave->trigger();
+
+    if ( p()->conduit.tumbling_waves->ok() && rng().roll( p()->conduit.tumbling_waves.percent() ) )
+    {
+      cooldown->reset( false );
```
wfarr: Not entirely sure this makes a difference in this particular case, but we can set this to true and it should be fine:

```suggestion
      cooldown->reset( true );
```

Bloodmallet (approved):

PR #5258, opened 2020-09-11
[shaman] Implement Echoes of Great Sundering legendary effect

wfarr:
When the player casts Earth Shock, a buff Echoes of Great Sundering
is triggered. While this buff is up, the next Earthquake cast does
X% more damage (currently 250%).

cc @Bloodmallet

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -328,6 +328,9 @@ public:
     buff_t* primordial_wave;
     buff_t* vesper_totem;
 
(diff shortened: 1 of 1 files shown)


wfarr:
Note the tooltip on wowhead is incorrect at this time. Elemental Blast does not proc this buff any longer.

scamille:
You probably want to decrement/expire the buff at the end of earthquake::execute if it only affects the *next* earthquake.

Bloodmallet (approved):

PR #5259, opened 2020-09-11
[Frost DK] Sqrt scaling for Howling Blast and BoS

Armadk:
BoS is pretty straight forward
howling blast is a bit different and I went with scaling it inside the spell itself, rather than via the flag.


engine/class_modules/sc_death_knight.cpp
```diff
@@ -5234,7 +5233,7 @@ struct howling_blast_aoe_t : public death_knight_spell_t
   frost_fever_t* frost_fever;
 
   howling_blast_aoe_t( death_knight_t* p, const std::string& options_str, frost_fever_t* frf ) :
-    death_knight_spell_t( "howling_blast_aoe", p, p -> find_spell( 237680 ) ),
+    death_knight_spell_t( "howling_blast_aoe", p, p -> spec.howling_blast ),
```
Melekus: If the aoe component is also associated with the main spell's ID now, you can just remove howling_blast_aoe_t and add aoe=-1 with reduced_aoe_damage=true to the main spell.

I looked a bit into spelldata and testing ingame, it seems the only remaining purpose of 237680 is to specify the radius of the aoe.  It's a simple double: action_t::radius.

Armadk: That was my original plan after noticing that the only thing 237680 was being used for was tooltip.

However looking at the execute method gave me pause.  We trigger a number of things in there in the execute method, and I wasn't familiar enough with the call stacks to feel confident that they would only trigger on the main target.

- killing_machine w/ obliteration
- avalanche 
- echoing howl azerite

We also trigger the runic power regen multiplier in the main spell.

If these only trigger once and the damage and impact trigger for each mob, than I am in full agreement we should simplify this section.

It can be a learning moment for me :)

Melekus: action_t::execute() is only done once through a spell cast. action_t::impact() is the one being duplicated for each target hit. There shouldn't be anything to worry about in Howling Blast's case

Armadk: Refactored per discussion.  Pre and post refactor output is available in the linked issue.

PR #5261, opened 2020-09-11
Expressions: factor out the lexer

nuoHep:
Refactor expression tokenization by factoring out a lexer to make it
easier to follow.

The lexer is non-allocating - tokens contain views into the input
string view.

Drop an intermidate `convert_to_unary()` step as the lexer produces
unary tokens for plus/minus, so we need to convert to binary, and it's
actually easier to do that right inside `parse_tokens()`. We never use
non-converted token streams anyway.

---

Plus a smol cleanup of token printing on top.

### DIFF
--- a/engine/sim/sc_expressions.cpp
+++ b/engine/sim/sc_expressions.cpp
@@ -45,15 +45,6 @@ struct lexer_t
     return false;
   }
 
-  bool match( util::string_view str, int consume_mod = 0 )
-  {
-    if ( util::str_prefix_ci( input.substr( current_len ), str ) ) {
-      current_len += ( str.size() + consume_mod );
-      return true;
-    }
-    return false;
-  }
-
   token_t yield_token( token_e type )
   {
     token_t token;
@@ -81,12 +72,6 @@ struct lexer_t
       }
       return yield_token( TOK_NUM );
     }
-
-    // special case floor & ceil "pseudo-functions"
-    if ( lowercase( ch ) == 'f' && match( "loor(", -1 ) )
-      return yield_token( TOK_FLOOR );
-    if ( lowercase( ch ) == 'c' && match( "eil(", -1 ) )
-      return yield_token( TOK_CEIL );
 
     // "strings" (identifiers, really)
     if ( is_alpha( ch ) ) {
@@ -1042,6 +1027,15 @@ std::vector<expr_token_t> parse_tokens( action_t* action, util::string_view expr
       token.type = TOK_SUB;
     if ( token.type == TOK_PLUS && prev_is_left_side )
       token.type = TOK_ADD;
+
+    // lexer never produces FLOOR/CEIL "pseudo-functions", match them here
+    if ( token.type == TOK_LPAR && prev_token == TOK_STR )
+    {
+      if ( util::str_compare_ci( tokens.back().label, "floor" ) )
+        tokens.back().type = TOK_FLOOR;
+      else if ( util::str_compare_ci( tokens.back().label, "ceil" ) )
+        tokens.back().type = TOK_CEIL;
+    }
 
     // optimization: fold minus followed by a number together into a single number token
     if ( token.type == TOK_NUM && prev_token == TOK_MINUS )
--- a/engine/sim/sc_expressions.cpp
+++ b/engine/sim/sc_expressions.cpp
@@ -104,6 +104,7 @@ struct lexer_t
       case '*': return yield_token( TOK_MULT );
       case '%':
       {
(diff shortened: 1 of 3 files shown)


engine/sim/sc_expressions.cpp
```diff
@@ -16,6 +16,145 @@ namespace
...
+      current_len++;
+      return true;
+    }
+    return false;
+  }
+
+  bool match( util::string_view str, int consume_mod = 0 )
+  {
+    if ( util::str_prefix_ci( input.substr( current_len ), str ) ) {
+      current_len += ( str.size() + consume_mod );
```
scamille: Maybe add an assert here that the resulting calculation is still in bounds of the input.
I understand that for the current match calls with consume_mod = -1 this should always be fine.

engine/sim/sc_expressions.cpp
```diff
@@ -16,6 +16,145 @@ namespace
...
+    switch ( ch )
+    {
+      case '(': return yield_token( TOK_LPAR );
+      case ')': return yield_token( TOK_RPAR );
+      case '+': return yield_token( TOK_PLUS );
+      case '-': return yield_token( TOK_MINUS );
+      case '*': return yield_token( TOK_MULT );
+      case '%':
+      {
+        if ( match( '%' ) )
```
scamille, line 93: So %% is mod, and a single % is division? Might be worth a comment since this is quite a bit unintuitive at first glance.

engine/sim/sc_expressions.cpp
```diff
@@ -1035,69 +1011,65 @@ bool is_binary( token_e expr_token_type )
...
+      token.type = TOK_SUB;
+    if ( token.type == TOK_PLUS && prev_is_left_side )
+      token.type = TOK_ADD;
 
-void print_tokens( std::vector<expr_token_t>& tokens, sim_t* sim )
-{
-  std::string str;
-  if ( !tokens.empty() )
-    str += "tokens: ";
+    // optimization: fold minus followed by a number together into a single number token
```
scamille, line 1040: Couldn't this be done inside the lexer?
Maybe move the TOK_NUM matching (line 75 to 83) to a helper function, and then also call that when you have a TOK_MINUS match.

scamille:
Great rewrite. Makes it much easier to understand the entirety of the simc lexing.

engine/sim/sc_expressions.cpp
nuoHep: I'd rather not tbh.
`-` can be both unary (`TOK_MINUS`) and binary (`TOK_SUB`). Right now the lexer does not care and always produces only the unary one. Moving this to the lexer will necessarily require the lexer to also handle binary `-` (as `)-1` should produce `[TOK_RPAR, TOK_SUB, TOK_NUM]`) which would require adding lookbehind.

I actually thought about handling `ceil` & `floor` outside of the lexer too (they can be easily matched outside), don't remember why I decided to leave it here (the commit is from July when I was playing with an interpreter for expressions).

engine/sim/sc_expressions.cpp
nuoHep: Yeah, been like that since d68425e69c0c032053b7cafe43cfa8727596c78c
Will add (I think it's kinda self documenting, but I see how it can be non-obvious when looking at it for the first time).

engine/sim/sc_expressions.cpp
scamille: Oh ok. So that's the main reason to restrict the lexer to unary is so that it works by just looking forward. And then it is easier to build binary SUB/ADD after the lexing pass and then optimize the leftover unary Minus+Num tokens.

Makes a lot of sense now that I fully understand it.

engine/sim/sc_expressions.cpp
nuoHep: Added a comment for why we use `%`/`%%` for div/mod in b941c44

engine/sim/sc_expressions.cpp
nuoHep: Right, outside of "linguistic purity" reasons, this is mostly about making the lexer simpler.

It's mostly a trade-off tbh. You still need to handle the "duality" of `-` (the fact that it's unary/binary depending on context) somewhere. It just so happens that handling it outside of the lexer is simpler - you have to handle the duality of `+` too and parsing is typically structured in such a way there is implicit "lookbehind".

If we used a more sophisticated parser this would be pretty obvious.

scamille: Thanks again for all the explanations.

PR #5269, opened 2020-09-13
[Priest] rework Shadowflame Prism

seanpeters86:
issue https://github.com/WarcraftPriests/sl-shadow-priest/issues/77

This adds support for the Shadowflame Prism legendary, along with a few helper functions to address some common use cases.
- Support for `pet.fiend.X` in the APL to reference either Mindbender or Shadowfiend depending on your talent
- Fixes some errors with `base.distance` for Shadow specifically
- Adds `get_current_main_pet()` for checking if the actor is currently using Mindbender or not, this is reused in a few places in the module
- Adds Shadowflame Prism
  - includes a `trigger_shadowflame_prism` method we can call to easy trigger this from Mind Blast and Shadow Word: Death (TODO: maybe we can trigger this from spell data instead of hard coding?)
  - Shadowflame Prism has an impact action for Shadowflame Rift that way we keep the spell data intact
  - Also increases the active pet by 1.5s each time this is called
  - APL line added for SW:D to be used if either bender/shadowfiend is active and you are using the legendary


engine/class_modules/priest/sc_priest.cpp
```diff
@@ -2034,12 +2033,24 @@ void priest_t::arise()
   buffs.whispers_of_the_damned->trigger();
 }
 
+void priest_t::trigger_shadowflame_prism( player_t* target )
```
scamille, line 2110: Some things which should probably still be improved here:
1. Check if current_pet is nullptr or not. Either add an assert, or if it is possible to trigger shadowflame prism without having a main pet (eg. no summon lines in the APL) just return without doing anything.

2. Move the assert current_pet->shadowflame_prism inside the !sleeping condition.

3. Why did you remove the set_target line? I would assume it works again after your other fixes.

engine/class_modules/priest/sc_priest.hpp
```diff
@@ -983,30 +982,48 @@ struct fiend_melee_t : public priest_pet_melee_t
...
+  }
+};
+
 // ==========================================================================
 // Shadowflame Prism
 // ==========================================================================
 struct shadowflame_prism_t final : public priest_pet_spell_t
 {
-  double shadowflame_increase;
+  double duration;
```
scamille: please use timespan_t here. That way the units are handled at the point of initialization and not during usage of the value.

engine/class_modules/priest/sc_priest.hpp
```diff
@@ -983,30 +982,48 @@ struct fiend_melee_t : public priest_pet_melee_t
...
+  void execute() override
   {
-    base_dd_min = base_dd_max = ( original_amount * shadowflame_increase );
-    player->sim->print_debug( "Triggered shadowflame prism damage on target {} at {} percent increase.", *target,
-                              shadowflame_increase );
+    priest_pet_spell_t::execute();
 
-    set_target( target );
-    execute();
+    auto current_pet = p().o().talents.mindbender->ok() ? p().o().pets.mindbender : p().o().pets.shadowfiend;
```
scamille: You should be able to use the get_current_main_pet() function here. Also check if the current_pet variable you retrieve is null or not.

scamille (changes requested):

lgkern (approved):

PR #5276, opened 2020-09-15
[Death Knight]: First pass for Unholy 

Khazakdk:
Implemented the following while trying to help Taeznak get his APL to work.  The APL changes below can be ignored since we'll just blow them away and replace them when Taez is ready to merge his barebones APLs.  

- ~Rank 2 for Death Coil~ (Reworked last build, done by Melekus), Festering Strike, Scourge Strike
- Renamed all references to Unholy Frenzy and implemented Unholy Frenzy's as a melee attack with wound application
- Reworked Outbreak to reflect its new behavior. (Copied Howling Blast's logic since beta Outbreak behaves like HB on live, press button, dot applies in a radius)


engine/class_modules/sc_death_knight.cpp
```diff
@@ -6115,7 +6060,7 @@ struct scourge_strike_t : public scourge_strike_base_t
     scourge_strike_shadow( new scourge_strike_shadow_t( p ) )
   {
     parse_options( options_str );
-
+    base_multiplier *= 1.0 + p -> spec.scourge_strike_2 -> effectN( 1 ).percent();
```
Khazakdk, line 6139: I hoped applying the multiplier to the base alone would be enough but it wasn't applying to the shadow component so I had to add the multiplier to each component.

scamille (approved):

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4364,6 +4361,13 @@ struct death_coil_t : public death_knight_spell_t
     p() -> cooldown.dark_transformation -> adjust( -timespan_t::from_seconds(
       p() -> spec.death_coil -> effectN( 2 ).base_value() ) );
 
+    // Reduces the cooldown on Apocalypse by 1.5s
+    if ( p() -> spec.death_coil_2 -> ok() )
+    {
+      p()->cooldown.apocalypse->adjust(
+        -timespan_t::from_seconds( p() -> spec.death_coil -> effectN( 2 ).base_value() / 10) );
```
Armadk: I think this should be / 1000 instead of / 10

While we are here at this section, we may want to patch the cooldown for dark_transformation from r1 as well, and apply it to apocalypse instead.  Though both of these may change this week when they revert the dark transformation merge.

As a final note, does rank 1 and rank 2 of death coil both apply a CD reduction?  It looks like in most spells things such as this were moved to the second spell, not duplicated to it.

scamille: if it is / 1000.0 then it is likely milliseconds and not seconds, and you can just use effectN( 2 ).time_value()

engine/class_modules/sc_death_knight.cpp
Melekus: Don't forget clawing shadows, it's also in the list of spells affected by rank 2.

engine/class_modules/sc_death_knight.cpp
Khazakdk: For whatever reason the cd reduction value is stored in the spell data for rank 1, not rank 2.  The spell data for Rank 2 just has a single blank dummy effect.  I'll touch Apoc and DT in a week or two after Blizz maybe figures out what they're doing with those spells.

scamille: We have seen that for many other rank2 spells as well. Quite a lot of the spell data for new stuff is horrible in shadowlands, in various ways.

But I think we just have to deal with it.

Melekus:
Looks good, I'll hold onto this PR until this week's beta/ptr patch to see if apoc/dt have to be changed

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4364,6 +4361,12 @@ struct death_coil_t : public death_knight_spell_t
     p() -> cooldown.dark_transformation -> adjust( -timespan_t::from_seconds(
       p() -> spec.death_coil -> effectN( 2 ).base_value() ) );
 
+    // Reduces the cooldown on Apocalypse by 1.5s
```
Melekus: With the return of DT, Apocalypse doesn't look like it's benefitting from death coil rank 2 on PTR.

Khazakdk: Yeah I'm gonna check beta as well but it looks like death coil 2 now applies to DT instead of Apoc.  I'm gonna go ahead and finish out the DT and Apoc changes/ranks today since it looks like they're maybe done with the reworks.

PR #5281, opened 2020-09-16
[Priest] DE can proc on cast or tick

seanpeters86:
Dissonant Echoes conduit can now proc on cast or on any tick. This duplicates the code put in the `execute()` block in the `tick()` as well since `mind_flay` does not have a tick_zero.


scamille:
Why are you not just creating a function to share the code?

seanpeters86:
@scamille that would require me to be intelligent

lgkern (approved):

scamille (approved):

PR #5284, opened 2020-09-19
[shaman] Implement Chains of Devastation legendary

wfarr:
Upon casting Chain Lighting, the player gets a buff that makes the
next Chain Heal instant cast.

Upon casting Chain Heal, the player gets a buff that makes the
next Chain Lightning instant cast.

I'm not able to test this on beta/PTR at the moment, so spell IDs
on the buffs are sourced from wowhead and should be confirmed first.


wfarr:
cc @bloodmallet @hawkcorrigan

scamille (approved):

PR #5285, opened 2020-09-19
Implement Static discharge

HawkCorrigan:
@wfarr @Bloodmallet 
Missing the Lightning Shield Implementation for Ele for now


engine/class_modules/sc_shaman.cpp
```diff
@@ -4905,9 +4907,53 @@ struct static_discharge_t : public shaman_spell_t
...
+    target_cache.is_valid          = false;
+    std::vector<player_t*> targets = target_list();
+    auto it                        = std::remove_if( targets.begin(), targets.end(), [ this ]( player_t* target ) {
+      return !this->td( target )->dot.flame_shock->is_ticking();
+    } );
+
+    targets.erase( it, targets.end() );
+    if ( targets.size() == 0 )
+      return;
+    player_t* target = targets[ static_cast<int>( p()->rng().range( 0, static_cast<double>( targets.size() ) ) ) ];
```
scamille: You should be able to directly call rng.range(targets.size()); to give you a number between 0 to targets.size()-1 inclusive.
There is a helper function in rng for exactly doing this without that much casting.

engine/class_modules/sc_shaman.cpp
```diff
@@ -4905,9 +4907,53 @@ struct static_discharge_t : public shaman_spell_t
...
+  { 
+    target_cache.is_valid          = false;
+    std::vector<player_t*> targets = target_list();
+    auto it                        = std::remove_if( targets.begin(), targets.end(), [ this ]( player_t* target ) {
+      return !this->td( target )->dot.flame_shock->is_ticking();
+    } );
+
+    targets.erase( it, targets.end() );
+    if ( targets.size() == 0 )
+      return;
```
scamille, line 5003: I think adding a print_debug statement for the case where no targets are available could help a lot here for any potential problems.

If this can happen in normal settings and not just edge cases you might even want to set up a proc_t for tracking how often static discharge got wasted because no target was available.

PR #5286, opened 2020-09-20
[Priest] mastery should only apply to certain spells

### LINKED ISSUE #47: Verify what mastery affects once it is more stable
Right now in simc we are treating the mastery as a flat modifying to all damage spells...not just priest ones. This is hard to test in game because the mastery is so buggy and inconsistent, so we should follow up with this later.

If Mastery continues to not work with trinkets but does work for all priest spells we need to verify if this is truely _all_ of our spells or not.

ACs:
- [ ] Check how Mastery interacts with trinkets/spells

### LINKED ISSUE #76: Eternal Call to the Void Rework
ACs:
- [x] Legendary reworked to work with Mind Sear to spawn void lasher
- [x] change threshold for mind sear when using the leggo, dont need the adjustment anymore
- [ ] change implementation to be a dot instead of direct damage (Base_td && tick_time && dot duration all not equal to 0)
- [x] un-hardcode spell data when fixed

seanpeters86:
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/76
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/47

Unfortunately there is no `affected_by` that applies to our new mastery, so we have to hard code this in every spell. On top of this, the mastery currently does not affect Mindbender/Shadowfiend actions but DOES affect EcttV Mind Flay's and Mind Sear's. As of now it also does not affect Covenant Abilities.

This PR also fixes an error with EcttV Mind Sear's not being a proper tick action and causing a crash and also reverts a previous commit that safe guarded some time to die checks since the bug was fixed.

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -992,8 +992,9 @@ struct void_tendril_mind_flay_t final : public priest_pet_spell_t
     : priest_pet_spell_t( "mind_flay", &p, p.o().find_spell( 193473 ) ),
       void_tendril_insanity( p.o().find_spell( 336214 ) )
   {
-    channeled    = true;
-    hasted_ticks = false;
+    channeled                  = true;
+    hasted_ticks               = false;
+    affected_by_shadow_weaving = true;
 
     // Merge the stats object with other instances of the pet
     auto first_pet = p.o().find_pet( p.name_str );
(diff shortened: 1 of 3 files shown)


scamille:
I think you should override composite_target_da/ta_multiplier functions instead.

lgkern:
I think we tried that, but ended up moving from it because we are flagging the actions themselves for mastery, so we weren't able to access that information from composite_target_da

seanpeters86:
yeah @scamille we only didnt do that because we wanted access to the `priest_spell_t` boolean that we set on every spell. Is there a way to do that with composite_target_da/ta_multiplier functions instead?

scamille:
I am talking about action_t functions not player_t ones.

I don't see why you can't override them while also accessing the priest_spell_t flag.

seanpeters86:
Okay refactored to using the composite_target mults instead

scamille (approved):

PR #5288, opened 2020-09-21
[Death Knight]: Sudden Doom, Army of the Damned Changes

Khazakdk:


engine/class_modules/sc_death_knight.cpp
```diff
@@ -3222,7 +3220,7 @@ struct apocalypse_t : public death_knight_melee_attack_t
       p() -> pets.apoc_ghouls.spawn( summon_duration, n_wounds );
     }
 
-    if ( p() -> azerite.magus_of_the_dead.enabled() )
```
Khazakdk, line 3225: Not 100% sure these ok() calls are the right way to go.  But since the duration is on a spell id separate from the talent id I wanted to make sure there was a reference to the talent for easy clean up if the talent is removed.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4801,6 +4799,14 @@ struct epidemic_t : public death_knight_spell_t
     aoe = 0;
   }
 
+  double cost() const override
```
Khazakdk, line 4802: Copied from Death Coil

engine/class_modules/sc_death_knight.cpp
```diff
@@ -9534,7 +9542,7 @@ struct death_knight_module_t : public module_t {
 
   void register_hotfixes() const override
   {
-    hotfix::register_spell( "Death Knight", "2019-03-12", "Incorrect cooldown for Magus of the Dead's Frostbolt.", 288548 )
```
Khazakdk, line 9537: Updated the existing hotfix instead of making a new one since I assume we have no need to leave the azerite trait hotfix enabled for Shadowlands

PR #5290, opened 2020-09-21
[Warlock] Finish Power Siphon

Azevara:


engine/class_modules/warlock/sc_warlock_demonology.cpp
```diff
@@ -660,6 +668,10 @@ struct power_siphon_t : public demonology_spell_t
     if ( imps.size() > max_imps )
       imps.resize( max_imps );
 
+    //TOCHECK: As of 9/21/2020, Power Siphon is bugged to always provide 2 stacks of the damage buff regardless of imps consumed
```
scamille, line 671: Please use YYYY-MM-DD date format next time, see https://github.com/simulationcraft/simc/wiki/CodingGuidelines#basic-rules
That way there is no confusion between european and american date formats :-)

PR #5291, opened 2020-09-21
[Frost DK] Add Biting cold Legendary

Armadk:
Fixed Frozen Tempest to also gain rime when hitting 3 enemies, it was using trigger, and I had previously used set_chance to set the proc rate, so it was not triggering properly.

In this merge I also moved all the legendaries to using .ok() instead of ->ok().  enabled() is also available, and ok() just calls enabled().  It looks like most azerite traits are using .enabled() within the file.  Is there a preference between the two for the class?  If we decide on one, I'll update them all prior to this being merged in.

#### No Legendary
```
   remorseless_winter                   Count=  15.1| 20.354sec  DPE=13507| 0.00%  DPET=10750  DPR=  0  pDPS=  0
    remorseless_winter_damage            Count= 201.7|  1.476sec  DPE= 1014|18.00%  DPET=   0  DPR=  0  pDPS=690  Miss= 0.00%  Hit=   243|   144|   461  Crit=   495|   289|   922|16.16%
   howling_blast                        Count=  36.4|  8.415sec  DPE= 8594| 9.88%  DPET=6943  DPR=  0  pDPS=377  Miss= 0.00%  Hit=   791|   120|  1865  Crit=  1646|   240|  3729|16.29%

   rime                     : start= 35.3 refresh=  0.9 interval=  8.4 trigger=  8.2 duration=  1.6 uptime= 19.09%  benefit= 96.34%
```
#### Biting Cold
```
    remorseless_winter                   Count=  15.2| 20.329sec  DPE=16615| 0.00%  DPET=13223  DPR=  0  pDPS=  0
    remorseless_winter_damage            Count= 207.7|  1.431sec  DPE= 1213|19.39%  DPET=   0  DPR=  0  pDPS=850  Miss= 0.00%  Hit=   289|   166|   533  Crit=   590|   332|  1066|16.22%
   howling_blast                        Count=  45.3|  6.595sec  DPE= 9375|11.99%  DPET=7556  DPR=  0  pDPS=525  Miss= 0.00%  Hit=   770|   294|  1865  Crit=  1596|   588|  3751|16.55%

   rime                     : start= 44.5 refresh=  3.2 interval=  6.6 trigger=  6.2 duration=  1.5 uptime= 22.09%  benefit= 97.67%
```

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -422,6 +422,7 @@ public:
   // Counters
   double eternal_rune_weapon_counter;
   bool triggered_frozen_tempest;
+  bool triggered_biting_cold;
   unsigned int km_proc_attempts;
 
   // Enemies affected by festering wounds
@@ -907,6 +908,7 @@ public:
     // Blood
 
     // Frost                                      // bonus_id
+    item_runeforge_t biting_cold;                 // 6945
     item_runeforge_t koltiras_favor;              // 6944
     item_runeforge_t rage_of_the_frozen_champion; // 7160
 
@@ -932,6 +934,7 @@ public:
     runeforge_expression_warning( false ),
     eternal_rune_weapon_counter( 0 ),
     triggered_frozen_tempest( false ),
+    triggered_biting_cold( false ),
     km_proc_attempts( 0 ),
     festering_wounds_target_count( 0 ),
     antimagic_shell( nullptr ),
@@ -5917,10 +5920,12 @@ struct raise_dead_t : public death_knight_spell_t
 struct remorseless_winter_damage_t : public death_knight_spell_t
 {
   double frozen_tempest_target_threshold;
+  double biting_cold_target_threshold;
 
   remorseless_winter_damage_t( death_knight_t* p ) :
     death_knight_spell_t( "remorseless_winter_damage", p, p -> spec.remorseless_winter -> effectN( 2 ).trigger() ),
-    frozen_tempest_target_threshold( 0 )
+    frozen_tempest_target_threshold( 0 ),
+    biting_cold_target_threshold( 0 )
   {
     background = true;
     aoe = -1;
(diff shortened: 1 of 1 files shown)


scamille:
@navv1234 those questions about the new item_runeforge_t class are best answered by you. If it is true that item_runeforge_t::ok() and its spell data ok() do not match that is also a bit dangerous in my opinion, but I am not sure if that is really the case here.

navv1234:
Ok() should be have identically, whether it's by the spell_data_t or from item_runeforge_t. If that's not the case then there's an issue somewhere that needs to be fixed.

Looking at the code, i'm not sure how it could enable the effect without the correct bonus id. In these cases, the item_runeforge_t structure is initialized with a spell_data_t::not_found() object, and no item is associated with the object.

item_runeforge-t::ok() should return false in this case. item_runeforge_t::operator-> should return the not_found spell data object, which would also have ok() return false.

Armadk:
I may not have been clear enough in my description.  They are all the same in the end and do the same thing.  It's more just for consistencies sake within the file.  I'd rather decide on one of the 3 available to use throughout the file for all legendaries to avoid confusion for anyone new coming in in the future.

navv1234:
I suspect it's more clear for the class module developers to use "ok()" or "enabled()" from the `item_runeforge_t`.

Using "->" actually uses the operator overload for -> which is somewhat esoteric. The intent for the accessor is to allow class module developers to easily use `item_runeforge_t` in conjunction with actions/buffs (where you want to either pass a `spell_data_t` pointer to the constructor, or you want to access a value in the spell data associated with the runeforge).

Also, using the accessors from `item_runeforge_t` allows us to change the enable functionality, for example if we want to add an expliclt "disable all item runeforges" option.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5947,6 +5957,11 @@ struct remorseless_winter_damage_t : public death_knight_spell_t
 
     m *= 1.0 + p() -> buffs.gathering_storm -> stack_value();
 
+    if ( p() -> legendary.biting_cold.ok() )
```
Melekus: You can even apply the modifier directly on base_multiplier during construction.

PR #5293, opened 2020-09-22
[Priest] rework Devouring Plague

### LINKED ISSUE #80: Devouring Plague Rework
Devouring Plague now handles refreshes considerably differently than what was there before. After chatting with folks in the simc discord we probably need something new to handle this.

Quick overview of how this works:
```
3 ticks without lust and no overlap
4 ticks without lust and overlap
5 ticks with lust and overlap
7 ticks with void torrent and lust
```

### Logs
- [A few examples](https://www.warcraftlogs.com/reports/ymtL71gdDQT2bJFR/#fight=last&type=damage-done&source=7&pins=0%24Separate%24%23244F4B%24damage%240%240.0.0.Any%241987432.0.0.Priest%24true%24144085.0.0.NPC%24false%24263165&ability=335467&view=events)
- [Some examples with lust](https://www.warcraftlogs.com/reports/QCayFgK7342Dmv1b/#fight=last&type=damage-done&source=1&pins=0%24Separate%24%23244F4B%24damage%240%240.0.0.Any%241987432.0.0.Priest%24true%240.0.0.Any%24false%24263165%5E0%24Separate%24%23909049%24auras-gained%240%240.0.0.Any%240.0.0.Any%24true%241987432.0.0.Priest%24false%24178207&ability=335467&view=events)

## Important Things to note
- The debuff on the target does seem to be "extending" to pandemic duration, but not changing the overall damage. This is potentially important because of Mastery/other things that rely on "is DP active on the target"
- Void Torrent is seemingly causing an extra tick or two in addition to the existing one with the ignite/refresh behavior.
- Devouring Plague ticks are still hasted, meaning that with lust we are getting more ticks (this is not how other ignite things work in game to my knowledge)

## Questions to answer
- What happens if you DP > wait for max pandemic > DP > Lust, does the damage from the first part now increases because of lust?

## Okay so wtf do we do?
Great question. We probably want to create something similar to ignite, which is using `residual_action` right now. Probably call it `hasted_residual_action` or something like that. So we should start by cloning residual action and then adding pieces into Devouring Plague.
- [ ] create `hasted_residual_action`
- [ ] remove DoT component from existing `devouring_plague_t` and add a tick action that triggers a `devouring_plague_dot` `hasted_residual_action`. This will need to be done through a helper trigger function, as a residual action cannot be executed directly
- [ ] ensure that the dot duration and dot ticks are matching in game (with and without haste increases)
- [ ] implement a way for Void Torrent to magically appear with more ticks? (honestly i have no idea here)

### LINKED ISSUE #86: SW:D should do more damage in execute
Need to check target health when casting SW:D and if they are below X% add the modifier

### LINKED ISSUE #91: Check void tendril insanity
insanity value is reported to not line up with ticks

lgkern:
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/86
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/80
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/91

Early DP implementation.

Certainties:
- When you cast DP with a DP already up, it will preserve Time to Next Tick and the Partial Tick and add it to the duration of the new dot
- Remaining ticks get added to the damage of the new DP

Assumptions:
- Damage added by extra ticks is divided evenly between ticks and partial ticks at the moment of application.

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -386,8 +386,13 @@ struct painbreaker_psalm_t final : public priest_spell_t
 
 struct shadow_word_death_t final : public priest_spell_t
 {
+  double execute_percent;
+  double execute_modifier;
+
   shadow_word_death_t( priest_t& p, util::string_view options_str )
-    : priest_spell_t( "shadow_word_death", p, p.find_class_spell( "Shadow Word: Death" ) )
+    : priest_spell_t( "shadow_word_death", p, p.find_class_spell( "Shadow Word: Death" ) ),
+      execute_percent( data().effectN( 2 ).base_value() ),
+      execute_modifier( data().effectN( 3 ).percent() )
   {
     parse_options( options_str );
 
@@ -414,6 +419,22 @@ struct shadow_word_death_t final : public priest_spell_t
       energize_resource = RESOURCE_INSANITY;
       energize_amount   = 10;
     }
+  }
+
+  double composite_da_multiplier( const action_state_t* state ) const override
+  {
+    double d = priest_spell_t::composite_da_multiplier( state );
+
+    if ( target->health_percentage() < execute_percent )
+    {
+      if ( sim->debug )
+      {
+        sim->print_debug( "{} below {}. Increasing Shadow Word: Death damage by {}", state->target->name_str, execute_percent, execute_modifier );
+      }
+      d *= 1 + execute_modifier;
+    }
+
+    return d;
   }
 
   void impact( action_state_t* s ) override
@@ -885,18 +906,98 @@ struct vampiric_touch_t final : public priest_spell_t
 // ==========================================================================
 // Devouring Plague
 // ==========================================================================
-struct devouring_plague_t final : public priest_spell_t
+struct devouring_plague_dot_t final : public priest_spell_t
 {
-  bool casted;
+  double rolling_dp_bonus;
 
-  devouring_plague_t( priest_t& p, bool _casted = false )
-    : priest_spell_t( "devouring_plague", p, p.find_class_spell( "Devouring Plague" ) )
+  devouring_plague_dot_t( priest_t& p )
+    : priest_spell_t( "devouring_plague", p, p.find_class_spell( "Devouring Plague" ) ), rolling_dp_bonus( 0.0 )
   {
-    casted                     = _casted;
     may_crit                   = true;
     tick_zero                  = false;
     tick_may_crit              = true;
     affected_by_shadow_weaving = true;
+    background                 = true;
+    hasted_ticks               = true;
+
+    base_dd_min = base_dd_max = spell_power_mod.direct = base_td = 0.0;
+  }
+
+  void init() override
+  {
+    priest_spell_t::init();
+
+    snapshot_flags |= STATE_CRIT | STATE_TGT_CRIT | STATE_HASTE;
+    update_flags |= STATE_CRIT | STATE_TGT_CRIT | STATE_HASTE;
+  }
+
+  // Make sure the dot doesn't also consume insanity
+  void consume_resource() override
+  {
+    return;
+  }
+
+  // TODO: see if we need this
+  // timespan_t composite_dot_duration( const action_state_t* s ) const override
+  // {
+  //   return dot_duration * ( tick_time( s ) / base_tick_time );
+  // }
+
+  void append_damage( player_t* target )
+  {
+    dot_t* dot = get_dot( target );
+    // get remaining full ticks left (not partials, these will be rolled over)
+    // calculate total damage
+    timespan_t remaining_dot = dot->remains();                                       // 2.5s
+    remaining_dot            = remaining_dot - dot->time_to_next_tick();             // 2.5 - .5s = 2s
+    int num_full_ticks = as<int>( std::ceil( remaining_dot / dot->time_to_tick ) );  // 2s / ~1.5 = 1.3 = 1 full tick
+
+    double total_damage = priest().tick_damage_over_time( num_full_ticks * dot->time_to_tick, dot );
+    double total_ticks  = dot->duration() / dot->time_to_tick;
+    if ( dot->time_to_next_tick() > timespan_t::from_seconds( 0 ) )
+    {
+      total_ticks += 1;
+    }
+
+    // TODO: make serge happy later
+    assert( 1 );
+
+    double partial_tick_coeff = ( remaining_dot - ( num_full_ticks * dot->time_to_tick ) ) /
+                                dot->time_to_tick;                           // 2s - ( 1 * 1.5 ) = .5 / 1.5 =~ 0.33
+    rolling_dp_bonus = total_damage / ( total_ticks + partial_tick_coeff );  // X / (( 1 + ticks_from_new_dp ) + 0.33)
+  }
+
+  double bonus_ta( const action_state_t* state ) const override
+  {
+    double t = priest_spell_t::bonus_ta( state );
+
+    if ( rolling_dp_bonus )
+    {
+      t += rolling_dp_bonus;
+    }
+
+    return t;
+  }
+};
+
+struct devouring_plague_t final : public priest_spell_t
+{
+  bool casted;
+  devouring_plague_dot_t* dot_spell;
+
+  devouring_plague_t( priest_t& p, bool _casted = false )
+    : priest_spell_t( "devouring_plague", p, p.find_class_spell( "Devouring Plague" ) ),
+      dot_spell( new devouring_plague_dot_t( p ) )
+  {
+    casted                     = _casted;
+    may_crit                   = true;
+    affected_by_shadow_weaving = true;
+
+    add_child( dot_spell );
+
+    // Turn off DoT
+    dot_duration = timespan_t::from_seconds( 0 );
+    base_td = base_td_multiplier = 0;
   }
 
   devouring_plague_t( priest_t& p, util::string_view options_str ) : devouring_plague_t( p, true )
@@ -929,10 +1030,26 @@ struct devouring_plague_t final : public priest_spell_t
   {
     priest_spell_t::impact( s );
 
+    dot_spell->rolling_dp_bonus = 0.0;
+
     // Damnation does not trigger a SA - 2020-08-08
     if ( casted )
     {
       priest().trigger_shadowy_apparitions( s );
+    }
+
+    if ( result_is_hit( s->result ) )
+    {
+      priest_td_t& td = get_td( s->target );
+      auto dp         = td.dots.devouring_plague;
+
+      if ( dp->is_ticking() )
+      {
+        dot_spell->append_damage( s->target );
+      }
+
+      dot_spell->target = s->target;
+      dot_spell->execute();
     }
   }
 };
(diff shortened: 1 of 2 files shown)


scamille (commented):
Ok so about the rolling dps bonus variable:

There could be problems with it when you cast multiple DP to e.g. 2 targets. I also think that if you cast 3 DP back to back you would get incorrect values if you are just resetting it to 0 on the third application.

So it might be worth it/necessary to use a custom action_state_t here.

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -885,18 +906,98 @@ struct vampiric_touch_t final : public priest_spell_t
...
+  void append_damage( player_t* target )
+  {
+    dot_t* dot = get_dot( target );
+    // get remaining full ticks left (not partials, these will be rolled over)
+    // calculate total damage
+    timespan_t remaining_dot = dot->remains();                                       // 2.5s
+    remaining_dot            = remaining_dot - dot->time_to_next_tick();             // 2.5 - .5s = 2s
+    int num_full_ticks = as<int>( std::ceil( remaining_dot / dot->time_to_tick ) );  // 2s / ~1.5 = 1.3 = 1 full tick
+
+    double total_damage = priest().tick_damage_over_time( num_full_ticks * dot->time_to_tick, dot );
```
scamille: Why are you not just passing remaining_dot to tick_damage_over_time function here? Does the num_full_ticks calculation really make a difference?

And if it does: what happens when between the last tick of the dot where time_to_tick was calculated and now you gain/loose haste? Does it still take the last calculated time_to_tick or do you need to re-evaluate that with current haste?

scamille (commented):
What exactly was the reason again why residual action did not match DP behavior?

engine/class_modules/priest/sc_priest_shadow.cpp
lgkern: We're not passing the remaining time because that's not how it is implemented in-game. Currently, only remaining full ticks become increase in power for the next DP. Time_to_next_tick and the leftover partial tick get added as increased time to the new Devouring Plague. 

I.e. you cast a second Devouring Plague and there is 1 second to next tick, a full 1.5s tick, and a 0.4s long partial tick left on the first one. What happens is that full tick increases the damage of the ticks of the new Devouring Plague, distributed evenly between the next ticks. The 1s to next tick and the 0.4s partial tick get added as extra duration to the new Devouring Plague.

Haste in this mess is... complicated, we haven't observed it enough (because the in-game implementation has some bugs to say the least) to know exactly what haste entails here, but I believe it will just change future tick frequency without changing the new dot duration.

lgkern:
> Ok so about the rolling dps bonus variable:
> 
> There could be problems with it when you cast multiple DP to e.g. 2 targets. I also think that if you cast 3 DP back to back you would get incorrect values if you are just resetting it to 0 on the third application.
> 
> So it might be worth it/necessary to use a custom action_state_t here.

Hmm, good point. How can we go about it here?

scamille:
I'll add something. Action state is a bit tricky to get right (not sure if I will succeed in the first try :D)

Do you have a test profile which does result in some DP refreshes, ideally in the first iteration when run with debug=1 ?

Edit: I pushed something. Not sure if it works at all. you should be able to see the ```rolling_dp_bonus=``` line in the debug log.

seanpeters86:
Simple test profile that does two DPs back to back:

```
priest="Base"
level=60
role=spell
position=back
talents=3121123
spec=shadow
covenant=kyrian

default_actions=0

actions=devouring_plague,if=insanity=100|dot.devouring_plague.ticking
actions+=/mind_flay

# head=confidants_favored_cap,id=183021,ilevel=226,bonus_id=6982
head=confidants_favored_cap,id=183021,ilevel=226,bonus_id=6983
neck=charm_of_eternal_winter,id=183040,ilevel=226
shoulder=shawl_of_the_penitent,id=183020,ilevel=233
back=crest_of_the_legionnaire_general,id=183032,ilevel=233
chest=robes_of_the_cursed_commando,id=182998,ilevel=233
wrists=acolytes_velvet_bindings,id=183017,ilevel=226
hands=impossibly_oversized_mitts,id=183022,ilevel=226
waist=shadewarped_sash,id=183004,ilevel=226
legs=leggings_of_lethal_reverberations,id=182981,ilevel=226
feet=slippers_of_the_forgotten_heretic,id=182979,ilevel=226
finger1=most_regal_signet_of_sire_denathrius,id=183036,ilevel=233
finger2=ritualists_treasured_ring,id=183037,ilevel=226
trinket1=vial_of_vampiric_essence,id=178810,ilevel=187
trinket2=soulletting_ruby,id=178809,ilevel=187
main_hand=sinbearers_absolution_staff,id=182418,ilevel=233
```

engine/class_modules/priest/sc_priest_shadow.cpp
scamille: Ok. I also somehow simplified the calculation in my commit, since there already is a dot state from which to calculate the tick_amount.

scamille:
The question about 3 DPs in a row is definitely a interesting one and should be resolved before we accidentally hit it in some bloodlust/voidform/legendary scenario.

Right now I think the code will add the remaining tick damage from DP1 onto DP2 ticks. and Then it will add the remaining tick damage from DP2 onto DP3, which will include the damage portion per tick from DP1 which made it into DP2.

lgkern:
That should be how it works. Can't test in-game because it is broken, casting a 3rd DP just eats off the first one right now.

image

seanpeters86:
Current status of back to back DP:

```
23.493 Player 'Base' schedules execute for Action devouring_plague
23.493 Player 'Base' direct amount for Action devouring_plague: amount=1452
23.493 Player 'Base' starts Dot devouring_plague on Player 'Fluffy_Pillow' with duration=6.000
23.493 New DoT Tick Event: Player 'Base' Dot devouring_plague tick 1-of3 time_to_tick=2.426

24.711 Player 'Base' schedules execute for Action devouring_plague
24.711 Player 'Base' direct amount for Action devouring_plague: amount=1636

24.711 Player 'Base' tick amount for Action devouring_plague on Player 'Fluffy_Pillow': amount=1161
24.711 Player 'Base' Action devouring_plague appended 788.7549808083191 bonus damage from previous dot. Bonus damage is now at 788.7549808083191
24.711 New DoT End Event: Player 'Base' Dot devouring_plague time_to_end=7.800
24.711 Player 'Base' refreshes Dot devouring_plague (1) on Player 'Fluffy_Pillow'. old_remains=4.782 refresh_duration=6.000 duration=7.800


tick 1/4: 25.919 Player 'Base' tick amount for Action devouring_plague on Player 'Fluffy_Pillow': amount=1309
tick 2/4: 28.345 Player 'Base' tick amount for Action devouring_plague on Player 'Fluffy_Pillow': amount=1309
tick 3/4: 30.771 Player 'Base' tick amount for Action devouring_plague on Player 'Fluffy_Pillow': amount=1309
tick 4/4: 32.511 Player 'Base' tick amount for Action devouring_plague on Player 'Fluffy_Pillow': amount=938
```

I think this is correct, the only thing I don't quite understand is this log:

```
24.711 Player 'Base' tick amount for Action devouring_plague on Player 'Fluffy_Pillow': amount=1161
```

I don't think the target actually loses health here (it shouldnt have ticked at all yet) and I think this is just telling us the base tick amount but not actually damaging the target.

scamille:
I added some more debug statements.

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -973,48 +973,38 @@ struct devouring_plague_dot_t final : public priest_spell_t
...
+    timespan_t existing_duration = dot->remains();
+    priest_spell_t::trigger_dot( s );
 
-    double partial_tick_coeff = ( remaining_dot - ( num_full_ticks * dot->time_to_tick ) ) /
-                                dot->time_to_tick;  // 2s - ( 1 * 1.5 ) = .5 / 1.5 =~ 0.33
+    // Calculate the modifier for the future ticks
+    auto dot_state           = debug_cast<devouring_plague_dot_state_t*>( dot->state );
+    double existing_modifier = dot_state->rolling_dp_modifier;
+    dot_state->rolling_dp_modifier =
+        ( existing_duration * existing_modifier + composite_dot_duration( s ) ) / dot->remains();
```
lgkern: This line right now isn't removing the next tick or the partial tick, which was calculated before. Probably need to add it back. 

A good test for this is to do DP > GCD > GCD > DP, which should lead just to more ticks, but no damage increase.

scamille:
@vituscze can you double-check the current implementation please?

@lgkern was there a reason to not use just the #full_ticks for the modifier calculation?

If there is nothing major remaining to be fixed we can also just merge this at some point. I assume shadow priests will want to create simulation charts eventually with it?
It is always possible that Blizzard adds some more changes to it anyway or the understanding of haste on it might get refined, at which point this can be improved upon.

lgkern:
@scamille yea it needs to use #full_ticks, I'll sit down with @seanpeters86 today to do that. After we fix that it should be ok.

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1002,27 +1001,56 @@ struct devouring_plague_t final : public priest_spell_t
...
+            num_full_ticks, tick_coefficient, new_num_ticks, new_tick_coefficient );
+      }
+
+      // figure out the increase for each new tick of DP
+      multiplier = 1 + ( ( ( num_full_ticks * tick_coefficient ) / new_num_ticks ) / new_tick_coefficient );
 
-      multiplier = 2.0; // TODO Calculate multiplier here
+      if ( sim->debug )
+      {
+        sim->print_debug(
```
scamille: You can simplify this to just:
```
        sim->print_debug(
            "{} {} updates modifier per tick from previous dot. Modifier per tick went "
            "from {} to {}.",
            *player, *this, old_multiplier, multiplier );
```
Both without the extra sim->debug check (no function calls or anything else here), and without adding Devouring Plague text into it (you already get that from the action through *this).

The other debug call above can be simplified as well.

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1002,27 +1001,56 @@ struct devouring_plague_t final : public priest_spell_t
...
+            "Devouring Plague calculations - num_full_ticks: {}, tick_coefficient: {}, new_num_ticks: {}, "
+            "new_tick_coefficient: {}",
+            num_full_ticks, tick_coefficient, new_num_ticks, new_tick_coefficient );
+      }
+
+      // figure out the increase for each new tick of DP
+      multiplier = 1 + ( ( ( num_full_ticks * tick_coefficient ) / new_num_ticks ) / new_tick_coefficient );
```
scamille: You might want to split this up into 3 steps to make it easier to comprehend and maintain.
First the ratio from how old ticks are distributed onto the new ones.
Then how coefficient ratio affects that.
And just when assigning adding the 1 to ensure the new dot also gets its base multiplier.

scamille (commented):
As far as I can tell this looks ok.
Did you discuss the multplier adjustments by the coefficient ratio with @vituscze ?
I am really not 100% sure if this is necessary here, since I thought the whole multplier thing instead of direct damage bonus would take care of it. You also did not clearly document / write in your comment why this is now necessary.

seanpeters86:
@scamille @lgkern @vituscze if im understanding the problem space correctly, we had to convert to use this modifier so that we respond correctly to the actor dynamically gaining intellect/SP in the middle of a dot tick. Based on what we've seen in game uses a modifier to achieve this exact behavior, so we are lining up with what blizzard is doing.

seanpeters86 (approved):

PR #5294, opened 2020-09-22
[Warlock] Implement From the Shadows

Azevara:


engine/class_modules/warlock/sc_warlock_demonology.cpp
```diff
@@ -69,6 +69,20 @@ struct demonology_spell_t : public warlock_spell_t
   double composite_target_multiplier( player_t* t ) const override
   {
     double m = warlock_spell_t::composite_target_multiplier( t );
+
+    auto td = this->td( t );
+
+    if (td->debuffs_from_the_shadows->check() && data().affected_by( td->debuffs_from_the_shadows->data().effectN( 1 ) ) )
```
scamille, line 75: To solve this more elegantly, I would recommend the following:

- Add a boolean variable ```bool is_affected_by_from_the_shadows;``` to demology_spell_t.
- Set that variable in the demonology_spell_t constructor based on the data().affected_by call or the id for Hand of Guldan.
- in the composite_target_multiplier function just use the variable to check if you should apply to modifier.


This not only unifies the code for affected_by as well as Hand of Guldan fix, it also moves the potentially expensive affected_by call into the contructor.

Azevara: I've added a TODO referencing this PR so we can get this suggestion implemented down the line

PR #5297, opened 2020-09-22
[Feral Druid] Fix Bloodtalons

indiche:
There was an issue with Bloodtalons triggering even without meeting all
the conditions. e.g.

Rake -> Shred -> Shred

Issue is that the trigger function is called each time we give dummy bt_
buff. Using example above following would happen:

Rake -> trigger is called, we don't have any buffs so if condition
evaluates to true (1 < 3)

Shred -> trigger is called, we have bt_rake from previous rake but
condition still evaluates to true (2 < 3)

Shred -> trigger is called, we have bt_rake, and bt_shred from previous
shred so condition now evaluates to false (3 < 3) giving us bloodtalons

This change will fix that by adding one to the entire expresion only if
we already don't have the buff that triggered the call. Credit goes to
Melekus for suggesting the fix.

### DIFF
--- a/engine/class_modules/sc_druid.cpp
+++ b/engine/class_modules/sc_druid.cpp
@@ -1586,7 +1586,7 @@ struct bt_dummy_buff_t : public druid_buff_t<buff_t>
 
   bool trigger( int s, double v, double c, timespan_t d ) override
   {
-    if ( !p().talent.bloodtalons->ok() || p().buff.bloodtalons->check() )
+    if ( !p().talent.bloodtalons->ok() )
       return false;
 
     if ( p().buff.bt_rake->check() + p().buff.bt_shred->check() + p().buff.bt_swipe->check() +
--- a/engine/class_modules/sc_druid.cpp
+++ b/engine/class_modules/sc_druid.cpp
@@ -1590,7 +1590,7 @@ struct bt_dummy_buff_t : public druid_buff_t<buff_t>
       return false;
 
     if ( p().buff.bt_rake->check() + p().buff.bt_shred->check() + p().buff.bt_swipe->check() +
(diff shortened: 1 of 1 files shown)


gastank:
I probably should've initially implemented this as an override of buff_t::start() to start with, since that is only called when the buff isn't currently up.

Declaration & definition for buff_t::start() are here:
https://github.com/simulationcraft/simc/blob/shadowlands/engine/buff/sc_buff.hpp#L405
https://github.com/simulationcraft/simc/blob/shadowlands/engine/buff/sc_buff.cpp#L1783

You could just leave `!p().talent.bloodtalons->ok()` condition in, and move the rest to a start() override. I don't know if the `p().buff.bloodtalons->check()` condition is necessary - can you proc bloodtalons again with 3 different spells while bloodtalons is already up?

indiche:
We've discussed moving it into buff_t::start() or even separate function on Discord. Melekus suggested something like this:

```cpp
druid_t::trigger_bloodtalons( buff_t* b )
{
  int current_buffs = p().buff.bt_rake->check() + p().buff.bt_shred->check() + p().buff.bt_swipe->check() +
         p().buff.bt_thrash->check() + p().buff.bt_moonfire->check() + p().buff.bt_brutal_slash->check();
  if ( ! b -> up() && current_buffs == 2 )
    p().buff.bloodtalons->trigger( p().buff.bloodtalons->max_stack() );
  else 
    b -> trigger();
}
```

I've decided not to do it until I get better understanding of the codebase.

Yes, you can proc bloodtalons while buff is already up. If you consumed a stack it will go up to 2 stack again, and if you already have 2 stacks it won't do anything.

gastank:
In that case, the `p().buff.bloodtalons->check()` condition should be removed, then we can check this in to fix the bug and revisit later if needed.

indiche:
Sure, I'll fix it later tonight.

gastank (approved):

PR #5303, opened 2020-09-23
[Warlock] Implement Rolling Agony Conduit

javedbartlett:


engine/class_modules/warlock/sc_warlock_affliction.cpp
```diff
@@ -149,6 +149,11 @@ struct agony_t : public affliction_spell_t
     pandemic_invocation_usable = false;  // BFA - Azerite
 
     dot_max_stack = as<int>( data().max_stacks() + p->spec.agony_2->effectN( 1 ).base_value() );
+
+    if ( p->conduit.rolling_agony->ok() && p->conduit.rolling_agony.value() > 0 )
+    {
+      dot_duration += timespan_t::from_millis( p->conduit.rolling_agony.value() );
+    }
```
nuoHep: You can replace this all with
```cpp
dot_duration += p->conduit.rolling_agony.time_value();
```
conduit `.value()` accessor always returns 0 for non-slotted/enabled conduits and `.time_value()` is just a wrapper around that that does `.from_millis()` for you

javedbartlett: That's great to know. Implemented the changes.

PR #5309, opened 2020-09-24
[Priest] dots auto benefit from mastery

### LINKED ISSUE #89: You get the mastery bonus when applying a dot that doesnt exist
If you dont have devouring plague active on the target, but cast a devouring plague you get the damage as if you had 1 mastery stack.

seanpeters86:
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/89

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -1178,6 +1178,7 @@ priest_t::priest_t( sim_t* sim, util::string_view name, race_e r )
     buffs(),
     talents(),
     specs(),
+    dot_spells(),
     mastery_spells(),
     cooldowns(),
     rppm(),
@@ -1721,6 +1722,11 @@ void priest_t::init_spells()
   specs.discipline = dbc::get_class_passive( *this, PRIEST_DISCIPLINE );
   specs.shadow     = dbc::get_class_passive( *this, PRIEST_SHADOW );
 
+  // DoT Spells
+  dot_spells.shadow_word_pain = find_class_spell( "Shadow Word: Pain" );
+  dot_spells.vampiric_touch   = find_class_spell( "Vampiric Touch" );
+  dot_spells.devouring_plague = find_class_spell( "Devouring Plague" );
+
   // Mastery Spells
   mastery_spells.grace          = find_mastery_spell( PRIEST_DISCIPLINE );
   mastery_spells.echo_of_light  = find_mastery_spell( PRIEST_HOLY );
@@ -2182,7 +2188,7 @@ void priest_t::remove_wrathful_faerie()
   }
 }
 
-int priest_t::shadow_weaving_active_dots( const player_t* target, const unsigned int id ) const
+int priest_t::shadow_weaving_active_dots( const player_t* target, const unsigned int spell_id ) const
 {
   int dots = 0;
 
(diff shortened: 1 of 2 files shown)


engine/class_modules/priest/sc_priest.cpp
```diff
@@ -2198,21 +2198,30 @@ int priest_t::shadow_weaving_active_dots( const player_t* target ) const
       const dot_t* vt  = td->dots.vampiric_touch;
       const dot_t* dp  = td->dots.devouring_plague;
 
-      dots = swp->is_ticking() + vt->is_ticking() + dp->is_ticking();
+      // You get mastery benefit for a DoT as if it was active, if you are actively putting that DoT up
+      // So to get the mastery benefit you either have that DoT ticking, or you are casting it
+      const spell_data_t* swp_spell = find_class_spell( "Shadow Word: Pain" );
```
scamille: definitely do not call find_class_spell here "during combat".

use some spell_data_t* variables stored in priest_t.

engine/class_modules/priest/sc_priest.hpp
```diff
@@ -583,8 +583,8 @@ struct priest_t final : public player_t
   void trigger_psychic_link( action_state_t* );
   void trigger_wrathful_faerie();
   void remove_wrathful_faerie();
-  int shadow_weaving_active_dots( const player_t* target ) const;
-  double shadow_weaving_multiplier( const player_t* target ) const;
+  int shadow_weaving_active_dots( const player_t* target, const unsigned int id ) const;
```
scamille: maybe rename this to action_id or spell_id, otherwise it is not clear what id you want here without looking at the implementation.

scamille (changes requested):

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -2198,21 +2198,30 @@ int priest_t::shadow_weaving_active_dots( const player_t* target ) const
...
+      const spell_data_t* vt_spell = find_class_spell( "Vampiric Touch" );
+      const spell_data_t* dp_spell = find_class_spell( "Devouring Plague" );
```
lgkern: This works, but there is probably a more elegant way. This is highly inefficient, as you're performing 3x 'find_class_spell' on every damage instance that benefits from Mastery.

Instead of sending the ID, maybe send the whole action pointer and do a class check if it is shadow_word_pain_t, vampiric_touch_t, or devouring_plague_t ?

lgkern (changes requested):
This works, but there is probably a more elegant way. This is highly inefficient, as you're performing 3x 'find_class_spell' on every damage instance that benefits from Mastery.

Instead of sending the ID, maybe send the whole action pointer and do a class check if it is shadow_word_pain_t, vampiric_touch_t, or devouring_plague_t ?

engine/class_modules/priest/sc_priest.cpp
scamille: using the action id is fine, as long as the lookup of spell data is moved out of here.

You just need to decide where the right place is to put those. Either just put it into priest.specs, or add a new struct.

Using the action type would require a dynamic_cast, which you can do but I would not prefer it as a solution.

seanpeters86: is there a way to pull it out of the `dot_t` object? I tried doing that but it didnt look like it populated correctly. Since we already do `const dot_t* swp = td->dots.shadow_word_pain`

EvanMichaels: > is there a way to pull it out of the `dot_t` object? I tried doing that but it didnt look like it populated correctly. Since we already do `const dot_t* swp = td->dots.shadow_word_pain`

dots.shadow_word_pain->current_action->internal_id and pass down the action->internal_id perhaps? But should be a fair amount of ways to use the current_action for comparisons.

lgkern (approved):

scamille (approved):

scamille:
I don't think dot_spell is the best name for the struct, but good enough :-)
I think we can also merge the structs active_spells and action. I'll do that at some point.

PR #5310, opened 2020-09-25
Priest eruption option

### LINKED ISSUE #90: Ascended Eruption counts you as an enemy when calculating damage
The sqrt falloff damage counts allies in the spell data, so our calculations were off.

seanpeters86:
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/90

with this we can easily tweak this in our sims how many targets it should be

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -602,14 +602,11 @@ struct ascended_eruption_t final : public priest_spell_t
 
   double composite_aoe_multiplier( const action_state_t* state ) const override
   {
-    double cam = priest_spell_t::composite_aoe_multiplier( state );
+    double cam  = priest_spell_t::composite_aoe_multiplier( state );
     int targets = state->n_targets;
-
-    if ( priest().options.priest_ascended_eruption_targets )
-    {
-      sim->print_debug( "{} {} sets damage multiplier as if it hit {} targets.", *player, *this, priest().options.priest_ascended_eruption_targets );
-      targets = priest().options.priest_ascended_eruption_targets;
-    }
+    sim->print_debug( "{} {} sets damage multiplier as if it hit an additional {} targets.", *player, *this,
+                      priest().options.priest_ascended_eruption_additional_targets );
+    targets += priest().options.priest_ascended_eruption_additional_targets;
     return cam / std::sqrt( targets );
   }
 };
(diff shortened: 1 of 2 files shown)


engine/class_modules/priest/sc_priest.cpp
```diff
@@ -603,7 +603,14 @@ struct ascended_eruption_t final : public priest_spell_t
   double composite_aoe_multiplier( const action_state_t* state ) const override
   {
     double cam = priest_spell_t::composite_aoe_multiplier( state );
-    return cam / std::sqrt( state->n_targets );
+    int targets = state->n_targets;
+
+    if ( priest().options.priest_ascended_eruption_targets )
+    {
+      sim->print_debug( "{} {} sets damage multiplier as if it hit {} targets.", *player, *this, priest().options.priest_ascended_eruption_targets );
+      targets = priest().options.priest_ascended_eruption_targets;
```
lgkern: This should be += instead of =. The option targets are added to the actual ones to replicate a raid or group env.

lgkern (changes requested):

engine/class_modules/priest/sc_priest.cpp
lgkern: and it should always add that option in. By default, it should be 10 to emulate raids (majority of sims that are run)

engine/class_modules/priest/sc_priest.hpp
```diff
@@ -414,6 +414,8 @@ struct priest_t final : public player_t
     // Fae Blessings CDR can be given to another player, but you can still get the insanity gen
     bool priest_self_benevolent_faerie = true;
 
+    // Ascended Eruption is currently bugged and counts allies as targets for the sqrt damage falloff
+    int priest_ascended_eruption_targets = 0;
```
lgkern: 10 instead of 0 to emulate a raid env

seanpeters86:
@lgkern oh I did it to set targets, not add targets. What do you prefer?

Also I think default should be for single target, since thats what most will sim for and then in our sims we can override with an option setting. Thoughts?

lgkern:
> @lgkern oh I did it to set targets, not add targets. What do you prefer?
> 
> Also I think default should be for single target, since thats what most will sim for and then in our sims we can override with an option setting. Thoughts?

While yes, most si for single target, their intent is to use it in a raid environment and not isolated in the world or solo scenarios, which is why I believe having 10 as default is the best option.

Also, add is probably the best operation, and not set, for the reasons above.

lgkern (approved):

PR #5312, opened 2020-09-25
[Priest] Void Torrent no longer refreshes DP

seanpeters86:
starts work on https://github.com/WarcraftPriests/sl-shadow-priest/issues/87

Making some assumptions until it is added in game


lgkern (commented):
With these changes are you able to VoiT>DP and get damaged carried over?

seanpeters86:
with this change in simc yes, unsure if thats how it works in game but i would assume so @lgkern

lgkern (approved):

PR #5315, opened 2020-09-25
[Warlock] Move Havoc handling to main module for Covenant usage

Azevara:


vituscze:
`destruction_spell_t::init()` should also be moved, otherwise you risk having spells with `use_havoc` which are not in the `havoc_spells` vector.

vituscze (changes requested):
`destruction_spell_t::init()` should also be moved, otherwise, you risk having spells with `use_havoc` which are not in the `havoc_spells` vector.

Azevara:
Easier to close and resubmit cleanly because man did i screw this up

PR #5316, opened 2020-09-25
[Priest] should roll over partials

seanpeters86:
Something changed or our data was wrong, Devouring Plague is now rolling over partials so we needed to adjust this calculation a bit

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -1031,16 +1031,16 @@ struct devouring_plague_t final : public priest_spell_t
       double old_multiplier   = cast_state( old_s )->rolling_multiplier;
 
       // figure out how many old ticks to roll over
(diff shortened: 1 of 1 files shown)


engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1031,16 +1031,16 @@ struct devouring_plague_t final : public priest_spell_t
       double old_multiplier   = cast_state( old_s )->rolling_multiplier;
 
       // figure out how many old ticks to roll over
-      int num_full_ticks = as<int>( std::floor( ( old_remains - time_to_tick ) / old_tick ) );
+      double num_ticks = std::floor( ( old_remains - time_to_tick ) / old_tick );
```
lgkern: Probably call this "num_munched_ticks" or something, to make it clear that those are just the ticks being added as damage.

lgkern (changes requested):
Rename tick variable for clarity.

PR #5317, opened 2020-09-26
[Priest] add bugged behavior for DP refresh

seanpeters86:


lgkern (approved):

lgkern (approved):

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1006,7 +1006,14 @@ struct devouring_plague_t final : public priest_spell_t
   timespan_t calculate_dot_refresh_duration( const dot_t* d, timespan_t duration ) const override
   {
     // when you refresh, you lose the partial tick
-    return duration + d->time_to_next_tick();
+    if ( priest().bugs )
```
scamille, line 1009: Maybe a comment here explaining both behaviors and why the first one is considered a bug would help.
Oh and and add a date to it as well.

PR #5320, opened 2020-09-26
[Elemental Shaman] Implement Ancestral Reminder

HawkCorrigan:
@simulationcraft/shaman 

Implemented Ancestral Reminder
Increases Duration of Bloodlust by 20seconds and effectiveness by 10%.


engine/class_modules/sc_shaman.cpp
```diff
@@ -6839,6 +6839,15 @@ void shaman_t::create_buffs()
   buff.stormkeeper = make_buff( this, "stormkeeper", talent.stormkeeper )
                          ->set_cooldown( timespan_t::zero() );  // Handled by the action
 
+  if ( legendary.ancestral_reminder->ok() )
+  {
+      
+    auto legendary_spell = find_spell( 336741 );
+    auto buff            = buffs.bloodlust;
+    buff->modify_duration( timespan_t::from_millis(legendary_spell->effectN( 1 ).base_value()) );
+    buff->modify_default_value( legendary_spell->effectN( 2 ).percent() + buff->value() );
```
vituscze: You definitely don't want to use `buff->value()` here, it returns 0 for buffs which are not up and does benefit tracking, none of which you want in sim setup.

HawkCorrigan: modify_default_value adds it to the base anyway so the value part is useless and incorrect anyway

wfarr (commented):
I wonder: does this implementation behave correctly with multi-actor sims, or does it give all players increased duration and value if there is a shaman with this legendary?

vituscze:
> 
> 
> I wonder: does this implementation behave correctly with multi-actor sims, or does it give all players increased duration and value if there is a shaman with this legendary?

Yes, this should work fine with multiactor sims.

PR #5322, opened 2020-09-26
[Warlock] Implement Decimating Bolt

javedbartlett:


engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -55,6 +55,73 @@ struct drain_life_t : public warlock_spell_t
...
+    double m = warlock_spell_t::composite_target_multiplier( target );
+
+    m *= 2.0 - target->health_percentage() * 0.01;
+
+    return m;
+  };
+
+  void impact( action_state_t* s ) override
+  {
+    warlock_spell_t::impact( s );
```
scamille: You can just remove the whole override section here if there is no custom code in it.

engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -55,6 +55,73 @@ struct drain_life_t : public warlock_spell_t
...
+    parse_options( options_str );
+    // can_havoc = true; NYI
+    travel_speed = p->find_spell( 327072 )->missile_speed();
+
+    add_child( decimating_bolt_dmg );
+  }
+
+  void impact( action_state_t* s ) override
+  {
+    if ( p()->talents.fire_and_brimstone->ok() )
```
scamille, line 118: Instead of having 2 lines triggering the buff I would only calculate the parameter you pass to the trigger function depending on whether you have FaB.
That way it is easier to understand that the buff always gets triggered, just with different values.

engine/class_modules/warlock/sc_warlock.hpp
```diff
@@ -508,6 +511,9 @@ struct warlock_t : public player_t
     gain_t* baleful_invocation;  // BFA - Azerite
 
     gain_t* memory_of_lucid_dreams;  // BFA - Essence
+
+    // SL
+    gain_t* decimating_bolt;
```
scamille: The gain here is not initialized and I can not see it used anywhere. Or is it supposed to go into the decimating_bolt_t::impact function?

javedbartlett: Forgot to remove it

javedbartlett: Forgot to remove it

engine/class_modules/warlock/sc_warlock.cpp
javedbartlett: Was going to use the impact override, but implemented the buff gain inside decimating_bolt_t instead. Forgot to remove it 😅

engine/class_modules/warlock/sc_warlock.hpp
javedbartlett: It just wasn't removed after testing, thanks

engine/class_modules/warlock/sc_warlock.cpp
javedbartlett: I do agree it's a bit verbose, but creating a trigger function doesn't seem necessary for this one off situation. I believe a middle ground would be to create shared variable that changes based off an if condition. This would shorten it down and still maintain the readability.

javedbartlett: > Instead of having 2 lines triggering the buff I would only calculate the parameter you pass to the trigger function depending on whether you have FaB.
> That way it is easier to understand that the buff always gets triggered, just with different values.

I do agree that it's a bit verbose, but creating a trigger function doesn't seem necessary for this one off situation. I believe a middle ground would be to create shared variable that changes based off an if condition. This would shorten it down and still maintain the readability.

scamille: Yeah that's what I meant, I just wrote it a bit confusingly.

scamille (approved):

Azevara:
Is ground_aoe_event_t usage going to cause decimating bolt to hit all targets in a multi-target sim?

javedbartlett:
> Is ground_aoe_event_t usage going to cause decimating bolt to hit all targets in a multi-target sim?

Just ran a test using `desired_targets=3` and decimating bolt only damaged 1 of them. I also ran a sim using Seed to make sure all 3 targets were stacked and it hit all 3 targets so it seems like Decimating bolt is working as intended! 🙂

nasirabtw (approved):
Will require editing for havoc but good enough for now to test ST stuff and aff/demo.

PR #5323, opened 2020-09-26
[Warlock] Implement Scouring Tithe

javedbartlett:


engine/class_modules/warlock/sc_warlock.hpp
```diff
@@ -508,6 +510,9 @@ struct warlock_t : public player_t
     gain_t* baleful_invocation;  // BFA - Azerite
 
     gain_t* memory_of_lucid_dreams;  // BFA - Essence
+
+    // SL
+    gain_t* scouring_tithe;
```
scamille, line 515: As far as I can tell, the gain variable is not initialized anywhere.

scamille (changes requested):

engine/class_modules/warlock/sc_warlock.hpp
javedbartlett: Nice catch, the init was deleted during a rebase!

scamille (approved):

PR #5324, opened 2020-09-27
[Warlock] Move havoc to main module for covenants

Azevara:


Azevara:
Resubmission of #5315 to deal with an incredibly silly set of git mistakes

PR #5325, opened 2020-09-27
dot_t rework

### LINKED ISSUE #5319: DoT Extension of Void Bolt and partial ticks
### Describe the bug
If Void Bolt extends a DoT that has only a partial tick remaining, the next tick of theDoT will be a partial tick, while it should be a full tick.

21.148 Dot vampiric_touch ticks (8 of 9). last_start=1.244, duration=21.000 time_to_tick=2.488
21.148 Player 'T25_Priest_Shadow' tick amount for Action vampiric_touch on Player 'Fluffy_Pillow': amount=823.7068797018447 initial_amount=411.85343985092237 base=0.0 bonus=0.0 s_mod=0.27 s_power=1176.0 a_mod=0.0 a_power=0.0 mult=1.297094481767833, tick_mult=1.0
21.148 New DoT Tick Event: Dot vampiric_touch tick 9-of9 time_to_tick=1.096
21.377 Player result for Action void_bolt_extension is hit.
21.377 Player extends duration of Dot vampiric_touch on Player 'Fluffy_Pillow' by 3.000 second(s).
21.377 Rescheduling event DoT End(10) from 22.244 to 25.244
**22.244 Dot vampiric_touch ticks (9 of 9). last_start=1.244, duration=24.000 time_to_tick=1.096
22.244 Player tick amount for Action vampiric_touch on Player 'Fluffy_Pillow': amount=362.8547990969541  <- this tick is a partial tick and should not happen at all**
24.732 Dot vampiric_touch ticks (10 of 11). last_start=1.244, duration=24.000 time_to_tick=2.488
24.732 Player tick amount for Action vampiric_touch on Player 'Fluffy_Pillow': amount=411.85343985092237 
25.244 Dot vampiric_touch ticks (11 of 11). last_start=1.244, duration=24.000 time_to_tick=0.512
25.244 Player 'T25_Priest_Shadow' tick amount for Action vampiric_touch on Player 'Fluffy_Pillow': amount=84.75440562848563

The extend_duration() function used to extend DoTs is incapable of rescheduling partial ticks, whereas it should to properly mirror ingame behaviour. ( it might extend past Void Bolt to a issue with the extend function)

### Expected behavior
Following log:https://www.warcraftlogs.com/reports/mG2H9CNyjdLMQV8t/#fight=2&source=17&type=damage-done&view=events

My VT should a total of 9 ticks, 8 full damage, the 9th being a partial (at 150 danage without extension)

At 22.559 my Voidbolt hit the target after the 8th VT tick, the next tick will happen at 24.667 seconds as a full tick.
At 26.334 seconds the last partial tick will happen with 264 damage (more then the value of 150 at no extension)

### To Reproduce
that's the apl i used:

actions.main+=/vampiric_touch,if=!dot.vampiric_touch.ticking
actions.main+=/void_eruption,if=dot.vampiric_touch.ticks_remain=3
actions.main+=/void_bolt,if=dot.vampiric_touch.ticks_remain=1,cooldown=500

with the following stats:
gear_intellect=726
gear_haste_rating=678
gear_mastery_rating=92
gear_versatility_rating=570

### Additional information
* SimulationCraft Version (shadowlands newest nightly build)

vituscze:
Simplified the internal DoT state by removing `last_start`, `prev_tick_time`, `extended_time`, `reduced_time`, `last_tick_factor`, `num_ticks`, `miss_time` and `time_to_tick`. New DoT state consists of just 6 members: `ticking`, `tick_event`, `end_event`, `tick_time`, `current_duration`, and `stack`. `extended_time` and `reduced_time` were replaced with a single member `extra_time`. `last_tick_factor`, `num_ticks`, and `time_to_tick` were replaced with derived member functions that calculate the previous values from the current DoT state.

The DoT now only schedules full ticks and partial ticks are handled by the end event. Tick events are scheduled even if the DoT would expire before the tick, which simplifies the internal state by a lot.

`extend_duration` and `reduce_duration` were simplified and merged into `adjust_duration` (the previous `reduce_duration` behavior is applied when called with negative timespans). Unused version of `copy` was removed (its behavior was identical to that of `copy( d->target, DOT_COPY_CLONE )`).

Opening this as a draft since it still requires more testing to make sure the behavior matches for some of the more complicated interactions. Some class code that depends on `dot_t` internals also needs to be updated.

### DIFF
--- a/engine/action/sc_dot.cpp
+++ b/engine/action/sc_dot.cpp
@@ -743,7 +743,7 @@ void dot_t::tick()
     }
   }
 
-  sim.print_debug( "{} ticks ({} of {}).  duration={} time_to_tick={}", *this, current_tick, num_ticks(),
+  sim.print_debug( "{} ticks ({} of {}). duration={} time_to_tick={}", *this, current_tick, num_ticks(),
                    current_duration, time_to_tick() );
 
   current_action->tick( this );
@@ -962,7 +962,6 @@ void dot_t::adjust_full_ticks( double coefficient )
     current_tick++;
     tick();
     new_tick_remains += new_tick_time;
-    rounded_full_ticks_left--;
   }
   event_t::cancel( end_event );
 
--- a/engine/action/sc_dot.cpp
+++ b/engine/action/sc_dot.cpp
@@ -18,7 +18,7 @@
 struct dot_t::dot_tick_event_t : public event_t
 {
 public:
-  dot_tick_event_t(dot_t* d, timespan_t time_to_tick);
+  dot_tick_event_t( dot_t* d, timespan_t tick_time );
 
 private:
   virtual void execute() override;
@@ -52,25 +52,19 @@ private:
 
 dot_t::dot_t( util::string_view n, player_t* t, player_t* s )
   : sim( *( t->sim ) ),
-    ticking( false ),
-    current_duration( timespan_t::zero() ),
-    last_start( timespan_t::min() ),
-    prev_tick_time( timespan_t::zero() ),
-    extended_time( timespan_t::zero() ),
-    reduced_time( timespan_t::zero() ),
-    stack( 0 ),
-    tick_event( nullptr ),
-    end_event( nullptr ),
-    last_tick_factor( -1.0 ),
+    ticking(),
+    current_duration(),
+    tick_time(),
+    extra_time(),
+    stack(),
+    tick_event(),
+    end_event(),
     target( t ),
     source( s ),
-    current_action( nullptr ),
-    state( nullptr ),
-    num_ticks( 0 ),
-    current_tick( 0 ),
-    max_stack( 0 ),
-    miss_time( timespan_t::min() ),
-    time_to_tick( timespan_t::zero() ),
+    current_action(),
+    state(),
+    current_tick(),
+    max_stack(),
     name_str( n )
 {
 }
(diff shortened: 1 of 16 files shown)


engine/action/sc_dot.cpp
```diff
@@ -743,7 +743,7 @@ void dot_t::tick()
     }
   }
 
-  sim.print_debug( "{} ticks ({} of {}).  duration={} time_to_tick={}", *this, current_tick, num_ticks(),
+  sim.print_debug( "{} ticks ({} of {}). duration={} time_to_tick={}", *this, current_tick, num_ticks(),
```
scamille, line 746: This should have a manual sim.debug check with the new function calls.

vituscze: There are quite a few old-style debug prints in the dot code. I'd rather do them all properly in a separate PR.

scamille: ok, makes sense not to mix in too many other changes .

nuoHep (approved):
I don't really know the code good enough to do an in-depth review, but fwiw it all looks much more comprehensible to me (like... I actually understand how dots work now :) in simc). The reduction of state and removal of mostly duplicate stuff is especially nice.

So lgtm.

vituscze:
Don't know why it requested a review from priests when I converted it from a draft.

navv1234:
Looked good to me also, and aligns much with what we have been discussing in the past about simplifying how the dot code works, and secondly, making it slightly less fragile in terms of the event stuff. Also added benefit is that it now aligns code-wise quite a bit better what what we have for ticking buffs.

scamille:
> 
> 
> Don't know why it requested a review from priests when I converted it from a draft.

I was trying out the https://github.com/simulationcraft/simc/blob/shadowlands/.github/CODEOWNERS file, which automatically adds reviewers when a PR touches certain files.
Definitely a neat thing for all the class-based PR we have from outside contributors right now, but the disadvantage is that it also triggers on commits like this which only touch a class module because of eg. a core API change.

PR #5326, opened 2020-09-27
[Death Knight] Implement remaining baseline and talent changes for Unholy

Khazakdk:
-Smart targeting when cleaving with Scourge Strike (Log snips demonstrating it works: https://pastebin.pl/view/raw/82e4d7df)
-Unholy Blight increasing minion damage
-Unholy Blight implementation reworked to be more like Remorseless Winter
-Unholy Pact (just renamed Helchains)
-Mastery now affects all pet damage instead of just pet shadow damage
-Soul Reaper rework


engine/class_modules/sc_death_knight.cpp
```diff
@@ -1113,6 +1115,9 @@ inline death_knight_td_t::death_knight_td_t( player_t* target, death_knight_t* p
 
   debuff.biting_cold       = make_buff( *this, "biting_cold", p -> find_spell( 337989 ) )
                            -> set_default_value( p -> conduits.biting_cold.percent() );
+
+  debuff.unholy_blight     = make_buff( *this, "unholy_blight", p -> find_spell( 115994 ) )
+                           -> set_default_value( p -> find_spell( 115994 ) -> effectN( 2 ).percent() );
```
Melekus: You can use ``set_default_value_from_effect`` to avoid repeating ``p -> find_spell( x )``
In this case, this would be ``set_default_value_from_effect( 2 )``
It's only a recent addition which is why it's not everywhere in the module yet

Melekus: You can also fetch the spell from ``p -> talent.unholy_blight -> effectN( 1 ).trigger()`` rather than finding the spell by ID.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -3957,14 +3961,8 @@ struct helchains_buff_t : public buff_t
...
 
-  // Helchains ticks twice on buff application and buff expiration, hence the following overrides
-  // https://github.com/SimCMinMax/WoW-BugTracker/issues/390
-  void execute( int stacks, double value, timespan_t duration ) override
-  {
-    buff_t::execute( stacks, value, duration );
-    damage -> execute();
-  }
-
+  // Unholy pact ticks twice on buff expiration, hence the following override
```
Melekus, line 3981: From my limited testing, it seems that unlike Helchains, Unholy Pact seems to tick two additional times at the end of the buff rather than one additional at the start and at the end.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -6139,6 +6137,24 @@ struct scourge_strike_base_t : public death_knight_melee_attack_t
   int n_targets() const override
   { return p() -> in_death_and_decay() ? as<int>( p() -> spec.scourge_strike -> effectN( 4 ).base_value() ) : 0; }
 
+  std::vector<player_t*>& target_list() const override // smart targeting to targets with wounds when cleaving SS
```
Melekus, line 6158: Does it always prioritize higher stack counts or just targets with wounds over targets that don't have wounds?

engine/class_modules/sc_death_knight.cpp
```diff
@@ -6139,6 +6137,24 @@ struct scourge_strike_base_t : public death_knight_melee_attack_t
...
+  {
+    std::vector<player_t*>& current_targets = death_knight_melee_attack_t::target_list();
+    if (current_targets.size() < 2 || !p() -> in_death_and_decay()) {
+      return current_targets;
+    }
+    
+    // first target, the action target, needs to be left in place
+    std::sort( current_targets.begin() + 1, current_targets.end(),
+      [this]( player_t* a, player_t* b) {
+        int a_stacks = p()->get_target_data( a )->debuff.festering_wound->check();
```
Melekus: You can use the simpler death_knight_action_t::td( target ). It's a shortcut for p() -> get_target_data( t ).
``int a_stacks = td( a ) -> ...``

engine/class_modules/sc_death_knight.cpp
```diff
@@ -6295,6 +6311,11 @@ struct unholy_blight_dot_t : public death_knight_spell_t
     tick_may_crit = background = true;
     may_miss = may_crit = hasted_ticks = false;
   }
+
+  void impact(action_state_t* state) override
```
Melekus, line 6348: Unholy Blight should probably be reworked into a buff that applies the stacking UB dot (tied to the debuff) and the VP dot every tick, similarly to Remorseless Winter.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -3939,14 +3943,14 @@ struct helchains_damage_t : public death_knight_spell_t
   }
 };
 
-struct helchains_buff_t : public buff_t
+struct unholy_pact_buff_t : public buff_t
```
Melekus, line 3961: This seems to be missing the strength buff of Unholy Pact, using spell ID = 319255 (or ``talent.unholy_pact``(319230) ``-> effectN( 1 ).trigger() ``(319232) ``-> effectN( 1 ).trigger()`` (319233) ``-> effectN( 4 ).trigger()`` (319255)

engine/class_modules/sc_death_knight.cpp
```diff
@@ -9263,6 +9284,8 @@ double death_knight_t::composite_player_pet_damage_multiplier( const action_stat
     m *= 1.0 + cache.mastery_value();
   }
 
+  m *= 1.0 + get_target_data( state->target )->debuff.unholy_blight->stack_value();
```
Melekus: This will increase damage based on the state's target's (aka the spell's target's) stacks. 

You should move this to death_knight_pet_t::composite_player_target_multiplier().
Unless this also affect non DK pet damage in which case I'm not sure how to handle this.

engine/class_modules/sc_death_knight.cpp
Khazakdk: I've been told it just picks wounds over no wounds but had issues getting the sort lambda to work when I tried to model that behavior.  Tried setting two bools based on if the targets had wounds and would return the first bool if it was true otherwise return the second.  This threw exceptions in the middle of an iteration, saying the lambda wasn't returning a bool.  Couldn't figure out why, I stepped through it working just fine several times but later in the sim it would fail.  After a couple of hours of fiddling with it I stripped it down to just return the result of comparison between the two wound counts.  Though even that first solution was imperfect because it favored targets that came earlier in the target list.  The most technically correct solution would be to make a vector of n_targets minus the first with and then use rng to pull 4 random ones with wounds out, and then assemble a new target list of the action target, the 4 randomly chosen targets, then the rest.  Hmm, might have just rubber ducked my way to an answer.

engine/class_modules/sc_death_knight.cpp
Khazakdk: Why is the spell target not what we want?

Melekus: If you cast a single aoe ability, there will be a single state object that contains the target of the spell cast.

Melekus: You want to make sure that the modifier uses the right target's data, which is why you should move this to a target_multiplier, that will affect the damage done to a specific target. (composite_player_pet_damage_multiplier affects damage based on the source).

engine/class_modules/sc_death_knight.cpp
Melekus: What happens if you just replace the a_stacks and b_stacks value by 0 if the debuff isn't up, and 1 if the buff is up?
Something like 
```   
int a_stacks = td( a )->debuff.festering_wound->up() ? 1 : 0;
int b_stacks = td( b )->debuff.festering_wound->up() ? 1 : 0;
return a_stacks > b_stacks; 
```
If your current implementation works I don't think there should be an issue with simplifying stack count to just 0s and 1s.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -7057,10 +7070,10 @@ void death_knight_t::trigger_soul_reaper_death( player_t* target )
 
   if ( td -> dot.soul_reaper -> is_ticking() )
   {
-    sim -> print_log( "Target {} died while affected by Soul Reaper, player {} gains Soul Reaper buff.",
+    sim -> print_log( "Target {} died while affected by Soul Reaper, player {} gains Runic Corruption buff.",
                       target -> name(), name() );
 
-    buffs.soul_reaper -> trigger();
+    buffs.runic_corruption -> trigger();
```
Melekus: The dk module uses death_knight_t::trigger_runic_corruption( double rpcost, double override_chance, proc_t* proc ) as a helper function to trigger the runic corruption, this is not only useful for RP spenders, but also to track the different sources of Runic Corruption (similarly to trigger_killing_machine() or trigger_festering_wound()). 

You'll have to create and init a new proc for runic corruption generated by Soul Reaper and send it to trigger_runic_corruption(), the function will handle the rest.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -6235,22 +6235,35 @@ struct scourge_strike_t : public scourge_strike_base_t
...
   }
 
-  void execute() override
+  void last_tick( dot_t* dot ) override
   {
-    death_knight_spell_t::execute();
-
-    p() -> replenish_rune( as<unsigned int>( p() -> talent.soul_reaper -> effectN( 2 ).base_value() ), p() -> gains.soul_reaper );
+    if (dot -> target -> health_percentage() < data().effectN( 3 ).base_value())
+    {
```
Melekus, line 6282: You should probably set the execute damage's target before executing it

engine/class_modules/sc_death_knight.cpp
```diff
@@ -6235,22 +6235,35 @@ struct scourge_strike_t : public scourge_strike_base_t
 
 // Soul Reaper ==============================================================
 
+struct soul_reaper_execute_t : public death_knight_spell_t
+{
+  soul_reaper_execute_t( death_knight_t* p, const std::string& options_str ) :
+    death_knight_spell_t( "soul_reaper_execute", p, p -> spell.soul_reaper_execute )
+  {
+    parse_options( options_str );
```
Melekus: Parsing options is only necessary for APL actions, you don't need it here (and you should set background to true)

engine/class_modules/sc_death_knight.cpp
Khazakdk: Ah that's a good idea.  That's a more straightforward way of what I tried to do on the first attempt.  It will still favor targets that appear first in the list but will be easier to understand and maintain than trying to randomly pick up to 4 extra targets with wounds and I'm not sure that the outcomes would be different if we did it that way.

Khazakdk: That did the job nicely https://pastebin.pl/view/raw/751b175b

engine/class_modules/sc_death_knight.cpp
```diff
@@ -745,7 +746,7 @@ struct death_knight_t : public player_t {
     const spell_data_t* mark_of_blood;
 
     const spell_data_t* voracious;
-    const spell_data_t* death_pact; // NYI
+    const spell_data_t* death_pact; // The ingame implementation is a mess with multiple unholy pact buffs on the player and multiple unholy pact damage spells. The simc version is simplified
```
Melekus: This comment in on the wrong "pact" talent spell

engine/class_modules/sc_death_knight.cpp
```diff
@@ -3957,16 +3978,11 @@ struct helchains_buff_t : public buff_t
...
-  // https://github.com/SimCMinMax/WoW-BugTracker/issues/390
-  void execute( int stacks, double value, timespan_t duration ) override
-  {
-    buff_t::execute( stacks, value, duration );
-    damage -> execute();
-  }
-
+  // Unholy pact ticks twice on buff expiration, hence the following override
+  // https://github.com/SimCMinMax/WoW-BugTracker/issues/675
   void expire_override( int expiration_stacks, timespan_t remaining_duration ) override
```
Melekus, line 3983: The last two hits aren't affected by the buff, but buff_t::expire_override is called during buff_t::expire() while the buff is still technically active. You can circumvent that by calling action_t::schedule_execute() instead of action_t::execute(), this will create damage events that will fire after a 0s delay, but after the current operations are done.

PR #5330, opened 2020-09-28
Update Windwalker profiles

scamille:


scamille:
@Hinalover there are 2 buffs 'weapons_of_order' which causes merge problems. One on the monk, the other in target_data.

~~Can you please rename one of them, maybe with a '_debuff' postfix?~~

~~(Feel free to fix it in the shadowlands branch, I can then rebase this one and check again)~~

Edit: Going to try fix it here, you can always rename it to something better afterswards.

scamille:
Ok so the crash is DungeonSlice only, and seems to be in crackling_tiger_lightning_t, a pet dot action with tick_action set. It complains about dot -> state -> target being asleep.

@navv1234 Since pet retargeting is your speciality, do you have a idea why this might occur here? The code at https://github.com/simulationcraft/simc/blob/adf04662b8865dc55d9d4b4eceef3d93d6bee0eb/engine/class_modules/sc_monk.cpp#L2285 looks innocent.

Hinalover:
fine by me. I can rename the actual debuff name later.

scamille:
@vituscze Do you have any idea why this is happening?

Might be related to dynamic_tick_action=true, or the pet in general. I just thought I'd ask since you have been working on dots recently as well, even if this might be more of a targeting issue.

vituscze:
I'll take a look but it's almost certainly not an issue in the dot code. I don't think the monk module does anything weird with dot internals either.

vituscze:
Yeah, this is definitely a targetting issue and it won't happen during first iteration, so no point trying to debug this with log=1 or debug=1.

Crackling Tiger Lightning action executes on a target that is sleeping (and the assert happens when that action's dot tries to tick for the first time). Looking at the target, it's one of the add waves but the action is executed before any of those add waves even happen, so it looks like the action keeps the target from the previous iteration.

vituscze:
Here's a debug output if you want to check.


[debug.5201515270900680107.log](https://github.com/simulationcraft/simc/files/5301800/debug.5201515270900680107.log)

scamille:
The assert it is running into is
```simc/engine/action/sc_action.cpp:1716: virtual void action_t::tick(dot_t*): Assertion `!d->target->is_sleeping()' failed.```

Should this even be an assert, or should it just return on sleeping targets?

scamille:
Fixed by 23cbd23fa984039fd22a880f3c879bde6bb97877 and 103bae30a4ceea8bfd3a3f9943910f5e6e941e3f
Thanks @navv1234

PR #5333, opened 2020-09-28
Try out bats test run

scamille:


nuoHep:
Nice, thanks for looking into this. Sadly don't have much time atm for this myself.

One thing I wanted to do to speed it up is to split up tests with a matrix.
`classes` / `enemies` is obvious. You can also go further and split it up by class / fight style.

scamille:
Well my current attempt is to just restore our old Jenkins setup, see how far that gets us.

I also thought about using github workflow stategy matrix, but that would definitely explode in number of combinations if we include all talents.

I don't really like this bats very much as well, but it seems to do the job.

nuoHep:
> I also thought about using github workflow stategy matrix, but that would definitely explode in number of combinations if we include all talents.

Oh, I guess I wasn't clear enough. I don't really propose doing talent comps or the like with a matrix. What I meant is to split it up at the macro level. So rn in this pr do matrix like
```yaml
strategy:
      matrix:
        test: [ enemies, classes ]

[...]

      - name: Run
        [ ... ]
        run: |
          cd simc-test
          bats --tap ${{ matrix.test }}.bats
```
Then you could also split up `classes.bats` into per class files. So the matrix would look smth like
```yaml
  test: [ druid, death_knight, demon_hunter, hunter, mage, monk, paladin, priest, rogue, shaman, warlock, warrior, enemies ]
```

---
We could also then either drop `enemies.bats` and do the tests as part of `ubuntu-clang-10-run` which already does patchwerk & dungeon_slice. Or fully move to bats for the tests. Dunno which is better tbh.

---

Also, I'm pretty sure our bfa travis setup used the tests from this repo (https://github.com/simulationcraft/simc/tree/shadowlands/tests) instead of the separate `simc-tests` one. And it's actually better if the tests are part of the repo.

scamille:
One thing here to note is that no matter how much we optimize the combination space, testing talents for all classes will take a lot of time, compared to the current quite fast workflow.
(Compared to the past, we are still quite good. Old macOS travis build took 30min or something like that)
Edit: The talent run here could be multiplied by more fight styles, and shadowlands could warrant covenant/conduit tests.

scamille:
Options to combat the CI runtime:
- reduce numbers of combinations we test
- Only run extended tests periodically, eg. nightly
- Optimize test runtime with other means? I am already only using 2 iterations, so as long as we use debug build there is probably not that much we can do.
- Set up a self-hosted runner (maybe @seriallos could help us out here?) with much more compute power to run the tests faster. (Which could be used to run even more tests like covenant stuff :D)

scamille:
It might really be worth it to add classes to the matrix, and create separate bats tests for talents which take the class as an argument.
That would then be more extendable to also adds tests for covenant combinations, etc.

scamille:
Ok so this should be ready for merging, runtime is down to 10m again after splitting it up into smaller jobs.

@nuoHep I also discussed the possibility of self-hosted runners with seriallos. The question is if there is even any advantage for us, compared to just ensuring no single job in our CI build takes too much time, so that we can take care of the 60 parallel workers available from GitHub. I previously thought i would be advantage to have some beefy machines available, when the solution was to just better split up the work.

seriallos:
I'm definitely up for helping if the need arises. My guess is that if you can get the splitting to work, the 60 workers from GitHub might be sufficient. If it ends up not being enough, the self-hosted route would almost certainly allow scaling beyond what Github's workers can provide.

nuoHep (commented):
Right, looks much better (and faster) with the per-class split.

One observation is that we use bats now as an essentially glorified bash loop over talent options now. I'll try to find some time later this week to look if we can completely drop it and simply run a custom python script for tests.

.github/workflows/main.yml
```diff
@@ -155,6 +155,49 @@ jobs:
           UBSAN_OPTIONS: print_stacktrace=1
         run: ${{ runner.workspace }}/b/ninja/simc output=/dev/null html=/dev/null json2=/dev/null json3=/dev/null ${{ matrix.simc_flags }}
 
+  ubuntu-clang-10-simc-tests:
+    name: simc-tests-${{ matrix.test }}-${{ matrix.tier }}-${{ matrix.fight_style }}-${{ matrix.class }}
```
nuoHep: Nit: I'd reshuffle it a bit to `simc-tests-${{ matrix.test }}-${{ matrix.tier }}-${{ matrix.class }}-${{ matrix.fight_style }}` (move class before fight style) and maybe even drop `simc-`.

Not sure if it's my setup with dark theme styler or a general problem with github actions layout but it currently looks like this to me:
image
which makes it really hard to find a specific per class run.

tests/talent_options
```diff
@@ -6,112 +6,112 @@ __max_tiers = 7
...
-        (1, -1),        # 30: All
-                        # 45: None
-        (3, -1),        # 60: All
-                        # 75: None
-        (5, -1),        # 90: All
-        (6, -1),        # 100: All
+        (1, -1),        # 25: All
+        (2, -1),        # 30: All
+        (3, -1),        # 35: All
+        (4, -1),        # 40: All
```
nuoHep: I'm not sure about the other classes, but for hunters all of the utility talents are not implemented. So I don't see much point in adding them to the talents test "matrix".

I'm pretty sure this holds true for all class utility rows tbh.

scamille:
Hehe yeah, had the same thought. My fear was that getting talent combinations back would be too complicated, and my scripting skills are not that good.

My plan is to add some more tests to the matrix later on, like covenant combinations. So it would be great if that is still somewhat simple to pull off.

tests/talent_options
scamille: Yes, the choice to activate this again was deliberate because talent tiers changed and should be reduced on a per class-basis again. Or are the utilty rows nowadays always in the same rows and can be disabled for all classes?
To be honest, I was just a bit lazy here to double-check everything.

.github/workflows/main.yml
scamille: Makes sense. I dropped the simc-tests- prefix as well, anything starting with the matrix.test name should be indication enough of what the job is doing.

scamille:
Ok question now is: Should I just merge this when everything is fixed, and you base your changes on the merged version, or should I push this to a simulationcraft/simc branch and create a new PR for it?

tests/talent_options
nuoHep: They didn't change from bfa to shadowlands at least (and from legion to bfa afair). And I'm pretty sure what we had here was up to date for all classes.

nuoHep:
> Ok question now is: Should I just merge this when everything is fixed, and you base your changes on the merged version, or should I push this to a simulationcraft/simc branch and create a new PR for it?

Just merge this. It's already miles better than what we have/had and I can easily do stuff on top when I have time.

tests/talent_options
scamille: Ok, I just removed my changes. The comments do not match, but that does not hurt since we still got the same number of talent rows.

PR #5337, opened 2020-09-29
[Priest] Fix CWC for Mindblast and Searing Nightmare

Saeldur:
Change CWC actions to use ready instead of usable while casting.
Move CWC to its own action group to stop mind sear and flay being
cancelled for instant spells.
Change mindblast to use only_cwc option (Currently) to make it use cwc.
This forces mindblast to only be usable while channeling. A separate
option is being used over use_while_channeling to differentiate the
slightly different functionality involved.

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -27,6 +27,7 @@ private:
   double whispers_bonus_insanity;
   const spell_data_t* mind_flay_spell;
   const spell_data_t* mind_sear_spell;
+  bool only_cwc;
 
 public:
   mind_blast_t( priest_t& player, util::string_view options_str )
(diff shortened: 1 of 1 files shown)


seanpeters86 (approved):

lgkern (approved):

scamille (approved):
Ok so only_cwc=false now also means never cast it during cast, right? Which means I can no longer have a single mind blast line which can be both cast normally and while casting. Which is ok I guess if you want those to separate anyway.

Please document this new option in the wiki Priest page later after the merge.

Saeldur:
> Ok so only_cwc=false now also means never cast it during cast, right? Which means I can no longer have a single mind blast line which can be both cast normally and while casting. Which is ok I guess if you want those to separate anyway.
> 
> Please document this new option in the wiki Priest page later after the merge.


Correct, only_cwc=false means it will never cast during a cast and only_cwc=true means it will only cast during a cast.
I was going to originally add support to keep use_while_casting as an additional option but it would then need to guard against being cast during other spells and I couldn't think of a clean way at the time of writing the original PR to go about it.
The reason for adding only_cwc (could be renamed if you deem it fit) was so that the cwc action is differentiated in APL so that it never tries to cast the wrong one unintentionally. I also chose not to overload use_while_casting so that the usage of "use_while_casting" it is the same no matter which class you're working with to avoid any confusion in the future.

scamille (approved):

seanpeters86:
added a note here https://github.com/simulationcraft/simc/wiki/Priests

PR #5339, opened 2020-09-30
[Warlock] Add Sacrolash legendary

Azevara:


Azevara:
There's some use of warlock_spell_t that should be affliction_spell_t, will correct tomorrow before this should be merged

PR #5344, opened 2020-10-01
[Warlock] Implementation Prolonged Decimation

khoepli:
Since I have not used C++ in a while and never worked on this project before, I took a recently approved pull request as example and hope my intuition did not fail me.
The variables for the soul conduit itself were already there, so all I did was apply them in the few spots where the buffs from 'Decimating Bolt' were comsumed and added a tracker similar to the one from 'Corrupting Leer'.

Please feel free to reach out to me on Discord (Failspawn#2108) if anything has really missed the mark!


Azevara:
As far as I can tell, this code looks fine and functional but the trigger implementation is a bit unorthodox. Was this copied/modeled on another similar behavior somewhere in the codebase?

Also, I assume these were tested by running sample profiles and they behaved as expected?

engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -761,7 +761,7 @@ void warlock_t::init_procs()
   procs.demonic_calling = get_proc( "demonic_calling" );
   procs.soul_conduit    = get_proc( "soul_conduit" );
   procs.corrupting_leer = get_proc( "corrupting_leer" );
-
+  procs.prolonged_decimation = get_proc( "prolonged_decimation" );
```
seanpeters86, line 764: this is a small thing, but clang-format here would want to line all these `procs` lines up by the `=`

khoepli: Thanks for pointing it out, I thought it would apply automatically after setting CLANG up. Now just used it for the sections with my changes manually.

scamille: I am not sure if the warlock module uses clang formatting. I think the priest module is one of the few which nowadays tries to use it somewhat consistently and if possible all the time. Which I think is great since I love reading formatted code but hate writing it :-)

But yeah for warlock just formatting the lines you worked on is probably the best, and it might need discussion with @Azevara if you guys want to do more formatting of existing code. It should in any case be in a separate commit and PR so as not to hide functional code changes.

engine/class_modules/warlock/sc_warlock_affliction.cpp
```diff
@@ -658,9 +658,6 @@ struct drain_soul_t : public affliction_spell_t
   void execute() override
   {
     dot_t* dot = get_dot( target );
-    if ( dot->is_ticking() && !p()->prolonged_decimation_trigger() )
-      p()->buffs.decimating_bolt->decrement();
```
vituscze, line 661: Do not remove this decrement here, otherwise you can chain cast Drain Soul without using any Decimating Bolt stacks.

engine/class_modules/warlock/sc_warlock_affliction.cpp
```diff
@@ -691,9 +688,10 @@ struct drain_soul_t : public affliction_spell_t
   {
     affliction_spell_t::last_tick( d );
     if ( !p()->prolonged_decimation_trigger() )
-      p()->buffs.decimating_bolt->decrement();
+      p()->buffs.decimating_bolt
+          ->decrement();  // Not sure if this should be a 'reset' instead since Decimating Bolt now no longer is
```
vituscze: `decrement` will call `expire` if there's only one stack, so you can leave it here.

khoepli: The issue isn't the decrement call so much as it is how the current implementation of Decimating Bolt always adds `p()->buffs.decimating_bolt->trigger( 3, value );` three stacks of the buff.
This does make sense since Drain Soul is the only time there is an exception to that rule and by implementing some kind of buff clearing we could handle it here instead of where the buffs are applied.

engine/class_modules/warlock/sc_warlock_affliction.cpp
khoepli: Thanks for the important input!
I just tested again on the SL Beta and the check ingame happens whenever the channel is aborted (regardless if finished until last tick or not). So is there maybe a better place to add this code to begin with?

engine/class_modules/warlock/sc_warlock_affliction.cpp
vituscze: If you trigger 3 stacks on a buff with 1 max stack, you'll only have 1 stack after, not 3. So this should be fine.

engine/class_modules/warlock/sc_warlock_affliction.cpp
vituscze: last_tick will always be called when you stop channeling, whether that's early interrupt or not. The issue is more with recasting Drain Soul while already channeling Drain Soul. The Drain Soul effect is technically never interrupted (so last_tick won't be called) but in game, it should consume a stack of the buff.

vituscze: Or, in other words, you need both the last_tick code and this code

engine/class_modules/warlock/sc_warlock_affliction.cpp
khoepli: It doesn't have a max stack of one if you are not talented into Drain Soul as affliction though. 
I couldn't even find the information on the spell itself, but there was this blue post that properly explained the new expected behaviour:
https://shadowlands.wowhead.com/news=318134/upcoming-changes-for-warlocks-on-the-shadowlands-beta-soul-rot-decimating-bolt

vituscze: Is that an issue? We're talking about code that will only be executed if you do have Drain Soul. Maybe I just misunderstood your point.

khoepli: Fair point, maybe my thought on that is overreaching the scope of this pullrequest.
I thought of it as a time saver due to making further changes to the buff application for the Drain Soul-special case unnecessary while achieving the same result.

engine/class_modules/warlock/sc_warlock_affliction.cpp
khoepli: Okay, I think I understand that, but wouldn't that leave an unresolved issue because the buff (which would then be consumed/decremented on the first tick) contains the needed info for the spell-multiplier?

vituscze: That's why the code is checking whether the "dot" (the channel) is ticking. If you're starting a fresh channel, it won't be ticking and therefore the buff won't be consumed.

engine/class_modules/warlock/sc_warlock.cpp
Azevara: Clang formatting is inconsistent in the module and any attempt to apply it right now has a high chance of creating a messy commit, so just something to be aware of. We'll need to revisit that issue once all the Shadowlands code is in.

engine/class_modules/warlock/sc_warlock_affliction.cpp
Azevara: Unless I'm misreading the PR code, this still needs to be added back in, correct?

engine/class_modules/warlock/sc_warlock_affliction.cpp
```diff
@@ -136,7 +136,10 @@ struct shadow_bolt_t : public affliction_spell_t
     if ( time_to_execute == 0_ms )
       p()->buffs.nightfall->decrement();
 
-    p()->buffs.decimating_bolt->decrement();
+    if ( rng().roll( p()->conduit.prolonged_decimation.percent() ) )
```
Azevara: Testing on Beta, I just noticed that the buff is currently consumed on impact, not execute. This is able to cause a buggy behavior, but we probably need to model this for now, so all these code sections in execute() regarding Decimating Bolt buff need to be moved unfortunately

engine/class_modules/warlock/sc_warlock_demonology.cpp
```diff
@@ -302,7 +302,10 @@ struct demonbolt_t : public demonology_spell_t
 
     p()->buffs.balespiders_burning_core->expire();
 
-    p()->buffs.decimating_bolt->decrement();
+    if ( rng().roll( p()->conduit.prolonged_decimation.percent() ) )
```
Azevara: Similar to above, this needs moving to impact()

engine/class_modules/warlock/sc_warlock_affliction.cpp
Azevara: Update to this: For some reason Incinerate is on cast finish, so it's just this one and Demonbolt that need updating.

engine/class_modules/warlock/sc_warlock_affliction.cpp
khoepli: Yes, it has been readded now.

engine/class_modules/warlock/sc_warlock_affliction.cpp
khoepli: Thanks for the input, I updated it accordingly now!

engine/class_modules/warlock/sc_warlock_demonology.cpp
khoepli: Done!

Azevara:
Going to close this PR since Blizzard decided to revise this conduit into something new: https://shadowlands.wowhead.com/spell=340268/fatal-decimation

PR #5345, opened 2020-10-01
[Priest] remove unused insanity drain code

seanpeters86:
first part of https://github.com/WarcraftPriests/sl-shadow-priest/issues/84

With the official removal of Legacy of the Void as a talent we no longer need this.


scamille:
I guess the question is whether we really want to remove it at this point or not?

And if you really want to remove it, why not just take the whole insanity_state with it, and handle everything through the normal player resource system?

seanpeters86:
Well if we ever want it back it would be easy to just refer back to this commit was the idea to add it back. Otherwise thats a lot of code being left in that we just don't use/don't maintain.

> why not just take the whole insanity_state with it, and handle everything through the normal player resource system?

This would be fine with me, just not sure how to do that piece, i.e. what can i remove safely/how do i do that.

engine/class_modules/priest/sc_priest.hpp
```diff
@@ -606,57 +600,21 @@ struct priest_t final : public player_t
...
-    void reset();
-
-    // Compute insanity drain per second with current state of the actor
-    double insanity_drain_per_second() const;
-
     // Gain some insanity
     void gain( double value, gain_t* gain_obj, action_t* source_action = nullptr );
 
     // Lose some insanity
     void lose( double value, gain_t* gain_obj, action_t* source_action = nullptr );
```
scamille: This seems to have no definition anyway, so the declaration should be removed.
(Has nothing to do with your changes)

scamille:
Well if you look at https://github.com/simulationcraft/simc/blob/6f1a2e939bb733562571577fb77180a6083a7c0f/engine/class_modules/priest/sc_priest_shadow.cpp#L1925 You can see that all that is left is just a wrapper to player.resource_gain call.

engine/class_modules/priest/sc_priest.hpp
seanpeters86: okay cool, didnt think so but figured id leave it till i got confirmation

scamille (approved):

PR #5346, opened 2020-10-01
[Warlock] Update Decimating bolt for 9.0.2.36086 and add more Malefic Rapture spells

javedbartlett:


scamille:
Are those value really nowhere to be found in spell data? Otherwise you will have to adjust on every data change, which we exactly try to avoid by using spell data values everywhere.

Of course if Blizzard hides stuff then there is not much you can do about it.

javedbartlett:
> Are those value really nowhere to be found in spell data? Otherwise you will have to adjust on every data change, which we exactly try to avoid by using spell data values everywhere.
> 
> Of course if Blizzard hides stuff then there is not much you can do about it.

Sadly no. I wasn't able to find them anywhere, @Azevara was also unable to find it in the spell data. Always fun when Blizz hides stuff 🙂

Azevara:
Unless there's some esoteric scaling table magic I'm not aware of, Blizzard is not reporting both ends of the range in the spell data. Fire and Brimstone 60% penalty is also not in spell data afaik.

PR #5347, opened 2020-10-01
(Priest) Fix Mindblast Dark Thoughts Bug

### LINKED ISSUE #35: Remove implemenation of Dark Thoughts Bug once fixed in-game
In the future the `ready()` method in `mind_blast_t` will need to be `overriden` to allow the casting of mindblast while you do not have a full charge if you have a buff. Currently it's being bypassed by the fact that if you have the buff you always have a blast charge since getting a proc currently resets the charge.

Saeldur:
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/35

This does two things.
A> Change Dark Thoughts to no longer reset Mind blast, as per the in game change.
B> Fix the ready condition on mind blast to support casting the spell if the dark thoughts buff is up as the current implementation sidesteps the Mind Blast cooldown during a dark thoughts proc. This was previously bypassed by Mind Blast always resetting cooldown from a dark thoughts proc.
The ready override is based off of the implementation in action_t::ready, if any changes are made to action_t::ready or if anything is added to priest_spell_t::ready that are important they will need to be manually added in the future.

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -72,60 +72,45 @@ public:
     apply_affecting_aura( player.find_rank_spell( "Mind Blast", "Rank 2", PRIEST_SHADOW ) );
   }
 
-  bool _ready()
-  {
-    // Check conditions that do NOT pertain to the target before cycle_targets
-    if ( cooldown->is_ready() == false && !priest().buffs.dark_thoughts->check() )
-      return false;
-
-    if ( internal_cooldown->down() && !priest().buffs.dark_thoughts->check() )
-      return false;
-
-    if ( player->is_moving() && !usable_moving() )
-      return false;
-
-    auto resource = current_resource();
-    if ( resource != RESOURCE_NONE && !player->resource_available( resource, cost() ) )
-    {
-      if ( starved_proc )
-        starved_proc->occur();
-      return false;
-    }
-
-    if ( usable_while_casting )
-    {
-      if ( execute_time() > timespan_t::zero() )
-      {
-        return false;
-      }
-
-      // Don't allow cast-while-casting spells that trigger the GCD to be ready if the GCD is still
-      // ongoing (during the cast)
-      if ( ( player->executing || player->channeling ) && gcd() > timespan_t::zero() &&
-           player->gcd_ready > sim->current_time() )
-      {
-        return false;
-      }
-    }
-
-    return true;
-  }
   bool ready() override
   {
     if ( only_cwc )
     {
       if ( !priest().buffs.dark_thoughts->check() )
         return false;
+
       if ( player->channeling == nullptr )
         return false;
-      if ( player->channeling->data().id() == mind_flay_spell->id() ||
-           player->channeling->data().id() == mind_sear_spell->id() )
-        return _ready();
-      return false;
+
+      if ( player->channeling->data().id() != mind_flay_spell->id() &&
+           player->channeling->data().id() != mind_sear_spell->id() )
+        return false;
+
+      if ( player->is_moving() && !usable_moving() )
+        return false;
+
+      auto resource = current_resource();
+      if ( resource != RESOURCE_NONE && !player->resource_available( resource, cost() ) )
+      {
+        if ( starved_proc )
+          starved_proc->occur();
+        return false;
+      }
+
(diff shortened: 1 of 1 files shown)


engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -74,21 +74,45 @@ struct mind_blast_t final : public priest_spell_t
 
   bool ready() override
   {
-    if ( only_cwc )
+    // Check conditions that do NOT pertain to the target before cycle_targets
+    if ( cooldown->is_ready() == false && !priest().buffs.dark_thoughts->check() )
+      return false;
+
+    if ( internal_cooldown->down() && !priest().buffs.dark_thoughts->check() )
+      return false;
```
seanpeters86, line 82: why do we need to check `cooldown` and `internal_cooldown`?

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -74,21 +74,45 @@ struct mind_blast_t final : public priest_spell_t
...
+      // Don't allow cast-while-casting spells that trigger the GCD to be ready if the GCD is still
+      // ongoing (during the cast)
+      if ( ( player->executing || player->channeling ) && gcd() > timespan_t::zero() &&
+           player->gcd_ready > sim->current_time() )
+      {
+        return false;
+      }
     }
+
+    return true;
```
seanpeters86, line 123: what is the benefit of not just using `return priest_spell_t::ready();` here? Dunno what @scamille thinks but I feel like we should have that

seanpeters86: Yeah we need ready here for this: https://github.com/simulationcraft/simc/blob/shadowlands/engine/class_modules/priest/sc_priest.hpp#L1383-L1386

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1836,15 +1861,15 @@ struct dark_thoughts_t final : public priest_buff_t<buff_t>
...
+  /*bool trigger( int stacks, double value, double chance, timespan_t duration ) override
   {
     bool r = base_t::trigger( stacks, value, chance, duration );
 
     // Currently dark thoughts also resets your mind blast.
     priest().cooldowns.mind_blast->reset( true, 1 );
 
     return r;
-  }
+  }*/
```
seanpeters86: we can just remove this instead of commenting it out, the commit history will hold this if we need it back

seanpeters86 (changes requested):

engine/class_modules/priest/sc_priest_shadow.cpp
Saeldur: return priest_spell_t::ready() will fail if the cooldown of mindblast is up but you have a dark thoughts proc. I'll manually add https://github.com/simulationcraft/simc/blob/shadowlands/engine/class_modules/priest/sc_priest.hpp#L1383-L1386 so it's compliant.
Both cooldown and internal_cooldown need to be checked because they are part of the base action_t::ready function. Internal cooldown is probably irrelevant though. If they don't get checked then the spell can just be cast since I never call priest_spell_t:ready()

engine/class_modules/priest/sc_priest_shadow.cpp
Saeldur: Very true, going to remove it.

engine/class_modules/priest/sc_priest_shadow.cpp
seanpeters86: hm okay, i see why you did this now, i wish there was a better way :/ @lgkern @scamille any ideas?

scamille:
Ok so I tried an alternative solution in #5354 after @seanpeters86 explained to me what this should do. Hope I got the essence of it right, haven't really tested it at all.

I understand why you had to completely replicate the action_t ready() implementation to get the cooldown checking line disabled during dark thoughts buff. There currently is no easy way to activate/deactive the cooldown of an action once it is started.

My somewhat hacky solution is to change the cooldown pointer of the action to a dummy cooldown before calling the actions execute() and ready() methods. This should hopefully ignore the normal Mind Blast cooldown.

@seanpeters86 can you test this, or maybe post a profile which uses dark thoughts?

Saeldur:
Closing pull request because there is a better solution available in the form of #5354

PR #5353, opened 2020-10-01
Try out covenants tests.

scamille:


scamille:
@nuoHep I tried the following strategy:
```
    strategy:
      fail-fast: false
      matrix:
        test: [ talents ]
        tier: [ Tier25 ]
        class: [ Death_Knight, Demon_Hunter, Druid, Hunter, Mage, Monk, Paladin, Priest, Rogue, Shaman, Warlock, Warrior ]
        fight_style: [ Patchwerk, DungeonSlice, HeavyMovement ]
        
        include:
          - test: covenants
            covenants: [ Kyrian, Venthyr, Night Fae, Necrolord ]
```
But it seems it does not allow nesting in that way.

What would be your suggestion to pull this off without duplicating the whole job?

nuoHep:
Hm, right, I don't think you can do a nested sequence like that. What do you want to do exactly? How the test runs should look like?

scamille:
I am not yet 100% sure what shadowland-specific tests should look like.

I wanted to start simple, and let each class (well all profiles for that class in Tier25 profile folder) run for the 4 available covenants. That could then be later extended to contain more covenant-specific combinations.

What we certainly don't want is to just add new dimensions into the matrix, or our combinations will explode. So we can't do class x fight_style x tier x covenants x soulbind x lengedary. That's why I wanted to create a new test, which takes the base profile (for now Tier25 profile set to level 60, later Preraid / Tier26 profile) and only goes through covenants x fight_style matrix.

nuoHep:
Right, Im pretty sure you can't do that in a single matrix and would have to add a separate job.

scamille:
On the other hand, the existing 'no covenants' is just another covenants choice besides the existing 4, so we could just add that to it.

There also needs to be a balance between the length of each job (which is currently 1-3min for going through all specs and all meaningsful talents for a given class) and number of jobs.

Scheduling a job for only 4 covenents per spec might be too short and the overhead per job becomes too large. We can probably go up to 5min for each job, so that an individual runner has a somewhat decent amount of work to do.

scamille:
Current idea is to let the strategy create a combination for each covenant.

The test would then iterate over all available conduits for that covenant and run a test for each one, with only that conduit being active.

That excludes conduit interactions, but should at least allow detecting basic implementation errors for each conduit/soulbind ability.

(Of course if the simulation time is so short for each job, we can also generate all possible conduit combinations to test all interactions, but that will very likely explode. And restricting it to only the set of possible combinations would make it quite complicated)

nuoHep:
> We can probably go up to 5min for each job, so that an individual runner has a somewhat decent amount of work to do.

We can always enable `-O1` for the build to speed up the tests. I didn't do it initially because it was a bad tradeoff for the limited amount of tests we had. But it may be a good tradeoff now.
It's basically a 2x slower compile for 2x faster binary. Considering most of recompiles are hot and should hit ccache (just by the virtue of most changes being the typical class dev/gear churn, even spell data updates is only like ~15 files to recompile) it may be worthwhile.

Also, we can't go faster than ~9 minutes for the whole CI run anyway as it's what docker & macos jobs take.

scamille:
Question now is how to get lists of soulbind abilities (which are generic and not class dependent, so I think I'll do them first) and after that conduits. Conduits are quite restricted by covenant and spec.

Raidbots does have JSON lists for all of them.
tests/talent_option is already a python script, which I am somewhat familiar with and in which it would be easy to build lists of conduits from a given json file.
I could just download the json file for each job from raidbots, or maybe we want to export (or already have) some list of soulbind/conduits directly in the simc project.

nuoHep:
Unless you also want to do the soulbinds, I'd do the covenants & conduits "matrix" in the test themselves. There is basically 1 conduit specific for each covenant and 4 potency conduits per spec which don't really interact with the covenants otherwise. So unless you want to explode the matrix with all possible combinations you have basically this per spec: [ conduit1, conduit2, conduit3, conduit4, Cov1, Cov1_conduit, Cov2, Cov2_conduit, Cov3, Cov3_conduit, Cov4, Cov4_conduit,  ].

nuoHep:
For soulbinds, I don't actually think there is any point in doing them for all class/specs tbh. They are all completely generic and shouldn't interact with classes much (or at all, unless there will be some apl shenanigans).
I'd actually do soulbinds in a random manner - for each soulbind ability grab a random spec and sim it with that covenant+soulbind combo.
It does mean our tests are going to be random somewhat, but for per-commit CI I think it should be ok tbh.

scamille:
I have managed to not really (want to) understand the conduit/soulbind system at all for the last months, and it still hasn't really clicked. Maybe I am not the right person for this :)

I am not really sure if random test inputs are a good solution. The soulbind abilities should be independent of the spec, but if there is a problem you can easily have false positives from a spec problem, and cannot deterministically replicate the problem.

scamille:
Ok so it seems that at least all classes run with all covenants, without crashing. That is a good thing :)

About the test matrix:

What would really be helpful is an easy way to replicate generating a test matrix inside the test script itself.

Probably the most helpful test would then be to either keep the current actions matrix of class & fight_style, or change it to 'spec' and move fight_style into the test as well.

the test could then be:
```
foreach fight_style in fight_styles:
   run test_talent
   foreach $covenant in covenants:
      foreach $conduit in [ conduit1, conduit2, conduit3, conduit4, Cov1, Cov1_conduit, Cov2, Cov2_conduit, Cov3, Cov3_conduit, Cov4, Cov4_conduit, ]:
         run covenant_test $covenant $conduit
   run trinket_sim
   .... run other tests
```

I know that this should all be possible with some scripting, question is just with what scripting language :)

Edit: Just to emphasize this some more: I think we should switch from 'class' to 'spec' in the test matrix, and use that as generator for jobs in the strategy matrix.
Each job should then run a set of tests for the given spec, with further combinations inside that jobs script.

Using spec is probably better representation of the unit of work in which a problem lets a test fail. Sure the specs also share a lot of class code, but differentiation between specs is still huge.

What do you think?

PR #5357, opened 2020-10-02
[Priest] fae fermata is no longer target capped

### LINKED ISSUE #100: Fae Fermata Wrathful Faerie can be on as many targets as you want
need to stop removing the faerie when you move it

seanpeters86:
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/100


scamille:
Is the ICD still there?

scamille (approved):

PR #5358, opened 2020-10-02
[Warlock] Implement Impending Catastrophe

javedbartlett:


scamille:
Does this really require a ground aoe event?

I am no warlock and have only read the spell description - but would it be possible to do like this:
- the main action is a normal simple action and you use impact_action variable instead of the custom code there.
- set the impact to aoe
- let the impact action trigger the dot in its impact function as a single target dot. (Will be triggered for all targets hit be the impact action)

Azevara:
Since we don't do any real things to handle damage done "in flight" this is being treated as doing that in flight damage right at the impact time.

The ground aoe event is unnecessary and can probably be scrapped in favor of just triggering the dot after the in flight damage trigger.

engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -55,6 +55,70 @@ struct drain_life_t : public warlock_spell_t
...
+
+    warlock_spell_t::impact( s );
+
+    // This is where we would implement Curse of Weakness/Tongues
+    make_event<ground_aoe_event_t>( *sim, p(), ground_aoe_params_t()
+        .n_pulses(1)
+        .target( s->target )
+        .action( impending_catastrophe_dot ),
+        true );
+
```
Azevara, line 134: As mentioned in the comments on this PR, the event code should be changed into something that triggers the action directly on each target, either through aoe or as part of the cycle through the target list that occurs for the inflight damage

Azevara (changes requested):

engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -55,6 +55,66 @@ struct drain_life_t : public warlock_spell_t
...
+struct impending_catastrophe_dot_t : public warlock_spell_t
+{
+  impending_catastrophe_dot_t( warlock_t* p )
+    : warlock_spell_t( "impending_catastrophe_dot", p, p->find_spell( 322170 ) )
+  {
+    background = true;
+    may_miss   = false;
+    dual       = true;
+    aoe        = -1;
+    dot_duration *= 1 + p->conduit.catastrophic_origin.percent();
```
nuoHep: Dunno if it works like that, but the conduit reads
  `Impending Catastrophe's damage over time and the inflicted Curse last |cFFFFFFFF${$s1}.1%|r longer against your primary target.`
If it actually does work like that, you'll have to change this so the dot duration isn't extended for all targets hit.

Pretty sure what you could do is to override `action_t::composite_dot_duration()` with smth like
```cpp
timespan_t composite_dot_duration( const action_state_t* s ) const override
{
  if ( s->chain_target = 0 ) // main target
    return dot_duration * ( 1 + p->conduit.catastrophic_origin.percent() );
  return dot_duration;
}
```

PR #5361, opened 2020-10-02
[Priest] add in Hungering Void talent

seanpeters86:
more work for https://github.com/WarcraftPriests/sl-shadow-priest/issues/84

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -1181,8 +1181,7 @@ priest_td_t::priest_td_t( player_t* target, priest_t& p ) : actor_target_data_t(
                                       ->set_duration( priest().conduits.fae_fermata.time_value() );
   buffs.hungering_void_tracking = make_buff( *this, "hungering_void_tracking", p.talents.hungering_void )
                                       ->set_quiet( true )
-                                      ->set_duration( timespan_t::from_seconds( 99 ) )
-                                      ->set_max_stack( 99 );
+                                      ->set_duration( timespan_t::from_seconds( 99 ) );
   buffs.hungering_void = make_buff( *this, "hungering_void", p.find_spell( 345219 ) );
 
   target->callbacks_on_demise.emplace_back( [ this ]( player_t* ) { target_demise(); } );
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -1270,7 +1270,6 @@ void priest_t::create_gains()
   gains.insanity_eternal_call_to_the_void_mind_sear =
       get_gain( "Insanity Gained from Eternal Call to the Void Mind Sear's" );
   gains.insanity_mind_sear = get_gain( "Insanity Gained from Mind Sear" );
-  gains.hungering_void_voidform_duration = get_gain( "Duration added to Voidform from Hungering Void" );
 }
 
 /** Construct priest procs */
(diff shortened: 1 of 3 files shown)


engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1213,6 +1216,27 @@ struct void_bolt_t final : public priest_spell_t
...
     }
+
+    if ( priest().talents.hungering_void->ok() && priest().buffs.voidform->check() )
+    {
+      priest_td_t& td = get_td( s->target );
+      // Check if this buff is active, every Void Bolt after the first should get this
+      if ( td.buffs.hungering_void_tracking->up() )
+      {
+        timespan_t seconds_to_add_to_voidform =
+            timespan_t::from_seconds( priest().talents.hungering_void->effectN( 3 ).base_value() );
```
lgkern: Should be done in the init, not here

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1213,6 +1216,27 @@ struct void_bolt_t final : public priest_spell_t
...
+        seconds_to_add_to_voidform +=
+            s->result == RESULT_CRIT
+                ? timespan_t::from_seconds( priest().talents.hungering_void->effectN( 4 ).base_value() )
+                : timespan_t::from_seconds( 0 );
```
lgkern: Same as before, init hv_vb_crit_bonus on init to clear up code.

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -1179,6 +1179,11 @@ priest_td_t::priest_td_t( player_t* target, priest_t& p ) : actor_target_data_t(
   buffs.wrathful_faerie_fermata = make_buff( *this, "wrathful_faerie_fermata", p.find_spell( 345452 ) )
                                       ->set_cooldown( timespan_t::zero() )
                                       ->set_duration( priest().conduits.fae_fermata.time_value() );
+  buffs.hungering_void_tracking = make_buff( *this, "hungering_void_tracking", p.talents.hungering_void )
+                                      ->set_quiet( true )
+                                      ->set_duration( timespan_t::from_seconds( 99 ) )
+                                      ->set_max_stack( 99 );
```
lgkern: Why 99 stacks?

lgkern (changes requested):
Added comments to the commit. Is HV bonus needed in 4 different locations? 

That code is starting to repeat itself in 4 different multipliers, maybe it should be turned into a function to put it all in one place only?

engine/class_modules/priest/sc_priest.cpp
seanpeters86: i did that just to be able to check the Nth void bolt at some point, but we dont really need it now so ill remove it

lgkern (approved):

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1133,27 +1133,24 @@ struct void_bolt_t final : public priest_spell_t
...
-          priest().procs.dissonant_echoes->occur();
-        }
-      }
     }
   };
 
   void_bolt_extension_t* void_bolt_extension;
   propagate_const<cooldown_t*> shadowfiend_cooldown;
   propagate_const<cooldown_t*> mindbender_cooldown;
+  timespan_t hungering_void_base_seconds;
```
scamille, line 1142: Just name the with _duration instead of seconds. It is a timespan type so you don't care what unit someone used to initialize its value.

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1213,6 +1222,24 @@ struct void_bolt_t final : public priest_spell_t
...
       void_bolt_extension->schedule_execute();
     }
+
+    if ( priest().talents.hungering_void->ok() && priest().buffs.voidform->check() )
+    {
+      priest_td_t& td = get_td( s->target );
+      // Check if this buff is active, every Void Bolt after the first should get this
+      if ( td.buffs.hungering_void_tracking->up() )
+      {
+        timespan_t seconds_to_add_to_voidform = s->result == RESULT_CRIT
```
scamille, line 1232: Same here use words like duration or extension instead of seconds.

scamille:
I don't really get the benefit of the tracking buff. Can you explain?

seanpeters86:
@scamille the tracking buff is just there to track the "2nd" void bolt in your voidform, I actually forgot to add a reset line but it should reset each voidform, in game its buggy though

PR #5369, opened 2020-10-03
[Priest] make mind sear and mind flay spec spells

seanpeters86:
Fixes some performance issues with calling `find_spell` in combat, and makes Mind Flay and Mind Sear priest spec spells so we can reduce some bloat we have in a few places.

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -1736,6 +1736,10 @@ void priest_t::init_spells()
   init_spells_discipline();
   init_spells_holy();
 
+  // Generic Spells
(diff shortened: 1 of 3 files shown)


scamille (commented):
Hmm I am not sure if you are not worsening performance with some of these changes.

The change in trigger_eternal_call_to_the_void is definitely important and needs to be made.
For the others, the id stored directly in the class that is probably more cache-friendly and has less indirections, so should be faster. I would say the mind_sear_id in priest_spell_t is the most micro-optimized one, even better than the ones in mind_blast_t.

seanpeters86:
@scamille ah ok, Saeldur thought the same but i wasnt sure. Removed the others for now then

scamille (approved):

PR #5370, opened 2020-10-04
[Priest] ECttV only gives insanity on the main target

### LINKED ISSUE #106: Implement bugged version of EcttV mind sear's
As of this bug: https://github.com/SimCMinMax/WoW-BugTracker/issues/687

the insanity is only generated from the main target hit, and not every enemy hit. Should implement this bug into simc with `priest().bugs()` checked.

seanpeters86:
implements https://github.com/SimCMinMax/WoW-BugTracker/issues/687
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/106

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -1082,8 +1082,9 @@ action_t* void_tendril_t::create_action( util::string_view name, const std::stri
 struct void_lasher_mind_sear_tick_t final : public priest_pet_spell_t
 {
   const double void_lasher_insanity;
+  player_t* main_target;
 
-  void_lasher_mind_sear_tick_t( void_lasher_t& p, const spell_data_t* s )
+  void_lasher_mind_sear_tick_t( void_lasher_t& p, const spell_data_t* s, player_t* _main_target )
     : priest_pet_spell_t( "mind_sear_tick", &p, s ),
       void_lasher_insanity( p.o().find_spell( 336214 )->effectN( 1 ).base_value() )
   {
@@ -1092,6 +1093,7 @@ struct void_lasher_mind_sear_tick_t final : public priest_pet_spell_t
     aoe        = -1;
     radius     = data().effectN( 2 ).radius_max();  // base radius is 100yd, actual is stored in effect 2
     affected_by_shadow_weaving = true;
+    main_target                = _main_target;
   }
 
   timespan_t composite_dot_duration( const action_state_t* ) const override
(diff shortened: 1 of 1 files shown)


scamille (commented):
Hmmm I am not sure if this is the correct solution, since you store the target only on construction and it can never change later.
Might be better to capture the target in ...tick_t::execute and then use that for your check in ...tick_t::impact.
@navv1234 what do you think?

navv1234:
I guess it depends on what "main target" is, but yeah, you cannot choose the target during construction time. The "main target" has to be chosen upon something run-time, the execute() for example.

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -1110,8 +1112,13 @@ struct void_lasher_mind_sear_tick_t final : public priest_pet_spell_t
   {
     priest_pet_spell_t::impact( s );
 
-    p().o().generate_insanity( void_lasher_insanity, p().o().gains.insanity_eternal_call_to_the_void_mind_sear,
-                               s->action );
+    // Currently bugged to only give insanity on the main target
+    // https://github.com/SimCMinMax/WoW-BugTracker/issues/687
+    if ( s->target == main_target || !p().o().bugs )
```
nuoHep: I *think* you can replace this with `s->target == parent_dot->target`

seanpeters86: looks great in testing that, thanks for the suggestion!

scamille (approved):

navv1234:
Btw, I'm not super sure if parent_dot works in these situations. From a quick look at the code, it seems very much a fixed construct, so it looks to me to be safe only if the target of the parent can never change. If that's the case here, then there's no issue whatsoever.

nuoHep:
Right, I kinda missed (or misread) it gets set only in action init and never updated after that. Which seems like a bug to me. But then... it's basically unused apart from some distance targeting logic for ground aoes (which also seems bugged, or not, it uses only original_x/y from the dot state). [ and a single use in the hunter module which will be gone in SL as its for azerite ].

So yeah, sorry, it potentially won't work in multi target sims with target switching (depending on what the pets actually do).

But, good news is that, looking at the code, it actually seems like you can simply use `s->target == target` as action_t updates the tick action target on each tick so it always points to the dot target.

navv1234:
Yep, I agree that it's a bug, however I have not had time to look at it more closely to see if the "bug" has any real implications on stuff.

If we can remove it, i'm all for it, but that probably requires some scouring of code.

PR #5372, opened 2020-10-05
Add trinket test

scamille:
For now just run all 3 trinket copy-profiles to see if this will be working or not.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -188,7 +188,7 @@ jobs:
           SIMC_ITERATIONS: 2
         run: |
           cd tests
-          bats --tap talents.bats covenants.bats
+          bats --tap classes.bats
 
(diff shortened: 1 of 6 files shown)


scamille:
@nuoHep I am closing this to bring it back later without bats.
trinket profiles files need better names first anyway.

scamille:
@nuoHep Bloodmallet adjusted the trinket profile names in b5f051f4e0724e1ef9623b557e5ada093dbafac5

I also asked him if it would be possible to export the trinkets in a different format, and he said that should be no problem and easy to do.
Maybe a list of tuple (name, item_id, ilevel) or something like that. That way we could let the CI test script orchestrate the iteration over the trinkets instead of running them in a single sim.

We could also use it directly in python with https://pypi.org/project/simc-support/

PR #5373, opened 2020-10-06
Warrior Shield Slam out of bounds fix.

Mwahi:
Changed effect #


scamille:
If I look at https://shadowlands.wowhead.com/spell=145045/shield-slam I see the power entry in effect `#2`, not `#1`

scamille:
@Melekus can I sometimes tag you for protection warrior reviews or are you not interested in looking after another tank spec?

PR #5378, opened 2020-10-06
Fix gui.pro for MacOS Builds

philipp-spiess:
Seems like this included some older references that aren’t used anymore.

Builds successfully after removing those:

Screenshot 2020-10-06 at 13 50 41


scamille:
I forgot to remove those when killing off the older profiles. Thanks.

PR #5379, opened 2020-10-06
[Soulbind] Necrolord volatile solvent

JosieBigler:
When running this code with the attached `profile.txt` I keep getting the following error: 

`Assertion failed: _mastery == player->composite_mastery(), file C:\Users\15027\Documents\Code\simc\engine\player\player_stat_cache.cpp, line 393`

[profile.txt](https://github.com/simulationcraft/simc/files/5335372/profile.txt)


scamille:
Feel free to rebase this on top of current shadowlands branch, since the problem should be fixed there.

if you also adjust that the buff is only triggered when the option is true instead of passing the option to the trigger function as @vituscze mentioned, this can be merged at some point :-)

JosieBigler:
I've addressed @vituscze comment and did a little cleanup for the spells tool-tip.

engine/player/soulbinds.cpp
```diff
@@ -821,7 +821,42 @@ void brons_call_to_action( special_effect_t& effect )
...
+    const spell_data_t* volatile_solvent_humanoid = effect.player->find_spell( 323491 );
+    const spell_data_t* volatile_solvent_beast = effect.player->find_spell( 323498 );
 
+    double amount = 0.02;
+    double mastery_amount = 2.0;
+
+    buff_t* buff_crit = make_buff( effect.player, "volatile_solvent_dragonkin", volatile_solvent_dragonkin )
+      ->set_pct_buff_type( STAT_PCT_BUFF_CRIT )
+      ->set_default_value( amount )
+      ->set_duration( timespan_t::from_seconds( 360 ) );
```
gastank: The duration info is already present in the spell data so these set_durations aren't necessary.

engine/player/soulbinds.cpp
```diff
@@ -821,7 +821,42 @@ void brons_call_to_action( special_effect_t& effect )
...
+    const spell_data_t* volatile_solvent_dragonkin = effect.player->find_spell( 323502 );
+    const spell_data_t* volatile_solvent_humanoid = effect.player->find_spell( 323491 );
+    const spell_data_t* volatile_solvent_beast = effect.player->find_spell( 323498 );
 
+    double amount = 0.02;
+    double mastery_amount = 2.0;
+
+    buff_t* buff_crit = make_buff( effect.player, "volatile_solvent_dragonkin", volatile_solvent_dragonkin )
+      ->set_pct_buff_type( STAT_PCT_BUFF_CRIT )
+      ->set_default_value( amount )
```
gastank: Unless it's not present in the spell data, it's best not to hardcode values like this and instead use set_default_value_from_effect_type(). For ex. the crit buff spell data shows `Apply Aura (6) | Modify Critical Strike% (290)`, so you'd want to use `set_default_value_from_effect_type( A_MOD_ALL_CRIT_CHANCE )`.

A_MOD_ALL_CRIT_CHANCE = 290 and other various A_MOD* enums can be found here https://github.com/simulationcraft/simc/blob/shadowlands/engine/dbc/data_enums.hh#L812

engine/player/soulbinds.cpp
```diff
@@ -821,7 +821,42 @@ void brons_call_to_action( special_effect_t& effect )
...
+      ->set_duration( timespan_t::from_seconds( 360 ) );
+
+    if ( effect.player->sim->shadowlands_opts.volatile_solvent_crit )
+      effect.player->register_combat_begin( [ buff_crit ]( player_t* p ) { buff_crit->trigger(); } );
```
gastank, line 832: You can move buff creation here within the block for the option check, since there's no need to create a buff that won't be used.

engine/player/soulbinds.cpp
```diff
@@ -821,7 +821,42 @@ void brons_call_to_action( special_effect_t& effect )
...
+      ->set_pct_buff_type( STAT_PCT_BUFF_AGILITY )
+      ->set_default_value( amount )
+      ->set_duration( timespan_t::from_seconds( 360 ) );
+
+    if ( effect.player->sim->shadowlands_opts.volatile_solvent_primary )
+      effect.player->register_combat_begin( [ buff_primary ]( player_t* p ) { buff_primary->trigger(); } );
+
+
+    buff_t* buff_mastery = make_buff( effect.player, "volatile_solvent_humanoid", volatile_solvent_humanoid )        
+      ->set_pct_buff_type( STAT_PCT_BUFF_MASTERY )
```
gastank: This looks like a ratings buff with `Apply Aura (6) | Modify Rating (189)` which is handled via the `stat_buff_t` template, which for single stat buffs automatically parses values & stat from the data. For ex. https://github.com/simulationcraft/simc/blob/shadowlands/engine/player/soulbinds.cpp#L441

engine/player/soulbinds.cpp
JosieBigler: I moved all buffs into their respective `if`` blocks

engine/player/soulbinds.cpp
JosieBigler: Made your suggested changes.

engine/player/soulbinds.cpp
JosieBigler: removed the buff duration.

PR #5382, opened 2020-10-06
Tests: switch to custom python runner from bats

nuoHep:
Out current tests are a bit arcane to setup and run because they use the
bats testing framework. While it's good and tested we use it only as
basically a glorified for loop, even calling to python for talent string
generation.

As we already use python for some parts of testing we may as well fully
switch to it. While it's true that it's another piece of code we have to
maintain, it's actually less code than we had with bats and has a bunch
of benefits:
* an actual language for setting up & running tests
* one less obscure dependency
* possibility to easily run tests locally

The "framework" is pretty limited & simple for now but seems to fully
replace what we have before. The main idea is that you simply add tests
into a list (with a possibility to group them) and then simply run them.

---
So that's what I have atm.

I'm not *entirely* sure about the way I handled setting up / adding tests here, but it seems to be pretty straightforward to me tbh. It's def is easier to follow than what we have currently with bats.

---

It also switches our matrix from class x fight_style jobs to per-spec ones. I feel like a spec is a more useful unit of work than a class, as specs per class is not homogenic. We can also always split it off back it tests start taking a lot of time individually.
As a bonus it fixes an issue where Hunter tests were also running Demon Hunter ones because of the way we globbed profiles.


scamille (approved):
Looks great.

I already had a look at it in your leaked push, and as far as I can tell everything is fine. I assume the error case will work just fine as well, and I am sure we will see it in action in the near future.

I definitely like that the script is a bit more verbose and transparent than bats, and allows for more output in non-error cases.

And while the backend is not that simple (but still easy enough to be understood and maintained by others with some python knowledge), it is very easy to add more tests to it.

nuoHep:
> I assume the error case will work just fine as well, and I am sure we will see it in action in the near future.

You can look for errors here: https://github.com/nuoHep/simc/actions/runs/293143138
I injected some stuff in hunter talents to check it.

> And while the backend is not that simple (but still easy enough to be understood and maintained by others with some python knowledge), it is very easy to add more tests to it.

Oh, I thought it's actually pretty simple :)
I did intend / do want to make it simple, so I'd really appreciate it if you could review it to point out the complex points.

scamille:
It is not really that complex - but without some python and programming knowledge it is definitely more complicated than eg. scripting the github actions workflow. But that is fine, and it really does provide value compared to the existing solution.

While we are at it: Would it be possible to set the error message so that it gets picked up by those annotations?
See https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-error-message

Edit:
Example error log:
```
 -- Process returned non-zero exit status 11                       --------------
['/home/runner/work/simc/b/ninja/simc', 'iterations=2', 'threads=2', 'cleanup_threads=1', 'default_actions=1', 'fight_style=HeavyMovement', '/home/runner/work/simc/simc/profiles/Tier25/T25_Hunter_Survival.simc', 'talents=0000003']
sc_hunter.cpp:4532:20: runtime error: member access within null pointer of type 'action_t'
sim_signal_handler: Segmentation fault! Iteration=0 Seed=1834724083625506936 TargetHealth=0.0

    #0 0x5b45aeb in (anonymous namespace)::attacks::chakrams_t::execute() /b/ninja/../../simc/engine/class_modules/sc_hunter.cpp:4532:20
    #1 0x4a116b3 in (anonymous namespace)::action_execute_event_t::execute() /b/ninja/../../simc/engine/action/sc_action.cpp:198:15
    #2 0x89acbb6 in event_manager_t::execute() /b/ninja/../../simc/engine/sim/event_manager.cpp:253:12
    #3 0x8d94e3e in sim_t::combat() /b/ninja/../../simc/engine/sim/sc_sim.cpp:1760:13
    #4 0x8dd26cc in sim_t::iterate() /b/ninja/../../simc/engine/sim/sc_sim.cpp:2877:5
    #5 0x8ddb885 in sim_t::run() /b/ninja/../../simc/engine/sim/sc_sim.cpp:3053:9
    #6 0x8fc2953 in sc_thread_t::native_t::execute(sc_thread_t*) /b/ninja/../../simc/engine/util/concurrency.cpp:47:10
    #7 0x8fc4ee6 in void std::__invoke_impl<void, void (*)(sc_thread_t*), sc_thread_t*>(std::__invoke_other, void (*&&)(sc_thread_t*), sc_thread_t*&&) /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/invoke.h:60:14
    #8 0x8fc4a7a in std::__invoke_result<void (*)(sc_thread_t*), sc_thread_t*>::type std::__invoke<void (*)(sc_thread_t*), sc_thread_t*>(void (*&&)(sc_thread_t*), sc_thread_t*&&) /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/invoke.h:95:14
    #9 0x8fc492d in void std::thread::_Invoker<std::tuple<void (*)(sc_thread_t*), sc_thread_t*> >::_M_invoke<0ul, 1ul>(std::_Index_tuple<0ul, 1ul>) /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/thread:244:13
    #10 0x8fc4737 in std::thread::_Invoker<std::tuple<void (*)(sc_thread_t*), sc_thread_t*> >::operator()() /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/thread:251:11
    #11 0x8fc3a9a in std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (*)(sc_thread_t*), sc_thread_t*> > >::_M_run() /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/thread:195:13
    #12 0x7f45714f0ce3  (/lib/x86_64-linux-gnu/libstdc++.so.6+0xd6ce3)
    #13 0x7f45712b1608 in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x9608)
    #14 0x7f45711aa102 in clone (/lib/x86_64-linux-gnu/libc.so.6+0x122102)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior sc_hunter.cpp:4532:20 in 
```
1. Is the ```-ffile-prefix-map=${{ runner.workspace }}/=/``` not working correctly or is it just unfortunate that you don't get nice paths in the UBSAN output?
2. I guess this would require some best-effort regex parsing, but might be doable to get file and line number for certain errors.

nuoHep:
> It is not really that complex - but without some python and programming knowledge it is definitely more complicated than eg. scripting the github actions workflow. But that is fine, and it really does provide value compared to the existing solution.

Ah, ok.
One thing I don't really like is how you have to manually add TestGroup/Tests into a list. I think we could do without that and just implicitly stuff them into a global list and go with that. I generally hate implicit global state like this but I think it could work fine here (and all testing frameworks I know of work like that tbh).

> While we are at it: Would it be possible to set the error message so that it gets picked up by those annotations?
> See https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-error-message

Right, I actually wanted to do that (see `IN_CI` global in the helper).
Doing simple annotations for failing tests is actually pretty easy: https://github.com/nuoHep/simc/actions/runs/293275267

> Edit:
> Example error log:
> 
> [ ... ]
> 
> 1. Is the `-ffile-prefix-map=${{ runner.workspace }}/=/` not working correctly or is it just unfortunate that you don't get nice paths in the UBSAN output?

It actually works, otherwise all paths would start with `/home/runner/work/simc`. The fact that `/b/simc/../../` is there is a bit unfortunate (it's a quirk of how cmake generates paths I believe) but I kinda cba'ed trying to clean that up.
> 2. I guess this would require some best-effort regex parsing, but might be doable to get file and line number for certain errors.

Right, it would be nice. But it would require parsing stderr for several different patterns (and sanitizers output is not exactly stable afaik). And most of the time you would still like to see a full error anyway.
What would actually be nice is if the annotations on the main page linked to the log. But alas.

nuoHep:
> > 1. Is the `-ffile-prefix-map=${{ runner.workspace }}/=/` not working correctly or is it just unfortunate that you don't get nice paths in the UBSAN output?
> 
> It actually works, otherwise all paths would start with `/home/runner/work/simc`. The fact that `/b/simc/../../` is there is a bit unfortunate (it's a quirk of how cmake generates paths I believe) but I kinda cba'ed trying to clean that up.

Oh, and, actually, this is a bit more complicated. I may be a bit off here, but as far as I remember `-ffile-prefix-map` is unused by Ubsan, and it uses `-fsanitize-undefined-strip-path-components=$UBSAN_STRIP_COUNT"` instead. Which is 🙄 tbh, but that's kinda historical.
So you have asserts/exceptions/asan using `-ffile-prefix-map` and ubsan using `-fsanitize-undefined-strip-path-components`. And they both strip only `/home/runner/work/simc` now.

scamille:
Adding the annotation with file and line id not only gets you the annotations, but I think github will also directly show it inside the pull request and mark the lines which fail.
And yes it would be nice if github would at least link the annotation to the failing job by default...

Anyway, adding error annotations is not really required and I fully understand that it would be a lot of work parsing several different stderr output patterns, depending on the error. I'd rather have the current setup working and keeping this as a "nice to have" addition for the feature.

Focusing on conduit tests probably has a greater value than the annotations.

Once this is in I personally would try to get some trinket test added, which should be a good way of seeing how usable the system really is.

> 
> 
> > It is not really that complex - but without some python and programming knowledge it is definitely more complicated than eg. scripting the github actions workflow. But that is fine, and it really does provide value compared to the existing solution.
> 
> Ah, ok.
> One thing I don't really like is how you have to manually add TestGroup/Tests into a list. I think we could do without that and  just implicitly stuff them into a global list and go with that. I generally hate implicit global state like this but I think it could work  fine here (and all testing frameworks I know of work like that tbh).

I also don't like global state if it can be avoided, and would just keep it like it is. You might be able to get away with a terser syntax by stuffing the tests as a generator into the TestGroup directly on construction, but that also increases the cognitive load when trying to understand what is going on.

PR #5386, opened 2020-10-07
Prot paladin shadowlands

OrcinusDrake:
On feature freeze to prep for merge.

~~Still tonnes of work to do so not expecting to merge yet, more an fyi I'm very slowly trying to do things~~


engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1590,12 +1580,15 @@ void paladin_t::init_gains()
...
   // Health
   gains.holy_shield                 = get_gain( "holy_shield_absorb" );
+  gains.first_avenger               = get_gain( "first_avenger_absorb" );
 
   // Holy Power
   gains.hp_templars_verdict_refund  = get_gain( "templars_verdict_refund" );
   gains.judgment                    = get_gain( "judgment" );
   gains.hp_cs                       = get_gain( "crusader_strike" );
   gains.hp_memory_of_lucid_dreams   = get_gain( "memory_of_lucid_dreams" );
+  gains.hp_avengers_shield             = get_gain( "avengers_shield" );
```
Melekus: This is not necessary, gains for relevant actions are automatically generated based on the action's tokenized name (especially if you're not using it anywhere).

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2091,6 +2085,15 @@ double paladin_t::composite_attribute_multiplier( attribute_e attr ) const
   if ( attr == ATTR_STAMINA )
   {
     m *= 1.0 + spec.protection_paladin -> effectN( 3 ).percent();
+    if ( buffs.redoubt -> up() )
```
Melekus, line 1954: You can use buffs.redoubt -> stack_value() as long as you set the default value to the value of -> effectN( 1 ).percent()

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2204,9 +2207,7 @@ double paladin_t::composite_bonus_armor() const
 
   if ( buffs.shield_of_the_righteous -> check() )
   {
-    shield_of_the_righteous_buff_t* sotr_buff = debug_cast<shield_of_the_righteous_buff_t*>( buffs.shield_of_the_righteous );
-
-    ba += sotr_buff -> value() * cache.strength() * ( 1.0 + sotr_buff -> avengers_valor_increase );
```
Melekus, line 2209: You probably don't want to remove the primary stat contribution from the armor buff on shield of the righteous.

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -117,40 +115,45 @@ struct blessed_hammer_tick_t : public paladin_spell_t
...
       num_strikes = 1;
-      sim -> error( "{} trying to hit less than one time with blessed_hammer, value changed to 1", p -> name() );
-    }
-    if ( num_strikes > 3 )
-    {
-      num_strikes = 3;
-      sim -> error( "{} trying to hit more than three times with blessed_hammer, value changed to 3", p -> name() );
+      sim -> error( "{} invalid blessed_hammer strikes, value changed to 1", p -> name() );
     }
+    // if ( num_strikes > 3 )
```
Melekus: You should probably keep some sort of upper bound for the number of ticks, even if you want to make it more flexible for the user.

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -300,42 +308,31 @@ struct hammer_of_the_righteous_t : public paladin_melee_attack_t
...
 {
-  timespan_t sotr_cdr; // needed for sotr interaction for protection
-
   judgment_prot_t( paladin_t* p, const std::string& options_str ) :
     judgment_t( p, options_str )
   {
     cooldown -> charges += as<int>( p -> talents.crusaders_judgment -> effectN( 1 ).base_value() );
-    cooldown -> duration *= 1.0 + p -> spec.protection_paladin -> effectN( 4 ).percent();
-
-    base_multiplier *= 1.0 + p -> spec.protection_paladin -> effectN( 12 ).percent();
```
Melekus, line 311: Looking at spelldata, this modifier is still there (although in effect#13)

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -300,42 +308,31 @@ struct hammer_of_the_righteous_t : public paladin_melee_attack_t
...
 
   // Special things that happen when Judgment damages target
   void impact( action_state_t* s ) override
   {
+    judgment_t::impact( s );
+
     if ( result_is_hit( s -> result ) )
     {
-      // Judgment hits/crits reduce SotR recharge time
       if ( p() -> sets -> has_set_bonus( PALADIN_PROTECTION, T20, B2 ) &&
```
Melekus: You can probably nuke the T20 code at this point

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -300,42 +308,31 @@ struct hammer_of_the_righteous_t : public paladin_melee_attack_t
...
       {
         p() -> cooldowns.avengers_shield -> reset( true );
       }
-    }
+      if ( p() -> spec.judgment_4 -> ok() )
+        td( s -> target ) -> debuff.judgment -> trigger();
 
-    judgment_t::impact( s );
+      if ( p() -> spec.judgment_3 -> ok() )
+        p() -> resource_gain( RESOURCE_HOLY_POWER, as<int>( p() -> find_spell( 220637 ) -> effectN( 1 ).base_value() ), p() -> gains.judgment );
```
Melekus: You should store the Holy Power value somewhere, it's not very efficient to run find_spell() on every impact() or execute().

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -535,6 +481,23 @@ struct shield_of_the_righteous_t : public paladin_melee_attack_t
   }
 };
 
+redoubt_buff_t::redoubt_buff_t( paladin_t* p ) :
```
Melekus: What is that for?

Melekus: Your expire_override doesn't do anything here and what's in the constructor could easily be swapped over to a simple make_buff() like in the previous implementation.

Melekus (changes requested):

engine/class_modules/paladin/sc_paladin_protection.cpp
OrcinusDrake: Yeah, but I get the right numbers without it. Could some automated system be taking it into account?

nuoHep: It's getting auto-applied in https://github.com/simulationcraft/simc/blob/8b90961d75ca46160701699bac8ff88dda2aff92/engine/class_modules/paladin/sc_paladin.cpp#L2718

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -315,23 +308,21 @@ struct judgment_prot_t : public judgment_t
     cooldown -> duration *= 1.0 + p -> spec.protection_paladin -> effectN( 5 ).percent();
   }
 
+  private:
```
Melekus: That value should be set during construction (it also doesn't really need to be specified private)

OrcinusDrake:
> You can check data().initial_stack() rather than hardcoding 3 in the case of moment of glory

Oops.

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -570,7 +569,10 @@ void paladin_t::target_mitigation( school_e school,
 
   if ( standing_in_consecration() )
   {
-    s -> result_amount *= 1.0 + ( cache.mastery() * mastery.divine_bulwark -> effectN( 2 ).mastery_value() );
+    double cons_multiplier = 1.0 + ( cache.mastery() * mastery.divine_bulwark_2 -> effectN( 1 ).mastery_value() );
```
Melekus: You shouldn't need to check for cons_2 -> ok(), effectN() will silently return 0 if the spell isn't found.

(it's also a bit weird to check for cons_2 but not for divine_bulwark_2 since the latter requires a higher level than the former)

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1217,6 +1217,7 @@ paladin_td_t::paladin_td_t( player_t* target, paladin_t* paladin ) :
   debuff.judgment_of_light = make_buff( *this, "judgment_of_light", paladin -> find_spell( 196941 ) );
   debuff.final_reckoning = make_buff( *this, "final_reckoning", paladin -> talents.final_reckoning );
   debuff.reckoning = make_buff( *this, "reckoning", paladin -> spells.reckoning );
+  debuff.vengeful_shock = make_buff( *this, "vengeful_shock", paladin -> find_spell( 340007 ) );
```
Melekus: Conduits don't change value throughout combat, you should be able to just set the default value here rather than calling it on every trigger.

Additionally, you should be able to replace the find_spell() with conduit.vengeful_shock->effectN( 1 ).trigger()

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2024,6 +2025,18 @@ double paladin_t::composite_melee_crit_chance() const
   return h;
 }
 
+double paladin_t::composite_player_target_multiplier ( player_t* target, school_e school ) const
+{
+  paladin_td_t* td = get_target_data( target );
+  double cptm = player_t::composite_player_target_multiplier( target, school );
+  if ( school == SCHOOL_HOLY && td -> debuff.vengeful_shock -> up() )
```
Melekus: Surprising that this is not a whitelist, although there probably aren't many generic Holy damage effects around (Light's Judgment maybe?).

OrcinusDrake: LJ was the main one I was thinking of. Not sure if there's any trinks or suchlike.

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -1082,18 +1081,168 @@ struct sanctified_wrath_t : public paladin_spell_t
...
 {
+  private:
+    typedef Base ab; // action base, eg. spell_t
+  public:
+    typedef holy_power_consumer_t base_t;
   bool is_divine_storm;
-  holy_power_consumer_t( const std::string& n, paladin_t* p, const spell_data_t* s ) :
-    paladin_melee_attack_t( n, p, s ),
+  holy_power_consumer_t( const std::string& n, paladin_t* player,
+      const spell_data_t* s = spell_data_t::nil() ) :
```
Melekus: I don't think defaulting spelldata to nil is a great idea, that should only be there for auto attacks.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -173,6 +172,7 @@ struct avenging_wrath_t : public paladin_spell_t
       background = true;
 
     harmful = false;
+    use_off_gcd = true;
```
Melekus: Actions that are off the global cooldown can be used before triggering the global cooldown with another action at the same timestamp. Setting ``use_off_gcd=true`` will make the action usable *at any time* including when the global cooldown isn't ready yet. This is not always necessary and can be set on each apl line individually for the same result as setting it in the constructor.
Unless you want to apply use_off_gcd to every Avenging Wrath line that will be added to the APL, I'd recommend not setting it in the constructor.

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -67,36 +55,98 @@ struct avengers_shield_t : public paladin_spell_t
...
+
+struct moment_of_glory_t : public paladin_spell_t
+{
+  moment_of_glory_t( paladin_t* p, const std::string& options_str ) :
+    paladin_spell_t( "moment_of_glory", p, p -> talents.moment_of_glory )
+  {
+    parse_options( options_str );
+
+    harmful = false;
+    use_off_gcd = true;
```
Melekus: Same as for Avenging Wrath. 
trigger_gcd shoulldn't need to be set to 0_ms if the correct spelldata is used.

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -117,40 +167,39 @@ struct blessed_hammer_tick_t : public paladin_spell_t
...
+      num_strikes = 2;
+      sim -> error( "{} invalid blessed_hammer strikes, value changed to 2", p -> name() );
     }
 
-    dot_duration = 0_ms; // The periodic event is handled by ground_aoe_event_t
+    // dot_duration = 0_ms; // The periodic event is handled by ground_aoe_event_t
+
     may_miss = false;
-    base_tick_time = 1667_ms; // Rough estimation based on stopwatch testing
+    base_tick_time = data().duration();
```
Melekus: Putting the duration in base_tick_time and using it to compute the actual pulse time of the ground aoe isn't very transparent, you should probably set both base_tick_time and dot_duration to 0 in the constructor and use data().duration() later in pulse time calculations.

engine/class_modules/paladin/sc_paladin.cpp
nuoHep: This check should be `dbc::is_school( school, SCHOOL_HOLY )` as far as I can tell. Otherwise it will miss composite schools like `HOLYFIRE`, `HOLYSTRIKE` and the like. Not sure if paladins themselves have spells with such but there were trinkets in the past.

Melekus (approved):

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -205,19 +210,23 @@ struct paladin_t : public player_t
     gain_t* judgment;
     gain_t* hp_cs;
     gain_t* hp_memory_of_lucid_dreams;
+    gain_t* hp_avengers_shield;
+    gain_t* hp_hammer_of_wrath;
```
Melekus: This isn't used and could get removed

PR #5387, opened 2020-10-07
Add trinket tests.

scamille:
Use argparse for run.py since I was not happy with the silent fail when invoking it without arguments.

simc-support is a helper library written by Bloodmallet which contains various game data as json and is easily accessible with python. It is a outside dependency, but I thought it was worth the effort. Can be disabled with --no-trinkets option.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -184,6 +184,9 @@ jobs:
             profiles
             tests
           key: ubuntu-clang-10-for_run-${{ github.sha }}
+          
+      - name: Install Python dependencies
+        run: pip install -r tests/requirements.txt
 
       - name: Run
         env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -95,9 +95,9 @@ jobs:
         run: cmake -H. -B'${{ runner.workspace }}/b/ninja' -GNinja -DBUILD_GUI=OFF
               -DCMAKE_BUILD_TYPE=Debug
               -DCMAKE_CXX_COMPILER=clang++-10 -DCMAKE_C_COMPILER=clang-10
-              -DCMAKE_CXX_FLAGS="-ffile-prefix-map=${{ runner.workspace }}/=/
+              -DCMAKE_CXX_FLAGS="-Og -ffile-prefix-map=${{ runner.workspace }}/=/
                 -fno-omit-frame-pointer -fsanitize=address,undefined
-                -fsanitize-undefined-strip-path-components=$UBSAN_STRIP_COUNT -Og"
+                -fsanitize-undefined-strip-path-components=$UBSAN_STRIP_COUNT"
               -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=lld -fsanitize=address,undefined"
               -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -97,7 +97,7 @@ jobs:
               -DCMAKE_CXX_COMPILER=clang++-10 -DCMAKE_C_COMPILER=clang-10
               -DCMAKE_CXX_FLAGS="-ffile-prefix-map=${{ runner.workspace }}/=/
                 -fno-omit-frame-pointer -fsanitize=address,undefined
-                -fsanitize-undefined-strip-path-components=$UBSAN_STRIP_COUNT"
+                -fsanitize-undefined-strip-path-components=$UBSAN_STRIP_COUNT -Og"
               -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=lld -fsanitize=address,undefined"
               -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -195,7 +195,7 @@ jobs:
           SIMC_PROFILE_DIR: ${{ github.workspace }}/profiles/${{ matrix.tier }}
           SIMC_THREADS: 2
           SIMC_ITERATIONS: 2
-        run: tests/run.py ${{ matrix.spec }}
+        run: tests/run.py ${{ matrix.spec }} --test-trinkets
 
   build-docker:
     name: docker
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -184,9 +184,9 @@ jobs:
             profiles
             tests
           key: ubuntu-clang-10-for_run-${{ github.sha }}
-          
+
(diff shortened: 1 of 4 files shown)


scamille:
Damn, it is quite hard to split our CLASS_SPEC string into class and spec.
There is eg. Death_Knight_Frost and also Hunter_Beast_Mastery

navv1234:
New naming scheme? class-spec?

scamille:
No I don't have plans to rename both our profiles and all the CI stuff :) Added a new function to simc-support, Bloodmallet said he will gladly accept PRs. Let's see if my python is good enough for him :P

scamille:
@nuoHep we talked about the optimization level for those tests in a previous issue.
I remember gcc having -Og which gave pretty good speed without sacrificing much for debugging. It is quite hard to find out what Cmake uses for optimization level for Debug builds, but I suspect it might be -O0. Maybe changing that would already give a huge speed improvement.

Edit: I'll just try it out and see if it makes a difference in runtime :-)

scamille:
88s vs 204s for Priest_Shadow. I think that's a good gain, and compile time should be lower again after the first run with ccache.

Bloodmallet:
PR was accepted, simc-support now has a function that allows gettting spec information (like trinkets) from the naming scheme simc uses.

Bloodmallet:
As a general question regarding testing simc via python, what is your opinion about slightly changing the set up of tests to use Unittests? In that case subtests for each trinket, whatever iterable you're working with could be reported about. and using Unittests would enable a whole lot of worksuites for test-reporting.

.github/workflows/main.yml
```diff
@@ -97,7 +97,7 @@ jobs:
               -DCMAKE_CXX_COMPILER=clang++-10 -DCMAKE_C_COMPILER=clang-10
               -DCMAKE_CXX_FLAGS="-ffile-prefix-map=${{ runner.workspace }}/=/
                 -fno-omit-frame-pointer -fsanitize=address,undefined
-                -fsanitize-undefined-strip-path-components=$UBSAN_STRIP_COUNT"
+                -fsanitize-undefined-strip-path-components=$UBSAN_STRIP_COUNT -Og"
```
nuoHep: Nit: I'd move it as the first flag instead of burying it after all the debugging/sanitizers related junk.

tests/run.py
```diff
@@ -22,14 +29,30 @@
...
+            get_wow_spec_from_combined_simc_name(klass))
+        fight_style = args.trinkets_fight_style
+        grp = TestGroup('{}/{}/trinkets'.format(profile, fight_style),
+                        fight_style=fight_style, profile=path)
+        tests.append(grp)
+        for trinket in trinkets:
+            Test('{} ({})'.format(trinket.name, trinket.item_id), group=grp, args=[
+                    ('trinket1', '{},id={},ilevel={}'.format(trinket.name, trinket.item_id, trinket.min_itemlevel))])
+else:
+    print("No profile found for {}".format(klass))
```
nuoHep: This seems to be printed out for all specs even if there actually is a profile.
I believe an `else` branch bound to a loop gets executed if there is no `break` in the loop. Or smth like that.

nuoHep:
> @nuoHep we talked about the optimization level for those tests in a previous issue.
> I remember gcc having -Og which gave pretty good speed without sacrificing much for debugging. It is quite hard to find out what Cmake uses for optimization level for Debug builds, but I suspect it might be -O0. Maybe changing that would already give a huge speed improvement.
> 
> Edit: I'll just try it out and see if it makes a difference in runtime :-)

Yeah, debug builds is basically `-O0`. `-Og` in clang is an alias for `-O1` (there only for compat with gcc), but better to use that in case we switch compilers or stmh.

It's generally around 2x slower compile for 2-3x faster runtime. And, yeah, given our typical commits ccache should do a good job in speeding it up.

nuoHep:
> As a general question regarding testing simc via python, what is your opinion about slightly changing the set up of tests to use Unittests? In that case subtests for each trinket, whatever iterable you're working with could be reported about. and using Unittests would enable a whole lot of worksuites for test-reporting.

I'll admit I don't know about `unittest` much. But skimming through https://docs.python.org/3.9/library/unittest.html I feel like it... does not really fit our use case? It's a nice unit testing framework for *python* code, sure, but we are not testing python here. All our tests are basically just data for a `subprocess.run()` call. Like... the only thing I can imagine we could replace in `helper.py` would be `def run(tests)`. And even then we'd still need to hook up return-code message and `stderr` pass-through to `unitest` somehow (and I have some plans on massaging that further with annotations and the like if I get time).
I may be totally wrong and maybe it would be 100% worth it, but I'd like to look at the code first. Because I'm just not seeing it right now and think it would be more complicated and more code.

Also, I don't expect people to run the tests locally (even if it's easier now). I assume the tests will still be ran only in CI and we don't need sophisticated test-reporting here.

tests/run.py
```diff
@@ -22,14 +29,30 @@
...
         tests.append(grp)
         for covenant in COVENANTS:
             Test(covenant, group=grp,
-                 args=[ ('covenant', covenant.lower()), ('level', 60) ])
+                 args=[('covenant', covenant.lower()), ('level', 60)])
 
-sys.exit( run(tests) )
+    if not args.no_trinkets:
+        from simc_support.game_data.WowSpec import get_wow_spec_from_combined_simc_name
+        from simc_support.game_data.Trinket import get_trinkets_for_spec
```
nuoHep: While I explicitly wanted to avoid external deps (because that was one of the problems with bats imo) I'm ok with it as long as it's optional. What I don't like is that you have to explicitly *disable* the trinket tests to not get yelled at for a missing dependency.

I think these imports should be wrapped into a `try: ... except ImportError:` block and these tests should be implicitly disabled if the import failed, with a suitable message.
But then... I don't really expect anyone to run the tests locally apart from myself.

scamille: Ok will do.
I do think it is a good thing that they can be run locally, and if it is only to reduce the number of broken commits I have to push. I also don't expect anyone running them regularly, only for development and other rare circumstances.
I definitely wouldn't put too much effort into it, but so far I think we are still keeping with a pretty standard python setup. Argparse fit in quite naturally and thank God python has a easy way to declare dependencies.

scamille:
I agree that a python unittest library is probably the wrong tool for the job here. But I also don't know much about the library itself. If you think it is better for the job of integration-testing a c++ application, and simplifies the code/scripting required then sure why not :-)

nuoHep (approved):

Bloodmallet:
@nuoHep @scamille Thank you for your responses. Once Shadowlands launched, or whenever I'll have some more free-time, I'd like explore unittest, or pytest for this setup in a fork. If I reach something nice, I'll poke you again about it.

PR #5389, opened 2020-10-08
[Priest] Hungering Void Rework

### LINKED ISSUE #84: New Talent: Hungering Void
This replaces Legacy of the Void.

> Your second Void Bolt cast on the same target will deal 25% more damage, cause the target to receive 15% more damage from you for until cancelled, and extend the duration of your Void Form by 1 sec. If your Void Bolt critically hits, your Void Form is extended by 1 sec.

TODO
- [x] figure out if its every second void bolt, or every void bolt after the first
- [x] until we get spell data, hard code a 6s duration or something for the buff on the target
- [x] what happens when you have an in flight void bolt that hits the target after voidform expires?
- [ ] does the damage modifier affect all spells? trinkets? mindbender?

ACs:
- [x] Remove Legacy of the Void (https://github.com/WarcraftPriests/sl-shadow-priest/issues/84)
- [x] Remove insanity drain code (https://github.com/WarcraftPriests/sl-shadow-priest/issues/84)
- [x] Add in Hungering Void
  - create a dummy debuff to track the "second" void bolt part (like shadow crash)
  - create a debuff to track the flat 15% modifier, should probably go in the same place as Mastery

seanpeters86:
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/84

the spell has been reworked:
  - no longer amps Void Bolt damage
  - moved damage increase to the buff spell
  - damage increase affects all priest spells, including trinkets, but nothing to do with pets
  - voidform extension on crits stored by itself as 2, no need to add 1+1
  - buff tracking was no longer needed, it now just always triggers the buff on every void bolt cast
  - added support for max targets

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -1503,6 +1503,11 @@ double priest_t::composite_player_target_multiplier( player_t* t, school_e schoo
     m *= 1.0 + target_data->buffs.schism->data().effectN( 2 ).percent();
   }
 
+  if ( hungering_void_active( t ) )
+  {
+    m *= ( 1 + talents.hungering_void_buff->effectN( 1 ).percent() );
+  }
+
   return m;
 }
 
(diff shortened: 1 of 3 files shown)


nuoHep:
> Because of that I didnt add anything for max_targets, I think its supported automatically.

We don't support stuff like that automatically, sadly. So you have to implement the `This effect can only be active on one target at any one time.` part yourself.

lgkern (approved):

PR #5390, opened 2020-10-08
Covenant + legendary ci

scamille:
I refactored the run.py / argparse a bit, adding a list of tests to run.

The whole thing is starting to get a bit large and probably needs some cleanup/refactoring at some point, but maybe we are in a good enough / finished state for shadowlands now.

### DIFF
--- a/tests/run.py
+++ b/tests/run.py
@@ -120,6 +120,8 @@ parser.add_argument('--trinkets-fight-style', default='DungeonSlice', type=str,
                     help='Fight style used for trinket simulations.')
 parser.add_argument('--soulbind-fight-style', default='DungeonSlice', type=str,
                     help='Fight style used for soulbind simulations.')
+parser.add_argument('--max-profiles-to-use', default='0', type=int,
+                    help='Maximum number of profiles to use per spec. 0 means use all available profiles')
 args = parser.parse_args()
 
 
@@ -129,6 +131,9 @@ print(' '.join(klass.split('_')))
 
 tests = []
 profiles = list(find_profiles(klass))
+if args.max_profiles_to_use != 0:
+    profiles = profiles[:args.max_profiles_to_use]
+
 if len(profiles) == 0:
     print("No profile found for {}".format(klass))
 
--- a/tests/run.py
+++ b/tests/run.py
@@ -2,6 +2,7 @@
 
 import sys
 import argparse
+import random
 
 from helper import Test, TestGroup, run, find_profiles
 from talent_options import talent_combinations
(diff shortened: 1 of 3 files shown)


scamille:
@nuoHep I did not hook up the soulbind tests to the CI so as not to run it for each spec.
Either need some input from you on how to do it, or you can do it later :-)

tests/run.py
```diff
@@ -2,41 +2,136 @@
...
+        for legendary in legendaries:
+            Test('{} ({} / {})'.format(legendary.full_name, legendary.id, legendary.bonus_id), group=grp, args=[
+                ('trinket1', '"{}",bonus_id={}'.format(legendary.full_name, legendary.bonus_id))])
+
+def test_soulbinds(klass: str, path: str):
+    try:
+        from simc_support.game_data.WowSpec import get_wow_spec_from_combined_simc_name
+        from simc_support.game_data.SoulBind import SOULBINDS
+    except ImportError:
+        print("Error importing simc-support library for legendary test. Skipping test.")
```
nuoHep: for *soulbinds* test

tests/run.py
```diff
@@ -2,41 +2,136 @@
...
+
+    spec = get_wow_spec_from_combined_simc_name(klass)
+    conduits = get_conduits_for_spec(spec)
+    for fight_style in FIGHT_STYLES:
+        grp = TestGroup('{}/{}/covenants'.format(profile, fight_style),
+                        fight_style=fight_style, profile=path)
+        tests.append(grp)
+
+        # Test covenants
+        for covenant in COVENANTS:
```
nuoHep: Considering this test requires simc_support library now anyway, I think we can drop our local `COVENANTS` array and use the one from the library.

tests/run.py
```diff
@@ -2,41 +2,136 @@
...
+            Test(talents, group=grp, args=[('talents', talents)])
+
+
+def test_covenants(klass: str, path: str):
+    try:
+        from simc_support.game_data.WowSpec import get_wow_spec_from_combined_simc_name
+        from simc_support.game_data.Conduit import get_conduits_for_spec
+    except ImportError:
+        print("Error importing simc-support library for covenant test. Skipping test.")
         return
```
nuoHep: Now that we have 4 places doing basically the same thing, I think we can move the import attempt of `get_wow_spec_from_combined_simc_name` (as it's used everywhere) to the top and add a bool like HAVE_SIMC_SUPPORT to check if we can run the tests.

Smth like
```python
try:
    from simc_support.game_data.WowSpec import get_wow_spec_from_combined_simc_name
    HAVE_SIMC_SUPPORT = True
except ImportError:
    HAVE_SIMC_SUPPORT = False
```
at the top and then
```python
if not HAVE_SIMC_SUPPORT:
    print("Error importing simc-support library for covenant test. Skipping test.")
    return
from simc_support.game_data.Conduit import get_conduits_for_spec
```
in this case for example.

nuoHep: Or, tbh, while I don't really like external deps, I think we can just require `simc_support` at this point.
Will also let us print human readable names for talent tests.

nuoHep:
> @nuoHep I did not hook up the soulbind tests to the CI so as not to run it for each spec.
> Either need some input from you on how to do it, or you can do it later :-)

Did you try running them? Mostly interested in how long do they take if you run the whole thing for each spec.

Hm... there are 12 soulbinds, ~8 soulbind abilities per soulbind. That's 96 more tests per fight style. We have ~180 with this PR already at 2-3 minutes per spec test job. So that's only like 50% more tests/time than we already have if we do them as trinkets with only a single fight style. Which seems pretty ok to me tbh.

Upd: and the tests will actually be faster once we transition them off Tier25 profiles as we are essentially stacking 2 expansions worth of systems on top of each other here (granted, without corruptions at least).

scamille:
Yes I did run them, and they were successful for Priest_Shadow.
96 tests is correct. We can just run them for a single fight style and it should still stay under 4min, max 5min for the worst specs.

So I will add them this way for now. Even if most are not tied to any specific class, it does not really hurt that much and should help gastank and others who are actively developing soulbinds for now to catch any problems with class interactions.

Thanks for the review.

scamille:
Soulbind test is enabled, run.py refactored to require simc-support include and only do it once at the top.

Please review get_talent_name addition. My python is a bit rusty, so I hope my talent_row calculations are ok.

scamille:
About test name padding because of too long solbind test names:
We could either just add 40 characters (I assume your current top limit is 80 chars, so that would still result in something <=120).
Or you could calculate the longest test name and use that, but that is probably not worth the effort.

scamille:
Oh and I also noticed Windwalker runtime is 9m. That is because there are 2 profiles in Tier25.
We could:
- Live with it.
- Only select the first profile
- Select a random profile

scamille:
Missing talent names ("Unknown") are a result of Bloodmallet/simc_support#5

tests/run.py
```diff
@@ -2,41 +2,134 @@
...
+        tests.append(grp)
+        for legendary in legendaries:
+            Test('{} ({} / {})'.format(legendary.full_name, legendary.id, legendary.bonus_id), group=grp, args=[
+                ('trinket1', '"{}",bonus_id={}'.format(legendary.full_name, legendary.bonus_id))])
+
+
+def test_soulbinds(klass: str, path: str):
+    fight_style = args.soulbind_fight_style
+    for covenant in Covenant.COVENANTS:
+        grp = TestGroup('{}/{}/soulbinds'.format(profile, fight_style),
```
nuoHep: Maybe move the covenant to the group name (considering TestGroups are per covenant)?
Everyone will know their soulbind names (if not soulbind abilities) in like a week anyway.
Should also help with test name lengths a bit.

tests/run.py
```diff
@@ -2,41 +2,134 @@
...
+                    Test('{} - {} ({}) Rank {}'.format(covenant.full_name, conduit.full_name, conduit.id, rank), group=grp, args=[
+                        ('covenant', covenant.simc_name), ('level', 60), ('soulbind', soulbind_argument)])
+
+        # test conduits available for all 4 covenants independent from covenant selection.
+        for conduit in conduits:
+            if len(conduit.covenants) == 4:
+                rank = random.choice(conduit.ranks) + 1
+                soulbind_argument = '{}:{}'.format(conduit.id, rank)
+                Test('{} ({}) Rank {}'.format(conduit.full_name, conduit.id, rank), group=grp, args=[
+                    ('level', 60), ('soulbind', soulbind_argument)])
```
nuoHep: One thought occurred to me while thinking about the profiles: this won't actually enable the conduits because there is no covenant set afaik. Which is a bummer...
Could just bite the bullet and just test all potency conduits with all covenants. Will add 12 more tests, but at least we'll know they def work.

nuoHep (commented):
> Please review get_talent_name addition. My python is a bit rusty, so I hope my talent_row calculations are ok.

I actually didn't imply you should add talent names here, just that it would be a nice possibility :)
Considering python doesn't seem to have a `find_first_not_of()` equivalent lgtm, and it works so 👍 

> Oh and I also noticed Windwalker runtime is 9m. That is because there are 2 profiles in Tier25.
> We could:
> 
> * Live with it.
> * Only select the first profile
> * Select a random profile

Dunno tbh. We could also
* Create profiles specific to CI.

We will inevitably switch to shadowlands profiles which will have both covenants, soulbinds/counduits & legendaries which may (or may not) interact with whatever we trying to test here in "interesting" ways class devs don't really expect at this moment (like having 2 legendaries or 2 of the same trinket equipped for example). So may be worth it.

scamille:
I don't think it is worth creating separate profiles. As long as we can reset whatever we want to test, then this should be good enough. But you are right in that can get difficult for items.
The other possibility is to create virtual, naked profiles for the CI tests, but that might be a bit unrepresentative for usual content.

PR #5398, opened 2020-10-09
[Rogue] Fix ilevel calculations for Spyglass

danlemmond:
Spyglass was calculating at ilevel 224, set to ilevel 130.


vituscze:
You also need to change the generator (https://github.com/simulationcraft/simc/blob/shadowlands/profiles/generators/Tier25/T25_Generate_Rogue.simc) otherwise your change will get overwritten next time profiles are regenerated.

PR #5402, opened 2020-10-11
[Priest] support proccing sephuz proclamation

seanpeters86:
starting work on https://github.com/WarcraftPriests/sl-shadow-priest/issues/103

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -1760,7 +1760,7 @@ void priest_t::init_scaling()
 void priest_t::init_finished()
 {
   base_t::init_finished();
-  
+
   buffs.sephuzs_proclamation = buff_t::find( this, "sephuzs_proclamation" );
 }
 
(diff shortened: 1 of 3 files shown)


scamille:
Maybe we should do this differently.

Let silence call a player_t function interrupt (not sure if we already have one?).

In there we can call debuffa.casting.expire() as well as that sephuz code, maybe guarded behind a option to also proc on non-casting targets.

What do you think @gastank @vituscze ?

@seanpeters86 your commit looks otherwise good. You could have created the variable directly in silence_t instead of in priest_t. The only thing that looks suspicious is target type enemy add. Your intention is to have something which expresses if the target is interruptible,. right?

seanpeters86:
Hm well. This implementation is kind of here to let me cheat. Yes we can proc sephuz on a casting add like casting patchwork. But I also want to be able to cast it off of silencing a mob arbitrarily even if they are not casting.

gastank:
If we want to expand out to allow other modules to always proc a successful kick regardless of the target, maybe the apl controllable option is the way to go. The check for casting happens in action_t::execute(), so we can add a 'always_succeed' option applicable to actions like silence with is_interrupt set true. So `actions=silence,if=blah` would only work on casting targets, but `actions=silence,always_succeed=1,if=blah` would work on all fight styles.

scamille:
My interpretation initially was that you wanted to fake a interrupt, without realizing that it procs just from silencing a mob.

As discussed in Discord I still think we should have some simple functions in player_t like silence() and maybe dispel() which then take care of all this. One could argue that is overkill, but it does keep class modules simple and offers to hookup whatever Blizzard comes up with next in a more central location.

scamille (approved):

PR #5403, opened 2020-10-11
Guardian APL changes

srenauld:
Catha worked a bit more on the guardian APL, and I implemented the changes:
- Optimize the rotation when incarn is chosen as a talent
- Optimize the use of adaptive swarm
- Add condition to allow the use of covenant abilities in AoE
- Optimize the use of galactic guardian
- Changes the handling of combo points in cat form to avoid losing combo points


srenauld:
(There's also a stray shadowpriest change, somebody forgot that the third param of `add_action` is the comment, not the second)

PR #5411, opened 2020-10-12
Protection Paladin (Phase 2)

OrcinusDrake:
On feature freeze in preparation for merge


engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -113,6 +118,10 @@ struct avengers_shield_t : public avengers_shield_base_t
       p() -> cooldowns.avengers_shield -> reset( false );
       p() -> buffs.moment_of_glory -> decrement( 1 );
     }
+    if ( p() -> legendary.holy_avengers_engraved_sigil -> ok() && rng().roll( holy_avenger_proc_chance ) )
+    {
+      p() -> cooldowns.avengers_shield -> reset( false );
+    }
```
Melekus: You may want to introduce a proc object similar paladin_t::procs.grand_crusader to track the AS resets from this legendary effect.

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -439,6 +448,7 @@ void shield_of_the_righteous_buff_t::expire_override( int expiration_stacks, tim
 
 struct shield_of_the_righteous_t : public holy_power_consumer_t<paladin_melee_attack_t>
 {
+  double punish_the_guilty_value;
```
Melekus: Is it really necessary to use a new value inside sotr instead of just calling p -> conduit.punish_the_guilty.percent() during runtime?

OrcinusDrake: Was in response to https://github.com/simulationcraft/simc/pull/5386#discussion_r501386063 but simplifying it appeals to me. I suppose in comparison to that earlier review the value should already be loaded up in a variable rather than having to look it up.

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -487,6 +489,14 @@ struct shield_of_the_righteous_t : public holy_power_consumer_t<paladin_melee_at
     {
       p() -> trigger_memory_of_lucid_dreams( 1.0 );
     }
+
+    if ( p() -> conduit.resolute_defender -> ok() && p() -> buffs.ardent_defender -> up() )
+    {
+      p() -> buffs.ardent_defender -> extend_duration( p(),
+        p() -> conduit.resolute_defender.percent() * ad_duration
```
Melekus: You should be able to call p() -> buffs.ardent_defender -> buff_duration() in runtime instead of re-storing the data's duration value in sotr.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2312,6 +2315,19 @@ double paladin_t::resource_gain( resource_e resource_type, double amount, gain_t
     {
       amount *= 1.0 + buffs.holy_avenger -> data().effectN( 1 ).percent();
     }
+
+    if (
+        legendary.from_dusk_till_dawn -> ok() &&
+        amount > 0 &&
+        resources.current[ RESOURCE_HOLY_POWER ] < resources.base[ RESOURCE_HOLY_POWER ] &&
```
Melekus: It's clearer to use resources.max[] than resources.max[] even if they return the same value.

Melekus: You could also move the whole block below player_t::resource_gain() and only check whether the amount returned by it is != 0 and if you're currently at max.

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -1104,6 +1112,15 @@ struct holy_power_consumer_t : public Base
...
 
+    if (
+        hp_used > 0 &&
+        p -> legendary.from_dusk_till_dawn -> ok() &&
+        p -> resources.current[ RESOURCE_HOLY_POWER ] == p -> legendary.from_dusk_till_dawn -> effectN( 2 ).base_value()
+      )
+    {
+      p -> buffs.blessing_of_dusk -> trigger();
+    }
+
```
Melekus: Similarly to blessing of dawn, this could be move to paladin_t::resource_loss() after player_t::resource_loss() and would only need to check whether the player is at 0 and if the player actually lost any resource this action.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -950,6 +965,13 @@ struct vanquishers_hammer_t : public holy_power_consumer_t<paladin_melee_attack_
     holy_power_consumer_t::impact( s );
 
     p() -> buffs.vanquishers_hammer -> trigger();
+
+    if ( p() -> conduit.righteous_might -> ok() )
+    {
+      r_m_heal -> base_dd_min = r_m_heal -> base_dd_max =
+        s -> result_amount * p() -> conduit.righteous_might.percent();
+      r_m_heal -> execute();
+    }
```
Melekus: Isn't the conduit value only for the damage increase on VH, with the conversion rate to healing being a static 100%?

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1003,6 +1002,15 @@ struct divine_toll_t : public paladin_spell_t
       p() -> active.divine_toll -> schedule_execute();
     }
   }
+
+  void execute() override
+  {
+    paladin_spell_t::execute();
+    if ( p() -> conduit.ringing_clarity -> ok() && rng().roll( p() -> conduit.ringing_clarity.percent() ) )
+      for ( int hits = 0; hits < p() -> conduit.ringing_clarity -> effectN( 2 ).base_value(); hits ++ )
+        p() -> active.divine_toll -> schedule_execute();
```
Melekus, line 1014: Might want to make sure that this uses the right target, as the tooltip mentions "on the main target".

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -1145,11 +1145,12 @@ struct holy_power_consumer_t : public Base
     // Free sotr from vanq does not proc RP 2020-09-10
     if ( p -> talents.righteous_protector -> ok() && !ab::background )
     {
-      timespan_t reduction = timespan_t::from_seconds(
-        // Why do I need to divide this by 10? Just give me sec or milli, what is this??
-         -1.0 * p -> talents.righteous_protector -> effectN( 1 ).base_value()
-         * num_stacks / 10
-       );
+      // Why is this in deciseconds?
+      timespan_t reduction = -100_ms * p -> talents.righteous_protector -> effectN( 1 ).base_value();
```
Melekus: Can probabkly just keep using the old format here ``timespan from seconds(value / 10)`` since it's the format that the tooltip code applies. Stupidformattinginterpretation#1 or stupidformattinginterpretation#2 doesn't make much difference anyway.

You should also probably use cost() instead of num_stacks since that one is specific to crusade code (unless you expanded it somewhere else)

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -522,6 +537,20 @@ struct shield_of_the_righteous_t : public holy_power_consumer_t<paladin_melee_at
       );
     }
 
+    if ( !background )
```
Melekus, line 543: Could probably simplify this to avoid repeating a lot of similar code.
Maybe something like
```
    {
      if ( p() -> buffs.shining_light_free -> up() || p() -> buffs.shining_light_stacks -> at_max_stacks() )
      {
        p() -> buffs.shining_light_stacks -> expire();
        p() -> buffs.shining_light_free -> trigger();
      }
      else
        p() -> buffs.shining_light_stacks -> trigger();
    }
```

engine/class_modules/paladin/sc_paladin.hpp
OrcinusDrake: Crusade, Righteous Protector, Relentless Inquisitor (the azerite version) (and I suppose fist of justice but NYI) all use the same logic of "only give value for holy power spent", so holy power reductions in general don't proc them. Hence using the same variable

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1403,7 +1403,7 @@ void paladin_t::init_base_stats()
   base.spell_power_per_intellect = 1.0;
 
   // Boundless Conviction raises max holy power to 5
-  resources.base[ RESOURCE_HOLY_POWER ] = 3 + passives.boundless_conviction -> effectN( 1 ).base_value();
+  resources.base[ RESOURCE_HOLY_POWER ] = resources.max[ RESOURCE_HOLY_POWER ] = 3 + passives.boundless_conviction -> effectN( 1 ).base_value();
```
Melekus: resources.max shouldn't need to be set here.

OrcinusDrake: It was set so it could be used in https://github.com/simulationcraft/simc/pull/5411#discussion_r505546437, but after restructuring the flow didn't even need it.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -909,27 +933,52 @@ struct hand_of_reckoning_t: public paladin_melee_attack_t
...
+    }
+};
+
 struct vanquishers_hammer_t : public holy_power_consumer_t<paladin_melee_attack_t>
 {
+  righteous_might_t* r_m_heal;
   vanquishers_hammer_t( paladin_t* p, const std::string& options_str ) :
-    holy_power_consumer_t( "vanquishers_hammer", p, p -> covenant.necrolord )
+    holy_power_consumer_t( "vanquishers_hammer", p, p -> covenant.necrolord ),
+    r_m_heal( new righteous_might_t( p ) )
```
Melekus: Shouldn't need to init r_m_heal if you don't have the conduit.

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -1126,28 +1127,39 @@ struct holy_power_consumer_t : public Base
...
     {
-      int num_stacks = as<int>( hp_used == 0 ? ab::base_costs[ RESOURCE_HOLY_POWER ] : hp_used );
       if ( p -> bugs && is_divine_storm && ( p -> buffs.empyrean_power_azerite -> up() || p -> buffs.empyrean_power -> up() ) )
-      {
-        num_stacks = 0;
-      }
+        ;
       else
-      {
         p -> buffs.crusade -> trigger( num_stacks );
```
Melekus: This should probably get reformulated.

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -113,6 +118,11 @@ struct avengers_shield_t : public avengers_shield_base_t
       p() -> cooldowns.avengers_shield -> reset( false );
       p() -> buffs.moment_of_glory -> decrement( 1 );
     }
+    if ( p() -> legendary.holy_avengers_engraved_sigil -> ok() && rng().roll( holy_avenger_proc_chance ) )
+    {
+      p() -> cooldowns.avengers_shield -> reset( false );
```
Melekus, line 122: The requires_reaction should be set to true here, since it's only a chance to proc.

OrcinusDrake: Yeah? Even though you have a whole gcd to react and know it's pretty likely?

Melekus: GCD duration is included in the reaction time. 
35% chance to proc is probably high enough to expect the occurence though, you're right.

engine/class_modules/paladin/sc_paladin.hpp
OrcinusDrake: Not sure how to make it clearer other than just adding comments. Or replacing `;` with`{ }`. Ret thing I don't really want to touch it

Melekus: You could just reverse the condition so instead of having if (...) {} else x, you'd have if ( ! ... ) { x }

OrcinusDrake: Annoying to read tho. But sure

PR #5413, opened 2020-10-12
[Priest] Disable Chorus Of Insanity

Saeldur:
Disable COI Trigger, for some reason the trigger is giving 100 stacks.
As this trait is currently irrelevant I'm just disabling the trigger outright.


Saeldur:
Closed because @seanpeters86 actually was bothered to fix the underlying bug despite it being prepatch

PR #5414, opened 2020-10-13
[Priest] put in bugs that dont appear in spelldata

seanpeters86:
These fixes are about adding things that are in PTR but not in beta to be removed once that data is reconciled.

### DIFF
--- a/engine/class_modules/priest/sc_priest.hpp
+++ b/engine/class_modules/priest/sc_priest.hpp
@@ -890,7 +890,15 @@ struct mindbender_pet_t final : public base_fiend_pet_t
   }
   double insanity_gain() const override
   {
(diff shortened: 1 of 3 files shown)


engine/class_modules/priest/sc_priest.hpp
```diff
@@ -890,7 +890,15 @@ struct mindbender_pet_t final : public base_fiend_pet_t
...
   double insanity_gain() const override
   {
-    return o().find_spell( 200010 )->effectN( 1 ).resource( RESOURCE_INSANITY );
+    // Currently not in beta, but in PTR data
+    if ( o().bugs )
+    {
+      return 5;
+    }
+    else
+    {
```
scamille, line 899: That looks like a find_spell call during combat. 
Same thing in shadowfiend pet.

PR #5419, opened 2020-10-14
Replace Dark thoughts cooldown hack by using charge system.

### LINKED ISSUE #137: Refactor Dark Thoughts to use the built in charges system
@scamille pointed out that the way we handle Dark Thoughts now has some issues and can potentially cause simc to crash given very specific scenarios. We should refactor this into the charges system.

scamille:
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/137


scamille:
@seanpeters86 I haven't really tested this in a way that dps matches some expected values. There is now a debug statement in the cooldown adjust code, which I tested with and gives the expected behavior. Assuming my understanding of how this should be implemented is correct.

seanpeters86:
@Saeldur found some interesting side effects with this PR IRT actual mind blast casts. This is getting....WAY more mind blasts off. 162 casts vs. 54 in the current implementation, which seems pretty off. Not sure if Sealdur had found a cause yet but you can see that with this profile:

```
priest="Base"
level=60
role=spell
position=back
talents=3121111
spec=shadow

default_actions=1

head=confidants_favored_cap,id=183021,ilevel=226,bonus_id=7103
neck=charm_of_eternal_winter,id=183040,ilevel=226
shoulder=shawl_of_the_penitent,id=183020,ilevel=233
back=crest_of_the_legionnaire_general,id=183032,ilevel=233
chest=robes_of_the_cursed_commando,id=182998,ilevel=233
wrists=acolytes_velvet_bindings,id=183017,ilevel=226
hands=impossibly_oversized_mitts,id=183022,ilevel=226
waist=shadewarped_sash,id=183004,ilevel=226
legs=leggings_of_lethal_reverberations,id=182981,ilevel=226
feet=slippers_of_the_forgotten_heretic,id=182979,ilevel=226
finger1=most_regal_signet_of_sire_denathrius,id=183036,ilevel=233
finger2=ritualists_treasured_ring,id=183037,ilevel=226
trinket1=glyph_of_assimilation,id=184021,ilevel=226
main_hand=sinbearers_absolution_staff,id=182418,ilevel=233

single_actor_batch=1
optimal_raid=0
desired_targets=1
iterations=10000
```

scamille:
@Saeldur @seanpeters86 please try again. I get 56 mind blast executions now, which seems more in line with what you expect.

seanpeters86 (approved):
LGTM

Saeldur:
I'm currently seeing this PR as a dps increase because it has an increase from 51.1 mindblast casts average to 60 mindblast casts average. I'm investigating if this PR or our original implementation was not working properly.

scamille:
Ok so there are still some outstanding issues here if the assert from sc_cooldown.cpp line ~416 is not removed.
Problem comes from - as far as I understand it - when mind blast start to cast but then on execute the cooldown is no longer ready / current_charge ist at 0.

I'll hope to get this working at some point, might take a few days. Thanks @vituscze for all the help so far.

scamille:
@vituscze I am still at a loss of resolving this without the removal of the assert, but I also think that the current solution does work fine as far as I can tell, since it will automatically handle the edge cases by just not decrementing past 0.

Would it be ok if we merge this, or do you think this really requires extra handling in action_t so that the assert in cooldown_t::start can be kept?

seanpeters86:
@Saeldur did you find any edge cases with this while testing?

vituscze:
@scamille Trying to decrement past 0 CD charges should definitely fail rather than just silently continue as if nothing happened. However, I'm not sure what's the issue with modifying update_ready in mind blast to not start the cooldown in that case? End result should be the same, but here you are being explicit about it rather than silently ignoring it for all actions.

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -205,8 +185,8 @@ struct mind_blast_t final : public priest_spell_t
 
     if ( priest().buffs.dark_thoughts->up() )
       priest().buffs.dark_thoughts->decrement();
-    else
-      priest_spell_t::update_ready( cd_duration );
+
+    priest_spell_t::update_ready( cd_duration );
```
vituscze: Is this right? If you decrement Dark Thoughts, that's gonna remove one charge and then update_ready will start the cooldown, removing another charge.

Saeldur:
Going to quickly confirm the merge happened properly since @seanpeters86 changed a bunch of the dark thoughts buff code.

seanpeters86 (approved):

seanpeters86:
LGTM pinging a few others if they want to take a look, @lgkern @scamille @vituscze

vituscze (approved):

vituscze (approved):

PR #5425, opened 2020-10-16
[Priest] adjust module from recent build

seanpeters86:
- rework painbreaker psalm for the "up to" insanity mechanic
- add bugged behavior for painbreaker only giving 5 per dot instead of the 7.5 in spell data
- remove old rabid shadows conduit effect
- fix rabid shadows conduit not working at all
- remove hard coded values now that they are in spell data
- remove target_ready for Shadow Crash as the min range is no longer a thing

### DIFF
--- a/engine/class_modules/priest/sc_priest.hpp
+++ b/engine/class_modules/priest/sc_priest.hpp
@@ -841,8 +841,6 @@ struct base_fiend_pet_t : public priest_pet_t
     resources.current = resources.max = resources.initial;
   }
 
-  double composite_melee_haste() const override;
-
   action_t* create_action( util::string_view name, const std::string& options_str ) override;
 };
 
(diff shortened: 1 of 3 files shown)


engine/class_modules/priest/sc_priest.hpp
```diff
@@ -955,7 +945,15 @@ struct fiend_melee_t : public priest_pet_melee_t
     if ( base_execute_time == timespan_t::zero() )
       return timespan_t::zero();
 
-    return base_execute_time * player->cache.spell_speed();
+    // Mindbender inherits haste from the player
+    timespan_t hasted_time = base_execute_time * player->cache.spell_speed();
+
+    if ( p().o().conduits.rabid_shadows->ok() )
```
scamille, line 951: If you want to be consistent you can just use .ok() here as well instead of doing -> indirection. But not really relevant.

scamille (approved):

PR #5426, opened 2020-10-16
[Priest] fix unfurling darkness calculation

seanpeters86:
VT spell data auto triggered the direct hit damage on every VT, and we just added more if using VT. This disables that completely and refactors UD to hit for the correct amount as a child action.

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -877,6 +877,7 @@ struct vampiric_touch_t final : public priest_spell_t
     if ( priest().talents.unfurling_darkness->ok() )
     {
       child_ud = new unfurling_darkness_t( priest() );
+      add_child( child_ud );
     }
   }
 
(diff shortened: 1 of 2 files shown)


engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -903,12 +906,11 @@ struct vampiric_touch_t final : public priest_spell_t
...
     }
 
-    // TODO: check if talbadars_stratagem can proc this
-    // Damnation does not proc Unfurling Darkness, but can generate it
-    if ( priest().buffs.unfurling_darkness->check() && casted )
+    if ( priest().buffs.unfurling_darkness->check() )
     {
       child_ud->target = s->target;
       child_ud->execute();
+      add_child( child_ud );
```
scamille: That is not the right place for add_child. Please move this up into vampiric_touch constructor, directly after you initialize child_ud.

scamille (changes requested):

scamille (approved):

lgkern (approved):

PR #5428, opened 2020-10-17
[Warlock] Soul Tithe Conduit

Azevara:


Azevara:
Leaving this as a PR for the moment while I try and verify if autoattacks are buffed for Demonology

engine/class_modules/warlock/sc_warlock.hpp
```diff
@@ -870,6 +871,9 @@ struct warlock_spell_t : public spell_t
   {
     double pm = spell_t::action_multiplier();
 
+    if ( p()->buffs.soul_tithe->check() && data().affected_by( p()->buffs.soul_tithe->data().effectN( 1 ) ) )
```
scamille: Calling affected_by during iteration is be a bit costly, as far as I can remember. You might want to add a flag for it and set it in the constructor.

Azevara: Mm this is a bad habit/practice that I seem to have inherited from this legacy code despite my best efforts. Fixed now for here, I've added a TODO in the module noting that this needs to be potentially cleaned up across the board.

PR #5429, opened 2020-10-17
[Priest] Shadow Crash does not have a fixed travel time

### LINKED ISSUE #123: Shadow Crash travel time

seanpeters86:
resolves https://github.com/WarcraftPriests/sl-shadow-priest/issues/123


scamille (commented):
Have you tried just removing the override, without manually setting travel_speed?
Check the html report if you get the correct value. I see no reason why this should not be picked up automatically.

seanpeters86:
@scamille ah yeah you are correct. fixing

lgkern (approved):

PR #5433, opened 2020-10-19
Buff initial stacks

gastank:
Implement parsing spell data for initial stack of buffs, and applying that number of stacks by default when the buff is called via trigger() or execute() with no explicit stack value.

For reverse buffs, behavior remains unchanged where a default value call will results in decrement(1).
Similarly, decrement() and increment() remain unchanged as expected behavior when those are called with default value is that the buff is adjusted by a single stack each time.

* buff_t::set_initial_stack( int ) and buff_t::modify_initial_stack( int ) can be used to explicitly set/modify the initial stacks, overriding what's read off the spell data. Can be chained at buff construction like the *_max_stack() counterparts.
* buff_t::initial_stack() will return the _initial_stack value.

### DIFF
--- a/engine/buff/sc_buff.hpp
+++ b/engine/buff/sc_buff.hpp
@@ -235,6 +235,7 @@ public:
   bool remains_gt( timespan_t time ) const;
   bool remains_lt( timespan_t time ) const;
   bool at_max_stacks( int mod = 0 ) const { return check() + mod >= max_stack(); }
+  int _resolve_stacks( int stacks ) { return stacks == -1 ? ( reverse ? _max_stack : _initial_stack ) : stacks; }
   bool trigger( action_t*, int stacks = -1, double value = DEFAULT_VALUE(), timespan_t duration = timespan_t::min() );
   bool trigger( timespan_t duration );
   bool trigger( int stacks, timespan_t duration );
--- a/engine/buff/sc_buff.hpp
+++ b/engine/buff/sc_buff.hpp
@@ -237,7 +237,7 @@ public:
   bool at_max_stacks( int mod = 0 ) const { return check() + mod >= max_stack(); }
   // For trigger()/execute(), default value of stacks is -1, since we want to allow for explicit calls of stacks=1 to
   // override using buff_t::_initial_stack
-  int _resolve_stacks( int stacks ) { return stacks == -1 ? ( reverse ? _max_stack : _initial_stack ) : stacks; }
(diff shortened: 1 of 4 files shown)


scamille (approved):

engine/buff/sc_buff.hpp
```diff
@@ -74,6 +74,7 @@ struct buff_t : private noncopyable
   const spell_data_t* trigger_data;
 
 public:
+  int _initial_stack;
```
nuoHep: Maybe move it adjacent to `_max_stack`?

scamille:
Can you maybe add some comments to the places where you introduced std::abs?

Without knowing the discussion in Discord, I would have a hard time knowing why you introduced those additions here.

I assume you added the abs adjustments to the trigger functions and not to the init function so that initial_stacks -1 is still a valid value in the buff, probably for reverse buffs?

gastank:
Yeah still working through. I don't think those abs() are even necessary. What do you mean by adjusting in the init function?

scamille:
What I meant is, why not ensure _initial_stacks is always > 0 already in the set_initial_stacks function instead of using abs functions in the trigger process. Unless of course negative values are perfectly valid, eg. for reverse buffs.

gastank:
That's already being done in`set_initial_stacks` to set `buff_t::_initial_stack` as 1 if the value in the spell data is negative.

The parts in trigger()/execute() is about handling the `stacks` parameter of those functions, which used to have default value of '1' but is being changed to '-1' to allow for calling `trigger(1)` to trigger a single stack instead of `_initial_stack`.

I am wondering however, since negative stacks in the spell data seems to indicate that the spell should have special/custom handling, it might be better to NOT guard against that in `set_initial_stack` and instead let it go through and assert out at execute(). So for those buffs, unless you:
1. explicitly set the initial stacks in buff construction, or
1. call trigger()/execute() with a stack value,

it will assert out to ensure that you do one of those things since the spell data indicates that the buff doesn't have a standard initial stack count.

scamille:
Sure why not. I assume the currently implemented buffs with -1 initial stack in spell data can be adjusted / already handle it, and you can add a little string literal inside the assert in trigger/execute explaining that this needs to be handled with an explicit stack value.

If it really only affects very few buffs and strongly correlates with "special handling required" and not just Blizzard having a mess in spell data, this sounds like a sensible idea.

PR #5438, opened 2020-10-20
[Fix] fury warrior st action list is broken

MhMadHamster:
Hey, i encountered a problem when importing character in the SimC and trying to simulate it, it failed because generated rotation is in an invalid format. When adding action to APL and passing `spell_data` and `action_name`, `action_name` is being passed further in raw format, In my case it was "Crushing Blow" and simc couldnt understand the spell cause of space inbetween the words. Example of code that uses add_action in such context: https://github.com/simulationcraft/simc/blob/shadowlands/engine/class_modules/sc_warrior.cpp#L6322

### DIFF
--- a/engine/class_modules/sc_warrior.cpp
+++ b/engine/class_modules/sc_warrior.cpp
@@ -6315,15 +6315,15 @@ void warrior_t::apl_fury()
   single_target->add_action( this, "Execute" );
   single_target->add_talent( this, "Bladestorm",  "if=prev_gcd.1.rampage" );
   single_target->add_action( this, "Bloodthirst", "if=buff.enrage.down|azerite.cold_steel_hot_blood.rank>1" );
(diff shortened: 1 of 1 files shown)


vituscze:
After taking a look at the functions in that file, I don't think the spell data and action name parameters are meant to be in any way related.

You could use this function to conditionally add an action to the APL based on a spec passive that doesn't need to share name with the action. You almost certainly don't want to ignore the `action_name` parameter.

There are two options: adding another overload that takes spell data but not action name (and constructs action name from the spell data) or changing the code in the warrior module. Either way, the `add_action` call in the warrior module is wrong.

MhMadHamster:
Thx for explanation, i fixed the problem in the warrior module. How come you cant just call the `add_action` for some spells, passing spell name, like for example fury warrior's Bloodbath/Crushing Blow or  DH's Death Sweep/Annihilation, aren't they supposed to be class/spec spells?

vituscze:
Not sure, it does seem like `add_action` with just the name should work there.

PR #5447, opened 2020-10-22
[Warlock] Implement Infernal Brand Conduit

Azevara:


Azevara:
It has been reported that the pets individually apply their own copies of the debuff instead of stacking it up. This may need some code tweaks in that situation so leaving as a PR for the moment.

PR #5450, opened 2020-10-23
Legacy pruning

gastank:
Remove inactive legendaries & set bonuses from past expansions


gastank:
Removed several old tier sets from warrior module, primarily protection but some that affect the dps specs. Shouldn't break anything in the warrior module, but tagging warrior review just to make sure.

PR #5464, opened 2020-10-26
[Warlock] Relic of Demonic Synergy

Azevara:


engine/class_modules/warlock/sc_warlock_pets.cpp
```diff
@@ -95,6 +114,24 @@ void warlock_pet_t::init_action_list()
...
+  syn_effect->spell_id = 337057;
+  syn_effect->execute_action = new demonic_synergy_pet_proc_t(this);
+  special_effects.push_back(syn_effect);
+
+  auto cb = new dbc_proc_callback_t(this, *syn_effect);
+
+  cb->initialize();
+
+  if ( !o()->legendary.relic_of_demonic_synergy->ok() || !is_main_pet )
+    cb->deactivate();
```
vituscze: If the effect isn't going to be used, it shouldn't even be created/initialized, same issue with the player one.

engine/class_modules/warlock/sc_warlock_pets.cpp
```diff
@@ -95,6 +114,24 @@ void warlock_pet_t::init_action_list()
...
 }
 
+void warlock_pet_t::init_special_effects()
+{
+  pet_t::init_special_effects();
+
+  auto const syn_effect = new special_effect_t(this);
+  syn_effect->name_str = "demonic_synergy_pet_effect";
+  syn_effect->spell_id = 337057;
+  syn_effect->execute_action = new demonic_synergy_pet_proc_t(this);
```
vituscze: You should probably use custom_buff here instead of an action whose only purpose is to trigger a buff.

engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -1189,6 +1192,19 @@ void warlock_t::init_special_effects()
       if ( new_ == 1 ) cb->activate();
       else cb->deactivate();
     } );
+
+  auto const syn_effect = new special_effect_t( this );
+  syn_effect->name_str = "demonic_synergy_effect";
+  syn_effect->spell_id = 337057;
+  syn_effect->execute_action = new warlock::actions::demonic_synergy_proc_t( this );
```
vituscze: Similar issue here, except that execute_action does more stuff. I'd suggest making a subclass of dbc_proc_callback_t and doing the logic in its execute method.

engine/class_modules/warlock/sc_warlock.hpp
```diff
@@ -974,6 +977,26 @@ struct grimoire_of_sacrifice_damage_t : public warlock_spell_t
...
+  demonic_synergy_proc_t( warlock_t* p )
+    : warlock_spell_t( "demonic_synergy_proc", p, p->find_spell( 337060 ) )
+  {
+    background = true;
+    proc = true;
+  }
+
+  void execute() override
+  {
+    if ( p()->warlock_pet_list.active )
```
vituscze: Leaving out the call to parent's execute is generally only safe with ACTION_OTHER (and similar) action types and even there, you need to be careful with it. I'd definitely advise against doing that in subclass of spell_t.

vituscze (changes requested):

vituscze (approved):

PR #5468, opened 2020-10-27
Add player_t::find_target_data

vituscze:
Since we're forced to use get_target_data in player_t::composite_player_target_multiplier, target data will be created for every target. This attempts to fix that issue by introducing find_target_data, which doesn't force the creation of the target data object.

### DIFF
--- a/engine/class_modules/sc_druid.cpp
+++ b/engine/class_modules/sc_druid.cpp
@@ -1002,6 +1002,7 @@ public:
   void recalculate_resource_max( resource_e, gain_t* g = nullptr ) override;
   void create_options() override;
   std::string create_profile( save_e type ) override;
+  const druid_td_t* find_target_data( const player_t* target ) const override;
   druid_td_t* get_target_data( player_t* target ) const override;
   void copy_from( player_t* ) override;
   void output_json_report( js::JsonOutput& /* root */ ) const override;
@@ -10604,6 +10605,14 @@ void druid_t::shapeshift( form_e f )
   }
 
   form = f;
+}
+
+// druid_t::find_target_data ================================================
+
+const druid_td_t* druid_t::find_target_data( const player_t* target ) const
+{
+  assert( target );
+  return target_data[ target ];
 }
 
 // druid_t::get_target_data =================================================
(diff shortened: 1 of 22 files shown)


vituscze:
Eventually, I'd like to remove the const from get_target_data so that we can avoid that ugly const_cast. But that requires far more changes (especially in the class modules), so I left it out of this PR.

scamille:
Last time I talked about this with @nuoHep his assessment was that performance wise this might not be as good as it sounds when used from outside of the class module with the final target data class, since the overhead of the virtual call is far worse than creating too many objects (which will only happen once).

In any case, it should not be worse and some class modules are already using this internally in some places.

engine/class_modules/sc_druid.cpp
```diff
@@ -10606,6 +10607,14 @@ void druid_t::shapeshift( form_e f )
   form = f;
 }
 
+// druid_t::find_target_data ================================================
+
+const druid_td_t* druid_t::find_target_data( const player_t* target ) const
+{
+  assert( target );
```
scamille, line 10614: Is there a specific reason that the druid override has an assert in it, and other do not?
I mean if we really care we should maybe at some point start annotation pointers we think should not be null. (gsl not_null maybe)

scamille:
About removing const from get_target_data: That is definitely a staged process.
- Add find_td to a classes main action-template as a companion to the get_td function we usually have there.
- Either manually check what can be changed to find_td there, or just remove const from the get_td function and adjust until it compiles.
- Repeat for next class.

Eg. https://github.com/simulationcraft/simc/blob/shadowlands/engine/class_modules/priest/sc_priest.hpp#L1169 is already not const, and since 99% of the usages of target data is from action_t, this means it is already in good shape even when the player_t::get_target_data eventually will be non-const.

engine/class_modules/sc_druid.cpp
vituscze: Wanted to keep it in line with the get_target_data stuff druids are doing. But no, it's not strictly necessary.

vituscze:
After the discussion on Discord, I'll leave get_target_data as is and leave find_target_data opt-in.

PR #5470, opened 2020-10-28
[Death Knight] Implement Gorefiends Domination Legendary

Armadk:
Tested in Beta.

Each cast of HS reduces CD of vamp blood by 2 seconds.

When DRW is up, reduces CD by 4 seconds.

Cleaving does not change the reduction, so it's per cast, not per target hit.

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -975,7 +975,7 @@ public:
     // Defensive/Utility
     // item_runeforge_t deaths_embrace; // 6947
     // item_runeforge_t grip_of_the_everlasting; // 6948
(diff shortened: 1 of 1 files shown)


Armadk:
Vampiric Blood looks like it may not be working correctly in trunk.  From debug output it looks to be casting much more often than it should be.  I'll look into what is not working with it and submiit a separate PR.

Armadk:
Ended up just fixing it in this PR as the fix was so small.  red thirst had the CD reduction changed from 1 to 1.5, and in the spelldata was changed from 1 to 150.

Also tested on beta, and the closed bug that was mentioned in the comment does not apply anymore.  CD reduction is 6 seconds when you have enough stacks for ossuary to kick in, and 6.75 seconds when stacks are low.

PR #5472, opened 2020-10-28
[Priest] rework Void Torrent

seanpeters86:
Void Torrent no longer refreshes SWP/VT and does not apply Devouring Plague. They also changed the duration and some SP modifiers.

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -1411,18 +1411,7 @@ struct void_torrent_t final : public priest_spell_t
   // DoT duration is fixed at 3s
   timespan_t composite_dot_duration( const action_state_t* ) const override
   {
-    return data().duration();
-  }
-
-  timespan_t tick_time( const action_state_t* ) const override
-  {
-    timespan_t t = base_tick_time;
-
-    double h = priest().composite_spell_haste();
-
-    t *= h;
-
-    return t;
+    return dot_duration;
   }
 
   void last_tick( dot_t* d ) override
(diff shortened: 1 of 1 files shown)


engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1408,13 +1406,12 @@ struct void_torrent_t final : public priest_spell_t
     energize_type     = action_energize::PER_TICK;
     energize_resource = RESOURCE_INSANITY;
     energize_amount   = insanity_gain;
-
-    child_dp->background = true;
   }
 
+  // DoT duration is fixed at 3s
   timespan_t composite_dot_duration( const action_state_t* ) const override
```
scamille, line 1412: Could this be achieved without the override, by maybe setting some hasted=false flags?

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1408,13 +1406,12 @@ struct void_torrent_t final : public priest_spell_t
...
   {
-    return timespan_t::from_seconds( 4.0 );
+    return data().duration();
```
Saeldur: This should just be `return dot_duration;`
Is there a reason why we override tick_time? The default implementation of tick_time appears to be identical except it gets the haste scaler from the action state instead of `priest().composite_spell_haste()`

engine/class_modules/priest/sc_priest_shadow.cpp
Saeldur: The spell is still hasted tickwise, just fixed duration channel.

seanpeters86: @scamille we need hasted ticks but fixed duration, dont think thats an option that i can see

engine/class_modules/priest/sc_priest_shadow.cpp
seanpeters86: good call, probably just leftover code

Saeldur (approved):

PR #5474, opened 2020-10-29
[DBC] chunk up item data

nuoHep:
Attempt to workaround compiler/linker limits (bugs?) when faced with
large arrays. We may want to increase the number of exported items and
it bumps into some problems.

Fortunately it's all pretty well encapsulated. We access the raw item
data array in essentially one place which is easily updated to do an
outer loop over "chunks". Searching by id simply requires a (linear)
prepass to find the relevant "chunk".

---

@navv1234 I don't have access to a windows box atm, so can't really test. If my theory is right re why msvc linker fails on this it should work with increasing the item count. I hope it does and we don't have to split items into even more tus.


navv1234 (commented):
Tested locally with removing ilevel restrictions on armor/weapons (thus increasing item database size by roughly 25%). Both Visual Studio 2017 and 2019 Win32 toolchains managed to compile and link the binary, where as doing it before caused either a compiler crash (internal compiler error), or compiler running out of heap space.

PR #5475, opened 2020-10-29
Adding a new action for Earth Ele EQ

HawkCorrigan:
@navv1234 @wfarr @Bloodmallet 

Sims were crashing when having actions for Earth Elemental and Earthquake because the Earthquake Damage was added as a child to both.

I just made a new Deeptremor Earthquake struct.


EvanMichaels:
I think for something like this I'd recommend trying to pull in the get_background_action template stuff that a few modules like Rogue/DH/Hunter/Druid are using nowadays.

Using the DH template as an example (but you can find similar things in a few modules):

```
  template <typename T, typename... Ts>
  T* get_background_action( util::string_view n, Ts&&... args )
  {
    auto it = range::find( background_actions, n, &action_t::name_str );
    if ( it != background_actions.cend() )
    {
      return dynamic_cast<T*>( *it );
    }

    auto action = new T( n, this, std::forward<Ts>( args )... );
    action->background = true;
    background_actions.push_back( action );
    return action;
  }
```

This would allow you to create a named, singleton background instance of a given ability class instead of duplicating. It stores the actions in a background_actions vector in the player class.

e.g.

```
rumble = p->get_background_action<earthquake_damage_t>( "earthquake_deeptremor" );
add_child( rumble );
```

This saves quite a lot of reimplementation as well as simplifying things a fair bit.

This does potentially require some minor refactoring of the ability constructors, since they have to be in a "standardized" form. Usually something like ability_blah_t( util::string_view name, class_type_t* p, ... )

HawkCorrigan:
Done.
Using strings instead of string_view cause the Shaman Base Spell still uses that for the token.

I also only changed the earthquake_damage_t constructor, cause it's the only one that is gonna use that for now.

navv1234:
Judging by the tooltip, the aura it has should follow the earth elemental. While for most purposes the implementation that is here will work (apart from having potential unintended consequences because it's inheriting from `shaman_spell_t`), it does place a "ground aoe" at the feet of the player's targeted enemy, instead of the earth elemental triggering a repeating damage thing. This means that if someone would conceive a situation where the enemy moves, the pets are presumed to follow their targets properly, however the ground aoe even will not, causing the earthquake aoe to no longer be in range.

Bloodmallet:
@navv1234 what would a fix to your mentioned issue look like?

navv1234:
I'd probably hack together some hidden buff on the pet that ticks the earthquake rumble while the pet is up.

PR #5479, opened 2020-10-31
[Warlock] Update Malefic Rapture

Azevara:


engine/class_modules/warlock/sc_warlock_affliction.cpp
```diff
@@ -592,14 +592,15 @@ struct malefic_rapture_t : public affliction_spell_t
 {
     struct malefic_rapture_damage_instance_t : public affliction_spell_t
     {
+      const spell_data_t parent_data;
```
vituscze: Don't store copies of spell data, use either a pointer or a reference.

engine/class_modules/warlock/sc_warlock_affliction.cpp
```diff
@@ -592,14 +592,15 @@ struct malefic_rapture_t : public affliction_spell_t
...
-          affliction_spell_t( "malefic_rapture_aoe", p, p->find_spell( 324536 ) )
+      malefic_rapture_damage_instance_t( warlock_t *p, const spell_data_t pd ) : 
+          affliction_spell_t( "malefic_rapture_aoe", p, p->find_spell( 324540 ) ),
+          parent_data( pd )
       {
         aoe = 1;
         background = true;
-        spell_power_mod.direct = data().effectN( 1 ).sp_coeff();
-        base_costs[ RESOURCE_SOUL_SHARD ] = 0;
+        spell_power_mod.direct = parent_data.effectN( 1 ).sp_coeff();
```
vituscze: Since you need the spell data just for spcoef, you could just have the constructor take a `double` param. No need to store anything then.

vituscze (changes requested):

PR #5483, opened 2020-11-01
[Priest] use correct spell references where possible

seanpeters86:
- always use `p` instead of `player`
- remove most duplicate `find_spell` entries
- fix error with incorrect Shadowy Apparition data ([causes a simc crash at low level](https://www.raidbots.com/simbot/report/jwS5FZg4GKYMmz6NjLsnmf/))

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -31,8 +31,8 @@ private:
   bool only_cwc;
 
 public:
-  mind_blast_t( priest_t& player, util::string_view options_str )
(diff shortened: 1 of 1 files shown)


engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -704,7 +704,7 @@ struct shadowy_apparition_damage_t final : public priest_spell_t
 
 struct shadowy_apparition_spell_t final : public priest_spell_t
 {
-  shadowy_apparition_spell_t( priest_t& p ) : priest_spell_t( "shadowy_apparitions", p, p.find_spell( 78203 ) )
```
seanpeters86, line 707: this was actually the _old_ SA so at low levels this would crash

Saeldur (approved):

PR #5484, opened 2020-11-02
Protection Paladin (Phase 3)

OrcinusDrake:
This isn't even the last PR :(


engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1023,9 +1061,31 @@ struct divine_toll_t : public paladin_spell_t
...
+struct hallowed_discernment_tick_t : public paladin_spell_t
+{
+  double aoe_multiplier;
+  // This should be using 340203 but I don't have its spell data.
+  hallowed_discernment_tick_t( paladin_t* p ) :
+    paladin_spell_t( "hallowed_discernment", p, p -> find_spell( 317221 ) )
+    {
+      base_multiplier *= p -> conduit.hallowed_discernment.percent();
+      background = true;
+      dual = true;
```
Melekus: Is this necessary? how does this effect show on logs?
If it's an additional AH tick it should probably call an AH tick, if not then its execution count shouldn't be hidden by dual=true

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -943,9 +983,7 @@ struct hand_of_reckoning_t: public paladin_melee_attack_t
 struct righteous_might_t : public heal_t
 {
   righteous_might_t( paladin_t* p ) :
-    // Should be 340193 but I don't have spell data for that.
-    // p -> conduit.righteous_might may just work when that's added.
-    heal_t( "righteous_might", p, p -> find_spell( 340192 ) )
+    heal_t( "righteous_might", p, p -> conduit.righteous_might )
```
Melekus: Shouldn't the ID be 340193 now that it's properly whitelisted and imported in spelldata?

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1035,31 +1095,50 @@ struct ashen_hallow_tick_t : public paladin_spell_t
...
-    {
-      return cam * (20.0 - state -> n_targets) / 15.0;
-    }
-    else if ( state -> n_targets <= 12 )
-    {
-      return cam * 2.0 / 3;
-    }
-    else // if ( state -> n_targets > 12 )
+    else
+      return cam * 7.8 * std::log( 0.112 * state -> n_targets + 1.346 ) / state -> n_targets;
```
Melekus: I'd personally rather have the previous version, even if it's not as exhaustive it is much easier to read.

Its place might also be in core, since it also affects other abilities across different classes (Keg Smash, Revenge, Thunder Clap, others?)

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1035,31 +1095,50 @@ struct ashen_hallow_tick_t : public paladin_spell_t
...
+        [] ( player_t * t ) { return t -> max_health() > 0; }
+      );
+      player_t* lowest_hp_target = *std::min_element(
+        targets.begin(), targets.end(), [ use_actual_hp ] ( player_t* lhs, player_t* rhs )
+        {
+          if ( use_actual_hp )
+            return lhs -> current_health() < rhs -> current_health();
+          return lhs -> health_percentage() < rhs -> health_percentage();
+        }
+      );
```
Melekus: I'm not actually sure what the purpose of all of this is. Could probably use some commenting.

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -1112,6 +1117,12 @@ struct holy_power_consumer_t : public Base
       return 0.0;
     }
 
+    if ( is_wog && ab::p() -> specialization() == PALADIN_PROTECTION )
+    {
+      if ( ab::p() -> buffs.shining_light_free -> check() || ab::p() -> buffs.royal_decree -> check() )
+        return 0.0;
+    }
+
```
Melekus: Does this have to be inside holy_power_consumer_t? It looks like it belongs inside word_of_glory_t

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -306,6 +306,9 @@ struct guardian_of_ancient_kings_t : public paladin_spell_t
     use_off_gcd = true;
     trigger_gcd = 0_ms;
     cooldown = p -> cooldowns.guardian_of_ancient_kings;
+
+    if ( p -> conduit.royal_decree -> ok() )
+      cooldown -> duration += timespan_t::from_millis( p -> conduit.royal_decree.value() );
```
Melekus: Can replace ``timespan_t::from_millis( p -> conduit.royal_decree.value() )`` with a simpler ``p -> conduit.royal_decree.time_value()``

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -838,7 +840,12 @@ void paladin_t::create_buffs_protection()
   buffs.ardent_defender = make_buff( this, "ardent_defender", find_specialization_spell( "Ardent Defender" ) )
         -> set_cooldown( 0_ms ); // handled by the ability
   buffs.guardian_of_ancient_kings = make_buff( this, "guardian_of_ancient_kings", find_specialization_spell( "Guardian of Ancient Kings" ) )
-        -> set_cooldown( 0_ms );
+        -> set_cooldown( 0_ms )
+        -> set_stack_change_callback( [ this ] ( buff_t*, int old, int curr )
```
Melekus, line 855: Is there a way to proc GoaK without casting it? royal decree trigger could just be added to guardian_of_ancient_kings_t::execute() if that's not the case (or if it's the case and procced goak don't trigger it)

Melekus (changes requested):
Looks good apart from the changes I noted, I'll try to think about something for the state of holy_power_consumer_t::execute() that is getting more and more overloaded as more mechanics have to be handled simultaneously. Hopefully the removal of azerite should make it lighter.

engine/class_modules/paladin/sc_paladin.hpp
OrcinusDrake: But mum, ret put all their divine storm stuff there, why don't I get to?

engine/class_modules/paladin/sc_paladin.cpp
OrcinusDrake: Oh you're right, I just checked whether it was working not whether it used the right spell id.

engine/class_modules/paladin/sc_paladin_protection.cpp
OrcinusDrake: "Reign of Endless Kings" Legendary procs goak for 4sec if you drop below 20% health, this procs royal decree.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -596,8 +597,19 @@ struct devotion_aura_t : public paladin_spell_t
   void execute() override
   {
     paladin_spell_t::execute();
-
-    p() -> buffs.devotion_aura -> trigger();
+    // If Devotion Aura is up, cancel it. Otherwise replace the current aura.
```
OrcinusDrake: This is just a draft.

engine/class_modules/paladin/sc_paladin.hpp
Melekus: I think that should be used as a last resort: for example if it has to be entertwined in the middle of various consumer-wide mechanics and may affect their behaviour.
For simple stuff like setting cost to 0 with specific buffs, it'd be much clearer if it was only done on WoG's side

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2508,6 +2509,18 @@ double paladin_t::resource_loss( resource_e resource_type, double amount, gain_t
...
+      resource_type == RESOURCE_HEALTH &&
+      legendary.reign_of_endless_kings -> ok() &&
+      ! buffs.reign_of_ancient_kings -> up() &&
+      health_percentage()/100 < legendary.reign_of_endless_kings -> effectN( 2 ).percent() &&
+      health_percentage()/100 + amount / max_health() >= legendary.reign_of_endless_kings -> effectN( 2 ).percent()
+    )
+  {
+    buffs.guardian_of_ancient_kings -> trigger( legendary.reign_of_endless_kings -> effectN( 2 ).trigger() -> duration() );
+    buffs.reign_of_ancient_kings -> trigger();
+  }
```
Melekus: This seems a bit circumvoluted... 

So the spelldata seems to indicate that the player gains the goak buff (ID 337851, with damage reduction data in the original goak buff ID 86657). Which one is on the player ingame?
You're creating a new buff called "reign of ancient kings" solely to handle the visible icd on the player, it might be better for simplicity and report clarity to make it the actual damage reduction buff (especially if it's the one used ingame) and just give it a cooldown duration equal to the one that is stored in 337852.

Last, did you make sure that the proc doesn't happen if the player is already below the threshold when taking damage (assuming the internal cooldown is up).

OrcinusDrake: 337851 is a 4sec buff that shows up in the log but not on the player's buffs, so I don't know what it's even doing. 337852 is the debuff that prevents it proccing for 45sec. 86659 is the actual goak buff itself you see on your character.
In-game and in the sim it won't proc if you're already below the threshhold when the debuff expires until you heal back over it.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1078,6 +1185,7 @@ struct ashen_hallow_t : public paladin_spell_t
     may_miss = false;
 
     add_child( damage_tick );
+    add_child( new hallowed_discernment_tick_t( p ) );
```
Melekus: This hallowed discernment tick isn't used except for setting the child action. You could send it to ashen hallow's constructor so you don't have to create it another time there.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1622,19 +1732,20 @@ void paladin_t::create_buffs()
...
         -> set_cooldown( 0_ms ); // handled by the ability
-  buffs.blessing_of_dawn = make_buff( this, "blessing_of_dawn", legendary.from_dusk_till_dawn -> effectN( 1 ).trigger() );
-  buffs.blessing_of_dusk = make_buff( this, "blessing_of_dusk", legendary.from_dusk_till_dawn -> effectN( 2 ).trigger() )
-        -> set_default_value( legendary.from_dusk_till_dawn -> effectN( 2 ).trigger() -> effectN( 1 ).percent() );
+  buffs.devotion_aura = make_buff( this, "devotion_aura", find_class_spell( "Devotion Aura" ) );
+
+  // Legendaries
+  buffs.blessing_of_dawn = make_buff( this, "blessing_of_dawn", legendary.of_dusk_and_dawn -> effectN( 1 ).trigger() );
+  buffs.blessing_of_dusk = make_buff( this, "blessing_of_dusk", legendary.of_dusk_and_dawn -> effectN( 2 ).trigger() )
+        -> set_default_value( legendary.of_dusk_and_dawn -> effectN( 2 ).trigger() -> effectN( 1 ).percent() );
```
Melekus: You can use set_default_value_from_effect() here

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2418,12 +2532,31 @@ double paladin_t::resource_loss( resource_e resource_type, double amount, gain_t
...
+      resources.current[ RESOURCE_HOLY_POWER ] == legendary.of_dusk_and_dawn -> effectN( 2 ).base_value()
     )
   {
     buffs.blessing_of_dusk -> trigger();
   }
+
+  if (
+      resource_type == RESOURCE_HEALTH &&
+      legendary.reign_of_endless_kings -> ok() &&
+      ! buffs.reign_of_ancient_kings -> up() &&
```
Melekus, line 2872: I'd rename this buff to make it clearer that it models the internal cooldown

Melekus: You could also just use a cooldown_t object instead, since you don't need the buff but only its duration

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2418,12 +2532,31 @@ double paladin_t::resource_loss( resource_e resource_type, double amount, gain_t
...
+      health_percentage()/100 < legendary.reign_of_endless_kings -> effectN( 2 ).percent() &&
+      // Won't trigger if you're below the threshhold when the debuff expires. You have to be healed back over it for it to be able to proc again.
+      health_percentage()/100 + amount / max_health() >= legendary.reign_of_endless_kings -> effectN( 2 ).percent()
```
Melekus: You could also store health_percentage() before resource loss somewhere

engine/class_modules/paladin/sc_paladin.cpp
OrcinusDrake: How about that. I'm surprised it knew it was a percentage.

engine/class_modules/paladin/sc_paladin.cpp
OrcinusDrake: But that's what it's called ingame D:

OrcinusDrake: I don't see how obfuscating the icd to just be a hidden trigger is cleaner than using the in-game functionality.

engine/class_modules/paladin/sc_paladin.cpp
Melekus: It defaults to the most reasonable multiplier based on the effect's type. You can find the logic behind it here: https://github.com/simulationcraft/simc/blob/shadowlands/engine/dbc/spell_data.cpp#L264
You can also specify the multiplier yourself if the default value isn't right.

engine/class_modules/paladin/sc_paladin.cpp
Melekus: Mostly because I think that it'd be confusing for the user. Seeing a buff simply called "Reign of Endless Kings" I think isn't clear enough that it represents the ICD (hence my first suggestion to rename it). 
My second suggestion is because I don't think having a visible buff for it is necessary and a simple cooldown is lighter to implement than a full buff.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -561,6 +584,48 @@ struct blinding_light_t : public paladin_spell_t
...
+    paladin_spell_t( n, p, s )
+  {
+    harmful = false;
+    aura_buff = nullptr;
+  }
+
+  void execute() override
+  {
+    paladin_spell_t::execute();
+    assert( aura_buff != nullptr && "Paladin auras must have aura_buff set to their appropriate buff" );
```
Melekus: You could move this assert to paladin_spell_t::init_finished() instead of running it every execute()

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1658,19 +1820,20 @@ void paladin_t::create_buffs()
...
        -> add_invalidate( CACHE_MASTERY )
        -> add_invalidate( CACHE_VERSATILITY )
        -> set_cooldown( 0_ms ); // let the ability handle the cooldown
-
   buffs.holy_avenger = make_buff( this, "holy_avenger", talents.holy_avenger )
         -> set_cooldown( 0_ms ); // handled by the ability
-  buffs.blessing_of_dawn = make_buff( this, "blessing_of_dawn", legendary.from_dusk_till_dawn -> effectN( 1 ).trigger() );
-  buffs.blessing_of_dusk = make_buff( this, "blessing_of_dusk", legendary.from_dusk_till_dawn -> effectN( 2 ).trigger() )
-        -> set_default_value( legendary.from_dusk_till_dawn -> effectN( 2 ).trigger() -> effectN( 1 ).percent() );
+  buffs.devotion_aura = make_buff( this, "devotion_aura", find_class_spell( "Devotion Aura" ) );
```
Melekus: You could set the default value here and use it instead of checking the buff, then using its damage reduction value from looking at the data.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2451,16 +2617,36 @@ double paladin_t::resource_gain( resource_e resource_type, double amount, gain_t
...
     buffs.blessing_of_dusk -> trigger();
   }
+
+  if (
+      resource_type == RESOURCE_HEALTH &&
+      legendary.reign_of_endless_kings -> ok() &&
+      ! buffs.reign_of_ancient_kings -> up() &&
+      health_percentage()/100 < legendary.reign_of_endless_kings -> effectN( 2 ).percent() &&
+      // Won't trigger if you're below the threshhold when the debuff expires. You have to be healed back over it for it to be able to proc again.
+      initial_hp/100 >= legendary.reign_of_endless_kings -> effectN( 2 ).percent()
```
Melekus: You could compare health_percentage() and initial_hp to the data's base_value() rather than dividing both by 100 and using percent()

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -655,6 +665,9 @@ void paladin_t::target_mitigation( school_e school,
     s -> result_amount *= 1.0 + buffs.blessing_of_dusk -> value();
   }
 
+  if ( buffs.devotion_aura -> up() )
+    s -> result_amount *= 1.0 + buffs.devotion_aura -> data().effectN( 1 ).percent();
```
Melekus: If you set devo aura's default value, you can just replace this with 
```suggestion
  s -> result_amount *= 1.0 + buffs.devotion_aura -> value();
```

Melekus (approved):

PR #5498, opened 2020-11-07
[Death Knight] Night Fae Covenant

Armadk:
- Implement Death's Due
- Implement Withering Ground conduit

As an implementation I went with merging the effects of death's due in the existing defile and death and decay functions, rather than copying both of them, and making them specific to DD.  I think there is still room for improvement here.  Open to any and all suggestions.

#### General
- It looks like over the patches, it was at some point set to replace death and decay, similar to what is done with defile.
- debuff that made mobs do less damage to you I did not implement.  In a quick glance, I couldn't find anything to make it so that the mob with the debuff would only do reduced damage to the DK that used DD.
#### Unholy
- withering ground does not increase damage if you take the defile talent
- Spell has a coefficient for damage, that matches Death and Decay, but nothing for defile
- In game, if you have defile, you do not get the death's due icon,  but you do get the graphic on the ground, and buff/debuff effects.
- In game interaction is odd, I think initially it didn't work with defile and it was later hacked in
#### Frost
- cleave gives resources on main hit only
- dw gives 4 stacks of buff per attack. 2 stacks from target 1, 2 stacks from target 2
- 2h gives 2 stacks of dd buff per attack. 1 stack from each target
- frost cleave seems like it was hacked in, nothing in spelldata for it
- added to the aoe apl
#### Blood
- DRW does not give an extra stack of the buff, only your own attacks

@nyterage I added death and decay into the frost aoe apl.  We may need to re-examine unholy apl in particular,  as with defile there was certain times when it was letting the buff drop as you varied the number of targets.


engine/class_modules/sc_death_knight.cpp
```diff
@@ -4552,7 +4573,7 @@ struct death_and_decay_t : public death_and_decay_base_t
 struct defile_t : public death_and_decay_base_t
 {
   defile_t( death_knight_t* p, const std::string& options_str ) :
-    death_and_decay_base_t( p, "defile", p -> talent.defile )
+    death_and_decay_base_t( p, p -> covenant.deaths_due -> ok() ? "deaths_due" : "defile", p -> talent.defile )
```
Melekus: Is the spell named Defile or Death's Due with the covenant active?

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4524,7 +4545,7 @@ struct death_and_decay_base_t : public death_knight_spell_t
 struct death_and_decay_t : public death_and_decay_base_t
 {
   death_and_decay_t( death_knight_t* p, const std::string& options_str ) :
-    death_and_decay_base_t( p, "death_and_decay", p -> spec.death_and_decay )
+    death_and_decay_base_t( p, p -> covenant.deaths_due -> ok() ? "deaths_due" : "death_and_decay", p -> spec.death_and_decay )
```
Melekus: Is the ability death and decay (from spec.death_and_decay) or death's due (covenant.deaths_due) with the covenant active?

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5924,6 +5946,17 @@ struct obliterate_strike_t : public death_knight_melee_attack_t
...
 
+  int n_targets() const override
+  {
+    // Unable to find the # of targets in spell data for death's due, but it does cause each obliterate hit to cleave. 2020-Nov-6
+    int num_targets = 0;
+    if ( p() -> covenant.deaths_due -> ok() )
+    {
+      num_targets = p() -> in_death_and_decay() ? 2 : 0;
+    }
+    return num_targets;
```
Melekus: Death's Due's tooltip references 315442 effect#2 for obliterate's target count increase. 
Moreover, I also noticed that the obliterate strike spells (222024 1H, 325461 2H, and 66198 OH) all have a chain_multiplier and chain_targets value of 1 in spelldata, so you should probably use that.

Melekus: Actually, it seems that Death and Decay contains data for Obliterate in its effect#4 
```#4 (id=871411)   : Apply Aura (6) | Add Flat Modifier (107): Spell Targets (17)
                   Base Value: 0 | Scaled Value: 0 | PvP Coefficient: 1.00000 | Misc Value: 17 | Target: Self (1)
                   Affected Spells: Obliterate (49020), Obliterate Off-Hand (66198), Obliterate (222024), Obliterate (325461)
                   Family Flags: 49
```
This value of 0 is increased to 1 by 315442, making the chain_targets value 2

engine/class_modules/sc_death_knight.cpp
Armadk: it's named defile with the covenant active.  Casting drops death's due graphic, with the death's due buffs.  It shows up in details as defile as well.  if I remember right, when they first put the spell in, it has no interaction with defile.

engine/class_modules/sc_death_knight.cpp
Armadk: It's an override.  Looks similar to how Val'kyr spell effect is handled.  It does have it's own spell, with it's own spell data, but everything matches, and even stuff within death's due is applied to death and decay.  Like the obliterate cleave as mentioned below.

I honestly wasn't sure what was going to be the cleanest implementation for these.  death's due could have been implemented as it's own, basically as a copy of death and decay, but we would still have to apply the manual overrides to defile.  This whole thing looks kinda like a mess, so I wouldn't be shocked to see it get an overhaul at some point.  Or it may stay this way all expansion since it will be tossed out in the next one.

Name             : Death's Due (desc=Night Fae) (id=315442) [Spell Family (15), Passive]
Class            : Death Knight
School           : Nature
Spell Type       : None
Attributes       : ......x. ........ ........ ........   ........ ........ ........ ........
                 : ........ ........ ........ ........   ........ ........ ........ ........
                 : ........ .......x ........ ........   ........ ........ ........ ........
                 : ........ ........ ........ ........   ........ ........ ........ ........
                 : ......x. ....x... ........ ........   ........ ........ ........ ........
                 : ........ ........ ........ ........   ........ ........ ........ ........
                 : ........ ........ ........ ........   ........ ........ ........ ........
                 : ........ ........ ........ ........
                 : Passive (6)
Effects          :
#1 (id=799201)   : Apply Aura (6) | Override Action Spell (Misc /w Base) (332)
                   Base Value: 324128 | Scaled Value: 324128 | PvP Coefficient: 1.00000 | Misc Value: 43265 | Misc Value 2: 0x3 | Target: Self (1)
#2 (id=871412)   : Apply Aura (6) | Add Flat Modifier (107): Spell Effect 4 (32)
                   Base Value: 1 | Scaled Value: 1 | PvP Coefficient: 1.00000 | Misc Value: 32 | Target: Self (1)
                   Affected Spells: Death and Decay (188290)
                   Family Flags: 118
Description      : Learn Death's Due, replacing Death and Decay.

Tooltip          : Drain power from each enemy within the area, reducing their damage done by 10% and increasing your Strength by 80 for 10 sec.

engine/class_modules/sc_death_knight.cpp
Armadk: Makes sense how it's all being done on blizz's end.  Is there a better way that we can accomplish this in simc than just adding 315442#2 to 325461#1.chain_target()?  It ends up skipping the interm 49020#4 (obliterate).

engine/class_modules/sc_death_knight.cpp
Melekus: Then i think the name here should be Defile as well, to better reflect what the users would see on logs/details.

engine/class_modules/sc_death_knight.cpp
Melekus: Well the spelldata driven way would be to return the damaging effect's chain_target() + the value from 49020#4 if inside dnd + the value from 315442#2 if inside dnd and death's due is talented, while making sure that if the result is 1, it returns 0 (to save computation time, "true" ST abilities have aoe=0 that let them skip target lists, target specific state, etc. ).

That said, your inital ``num_targets = p() -> in_death_and_decay() ? 2 : 0;`` works very well too and I doubt that it's something Blizzard will change in the near future. If you're afraid of that you could also compute the total value during construction and return either that or 0 on runtime based on death and decay/covenant.

engine/class_modules/sc_death_knight.cpp
Melekus: I was just wondering if it's a good idea to keep using spec.death_and_decay for the spell's data instead of covenant.deaths_due, since it is, I assume, the actual ability used by the player.

engine/class_modules/sc_death_knight.cpp
Armadk: I can go either way on this, but originally I had gone in this direction for the naming due to how it looks in the html report.

image


Text report however only reports on the one name:  deaths_due
image

I felt the html report was more clear for defile this way, and left no questions about it,  at the expense of the text report.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4379,6 +4372,51 @@ struct defile_damage_t : public death_and_decay_damage_base_t
   }
 };
 
+struct deaths_due_damage_t : public death_and_decay_damage_base_t
```
Melekus: If it works with pestilence just like regular death and decay, you could make it a ``death_and_decay_damage_t`` instead of a ``death_and_decay_damage_base_t``

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4590,6 +4628,34 @@ struct defile_t : public death_and_decay_base_t
...
+  }
+
+  void execute() override
+  {
+    debug_cast<deaths_due_damage_t*>( damage ) -> pestilence_procs_per_cast = 0;
+
+    death_and_decay_base_t::execute();
+  }
+
+  bool ready() override
```
Melekus, line 4681: You should make sure that this only returns true with the covenant selected (spelldata validation should do it automatically) and that death_and_decay_t::ready() returns false for night fae.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5935,6 +6006,10 @@ struct obliterate_strike_t : public death_knight_melee_attack_t
     may_miss = false;
     weapon = w;
 
+    deaths_due_cleave_targets = as<int>(p -> spell.deaths_due -> effectN( 2 ).base_value()) + 
+                                  s_data -> effectN ( 1 ).chain_target() +
```
```suggestion
                                  data().effectN ( 1 ).chain_target() + 
```

The module usually uses data() instead of directly calling s_data.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5935,6 +6006,10 @@ struct obliterate_strike_t : public death_knight_melee_attack_t
     may_miss = false;
     weapon = w;
 
+    deaths_due_cleave_targets = as<int>(p -> spell.deaths_due -> effectN( 2 ).base_value()) + 
+                                  s_data -> effectN ( 1 ).chain_target() + 
+                                  as<int>(p -> find_spell( 49020 ) -> effectN ( 4 ).base_value());
```
Melekus: 49020 is spec.obliterate, the data you're looking for is in 188290#4 (or spell.dnd_buff -> effectN( 4 ))

engine/class_modules/sc_death_knight.cpp
```diff
@@ -9064,7 +9151,7 @@ void death_knight_t::default_apl_frost()
   aoe -> add_talent( this, "Glacial Advance", "if=talent.frostscythe.enabled" );
   aoe -> add_action( this, "Frost Strike", "target_if=max:(debuff.razorice.stack+1)%(debuff.razorice.remains+1)*death_knight.runeforge.razorice,if=cooldown.remorseless_winter.remains<=2*gcd&talent.gathering_storm.enabled" );
   aoe -> add_action( this, "Howling Blast", "if=buff.rime.up" );
-  aoe -> add_action( this, "Death and Decay", "if=covenant.night_fae&spell_targets.death_and_decay>=2");
+  aoe -> add_action( "any_dnd,if=covenant.night_fae&spell_targets.death_and_decay>=2");
```
Melekus: In my opinion, calling death's due specifically and removing the night fae covenant condition would be just as good here.

engine/class_modules/sc_death_knight.cpp
Armadk: I'm not really sure how I managed to do that one... but this is why we do review ;)

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4635,16 +4639,6 @@ struct deaths_due_t : public death_and_decay_base_t
...
-  bool ready() override
-  {
-    if ( p() -> talent.defile -> ok() )
-    {
-      return false;
-    }
-
-    return death_and_decay_base_t::ready();
-  }
 };
```
Melekus, line 4689: What's the plan with defile talented? This implementation would make both cast of Defile and death's due available (athough any_dnd handles it, that will not work for specific deaths_due or defile action names).

engine/class_modules/sc_death_knight.cpp
```diff
@@ -9066,9 +9060,11 @@ void death_knight_t::default_apl_blood()
   standard -> add_talent( this, "Bonestorm", "if=runic_power>=100&!buff.dancing_rune_weapon.up" );
   standard -> add_action( this, "Death Strike", "if=runic_power.deficit<=(15+buff.dancing_rune_weapon.up*5+spell_targets.heart_strike*talent.heartbreaker.enabled*2)|target.1.time_to_die<10" );
   standard -> add_action( this, "Death and Decay", "if=spell_targets.death_and_decay>=3" );
+  standard -> add_action( "deaths_due,if=spell_targets.death_and_decay>=3" );
   standard -> add_action( this, "Heart Strike", "if=buff.dancing_rune_weapon.up|rune.time_to_4<gcd" );
   standard -> add_action( this, "Blood Boil", "if=buff.dancing_rune_weapon.up" );
   standard -> add_action( this, "Death and Decay", "if=buff.crimson_scourge.up|talent.rapid_decomposition.enabled|spell_targets.death_and_decay>=2" );
+  standard -> add_action( "deaths_due,if=buff.crimson_scourge.up|talent.rapid_decomposition.enabled|spell_targets.death_and_decay>=2" );
```
Melekus: You could merge the old death and decay actions and the new death's due actions if you just plan to copy the conditions.

Armadk: I'll use any_dnd on these ones for now, but we may want to split them in the future when we have time to check into apl usage of them more.  It will come down to if there is more value in maintaining 100% uptime of the buff or not.  

Simply removing the target check for death's due was a slight (10) dps loss.
```
5 Target Patch
deaths_due                   : start=  5.3 refresh=291.9 interval= 58.4 trigger=  1.0 duration= 51.8 uptime= 93.35%
1 Target Patch - original apl
deaths_due                   : start=  8.1 refresh= 40.6 interval= 37.2 trigger=  5.8 duration= 27.2 uptime= 73.59%
1 Target Patch - no num_targets check
deaths_due                   : start=  5.9 refresh= 56.5 interval= 53.1 trigger=  4.8 duration= 47.2 uptime= 93.00%
```

engine/class_modules/sc_death_knight.cpp
Melekus: Yeah the HTML report will fetch tooltip for the spell ID and show the name associated with it regardless of the actual name in the c++ code. Whether the spell and its damage component are called defile, death's due, or something else is up to you but I think calling it Defile would at least fit what it's doing ingame (just a Defile spell and damage, with death's due mechanics band-aided on it if I got it right).

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5938,6 +6039,16 @@ struct obliterate_strike_t : public death_knight_melee_attack_t
     }
   }
 
+  int n_targets() const override
+  {
+    int num_targets = 0;
+    if ( p() -> covenant.deaths_due -> ok() )
+    {
+      num_targets = p() -> in_death_and_decay() ? deaths_due_cleave_targets : 0;
```
Melekus, line 6047: if num_targets inits at 0, you could just replace this with 
```
      if ( p() -> covenant.deaths_due -> ok() && p() -> in_death_and_decay() )
        return deaths_due_cleave_targets;
      return death_knight_melee_attack_t::n_targets();
```

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5938,6 +6039,16 @@ struct obliterate_strike_t : public death_knight_melee_attack_t
...
+    }
+    return num_targets;
+  }
```
Melekus: You could just replace this with (or even with 0 instead of calling death_knight_melee_attack_t::n_targets())
```suggestion
  int n_targets() const override
  {
      if ( p() -> covenant.deaths_due -> ok() && p() -> in_death_and_decay() )
        return deaths_due_cleave_targets;
      return death_knight_melee_attack_t::n_targets();
  }
```

Melekus:
You should add a note somewhere that the damage reduction debuff is currently not yet implemented.

engine/class_modules/sc_death_knight.cpp
Armadk: Yes, defile spell with death's due mechanics band-aided on.  Updated to report it as defile.

engine/class_modules/sc_death_knight.cpp
Armadk: Just re-tested to make sure, and yes, pestilence works with it.  Subclassing from death_and_decay_damage_t instead as per suggestion.

Armadk:
```
Pre refactor
   deaths_due                  Count=   6.6| 34.786sec  DPE=3240| 2.15%  DPET=2399  DPR=   0  pDPS= 72  Miss= 0.00%  Hit=    58|    55|    61  Crit=   116|   109|   122|11.62%
    deaths_due                : start=  6.6 refresh= 82.2 interval= 35.1 trigger=  2.3 duration= 17.7 uptime= 39.31%

Post refactor
   deaths_due                  Count=   6.5| 34.915sec  DPE=3246| 2.15%  DPET=2404  DPR=   0  pDPS= 72  Miss= 0.00%  Hit=    58|    55|    61  Crit=   116|   109|   122|11.68%
    deaths_due                : start=  6.5 refresh= 82.4 interval= 35.1 trigger=  2.3 duration= 17.7 uptime= 39.24%
```

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4390,49 +4397,16 @@ struct defile_damage_t : public death_and_decay_damage_base_t
...
   { }
-
-  double composite_da_multiplier( const action_state_t* state ) const override
-  {
-    double m = death_knight_spell_t::composite_da_multiplier( state );
-
-    m *= 1.0 + p() -> conduits.withering_ground.percent();
-
-    return m;
-  }
```
Melekus: !

Melekus: (You can also remove the pestilence related variables, it should all be handled by ``death_and_decay_damage_t`` now)

PR #5505, opened 2020-11-09
[Death Knight] Implement Superstrain

Khazakdk:
Also removes the extra base AP on shambler that was recently fixed.  

Confirmed the following between Ceriloch and myself:
FF and VP are applied at 37.5% effectiveness, BP at 75%.
RP gain from frost fever and VP eruption on death seem unaffected
Death's Caress and Unholy Blight apply all dots, DRW does not

This isn't the sexiest implementation, I don't really like how copy/paste it feels but its straightforward and the xpac is out in 2 weeks so in the spirit of "the perfect is the enemy of the good" I can live with this.  

The one thing where I would extra appreciate any alternate ideas is where I had to switch the spell data passed into VP from the spell associated with effect #2 on Outbreak.  I had to do this because while VP is tagged for all 3 specs, Outbreak is tagged for Unholy only and so when it tried to apply VP as Frost/Blood it couldn't because the spell data for Outbreak returned null and thus it couldn't find VP.

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -6736,6 +6736,9 @@ struct unholy_blight_dot_t : public death_knight_spell_t
 
     td( state->target ) -> debuff.unholy_blight -> trigger();
 
+    p() -> active_spells.virulent_plague -> set_target( state -> target );
+    p() -> active_spells.virulent_plague -> execute();
+
     if ( p() -> legendary.superstrain -> ok() )
     {
       p() -> active_spells.frost_fever -> set_target( state -> target );
@@ -6749,18 +6752,15 @@ struct unholy_blight_dot_t : public death_knight_spell_t
 struct unholy_blight_buff_t : public buff_t
 {
   unholy_blight_dot_t* dot;
-  virulent_plague_t* vp;
 
   unholy_blight_buff_t( death_knight_t* p ) :
     buff_t( p, "unholy_blight", p -> talent.unholy_blight ),
-    dot( new unholy_blight_dot_t( p ) ),
-    vp( new virulent_plague_t( p ) )
+    dot( new unholy_blight_dot_t( p ) )
   {
     cooldown -> duration = 0_ms;
     set_tick_callback( [ this ]( buff_t* /* buff */, int /* total_ticks */, timespan_t /* tick_time */ )
     {
       dot -> execute();
-      vp -> execute();
     } );
   }
 };
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -8035,6 +8035,7 @@ void death_knight_t::create_actions()
 
     if ( spec.outbreak -> ok() )
     {
+      active_spells.virulent_plague = new virulent_plague_t(this);
       active_spells.virulent_eruption = new virulent_eruption_t( this );
     }
   }
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -7590,11 +7590,6 @@ void death_knight_t::trigger_virulent_plague_death( player_t* target )
     return;
   }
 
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_death_knight.cpp
```diff
@@ -6694,6 +6735,14 @@ struct unholy_blight_dot_t : public death_knight_spell_t
     death_knight_spell_t::impact( state );
 
     td( state->target ) -> debuff.unholy_blight -> trigger();
+
+    if ( p() -> legendary.superstrain -> ok() )
```
Khazakdk, line 6742: I attached the application of the off specs dots here as the tick callback doesn't have state as a parameter and I wanted to make sure the target was set explicitly since the offspec dots aren't child actions of UB like VP is.  Maybe they should be?

Armadk:
- Damage looks good, profiles match my in game testing
- frost fever RP gain looks good
- I think this implementation is still missing the virulent eruption implementation.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -3604,6 +3590,14 @@ struct blood_plague_t : public death_knight_spell_t
     may_miss = may_crit = hasted_ticks = false;
 
     base_tick_time *= 1.0 + p -> talent.rapid_decomposition -> effectN( 1 ).percent();
+
+    if ( p -> legendary.superstrain -> ok() && (p -> specialization() == DEATH_KNIGHT_FROST || p -> specialization() == DEATH_KNIGHT_UNHOLY) )
```
Melekus: Can probably check if the spec isn't Blood, rather than checking if it's either of the other two.

Melekus:
```suggestion
    if ( p -> legendary.superstrain -> ok() && p -> specialization() != DEATH_KNIGHT_BLOOD )
```

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5645,6 +5655,13 @@ struct frost_fever_t : public death_knight_spell_t
       // There's a 0.98 modifier hardcoded in the tooltip if a 2H weapon is equipped, probably server side magic
       base_multiplier *= 0.98;
     }
+
+    if ( p -> legendary.superstrain -> ok() && ( p -> specialization() == DEATH_KNIGHT_BLOOD || p -> specialization() == DEATH_KNIGHT_UNHOLY ) )
```
```suggestion
    if ( p -> legendary.superstrain -> ok() && p -> specialization() != DEATH_KNIGHT_FROST )
```

engine/class_modules/sc_death_knight.cpp
```diff
@@ -6120,6 +6145,14 @@ struct virulent_plague_t : public death_knight_spell_t
 
     tick_may_crit = background = true;
     may_miss = may_crit = hasted_ticks = false;
+
+    // when taking superstrain out, set the spell data back to spec.outbreak -> effectN( 2 ).trigger()
+    if ( p -> legendary.superstrain -> ok() && (p -> specialization() == DEATH_KNIGHT_FROST || p -> specialization() == DEATH_KNIGHT_BLOOD) )
```
```suggestion
    if ( p -> legendary.superstrain -> ok() && p -> specialization() != DEATH_KNIGHT_UNHOLY )
```

engine/class_modules/sc_death_knight.cpp
Melekus: Well it's more that unlike VP, they're not applied in aoe.

Melekus: I think adding them to blight's dot component's impact override is good, moving VP application there might also be a good idea.

Melekus:
> * I think this implementation is still missing the virulent eruption implementation.

That is correct, the Virulent Eruption on enemy death is gated behind ```  if ( ! spec.outbreak -> ok() )
  {
    return;
  }``` which will obviously be invalid for non unholy.
Adding superstrain as an alternate condition inside death_knight_t::trigger_virulent_plague_death() should fix it.

PR #5510, opened 2020-11-11
[Priest] create pet.fiend.down expression

seanpeters86:
Resolves the similar problem we've had else where conditions break depending on your fiend talent.

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -1367,6 +1367,11 @@ std::unique_ptr<expr_t> priest_t::create_expression( util::string_view expressio
               };
             } );
           }
+          else if ( splits[ 2 ] == "down" )
+          {
+            bool down = get_cooldown( pet->name_str )->down();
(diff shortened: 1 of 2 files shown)


engine/class_modules/priest/sc_priest.cpp
```diff
@@ -1367,6 +1367,11 @@ std::unique_ptr<expr_t> priest_t::create_expression( util::string_view expressio
               };
             } );
           }
+          else if ( splits[ 2 ] == "down" )
+          {
+            bool down = get_cooldown( pet->name_str )->down();
+            return make_fn_expr( expression_str, [ pet, down ] { return down; } );
```
nuoHep, line 1373: As far as I can tell, this is basically just a constant expression that will always return false.

I *think* what you actually want to do here is something along the lines of
```cpp
cooldown_t* cd = get_cooldown( pet->name_str );
return make_fn_expr( expression_str, [ cd ] { return cd->down(); } );
```

seanpeters86: Hm I can try that instead. But the condition is returning true in my testing I did

PR #5511, opened 2020-11-11
Update APL for Vengeance

Musouv:
Vengeance APL is severely out of date, this should work for both brand and non-brand builds

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -5800,16 +5800,19 @@ void demon_hunter_t::apl_vengeance()
   action_priority_list_t* apl_default = get_action_priority_list( "default" );
 
   apl_default->add_action( "auto_attack" );
+  apl_default->add_action( "variable,name=brand_build,value=talent.agonizing_flames.enabled&talent.burning_alive.enabled&talent.charred_flesh.enabled" );
+  apl_default->add_action( this, "Disrupt" );
   // Only triggers if there is something to steal
   apl_default->add_action( this, "Consume Magic" );
-  apl_default->add_action( "call_action_list,name=brand" );
+  apl_default->add_action( this, "Throw Glaive", "if=buff.fel_bombardment.stack=5&(buff.immolation_aura.up|!buff.metamorphosis.up)" );
+  apl_default->add_action( "call_action_list,name=brand,if=variable.brand_build" );
   apl_default->add_action( "call_action_list,name=defensives" );
   apl_default->add_action( "call_action_list,name=cooldowns" );
   apl_default->add_action( "call_action_list,name=normal" );
 
   action_priority_list_t* apl_defensives = get_action_priority_list( "defensives", "Defensives" );
   apl_defensives->add_action( this, "Demon Spikes" );
-  apl_defensives->add_action( this, "Metamorphosis" );
+  apl_defensives->add_action( this, "Metamorphosis", "if=!(talent.demonic.enabled)&(!covenant.venthyr.enabled|!dot.sinful_brand.ticking)|target.time_to_die<15" );
   apl_defensives->add_action( this, "Fiery Brand" );
 
   action_priority_list_t* cooldowns = get_action_priority_list( "cooldowns" );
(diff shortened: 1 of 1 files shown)


EvanMichaels:
I know we had a few people looking at potential APL updates, do you have some idea of the level of improvements here? (e.g. raidbots comparisons or something)

If it's looking good, should be able to pull it in though, since you are right that the current version is quite out of date. But don't know how it will compare to some of the WIP ones.

Musouv:
It's not really better DPS, but it more accurately reflects how VDH will be playing in SL. The primary change is adding Fel Devastation into the APL, triggering Demonic and the Legendaries.
Here's a sample comparison of the old APL and the new one:
https://www.raidbots.com/simbot/report/3yjrZ1mtvL69yhu8wVSwfd
https://www.raidbots.com/simbot/report/fWb2z1VXcoK2iT7tpMai42

New APL has slightly lower DPS, but it's because this specific setup with Razelikh's incentivizes using all available Fury on Soul Cleave to get more Elysian Decree casts off.

Without the Razelikh's legendary, there's a slight gain:
https://www.raidbots.com/simbot/report/oJ6sLvmiasTSVv19zZt9w2 (old)
https://www.raidbots.com/simbot/report/sDdvG6rSFkdJGraiSqk446 (new)

There may be further improvements to be made to the APL but I think this is a good starting point.

EvanMichaels:
Ok thanks! I see you made some more changes today. You happy with pulling this in now, or still working on stuff?

Musouv:
Should be good to go for now, some of the logic may change if we get further hotfixes, but I'm not anticipating any. Only other major changes would be consumables, but I haven't figured out which ones are optimal yet. This is just a baseline change for the rotation.

EvanMichaels:
Hi @Musouv would you be able to revisit one of these lines and clarify the intent?

soul_cleave,if=((talent.spirit_bomb.enabled&soul_fragments=0)|!talent.spirit_bomb.enabled)&((talent.fracture.enabled&fury>=55)|(!talent.fracture.enabled&fury>=70)|cooldown.fel_devastation.remains>target.time_to_die|(buff.metamorphosis.up&((talent.fracture.enabled>=5)|(!talent.fracture.enabled&fury>=20))))

The "talent.fracture.enabled>=5" part seems in error.

I think for the sake of readability, I think this line should be split into two soul_cleave lines. One with !talent.fracture.enabled at the very start, and one with talent.fracture.enabled at the start. Would make this much easier to parse I think.

EvanMichaels:
As a note, this APL seems to get stuck during Fel Devastation Meta, which is how this was noticed: https://www.raidbots.com/simbot/report/kbi1TL7dUHrnmgopzeHX3d

Musouv:
Ah yeah, that's a typo. It should be (talent.fracture.enabled&fury>=35)
I can try splitting it into multiple lines. There are various conditions whether the player has Spirit Bomb and Fracture talented or not.
Effectively:
1. We never want to cleave if the player has SpB and any souls out.
2. If the player doesn't have SpB, we want to Soul Cleave while maintaining enough Fury to cast Fel Devastation (50 Fury cost) as soon as it's off cooldown.
3. While maintaining that Fury, we want to make sure we aren't overcapping, and the Fury generation is different depending on whether Fracture is talented or not, so we check both conditions.
- 3a. Meta is not up with Fracture, we need 55 Fury, leaving us with 25 and the next Fracture brings us to 50.
- 3b. Meta is not up without Fracture, we need need 70 Fury, leaving us with 40 and the next Shear brings us to 50.
- 3c. Meta is up with Fracture, we need 35 Fury, leaving us with 5 and the next Fracture brings us to 50.
- 3d. Meta is up without Fracture, we need 50 Fury, leaving us with 20 and the next Shear brings us to 50.

If splitting it into multiple lines, there will be some redundant checks, but it would improve readability, if that is a priority.

What it would look like is something like this
`apl_normal->add_action( this, "Soul Cleave", "if=talent.fracture.enabled&((talent.spirit_bomb.enabled&soul_fragments=0)|!talent.spirit_bomb.enabled)&(fury>=55|cooldown.fel_devastation.remains>target.time_to_die|(buff.metamorphosis.up&fury>=35)))" );`
`apl_normal->add_action( this, "Soul Cleave", "if=!talent.fracture.enabled&((talent.spirit_bomb.enabled&soul_fragments=0)|!talent.spirit_bomb.enabled)&(fury>=70)|cooldown.fel_devastation.remains>target.time_to_die|(buff.metamorphosis.up&fury>=50)))" );`

I've made a fix to that, as well as some other improvements here: https://github.com/simulationcraft/simc/pull/5514

PR #5516, opened 2020-11-13
[Death Knight] Phearomones

Khazakdk:
Confirmed with Taez that pets double dip on the haste.  I'm not 100% on the way I set the default value.  It works properly for each spec, but its a bit ugly and uses the same magic number id 3 times.  

Made a couple of assumptions:

- We don't care about modeling the CC effect

- The player will always be standing in their DnD.  This means instead of trying to turn the buff on and off based on in_death_and_decay() I just gave the buff the same duration as DnD

Included is a fix for Outbreak not acting as a proper AoE spell.   It was causing an issue with Superstrain if you didn't have Unholy Blight talented where only the main target got all 3 dots.  With Outbreak reworked to behave similar to Howling Blast I moved the aoe flag to Outbreak instead of having VP apply the other 2 dots itself.

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -501,6 +501,7 @@ public:
 
     // Legendaries
     buff_t* frenzied_monstrosity;
+    buff_t* death_turf;
   } buffs;
 
   struct runeforge_t {
@@ -956,7 +957,7 @@ public:
   struct legendary_t
   { // Commented out = NYI                        // bonus ID
     // Shared
-    // item_runeforge_t phearomones; // 6954
+    item_runeforge_t phearomones; // 6954
     item_runeforge_t superstrain; // 6953
 
     // Blood
@@ -1658,6 +1659,30 @@ struct death_knight_pet_t : public pet_t
     }
 
     return m;
+  }
+
+  double composite_melee_haste() const override
+  {
+    double haste = pet_t::composite_melee_haste();
+
+    if (o() -> legendary.phearomones -> ok())
+    {
+      haste *= 1.0 / (1.0 + o() -> buffs.death_turf -> check_value());
+    }
+
+    return haste;
+  }
+
+  double composite_spell_haste() const override
+  {
+    double haste = pet_t::composite_spell_haste();
+
+    if (o() -> legendary.phearomones -> ok())
+    {
+      haste *= 1.0 / (1.0 + o() -> buffs.death_turf -> check_value());
+    }
+
+    return haste;
   }
 };
 
@@ -4478,6 +4503,12 @@ struct death_and_decay_base_t : public death_knight_spell_t
     {
       p() -> active_spells.bone_spike_graveyard -> set_target( execute_state ->target );
       p() -> active_spells.bone_spike_graveyard -> execute();
+    }
+
+    if ( p() -> legendary.phearomones -> ok() )
+    {
+      p() -> buffs.death_turf -> trigger();
+      p() -> buffs.death_turf -> set_duration(data().duration() + 500_ms);
     }
 
     make_event<ground_aoe_event_t>( *sim, player, ground_aoe_params_t()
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_death_knight.cpp
```diff
@@ -1659,6 +1660,30 @@ struct death_knight_pet_t : public pet_t
...
+  {
+    double haste = pet_t::composite_spell_haste();
+
+    if (o() -> legendary.phearomones -> ok())
+    {
+      haste *= 1.0 / (1.0 + o() -> buffs.death_turf -> check_value());
+    }
+
+    return haste;
+  }
```
Melekus: You may want to gate that behind a owner -> bugs check and add comments about the double dipping.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -9449,6 +9490,10 @@ void death_knight_t::create_buffs()
   buffs.frenzied_monstrosity = make_buff( this, "frenzied_monstrosity", find_spell ( 334896 ) )
     -> add_invalidate( CACHE_ATTACK_SPEED )
     -> add_invalidate( CACHE_PLAYER_DAMAGE_MULTIPLIER );
+
+  buffs.death_turf = make_buff( this, "death_turf", find_spell ( 335180) )
+    -> set_default_value( specialization() == DEATH_KNIGHT_BLOOD ? find_spell( 335180 ) -> effectN(1).percent() / 2 : find_spell( 335180 ) -> effectN(1).percent() )
+    -> add_invalidate( CACHE_HASTE );
```
Melekus: You can use set_pct_buff_type() so you don't have to add the buff to the player's composite melee/spell haste methods manually. 
Default value could be set to that of effect#1, and halved for blood after that, the end result would look like:
```
buffs.death_turf = make_buff( this, "death_turf", find_spell ( 335180 ) )
    -> set_default_value_from_effect( 1 )
    -> set_pct_buff_type( STAT_PCT_BUFF_HASTE );
if ( specialization() == DEATH_KNIGHT_BLOOD )
  buffs.death_turf -> default_value /= 2;
```

engine/class_modules/sc_death_knight.cpp
```diff
@@ -6143,6 +6176,19 @@ struct virulent_plague_t : public death_knight_spell_t
       base_multiplier *= .375;
     }
   }
+
+  void impact( action_state_t* state ) override
```
Khazakdk, line 6180: I got stumped on how to switch all 3 specs to a pattern of putting superstrain in the dots themselves instead of the applicators but this fixes superstrain not working for unholy in aoe and leaves outbreak as a ST spell.  Might be worth revisiting for cleanliness but it gets something working into Taez's hands for now.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -9492,8 +9489,11 @@ void death_knight_t::create_buffs()
     -> add_invalidate( CACHE_PLAYER_DAMAGE_MULTIPLIER );
 
   buffs.death_turf = make_buff( this, "death_turf", find_spell ( 335180) )
-    -> set_default_value( specialization() == DEATH_KNIGHT_BLOOD ? find_spell( 335180 ) -> effectN(1).percent() / 2 : find_spell( 335180 ) -> effectN(1).percent() )
-    -> add_invalidate( CACHE_HASTE );
+    -> set_default_value_from_effect( 1 )
+    -> set_pct_buff_type( STAT_PCT_BUFF_HASTE );
+  if ( specialization() == DEATH_KNIGHT_BLOOD ) {
```
```suggestion
// According to tooltip data and ingame testing, the buff's value is halved for blood
  if ( specialization() == DEATH_KNIGHT_BLOOD ) 
  {
```

engine/class_modules/sc_death_knight.cpp
Melekus: A solution would be to add a boolean variable to the diseases and create different objects for superstrain usage. But this implementation works fine as is.

PR #5523, opened 2020-11-14
[Death Knight] Kyrian Covenant

Armadk:
- Implement Shackle the Unworthy
- Implement Proliferation conduit

I tested the abilities that trigger it on beta.  Abilities tested, as well as what worked and what didn't can be found at: https://github.com/Ceriloch/simc/issues/37

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -9037,7 +9037,6 @@ void death_knight_t::default_apl_blood()
   essences -> add_action( "ripple_in_space,if=!buff.dancing_rune_weapon.up" );
 
   // Single Target Rotation
-  standard -> add_action( "shackle_the_unworthy" );
   standard -> add_action( this, "Death Strike", "if=runic_power.deficit<=10" );
   standard -> add_talent( this, "Blooddrinker", "if=!buff.dancing_rune_weapon.up" );
   standard -> add_action( this, "Marrowrend", "if=(buff.bone_shield.remains<=rune.time_to_3|buff.bone_shield.remains<=(gcd+cooldown.blooddrinker.ready*talent.blooddrinker.enabled*2)|buff.bone_shield.stack<3)&runic_power.deficit>=20" );
@@ -9183,7 +9182,6 @@ void death_knight_t::default_apl_frost()
   obliteration -> add_action( this, "Obliterate", "target_if=max:(debuff.razorice.stack+1)%(debuff.razorice.remains+1)*death_knight.runeforge.razorice" );
 
   // Standard rotation
-  standard -> add_action( this, "Shackle the Unworthy" );
   standard -> add_action( this, "Remorseless Winter", "if=talent.gathering_storm.enabled|conduit.everfrost.enabled|runeforge.biting_cold.equipped", "Standard single-target rotation" );
   standard -> add_talent( this, "Glacial Advance", "if=!death_knight.runeforge.razorice&(debuff.razorice.stack<5|debuff.razorice.remains<7)" );
   standard -> add_action( this, "Frost Strike", "if=cooldown.remorseless_winter.remains<=2*gcd&talent.gathering_storm.enabled" );
@@ -9295,7 +9293,6 @@ void death_knight_t::default_apl_unholy()
   cooldowns -> add_action( this, "Sacrificial Pact", "if=active_enemies>=2&!buff.dark_transformation.up&!cooldown.dark_transformation.ready" );
 
   // General Single Target Priority
-  generic -> add_action( this, "Shackle the Unworthy" );
   generic -> add_action( this, "Death Coil", "if=buff.sudden_doom.react&!variable.pooling_for_gargoyle|pet.gargoyle.active", "Single Target" );
   generic -> add_action( this, "Death Coil", "if=runic_power.deficit<13&!variable.pooling_for_gargoyle" );
   generic -> add_talent( this, "Defile", "if=cooldown.apocalypse.remains" );
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -6679,6 +6679,7 @@ struct scourge_strike_t : public scourge_strike_base_t
 };
 
 // Shackle the Unworthy =====================================================
+// Damage reduction debuff is NYI
 
 struct shackle_the_unworthy_t : public death_knight_spell_t
 {
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -962,7 +962,6 @@ public:
     // const spell_data_t* deaths_due; // Night Fae
     const spell_data_t* shackle_the_unworthy; // Kyrian
     // const spell_data_t* swarming_mist; // Venthyr
-    // const spell_data_t* swarming_mist_energize; // Venthyr, Runic power gain
   } covenant;
 
   struct legendary_t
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_death_knight.cpp
```diff
@@ -3004,6 +3021,41 @@ struct death_knight_melee_attack_t : public death_knight_action_t<melee_attack_t
...
   void trigger_icecap( const action_state_t* state ) const;
   void trigger_razorice( const action_state_t* state ) const;
+
+  double cost() const override
+  {
+    double c = death_knight_action_t::cost();
+    if ( this -> current_resource() == RESOURCE_RUNE )
+    {
+      // If we spend a rune, we have a chance to spread the dot
+      if ( this->triggers_shackle_the_unworthy && as<int>(c) > 0 && p() -> covenant.shackle_the_unworthy -> ok() && p() -> cooldown.shackle_the_unworthy_icd -> is_ready() )
```
scamille: I think execute() is a better place for this. Place it after the action_t::execute call, and you can access last_resource_cost to get the last cost used.

Edit: Not just is it cleaner looking. Putting it into cost() here means that it gets called each time cost() is called, which is more than just once per execution. It can be called many times from action ready checks, including when used through the expression system.

Melekus: Serge is right, cost() is definitely not its place.

I have a few things to add:
If it's the same code for death_knight_melee_attack_t and death_knight_spell_t, you could just override death_knight_action_t::execute(). You could also add it inside death_knight_t::resource_loss() as it receives the action triggering it (although you will have to type it to death_knight_action_t to get the shackles boolean).

Does it also proc on rune spenders when their cost is free (Howling Blast with Rime, DRW abilities?).

Can the dot trigger itself on a target that was already affected before (ie. can you bounce the dot back between two targets indefinitely?)

nyterage: @Melekus Yes it can proc on reduce cost/free rune spenders. It also spreads at its current remaining time (mostly, there is a few ms variance in spread duration...), never spreading to a target its already active on, so its impossible to maintain it indefinitely.

Armadk: Tested with howling blast w/ rime and it definately does spread.  DRW I couldn't tell.  It attributes the buff refreshing by me and comes after my attacks, but before drw's clone attack, so it looks like it is only my attacks that can cause it.  Any way of more reliably testing this?

Armadk: Thanks, yeah I wasn't sure about having it in cost, it's been moved into death_knight_action_t now.  Using howling blast, I also can't get it to proc on more than 1 mob per cast, fyi.  Which should be expected behavior.

Also added a note that the damage reduction portion is not implemented yet.  Is there a way to build a debuff so that the damage reduction is only applied to the caster?  I'm not sure that it actually matters in practical terms for the sim, but wanted to check in to see if there was a way to do it to match in game behavior, before applying a blanket reduction.

Melekus: I think the only reliable way to check with DRW is if you observe a double spread from a single heart strike hit, so a >1 spread with one enemy affected, or >2 spread if your heart strike hits two enemies affected by shackles.
edit: never mind, there's an internal cooldown, so you'd have to look at the actual spread rate vs expected rate from casting HS during DRW on a shackled target

Melekus: For your second question: debuffs are based on actor pairs and as long as you get the right player's target data you should only see the debuffs that they applied themselves.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -2999,6 +2998,38 @@ struct death_knight_action_t : public Base
...
+  void execute() override
+  {
+    action_base_t::execute();
+    // If we spend a rune, we have a chance to spread the dot
+    if ( this->triggers_shackle_the_unworthy && p() -> covenant.shackle_the_unworthy -> ok() && p() -> cooldown.shackle_the_unworthy_icd -> is_ready() )
+    {
+      if ( p() -> rng().roll( p() -> covenant.shackle_the_unworthy -> effectN( 5 ).percent() ) )
+      {
+        death_knight_td_t* source = p() -> get_target_data( p()->target );
+        if ( source -> dot.shackle_the_unworthy -> is_ticking() )
```
Melekus: You could group most of these conditions, and switch from get_target_data( p() -> target ) to get_target_data(target)
edit: store the initial target's shackle dot for copy usage later in the code.
```
dot_t* source_dot = p() -> get_target_data( target ) -> dot.shackle_the_unworthy;
 if ( p() -> covenant.shackle_the_unworthy -> ok() && this->triggers_shackle_the_unworthy && 
      source_dot -> is_ticking() &&  p() -> cooldown.shackle_the_unworthy_icd -> is_ready() &&
      p() -> rng().roll( p() -> covenant.shackle_the_unworthy -> effectN( 5 ).percent() ) )
    {
```

engine/class_modules/sc_death_knight.cpp
```diff
@@ -2999,6 +2998,38 @@ struct death_knight_action_t : public Base
...
+        {
+          for ( auto destination : action_t::target_list() )
+          {
+            death_knight_td_t* destination_td = p() -> get_target_data( destination );
+            if ( p()->target == destination || destination_td -> dot.shackle_the_unworthy -> is_ticking() )
```
Melekus: You should check the target of the action instead of the player's.
Is the spread order random or somewhat deterministic? (range, enemy ID, etc.)

Armadk: Spread does seem to be random.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -3015,9 +3015,12 @@ struct death_knight_action_t : public Base
...
         }
-        action_t::sim->out_log.printf("%s spreads shackle the unworthy with %s from %s to target %s (remains=%.3f)",
-                action_t::player->name(), this->name(), action_t::target->name(), destination->name(),
-                source_dot->remains().total_seconds() );
+        if ( action_t::sim->log )
+        {
+          action_t::sim->out_log.printf("%s spreads shackle the unworthy with %s from %s to target %s (remains=%.3f)",
+                  action_t::player->name(), this->name(), action_t::target->name(), destination->name(),
+                  source_dot->remains().total_seconds() );
+        }
```
Melekus: Although this works, I encourage you to switch to sim -> print_log() that contains the sim -> log check and uses a slightly different (and simpler) formatting:
```suggestion
        sim -> print_log("{} spreads shackle the unworthy with {} from {} to target {} (remains={} )",
                  player->name_str, name_str, target->name_str, destination->name_str,
                  source_dot->remains().total_seconds() );
```

engine/class_modules/sc_death_knight.cpp
```diff
@@ -2982,6 +2998,36 @@ struct death_knight_action_t : public Base
...
+        death_knight_td_t* destination_td = p() -> get_target_data( destination );
+        if ( action_t::target == destination || destination_td -> dot.shackle_the_unworthy -> is_ticking() )
+        {
+          continue;
+        }
+        if ( action_t::sim->log )
+        {
+          action_t::sim->out_log.printf("%s spreads shackle the unworthy with %s from %s to target %s (remains=%.3f)",
+                  action_t::player->name(), this->name(), action_t::target->name(), destination->name(),
+                  source_dot->remains().total_seconds() );
```
scamille: Even shorter:
```suggestion
          action_t::sim -> print_log("{} spreads shackle the unworthy with {} from {} to {} (remains={} )",
                  *action_t::player, *this, *action_t::target, *destination, source_dot->remains() );
```

PR #5524, opened 2020-11-14
[Priest] Add a Second Power Infusion Cooldown

Saeldur:
This gives access to APL to a second power infusion that still obeys all regular cooldowns.
This will let us sim a priest with two power infusions, e.g. a pair of shadows trading with twin suns legendary.


Saeldur:
This is slightly more useful than just using the external PI variable on a timer due to them both respecting their own individual cooldowns properly giving us a bit more flexibility when both get used.

seanpeters86 (approved):

seanpeters86:
@Saeldur so is the plan in a future PR to add some option that inserts a new line into the APL or did you want to do that in this PR?

PR #5528, opened 2020-11-15
Initial implementation of parts of Blessing of Seasons

### LINKED ISSUE #5596: Simulation prot pala bug with night fae.
Hello, when I sim using night fae paladin prot, it's not using the blessings at all and then do not gain the 8 mastery bonus stacks from niya.

Thank you.

mserrano:
resolves https://github.com/simulationcraft/simc/issues/5596

This one is a bit of a doozy. I got semi-started but haven't made a ton of progress yet, and a lot of the behavior I'm having trouble verifying.

Some current issues:
- Blessing of Summer does nothing; I need to figure out what the right way to implement that is (blessing of might and such were extra login in action_t derivatives for Paladins; if we want this to be generic, does that mean we do it in the global action_t?)
- Blessing of Winter seems to proc off of itself, and does no damage (the formula for the damage it does relies on a dummy effect). Maybe relying on this weird special effect stuff is the wrong move here - thoughts?
- Blessing of Spring should be fine
- Blessing of Summer needs verification on whether it's filtering to the right cooldowns.
- There has to be a better way to do the cooldown category stuff; why are these not sharing a cooldown from DBC data?

Anyway I'll keep digging later but wanted to share this since (a) I'm super rusty, and (b) you probably know of better ways to work around these problems


mserrano:
(adding @navv1234 since I figure they'll have advice on how to get the autodetection stuff to work)

Melekus:
First of all, is there actually a need for this to be generic?
It would be much simpler to implement it specifically for paladins (at least for now).

I don't really know for Blessing of Summer, I think you can make it a proc callback and trigger based on the amount of damage done somewhere? Is it actually able to proc on anything matching this?
> White Melee, Yellow Melee, White Ranged, Yellow Ranged, Generic Hostile Spell, Magic Hostile Spell, Periodic

Winter seems to be using another spell for the actual damage and debuff, spell ID 328506 contains a damage effect with a 0.05 coefficient, and buff data for the attack speed/movement speed debuff. I'm not familiar with creating special effects from scratch though, what is SPECIAL_EFFECT_SOURCE_COVENANT supposed to do?

Finally, looking at spelldata, the blessings seem to all share the same cooldown category
> Category         : 1890
Category Cooldown: 45 seconds

mserrano:
> First of all, is there actually a need for this to be generic?
> It would be much simpler to implement it specifically for paladins (at least for now).

I agree. I thought there might be some value in it being generic, since people might want to be able to model which classes are best to put the various buffs on, or classes may have varying priorities based on whether they get Autumn or not. So I think it's worth working on this; I'll put together a separate PR to at least have the paladin-only version working since I think that'll be easier, but I think we should keep this one around.

> I don't really know for Blessing of Summer, I think you can make it a proc callback and trigger based on the amount of damage done somewhere? Is it actually able to proc on anything matching this?
> 
> > White Melee, Yellow Melee, White Ranged, Yellow Ranged, Generic Hostile Spell, Magic Hostile Spell, Periodic

That is my impression (I don't have beta, so I can't really test myself). Others in the Ret discord seem to think so. Do we have examples of similar proc callbacks? I'm not particularly familiar with the callback system in general, hence the confusion in this PR. For the paladin specific version I can just implement it the same way we did with Greater Blessing of Might in early Legion, but that doesn't seem like the right approach for this more generic system.

> Winter seems to be using another spell for the actual damage and debuff, spell ID 328506 contains a damage effect with a 0.05 coefficient, and buff data for the attack speed/movement speed debuff. I'm not familiar with creating special effects from scratch though, what is SPECIAL_EFFECT_SOURCE_COVENANT supposed to do?

Yes. The `special_effect_t` / `dbc_proc_callback_t` pairing, to my understanding, is meant to pull this out of spelldata. See for instance Faulty Countermeasure - https://github.com/simulationcraft/simc/blob/shadowlands/engine/player/unique_gear_legion.cpp#L4547-L4551 - which uses the driver spell ID for the effect ID there rather than the actual damage proc ID. When I had all of this in player_t::create_buff, it seemed to work - IIRC it would proc damage events with spell ID 328506. However, if you look closely at the spelldata for 328506, you'll note that the coefficients on the damage effect are actually 0:

```
#1 (id=821066)   : School Damage (2): frost
                   Base Value: 0 | Scaled Value: 0 (coefficient=0, delta coefficient=0.05) | PvP Coefficient: 1.00000 | Target: Enemy (6)
```

and with debug logs this is replicated, with the proc consistently doing 0 damage. I'm not sure how to properly override this to use a coefficient of 5% with the maximum of spell power or attack power, or how to have the callback trigger an action with the right functions overriden instead of the autogenerated action.

I can "fix" this for paladins alone by just implementing it by hand, but it feels wrong not to be able to have the DBC-derived autodetection work here. Apart from this `max(AP, SP)` thing, none of the spelldata here is *that* weird.

> Finally, looking at spelldata, the blessings seem to all share the same cooldown category
> 
> > Category         : 1890
> > Category Cooldown: 45 seconds

I agree! This is why I was surprised I had to manually trigger the cooldowns instead of the core handling this automatically (like it looks like it does for Shaman shocks, for instance). It's possible I messed that up somehow and misinterpreted some log output, though.

navv1234:
These things would need to be global at some point at least. How these are applied globally is I suppose up for discussion, it does seem like there needs to be some sort of user-interface (set of options?) for this.

As for how to implement global/generic "procs", there really isn't super good way to do it, honestly. Depending on what actually procs the effect, it could be implemented through a normal special_effect_t/dbc_proc_callback_t means (initialized in a suitable place), or it might need to be implemented as part of the assessors (if it can proc from trinkets and various other procs).

We don't have automatic support for category cooldowns in the action initialization, as far as I know. Shaman shocks are done manually at the moment for example.

mserrano:
I merged the paladin-only version, and will rebase this pull request to attempt to make the implementation more global soon (tm). I figure it's useful for at least paladins to be able to use this while we figure out how to make it useful for everyone.

I am evidently blind; I don't know how I missed the cooldown initializations in `frost_shock_t`. I'll also edit this to take a similar approach.

The big questions I think that remain, once I rebase on top of that implementation, are:
- how should we set this up option-wise for non-paladins?
- how should we implement Summer for non-paladins? Do I just hook into action_t::impact?
- where should the code for the requisite custom actions (blessing of winter procs, blessing of summer procs) live?

navv1234:
I suspect you have different possibilities for the option(s). If I was to implement some global override option for this, all implementation choices that I can think of would probably revolve around some sort of repeating trigger of the buff on a specific cycle. Thus, concrete implementation would be a repeating event on a specific interval to trigger the buff. So maybe something like, `override.blessing_of_x=<interval>` to trigger `x` for all sim profile actors every `<interval>` seconds. I don't know if it should have some default interval to enable it, or if it should be disabled as a default value. Gut feeling says latter.

Depending on what actually procs the additional holy damage from Blessing of Summer, you could go with something similar to what I did for the global windfury totem, i.e., add a global switch to enable/disable that controls the proc activation/deactivation (in this case via buff stack change callback), and then just implement a generic special-effect driven callback. First phase initialization can be handled in the same place the rest of the generic special effects are done in player_t. This also lets you create the custom actions just like we do for other procs driven by special_effect_t interface/functionality.

Note that if the additional holy damage and/or the cold damage can proc from stuff like "other procs" and trinkets (other item effects more generally), then we need to move the hooking into assessors. They are the "lowest level" action state manipulation interface in the simulator, and the "damage resolution" for every single execution travels through the simple pipeline.

EDIT: Blessing of autumn may be the problem point here. We currently do not have any kind of "global" (i.e., actor-level) recharge multiplier functionality, all of it is in action_t. This autumn thing may need such an interface implemented.

vituscze:
@Dorovon implemented the blessing (for everyone) some time ago, so I'm gonna close this one.

PR #5536, opened 2020-11-17
[Paladin] Implement Blessing of Seasons for paladins only

mserrano:
Figure I should use the PR system for this since it affects Prot too. I'd rather do the other PR long-term but for now I think this should work (pending someone regenerating spelldata with hotfixes to extract the damage spell for Summer).


mserrano:
(I'm actually just going to merge this, we can revert if we don't like it in preparation to do the other version of the PR)

PR #5537, opened 2020-11-17
[Death Knight] Necrolord Covenant

Armadk:
implement abomination limb
implement brutal grasp conduit

From testing, the resource gain is happening on first damage tick, not on pull mechanic.  I tested it against a boss, and while it can not pull them, once they get in range and the aoe ticks you will gain the resources.  So no need to be out of range to gain the resources.

I skipped implementing the grasp portion per a discussion from acherus.  While it does do damage, it can not damage things that are immune, such as bosses.  It also requires the mob be at least 8 yards away from you.  So once mobs are grouped up on you, there is no good way to trigger it, except running away from the pack, then back in.


engine/class_modules/sc_death_knight.cpp
```diff
@@ -3374,6 +3387,80 @@ struct auto_attack_t : public death_knight_melee_attack_t
...
+    }
+  }
+
+  void execute() override
+  {
+    death_knight_spell_t::execute();
+
+    p() -> buffs.abomination_limb -> trigger();
+
+    // Through testing we find that
```
Melekus: ? 
Might use a comment to explain that abom limb's implementation is simplified to trigger the gains on cast rather than the first time it deals damage as the tooltip says.
You should also mention that the pull effect is NYI

PR #5539, opened 2020-11-19
[Soulbind] fix defaults for Lead by Example

seanpeters86:


engine/player/soulbinds.cpp
```diff
@@ -1031,7 +1031,7 @@ void lead_by_example( special_effect_t& effect )
 
     buff = make_buff( effect.player, "lead_by_example", s_data )
       ->set_default_value_from_effect( 2 )
-      ->modify_default_value( s_data->effectN( 2 ).percent() * effect.player->sim->shadowlands_opts.lead_by_example_nearby )
+      ->modify_default_value( s_data->effectN( 1 ).percent() * effect.player->sim->shadowlands_opts.lead_by_example_nearby )
```
Saeldur: This is wrong, `s_data->effectN( 2 )` still contains this effect.
It's `s_data->effectN( 1 )` for baseline value, I'm not sure if `set_default_value_from_effect( 2 )` is using the same spelldata.

Saeldur (approved):
LGTM, this still leaves the default at 2 which seems fair given its potentially awful 8yd range.
This may still end up being an average 4 for raid environments for casters but melees may see less and dungeons are definitely less.

PR #5546, opened 2020-11-21
[Death Knight] fix voracious

Armadk:
Voracious calculation when using bugs std:floor was returning 0, this fixes it to return the proper value with voracious.


Armadk:
FYI this one is included in the blood PR I just put up.  It was noticed when I was implementing Bryndaor's.

PR #5549, opened 2020-11-22
[Death Knight] Blood conduits/legendaries

Armadk:


engine/class_modules/sc_death_knight.cpp
```diff
@@ -2528,6 +2536,16 @@ struct dancing_rune_weapon_pet_t : public death_knight_pet_t
     int n_targets() const override
     { return p() -> o() -> in_death_and_decay() ? aoe + as<int>( p() -> o() -> spec.death_and_decay_2 -> effectN( 1 ).base_value() ) : aoe; }
 
+    double composite_da_multiplier( const action_state_t* state ) const override
+    {
+    double m = drw_attack_t::composite_da_multiplier( state );
+    if ( p() -> o() -> conduits.withering_plague -> ok() && target ->get_dot( "blood_plague", p()) -> is_ticking() )
```
scamille: 1. I think there is some composite_target_da_multiplier function for this, which I recommend using.
2. This does a string-based dot lookup during combat. Please use death_knight_target_data_t instead. Something like
auto td = get_td(target);
if (td->blood_plague->is_ticking())
  ...

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5788,6 +5830,16 @@ struct heart_strike_t : public death_knight_melee_attack_t
   int n_targets() const override
   { return p() -> in_death_and_decay() ? aoe + as<int>( p() -> spec.death_and_decay_2 -> effectN( 1 ).base_value() ) : aoe; }
 
+  double composite_da_multiplier( const action_state_t* state ) const override
+  {
+    double m = death_knight_melee_attack_t::composite_da_multiplier( state );
+    if ( p() -> conduits.withering_plague -> ok() && td(state -> target) -> dot.blood_plague -> is_ticking() )
```
scamille: Same thing here, there should be a function composite_target_da_multiplier (or similar) which you can use.

The usgae of target data looks much better than at the other place.

scamille (changes requested):

engine/class_modules/sc_death_knight.cpp
```diff
@@ -946,9 +950,9 @@ struct death_knight_t : public player_t {
     // conduit_data_t spirit_drain; Finesse trait, 70
 
     // Blood
-    // conduit_data_t debilitating_malady; // 123
-    // conduit_data_t meat_shield; // Endurance trait, 121
-    // conduit_data_t withering_plague; // 80
+    conduit_data_t debilitating_malady; // 123
+    conduit_data_t meat_shield; // Endurance trait, 121
+    conduit_data_t withering_plague; // 80
```
Melekus, line 955: So Debilitating Malady doesn't affect DRW's blood plague damage, but Withing Plague increases damage done by your DRW's Heart Strike? that's peculiar.

engine/class_modules/sc_death_knight.cpp
Melekus: You can use composite_target_multiplier, just like in your Debilitating Malady implementation.
Is the damage increase dependant on the player's blood plague being up on the target, or on dancing rune weapon's blood plague? For the former you can fetch the player's target data, for the latter you will have to create a simple target data structure containing the blood plague dot (check how it's done for Magus of the Dead's frostbolt debuff).

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5073,6 +5107,14 @@ struct death_strike_heal_t : public death_knight_heal_t
   {
     death_knight_heal_t::impact( state );
 
+    if ( state -> result_total > player -> resources.max[ RESOURCE_HEALTH ] * p() -> legendary.bryndaors_might -> effectN( 2 ).percent() )
+    {
+      // Last resource cost doesn't return anything so we have to get the cost of Death Strike
+      double c = p() -> find_action( "death_strike" ) -> cost();
```
Melekus: Finding action by name during runtime doesn't sound like a great idea, even if it was like that for the Legion implementation. I think it would be better to simply store the parent death strike action somewhere and call its cost() when needed.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -10067,6 +10149,9 @@ void death_knight_t::bone_shield_handler( const action_state_t* state ) const
   {
     buffs.bones_of_the_damned -> expire();
   }
+
+  // Value is stored in the spelldata as negative, so multiply by a positive
+  cooldown.dancing_rune_weapon -> adjust( 1.0 * legendary.crimson_rune_weapon -> effectN( 1).time_value() );
```
```suggestion
  cooldown.dancing_rune_weapon -> adjust( legendary.crimson_rune_weapon -> effectN( 1 ).time_value() );
```
You don't need any special handling of the data here

engine/class_modules/sc_death_knight.cpp
```diff
@@ -9844,11 +9896,28 @@ void death_knight_t::create_buffs()
...
+
   buffs.unleashed_frenzy = make_buff( this, "unleashed_frenzy", conduits.unleashed_frenzy->effectN( 1 ).trigger() )
       -> add_invalidate( CACHE_STRENGTH )
       -> set_default_value( conduits.unleashed_frenzy.percent() );
 
   // Legendaries
+
+  buffs.crimson_rune_weapon = make_buff( this, "crimson_rune_weapon", find_spell( 334526 ) )
+      -> set_default_value( find_spell( 334526 ) -> effectN( 1 ).percent() );
+
```
Melekus: Just like Runic Corruption or the T21 set bonus that crimson rune weapon is ~~copied from~~ based on, you also have to update the rune regen coefficient whenever the buff's state changes and make sure the sim knows they affect regeneration rates with set_affects_regen( true ).
https://github.com/simulationcraft/simc/blob/legion-dev/engine/class_modules/sc_death_knight.cpp#L4449

```suggestion
      -> set_default_value( find_spell( 334526 ) -> effectN( 1 ).percent() )
      -> set_affects_regen( true )
      -> set_stack_change_callback( [ p ]( buff_t*, int, int )
           { p -> _runes.update_coefficient(); } );
```

engine/class_modules/sc_death_knight.cpp
Armadk: Tell me about it.  From my testing notes, Withering Plague DRW damage amp is based on it's own BP.

DRW's blood boil is the one that amps the damage.
Casting BB -> DRW -> HS does not amp DRW-HS damage.
Casting DRW -> BB -> HS does amp DRW-HS damage

For debilitating malady, your BB applies a debuff to the mob which ramps the damage, but DRW's BB does not apply the debuff.  I could not observe any change in damage with DRW, but to be fair, the amounts are so small it might be a rounding error.  This ability ramps damage by 1% per stack, which you generally get 2, maybe 3 at most stacks due to BB cooldown.  With my personal BP I can see the damage go from 93 to 95, and at 3 stacks a 96.  DRW's BB is only 31 damage though.  With 2 stacks it should be enough to push it to 32, but I have only observed 31 from it.  Yes, this conduit actually only increases BP damage by 1% per stack at rank 1.

engine/class_modules/sc_death_knight.cpp
Armadk: It is based on DRW's blood plague.  Thanks for the suggestion.  I'll go that route.

scamille (approved):

engine/class_modules/sc_death_knight.cpp
scamille: Ah I did not realize that this was about a pet action, and thus no target data was set up. Looks good now with the added target data class.

PR #5550, opened 2020-11-22
Updated the affliction warlock APL

PiBeta:
Updated the affliction warlock APL to use the new covenant abilities and an APL optimized around said covenants and their soulbinds. This includes information for all four covenants. I specifically focused on optimizing the Night Fae APL, since that seemed most in need of optimization and was the only one likely to overtake Kyrian.

Results using the current APL: https://www.raidbots.com/simbot/report/af6aHVjqwWRPfEWan2QnYJ

Results using the new APL: https://www.raidbots.com/simbot/report/736NKkJvAYijjo2F7ZjYQj

NOTE: The new APL did not use any consumables in the Raidbots sim for unknown reasons. I'm probably missing a configuration option somewhere, but the new APL is still obviously higher DPS.
NOTE: This does disable BFA-specific actions like essences and azerite traits, since making that work with the new covenants is both redundant and unnecessarily complex.

### DIFF
--- a/engine/class_modules/warlock/sc_warlock_affliction.cpp
+++ b/engine/class_modules/warlock/sc_warlock_affliction.cpp
@@ -1044,44 +1044,105 @@ void warlock_t::create_apl_affliction()
   action_priority_list_t* se    = get_action_priority_list( "se" );
   action_priority_list_t* aoe   = get_action_priority_list( "aoe" );
   action_priority_list_t* item  = get_action_priority_list( "item" );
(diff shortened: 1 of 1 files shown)


Azevara:
The APL this was built on has been merged into SimC. There may not actually be real merge conflicts, but the formatting in this PR needs to be adjusted to match the formatting used elsewhere (single lines only) in order to easily diff this.

Azevara:
It appears there are no functional differences remaining between the APL in the first comment and the one currently used by simc, closing this.

PR #5552, opened 2020-11-22
[Death Knight] Unholy blight dot no longer pandemics

Armadk:
the unholy blight dot damage portion no longer pandemics, virulent still does.

Is there a better way to do this?  The other refresh options for dot's didn't seem to match what I needed for behavior.

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -7098,6 +7098,12 @@ struct unholy_blight_dot_t : public death_knight_spell_t
     aoe = -1;
   }
 
+  timespan_t calculate_dot_refresh_duration( const dot_t* dot, timespan_t triggered_duration ) const override
+  {
+    // No longer pandemics
+    return triggered_duration;
+  }
+
   void impact(action_state_t* state) override
   {
     death_knight_spell_t::impact( state );


Melekus:
Looking at the code around action_t::dot_behavior, it seems to be pretty barebones and the only thing that it does is cancel the dot when trying when it's already on the target and the behavior is DOT_CLIP (which may fit what you want to do here?).

I remember that @vituscze was working on dots recently, maybe he'll have a better idea?

Melekus:
Looking at how dot behavior is implemented, it could certainly use some fleshing out, and your current implementation seems fine for Unholy Blight so I'll push this and make a note to look into dot behavior later on.

navv1234:
We do not have explicit controls to determine refresh duration, you'll need to manually put in the correct duration. For "old style" dots, it would be refresh_duration + ongoing-tick-time (`return dot->time_to_next_tick() + triggered_duration`), in some very rare cases, it is/has been `triggered_duration` (i.e., the dot refreshes to the maximum duration and ignores the ongoing tick).

Melekus:
In this case, the dot just refreshes to its default duration (14s) on every application (every 1s for 6s at the start) and keeps ticking every 2s regardless of haste (including one initial tick on the first application, but not on refresh) for a total duration of 20s.
Applying the dot for only a portion of the 6s just results in a lower overall duration dot, with a possible partial tick.

PR #5554, opened 2020-11-23
Manual override of Death Knight Night Fae ability lookup

Armadk:
manually add deaths due covenant ability so that soulbind callbacks can correctly locate it.

dbc/generated/covenant_data.inc and dbc/generated/covenant_data_ptr.inc have 0 in the field meant for class for Death's Due, when the value should be 6.  As a result, we are not able to automatically detect the DK covenant ability, so callbacks from the various soulbinds do not work.

This manually sets it, allowing callbacks to correctly trigger.


Armadk:
While this does get the soulbind callbacks to start working, the html report still tries to pull the ability name from the file it seems, and so is unable to find it, listing 'undefined' in the html report.

Melekus:
Should be fixed by https://github.com/simulationcraft/simc/commit/c656dba13ddd191e807f5f17324d2eacb1e8a0cb

(waiting for data pull to see it in effect before closing)

Armadk:
Verified on latest build.  This PR is no longer required.

PR #5567, opened 2020-11-24
Update Warlock APL to prevent PS issue

Strahbehry:
Currently PS will accidentally be used just after Darkglare, waiting for it for 15 seconds max completly prevents this issue. And then calling it in the Darkglare prep. Depending on haste it may or may not occur with a 5min sim. Here is a comparison on a 10min 2t. 

https://www.raidbots.com/simbot/report/7f6uhMWwk6GFZTX96mkiXh

https://www.raidbots.com/simbot/report/iRZk4rpC9SiY5wRzS4Ddvn

1t:

https://www.raidbots.com/simbot/report/bWKc6JE1pZxezRs7Ezw8YS

https://www.raidbots.com/simbot/report/ozq7Jnr4csNgWKEEn3yG1Q

### DIFF
--- a/engine/class_modules/warlock/sc_warlock_affliction.cpp
+++ b/engine/class_modules/warlock/sc_warlock_affliction.cpp
@@ -1046,7 +1046,7 @@ void warlock_t::create_apl_affliction()
   action_priority_list_t* item  = get_action_priority_list( "item" );
 
   def->add_action( "call_action_list,name=aoe,if=active_enemies>3" );
-  def->add_action( "phantom_singularity,if=time>30" );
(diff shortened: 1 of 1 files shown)


Azevara:
I have left a comment in the Warlock discord's #theorycrafting channel regarding this - there are some issues when using Darkcaller that appear to cause losses. This would need correcting before this could be incorporated. The gains mostly appear limited to situations where Haunt is talented, perhaps only implementing it for that talent would be a good iterative step.

seanpeters86:
@Strahbehry any update on this?

PR #5590, opened 2020-12-01
[Paladin] Add prot pala preraid profile

srenauld:
Adding a relatively basic pre-raid prot paladin profile. Since stat weights were so close to ret's, most of it is the same.

Changes:
- Venthyr with nadjia
- Talents were updated to match SL recommended talents
- Different MH/OH
- Different legendary (relentless inquisitor, replaced on legs, changed boots to muckwallow stompers)


srenauld:
Woli is coming with a better PR. Closing this.

Issue #5586, opened 2020-11-30
GUI load failed because SimulationCraft.ini cannot load with CJK user name


EKE00372:
### Describe the bug
GUI load failed because SimulationCraft.ini cannot load with CJK user name

### Expected behavior
When user name was named as CJK character, SimulationCraft.ini will load failed and cause SimGUI won't start/cannot load/crash, it probably treated as an unknow issue (in wiki question 3 and 7 and maybe #5466). Also, this issue feedback from a chinese language user, so i'm not sure it's only happened on chinese Character only, or CJK, or even non-english, but i think CJK is the most likely happened.

Source feedback [here (chinese forum)](https://bbs.nga.cn/read.php?tid=24382103)

### To Reproduce
Create a new user account in PC and name as CJK character.
C:\Users\ **UserName** \AppData\Roaming\SimulationCraft

### Additional information
* SimulationCraft Version (required)
Since version 810
below are able to open.

suujonH:
Same problem.
This issue inspired me, so I used [Locale-Emulator](https://github.com/xupefei/Locale-Emulator), set the location to **en-us**, and then successfully started.

scamille:
Thanks a lot for the report. That might indeed be the cause of some of the other issues reporting that they can not start the GUI.

How exactly did you determine that the failure is while loading Simulationcraft.ini?

EKE00372:
First, he tried compare another PC's file which can run simc normally, and copy it's Simulationcraft.ini to his pc but still crash. and then he started try old version, notice this problem happened since 810 and work fine with 801, but when he edit his ini file and start with upper version, Simulationcraft.ini in AppData didn't have any change, so finally found issue.

I don't have this issue......my user name is english. but as my opinion, CJK support is also a usually problem on program, not only simc.

for example it even happened on WOW client, 6 years ago, WOD beta can crash by this issue, "do not use CJK char on folder path which your client installed" is still in my beta guide until now, although it never happened in later beta again.

suujonH:
#5595

serrufs:
> Same problem.
> This issue inspired me, so I used [Locale-Emulator](https://github.com/xupefei/Locale-Emulator), set the location to **en-us**, and then successfully started.

Did you download SimulationCraft.exe for Local-Emulator? This Locale-Emulator does not work for SimulationCraft.exe
[...]

PR #5595, opened 2020-12-01
Windows platform: Convert the environment string's encoding to utf8 when using io::ofstream.

suujonH:
#5586 
In io::ofstream, the path is always converted from UTF-8, but in the function token_paths(), obtained text encoding is ANSI.

### DIFF
--- a/engine/interfaces/bcp_api.cpp
+++ b/engine/interfaces/bcp_api.cpp
@@ -104,9 +104,13 @@ std::vector<std::string> token_paths()
 
   if ( const char* home_drive = getenv( "HOMEDRIVE" ) )
   {
-    if ( const char* home_path = getenv( "HOMEPATH" ) )
+    if ( const char* home_path_ansi = getenv( "HOMEPATH" ) )
     {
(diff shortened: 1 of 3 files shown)


scamille:
Builds fine with the qmake created VS solution, and I think it works fine at least on my plain old ascii username.

I don't really know if this is the correct way to convert the windows getenv values to utf8, or if it is at least good enough for the intended platform. Would be great if someone else more knowledgeable with this could have a look as well :)

It could also be adjusted to use vectors instead of new[]/delete[] , but that shouldn't really have any impact on the functionality.

scamille (approved):

navv1234:
I'm no expert either, really. I guess my only question is whether this works when the code page is already UTF8 (ie., GetACP() returns CP_UTF8 (or 65001))?

suujonH:
> I'm no expert either, really. I guess my only question is whether this works when the code page is already UTF8 (ie., GetACP() returns CP_UTF8 (or 65001))?

It's work(In my case Win10 1904).

scamille:
Thanks again, this seems to have resolved problems for a lot of people not being able to start SimulationCraft.exe

PR #5599, opened 2020-12-01
update to legendary for preraid profile

Tomslack:
because people are psychotic linking stack sims in my class discord


seanpeters86 (approved):

Melekus (changes requested):
Make sure that your pull request doesn't accidentally revert the new Guardian PR profile

PR #5627, opened 2020-12-05
[Arcane] Shifting Power Use In AOE

The-PaleRider:
Base SP use around Orb CD in AOE

https://www.raidbots.com/simbot/report/qn5S5Hk4y49Yxe6Mc9b8nh/simc


Sergrand1:
Tested this with various talent/conduit/target count setups, and it was a gain across them all. It might be better to move that hardcoded 5 to a variable in the future.

PR #5628, opened 2020-12-05
[shaman] Add default_temporary_enchant

wfarr:
This sets the weapon oil or other default enchant string.

* For now, only set Elemental to use shadowcore oil.
* I'm sure Enhance may need something special for their weapon enchants.
* Resto is probably going to use shadowcore oil for dps too but I'll let
  someone keen on Resto DPS sims worry about that.


navv1234:
For now, enhance can have both enabled if you set the temporary_enchant string, but I'll try to do some workaround there to ensure that if temporary enchant is enabled on mainhand, windfury is not.

PR #5633, opened 2020-12-06
[Death Knight]  T26 Blood Profile

Armadk:
Regenerated branch and PR


Melekus (approved):
Runs fine through the generate script and I don't see any issue with it at glance (but I didn't check the individual items)

PR #5646, opened 2020-12-08
Discipline add back to CI.simc; Disable Multi Target simulations

scamille:


seanpeters86:
@scamille looks like for this we would have to remove from CI? or change how it errors out

scamille:
Yeah disabling multi target sims is not really something that works great with our CI setup - or is easy to get around.

I'll try to salvage the other parts of this, without disabling it.

PR #5659, opened 2020-12-10
Added precombat option for stealth

glacialz:
[Rogue] Gave stealth the precombat option to take advantage of master of shadows.

### DIFF
--- a/engine/class_modules/sc_rogue.cpp
+++ b/engine/class_modules/sc_rogue.cpp
@@ -1399,6 +1399,7 @@ public:
   void trigger_find_weakness( const action_state_t* state, timespan_t duration = timespan_t::min() );
   void trigger_grand_melee( const action_state_t* state );
   void trigger_master_of_shadows();
+  void trigger_master_of_shadows( timespan_t precombat_seconds_lost );
   void trigger_akaaris_soul_fragment( const action_state_t* state );
   void trigger_bloodfang( const action_state_t* state );
   void trigger_count_the_odds( const action_state_t* state );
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_rogue.cpp
```diff
@@ -4102,9 +4102,15 @@ struct vendetta_t : public rogue_spell_t
 
 struct stealth_t : public rogue_spell_t
 {
+  double precombat_seconds;
```
nuoHep, line 4106: Without evaluating if it's needed in the first place or not, times in simc are generally stored as `timespan_t`s.

engine/class_modules/sc_rogue.cpp
```diff
@@ -4117,6 +4123,14 @@ struct stealth_t : public rogue_spell_t
 
     p()->buffs.stealth->trigger();
     trigger_master_of_shadows();
+
+    // trigger_master_of_shadows() doesn't trigger out of combat, must be applied manually
+    if ( precombat_seconds && !p()->in_combat )
```
EvanMichaels: I think either trigger_master_of_shadows() needs to be in an else case here, or this logic needs to be moved into the trigger_master_of_shadows() function itself with a precombat_seconds optional parameter or something. 

Think it's a little unclear to someone browsing this code as to why we are triggering MoS twice here or doing so manually when there is already a function here.

engine/class_modules/sc_rogue.cpp
glacialz: It's cast to `timespan_t` after the conditional, similar to how precombat is implemented on the other spells in this file.

nuoHep: Yes, I saw that. I'm not going to press it (not "my" module etc) but I'd argue the other spells are doing it "wrong" too.

navv1234: I'd agree with that sentiment. Time is intended to be represented by `timespan_t` to avoid ambiguity between units.

Mystler (commented):
I'm not really a fan of either the purpose nor the implementation. 1s prepull stealth is some pretty risky maneuver that I would call unrealistic, all for a super super minor energy minmax. As for the implementation, a copy of the trigger MoSh function is not really necessary and this could have been done with a default argument.

PR #5663, opened 2020-12-11
Shadowcore Oil Fix

liealgebras:
Shadowcore oil damage is being applied to Enhancement Shaman simulations even though it's impossible to stack both weapon imbues (Windfury/Flametongue) with shadowcore oil.

Problem

Fix


navv1234:
This isn't really fixing the core issue (allowing both to be up at the same time). I'll fix over the weekend so that temporary_enchant= will override imbues properly. I also need to allow double flametongue imbues, and tie imbues to their respective actions anyhow.

navv1234:
Should now be fixed in commits a0f4348 and f020e7d.

PR #5671, opened 2020-12-15
Change to Race for VDW

Ruby437:
Since both troll outsims goblin in all the sims I've run (standard patchwerk, patchwerk 3 targets, hectic add cleave, dungeon slice) I suggest changing the default race to troll.


scamille:
cc @simulationcraft/shaman 

@aethys256 is this something which might apply to more profiles than just this one?

PR #5672, opened 2020-12-15
windwalker APL update

scurch1:
calculated strikes works with mastery so need to weave again
Fist of the White Tiger corrected to Fists of Fury


scamille:
cc @simulationcraft/monk

PR #5677, opened 2020-12-16
Priest ecttv refactor

seanpeters86:
The current implementation just tosses the leggo effects under the spells that proc it, rather than under a EcttV action like WCL does. This should make it easier to distinguish the power of the legendary by looking at the result table.

new ecttv

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -1286,6 +1286,18 @@ struct psychic_horror_t final : public priest_spell_t
 };
 
 // ==========================================================================
+// Eternal Call to the Void (Shadowlands Legendary)
+// ==========================================================================
+struct eternal_call_to_the_void_t final : public priest_spell_t
+{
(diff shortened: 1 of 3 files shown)


scamille (approved):

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1285,6 +1285,17 @@ struct psychic_horror_t final : public priest_spell_t
...
 };
 
+// ==========================================================================
+// Eternal Call to the Void (Shadowlands Legendary)
+// ==========================================================================
+struct eternal_call_to_the_void_t final : public priest_spell_t
+{
+  eternal_call_to_the_void_t( priest_t& p )
+    : priest_spell_t( "eternal_call_to_the_void", p, p.find_spell( p.legendary.eternal_call_to_the_void->id() ) )
+  {
```
scamille, line 1295: Add background=true;

scamille (changes requested):

PR #5680, opened 2020-12-18
Adding doomwinds to the aoe rotation.

Ruby437:
Prioritizing windfury proccing skills during the doomwinds window. Reordered to prioritize Crash Lightning. Fitting sundering in the last GCD becomes better the more targets are present, it's equal at 3 targets.


Pewtro:
If this is first equal at 3 targets, then you should ensure this is only run on 3+ targets. Right now it looks to be run on 2 targets too, which would be a DPS loss in that situation, and there are situations like that in DSlice etc too that you may want to account for. 

cc @simulationcraft/shaman since they'll know more - this is just basic suggestions and considering you should address, since I don't play shaman.

Ruby437:
Just ran a sim at 2 targets in a few settings, its equal or better in 2 targets too. This is a straight copy and minor optimization to the single target doom winds adjustments so there's no weird side effects that could occur as far as i'm aware.

Issue #5685, opened 2020-12-20
Update GUI realm lists


JoaoFranciscoNeto:
### Describe the bug
Unable to load characters because of incorrect slug
Slug is aggra-portugues but should be aggra-português.

### To Reproduce
Please include any steps necessary to recreate the problem, or include logs/data to show what is occurring in game if there is an issue with how something is modeled.

### Additional information
* SimulationCraft Version simc-902-01

scamille:
Do I understand this correct that the server name used by SimC is incorrect, and thus the armory import does not work?

Please provide a armory link to one of your characters on Aggra so we can more easily replicate the problem.

cc @navv1234

JoaoFranciscoNeto:
That is correct, the slug should have the special character
Armory link:
https://worldofwarcraft.com/en-us/character/eu/aggra-portugu%C3%AAs/smashstein

Profile request:
https://eu.api.blizzard.com/profile/wow/character/aggra-portugu%C3%AAs/smashstein?namespace=profile-eu&locale=en_GB

Created a Pull Request with a fix.

scamille:
Please check with the latest nightly that it is working, and then this can be closed. Thanks for the pull request :-)

navv1234:
Just fyi @scamille there are several other realms that are also broken like that. I'll try to re-generate all our realmlist data at some point soon.

JoaoFranciscoNeto:
> Please check with the latest nightly that it is working, and then this can be closed. Thanks for the pull request :-)

No, thank you for the fantastic work you have been putting on the project.
I just did what was probably the shortest Pull Request in Git history 😄

PR #5686, opened 2020-12-20
Update eu-realmlist.json

JoaoFranciscoNeto:
Fixing issue:
Incorrect slug for Aggra (EU realm) #5685


navv1234 (approved):
This should be ok, however we should rewrite some of this stuff for the new BCP API, since the realm endpoint json has changed apparently. Those files are direct dumps from their (now defunct) API anyhow.

PR #5689, opened 2020-12-22
Added optimization for Firestorm Legendary in 2 target with Flamepatch.

Preheat:
While Firestorm is active on more than one target with Flame Patch, use Flamestrike instead of Pyroblast


Melekus:
This PR only updates the T26 default profile that will be overriden the next time someone regenerates default profiles.

If you want your changes to stick you will have to update the cpp file that generates the default APL:
https://github.com/simulationcraft/simc/blob/shadowlands/engine/class_modules/apl/mage.cpp#L273

PR #5690, opened 2020-12-22
[GUI] Check in options for client id / secret during import process

### LINKED ISSUE #5279: "Unable to download JSON" When Importing
### Describe the bug
I tried several times to import my character with the import tab, I always get the same result in the simulation
# Unable to generate profile from: 
# Unable to download JSON from 'https://eu.api.blizzard.com/profile/wow/character/kelthuzad/dabu?namespace=profile-eu&locale=en_GB' 

It works when I copy paste the data from the wow addon

### Expected behavior
It importing my Character like it says it would

### To Reproduce
Open simulation craft, change to import, enter  EU Kel'Thuzad Dabu click import

### Additional information
* SimulationCraft Version (required)
830-01

doesnt even begin simulating

elegos:
Hello!

I've made this PR to fix #5279 

I'm only unsure whether:

- put the option names into constants, but I see no such pattern in the code (and thus wouldn't be part of this PR)
- use `sc_sim.cpp`'s `validate_api_key` function, but it's currently in an anonymous namespace and didn't want to refactor

(my C++ skills are... rusty)

Checked under Linux.


scamille:
I moved validate_api_key for you in ad2d7e41c3b13a9317fa563337dd2952136a32f6

I think you can just completely remove SC_MainWindow::startImport, as I can't find any reference to it. That leaves only the SC_MainWindow::startNewImport function to implement your code changes.

qt/MainWindow.cpp
```diff
@@ -661,6 +661,21 @@ void SC_MainWindow::startNewImport( const QString& region, const QString& realm,
   }
   simProgress = 0;
   import_sim  = initSim();
+
+  #ifndef SC_NO_NETWORKING
+  // Try to import the API key from the GUI options
+  if (import_sim->apikey == "") {
+    QSettings settings;
+    std::string apikey = settings.value("options/api_client_id").toString().toStdString() +
```
scamille: ~~I think you should get the value not from the settings object, but from optionsTab->api_client_id->currentText() and optionsTab->api_client_secret->currentText().~~ (Edit: There is a function SC_OptionsTab::get_api_key() for you to get the api key.) Otherwise this will not pick up the value a user just entered into the QLineEdit control. Settings are only saved when a simulation starts / gui is closed.

To check if a std::string is empty you can just call its empty() member function instead of comparing to "".

elegos: Ok I've resolved this one :)

Getting into projects is always quite a shock, but I'll kick in (I hope :D )

elegos:
@scamille between my job and my PR, I think I closed all the points (eventually!)

scamille (approved):

scamille:
Yeah looks good now. I take it you tested the gui with it and it actually helps resolving the armory download with a custom api key entered through the GUI options?

elegos:
Yup exactly, tested at each change :) works fine for me, deleting the `.simc_apikey` file during the tests obviously.

scamille:
Btw you should be able to clang-format a subset of files in SimC to take care of formatting, for example this one. Just make sure to separate formatting from content changes if it affects more than what you are adding.

PR #5696, opened 2020-12-24
[Monk] Implement Xuen Rank 2

Hinalover:


engine/class_modules/sc_monk.cpp
```diff
@@ -4331,6 +4551,23 @@ struct monk_action_t : public Base
     }
   }
 
+  void trigger_empowered_tiger_lightning( action_state_t* s )
+  {
+    if ( p()->spec.invoke_xuen_2->ok() )
+    {
+      // Make sure Xuen is up and the action is not the Empowered Tiger Lightning itself
+      // Touch of Karma (id = 124280) does not contribute to Empowered Tiger Lightning
+      if ( p()->buff.invoke_xuen->up() && s->result_total > 0
```
scamille, line 4560: Instead of checking for ids, would it be possible to add a flag
bool triggers_invoke_xuen_2; to monk_action_t, which you could then disable on specific actions in the constructor by adding triggers_invoke_xuen_2 = false;

engine/class_modules/sc_monk.cpp
```diff
@@ -2347,15 +2370,87 @@ struct xuen_pet_t : public pet_t
...
+        }
+      }
+
+      melee_attack_t::impact( s );
+    }
   };
 
   struct crackling_tiger_lightning_tick_t : public spell_t
   {
+    monk_t* owner;
```
scamille, line 2412: There is no need to add a owner variable here. xuen_pet_t* p which you pass to the action already has a owner, which you can access through its o() member function.

engine/class_modules/sc_monk.cpp
```diff
@@ -2347,15 +2370,87 @@ struct xuen_pet_t : public pet_t
...
-      : spell_t( "crackling_tiger_lightning_tick", p, p->o()->passives.crackling_tiger_lightning )
+      : spell_t( "crackling_tiger_lightning_tick", p, p->o()->passives.crackling_tiger_lightning ), 
+        owner( p->o() )
     {
       dual = direct_tick = background = may_crit = true;
     }
+
+    void impact( action_state_t* s ) override
+    {
+      if ( owner->spec.invoke_xuen_2->ok() )
```
scamille, line 2422: This should definitely go into a shared trigger_empowered_tiger_lightning function. Too much copy-paste for the pet code.

You could either make it freestanding and just do its job, or maybe it would be helpful to add a xuen_pet_action_t template, similar to monk_action_t

PR #5710, opened 2020-12-31
Monk: Remove Azerite

scamille:
Also switch to using util::string_view in some places.


scamille:
deterministic sim gives the same result, so looks like this really had no impact on current windwalker simulations.

PR #5712, opened 2020-12-31
Priest: Remove c++ if in the shadow APL code; merge disc/holy apls in…

scamille:
…to dps only.

Disable healing role for priests. This will also disable default holy sims unless dps role is actively specified.

Disable windfury totem callback for (healing) enemy. This fixes the assert for holy priest and its associated healing enemy, which is only set to quiet even if disabled.

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -1974,15 +1974,12 @@ void priest_t::generate_apl_shadow()
                         "Default fallback for usable items: Use on cooldown in order by trinket slot." );
 
   // CDs
-  if ( options.priest_self_power_infusion )
-  {
-    cds->add_action( this, "Power Infusion",
-                     "if=buff.voidform.up|!soulbind.combat_meditation.enabled&cooldown.void_eruption.remains>=10|fight_"
-                     "remains<cooldown.void_eruption.remains",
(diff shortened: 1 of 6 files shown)


engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1974,15 +1974,12 @@ void priest_t::generate_apl_shadow()
...
-  if ( options.priest_self_power_infusion )
-  {
-    cds->add_action( this, "Power Infusion",
-                     "if=buff.voidform.up|!soulbind.combat_meditation.enabled&cooldown.void_eruption.remains>=10|fight_"
-                     "remains<cooldown.void_eruption.remains",
-                     "Use Power Infusion with Voidform. Hold for Voidform comes off cooldown in the next 10 seconds "
-                     "otherwise use on cd unless the Pelagos Trait Combat Meditation is talented, or if there will not "
-                     "be another Void Eruption this fight." );
-  }
+  cds->add_action( this, "Power Infusion",
```
seanpeters86, line 1977: I did it this way to make the apl more readable when you have the option disabled. Felt weird having it in but not casted

scamille: I guess this is the old discussion we also had in the lounge a few times: Should the default APL observe the player state when being built, or should it always be 100% the same so you can edit the player while keeping the already generated APL the same.

That was the intention to this change. I can also just revert this part and keep it the way it is.

seanpeters86 (approved):

PR #5713, opened 2020-12-31
Priest: Split pets into their own file.

scamille:


seanpeters86:
Would you mind running a few quick sims before/after this change to make sure the following still work?
- Shadowflame Prism/Mindbender/Shadowfiend ( `head=confidants_favored_cap,id=183021,ilevel=226,bonus_id=6982`)
- Eternal Call to the Void `head=confidants_favored_cap,id=183021,ilevel=226,bonus_id=6983`

scamille:
All 3 sims (T26_Priest_Shadow.simc and one for each of the head pieces you listed) gives the same deterministic results.
It definitely made sense to double-check this, but I also did not anticipate any problems there. The likeliest problem would have been in the pet expression code.

seanpeters86 (approved):

PR #5723, opened 2021-01-04
[Priest] add damage and healing reversal effects for Mindgames

seanpeters86:
Mindgames has two variable formulas to calculate the healing and damage reversal parts of the spell. By default in SimC we only have `priest_mindgames_damage_insanity` enabled by default.

From this refactor I changed the options to reflect the whole reversal instead, they will create the proper healing and/or damage events based on the reversal options set.

Specifically activating `priest_mindgames_healing_reversal=1` create a child damage action on Mindgames that does the reversal damage. `priest_mindgames_damage_reversal=1` create a healing event that heals the actor. Both of these follow specific formulas for the damage/healing found in the parent spell.

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -559,8 +559,7 @@ struct mindgames_t final : public priest_spell_t
       insanity += insanity_gain;
       if ( !ignore_healing )
       {
-        child_mindgames_healing_reversal->target = player;
-        child_mindgames_healing_reversal->execute();
+        child_mindgames_damage_reversal->execute();
       }
     }
 
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -470,7 +470,7 @@ struct unholy_nova_t final : public priest_spell_t
 // ==========================================================================
 struct mindgames_healing_reversal_t final : public priest_spell_t
 {
-  mindgames_healing_reversal_t( priest_t& p)
+  mindgames_healing_reversal_t( priest_t& p )
     : priest_spell_t( "mindgames_healing_reversal", p, p.covenant.mindgames_healing_reversal )
   {
     background        = true;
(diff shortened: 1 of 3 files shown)


engine/class_modules/priest/sc_priest.cpp
```diff
@@ -468,13 +468,43 @@ struct unholy_nova_t final : public priest_spell_t
 // ==========================================================================
 // Mindgames - Venthyr Covenant
 // ==========================================================================
+struct mindgames_healing_reversal_t final : public priest_spell_t
+{
+  const spell_data_t* mindgames_spell;
```
scamille: There is no need to store the spell data when you are only using its value in the constructor to initialize base_dd_min/max.

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -484,6 +514,23 @@ struct mindgames_t final : public priest_spell_t
...
+
+  void trigger_heal()
+  {
+    if ( ignore_healing )
+    {
+      return;
+    }
+    // Formula found in spelldata for $damage
+    // $damage=${($SPS*$s2/100)*(1+$@versadmg)*$m3/100}
+    double amount_to_heal = ( priest().intellect() * ( data().effectN( 2 ).base_value() / 100 ) ) *
```
scamille: As mentioned you could do this with a priest_heal_t, similar to mindgames_healing_reversal_t. But this is probably fine as well.

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -468,13 +468,43 @@ struct unholy_nova_t final : public priest_spell_t
...
+    may_crit          = false;
+    energize_type     = action_energize::NONE;  // disable insanity gain (parent spell covers this)
+    energize_amount   = 0;
+    energize_resource = RESOURCE_NONE;
+
+    spell_power_mod.direct = 1.0;
+
+    // Formula found in spelldata for $healing
+    // $healing=${($SPS*$s7/100)*(1+$@versadmg)*$m3/100}
+    base_dd_min = base_dd_max = ( priest().intellect() * ( mindgames_spell->effectN( 7 ).base_value() / 100 ) ) *
```
scamille: The problem here is that you are snapshotting all those values during the action constructor. Which will be at some point before combat.
The values for intellect and versatility are completely undefined here.

scamille: I am also not really sure what this action intends to do. Isn't the reversal supposed to be the enemy attacking the priest? This looks like you are just adding some more dps for the priest.

scamille (changes requested):

engine/class_modules/priest/sc_priest.cpp
seanpeters86: Going to leave it like this for now just to mimic the other things we have, will revisit more healing things in a future PR

engine/class_modules/priest/sc_priest.cpp
scamille: You also might want to use cache.intellect()

seanpeters86: This was done to mimic what damage meters and logs do in game. If an enemy heals himself you reverse that into damage instead get an extra bit of damage attributed to the priest.

Moving the calculation to impact to solve this.

scamille (approved):

PR #5727, opened 2021-01-05
[Priest] implement unholy transfusion leech healing

seanpeters86:
- implement unholy transfusion leech healing for the priest
- add healing for devouring plague ticks

TODOs:
- verify if unholy transfusion leech healing can proc things like celestial guidance
- implement unholy nova heal

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -245,12 +245,12 @@ struct smite_t final : public priest_spell_t
       double hf_proc_chance = holy_fire_rank2->effectN( 1 ).percent();
       if ( rng().roll( hf_proc_chance ) )
       {
-        sim->print_debug( "{} reset holy fire cooldown, using smite. ", priest() );
+        sim->print_debug( "{} reset holy fire cooldown, using smite.", priest() );
         priest().cooldowns.holy_fire->reset( true );
       }
     }
 
-    sim->print_debug( "{} checking for Apotheosis buff and Light of the Naaru talent. ", priest() );
+    sim->print_debug( "{} checking for Apotheosis buff and Light of the Naaru talent.", priest() );
     auto cooldown_base_reduction = -timespan_t::from_seconds( holy_word_chastise->effectN( 2 ).base_value() );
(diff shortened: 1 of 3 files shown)


engine/class_modules/priest/sc_priest.hpp
```diff
@@ -849,20 +852,29 @@ struct priest_spell_t : public priest_action_t<spell_t>
...
-      if ( priest().specialization() == PRIEST_SHADOW &&
-           ( s->result_type == result_amount_type::DMG_DIRECT || priest().bugs ) && s->result_amount > 0 )
+      if ( s->result_amount > 0 )
       {
         const priest_td_t* td = find_td( s->target );
-        if ( td && td->buffs.wrathful_faerie->check() )
+        // Wrathful Faerie works for any direct attacks by anyone, bugging this for now
+        // TODO: maybe rework this to just be a buff that gives insanity every tick instead?
+        // https://github.com/SimCMinMax/WoW-BugTracker/issues/777
+        if ( priest().specialization() == PRIEST_SHADOW &&
```
scamille: is there a particular reason to check the spec here? I would assume that it is really the problem of the buffs you are checking whether they can be triggered in the first place, and not of the code here.

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -431,6 +431,33 @@ struct unholy_transfusion_t final : public priest_spell_t
...
+  bool ignore_healing;
+
+  unholy_transfusion_healing_t( priest_t& p )
+    : priest_heal_t( "unholy_transfusion_healing", p,
+                     p.covenant.unholy_nova->effectN( 2 ).trigger()->effectN( 2 ).trigger() ),
+      ignore_healing( p.options.priest_ignore_healing )
+  {
+    background             = true;
+    harmful                = false;
+    spell_power_mod.direct = data().effectN( 1 ).sp_coeff();
```
scamille: Did you check that this did not get picked up automatically?

engine/class_modules/priest/sc_priest.hpp
seanpeters86: i think it was a matter of not wanting to impact performance of disc/holy sims at the time. But not sure that really matters

PR #5728, opened 2021-01-05
Priest: Refactor option variables and use priest. prefix

scamille:
Makes options consistent with other modules (hunter, mage and others who have started switching to class name prefix), as well as with priest apl options (priest.self_power_infusion, priest.fiend, etc.)

Remove the priest_ prefix from variable names, since there really is no point of doing that. They are inside a option struct inside priest_t already.

cc @seriallos

### DIFF
--- a/engine/class_modules/priest/sc_priest.cpp
+++ b/engine/class_modules/priest/sc_priest.cpp
@@ -8,6 +8,7 @@
 
 #include "class_modules/apl/apl_priest.hpp"
 #include "sc_enums.hpp"
+#include "sim/sc_option.hpp"
 #include "tcb/span.hpp"
 
 #include "simulationcraft.hpp"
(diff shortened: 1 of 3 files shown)


engine/class_modules/priest/sc_priest.cpp
```diff
@@ -1740,18 +1741,30 @@ void priest_t::create_options()
...
+  add_option( opt_deprecated( "priest_use_ascended_eruption", "priest.use_ascended_eruption" ) );
+  add_option( opt_deprecated( "priest_mindgames_healing_reversal", "priest.mindgames_healing_reversal" ) );
+  add_option( opt_deprecated( "priest_mindgames_damage_reversal", "priest.mindgames_damage_reversal" ) );
+  add_option( opt_deprecated( "priest_self_power_infusion", "priest.self_power_infusion" ) );
+  add_option( opt_deprecated( "priest_self_benevolent_faerie", "priest.self_benevolent_faerie" ) );
   add_option(
-      opt_int( "priest_ascended_eruption_additional_targets", options.priest_ascended_eruption_additional_targets ) );
-  add_option( opt_int( "priest_cauterizing_shadows_allies", options.priest_cauterizing_shadows_allies ) );
+      opt_deprecated( "priest_ascended_eruption_additional_targets", "priest.ascended_eruption_additional_targets" ) );
+  add_option( opt_deprecated( "priest_cauterizing_shadows_allies", "priest.cauterizing_shadows_allies" ) );
```
seanpeters86, line 1755: any reason in particular that you think we need these?

scamille: To yell at people using the old ones. I think it is polite to hint at the new options.

Those can of course be removed at a later time.

seanpeters86 (approved):
lets make sure to update the wiki after this

scamille:
> 
> 
> lets make sure to update the wiki after this

done

PR #5740, opened 2021-01-09
Clang tidy

scamille:


scamille:
@navv1234 @vituscze @nuoHep I wanted to play around with clang-tidy and its various checks. These are some changes that resulted from it.

I started with running the default analyzers, from which I was able to fix some or at least put  stronger asserts in there. There are still some left, so that's something which we might want to revisit.

I then drifted into those various other checks like performance-*, bugprone-* and readability-*. Some have some been auto-fixed, others I adjusted manually.

Let me know if you have strong objections against any of them. I can just revert them if we really don't want them. Most things are not that important, but since most checks can be fixed automatically, I thought it can't hurt, and I wanted to see what was found in our codebase.

I intentionally skipped some checks like readability-redundant-member-init. Others aren't mentioned since they did not find anything.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -1149,46 +1150,44 @@ inline void rune_event_t::execute_event()
...
+
+  debuff.debilitating_malady = make_buff( *this, "debilitating_malady", player->find_spell( 338523 ) )
+                                   ->set_default_value( player->conduits.debilitating_malady.percent() );
+
+  debuff.everfrost = make_buff( *this, "everfrost", player->find_spell( 337989 ) )
+                         ->set_default_value( player->conduits.everfrost.percent() );
+
+  debuff.unholy_blight =
+      make_buff( *this, "unholy_blight_debuff", player->talent.unholy_blight->effectN( 1 ).trigger() )
+          ->set_default_value_from_effect( 2 );
```
Melekus, line 1190: I fail to see how this is an improvement over the previous version. No alignement and inconsistent line breaks.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -7810,70 +7812,71 @@ action_t* death_knight_t::create_action( util::string_view name, const std::stri
 // death_knight_t::create_expression ========================================
 
 // Equipped death knight runeforge expressions
-std::unique_ptr<expr_t> death_knight_t::create_runeforge_expression( util::string_view name, bool warning = false )
+std::unique_ptr<expr_t> death_knight_t::create_runeforge_expression( util::string_view expression_str,
+                                                                     bool warning = false )
```
Melekus, line 7816: The variable was called 'name' for a very good reason : because it is the name of the runeforge.

engine/class_modules/sc_death_knight.cpp
scamille: The main idea was to have the function argument names consistent to its declaration. But I guess relying on auto-formatting after applying those changes was a bad idea to do in the DK module?

That said, having the second argument aligned for the get_dot function - and is definitely not consistent with how make_buff function calls are performed.

Things like mark of blood and apocalypse_famine and are a bit unfortunate, I agree.

PR #5741, opened 2021-01-09
[Priest] convert DP and VT healing to priest_heal_t

seanpeters86:
- remove unused heal on VT impact (Unfurling Darkness dmg does not trigger this heal)
- convert manual VT healing to a priest_heal_t struct
- convert manual DP healing to a priest_heal_t struct

### DIFF
--- a/engine/class_modules/priest/sc_priest.hpp
+++ b/engine/class_modules/priest/sc_priest.hpp
@@ -42,6 +42,8 @@ struct wrathful_faerie_fermata_t;
 struct psychic_link_t;
 struct eternal_call_to_the_void_t;
 struct unholy_transfusion_healing_t;
+struct vampiric_touch_heal_t;
+struct devouring_plague_heal_t;
 }  // namespace spells
 namespace heals
 {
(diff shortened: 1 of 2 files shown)


scamille:
Hmm I wonder if is really worth creating extra heal spells for stuff which just heals for a fixed amount?
Is there truly a separate spell for the healing part ingame, or does the main dot action just do both damage and healing? If it is the same, we might just want to stick with the old implementation.

If we do want to got with the separate heal, it might be worth it to create a base class priest_heal_with_fixed_healing_amount_t or similar, which handles all that extra code of settings multipliers to 1, and causing a heal to be triggered with a specific amount.

seanpeters86:
The main dot does the damage and healing, which is why I had to disable all that stuff specifically for each one. I prefer the new implementation just to be able to see all healing in the same table, rather than have it be split. Also making them actual heal spells gives it the correct chance to proc various things.

I thought about making a generic thing for it, but thought that would end up being more confusing from a maintenance standpoint.

scamille (approved):
Looks good to merge this. I still want to create some shared code for the heal triggered with a fixed value - but that can wait for later.

PR #5745, opened 2021-01-13
Retribution APL Optimizations

keyboardmash:
The changes are mostly for optimizing Execution Sentence, although there are some other small increases as well. Also cleaned up references to BfA systems that are no longer relevant.


scamille:
Thanks for your contribution.
The problem here is that the profile files itself will automatically generated and will be overridden when next generated.
See https://github.com/simulationcraft/simc/blob/shadowlands/CONTRIBUTING.md section profiles and APL on where would be the best place to change the APL. 
https://github.com/simulationcraft/simc/blob/shadowlands/engine/class_modules/paladin/sc_paladin_retribution.cpp#L869 would be the appropriate place I think.

cc @mserrano to discuss how you want to resolve this.

keyboardmash:
Ok, I attempted to make the changes there as well.

engine/class_modules/paladin/sc_paladin_retribution.cpp
```diff
@@ -932,19 +934,20 @@ void paladin_t::generate_action_prio_list_ret()
...
-  cds -> add_action( "use_item,name=dreadfire_vessel" );
-  cds -> add_action( "use_item,name=darkmoon_deck_voracity,if=buff.avenging_wrath.up|buff.crusade.up&buff.crusade.stack>=10" );
-  cds -> add_action( "use_item,name=overwhelming_power_crystal,if=buff.avenging_wrath.up|buff.crusade.up&buff.crusade.stack>=10" );
-  cds -> add_action( "use_item,name=spare_meat_hook" );
-  cds -> add_action( "use_item,name=grim_codex" );
   cds -> add_action( "use_item,name=inscrutable_quantum_device,if=buff.avenging_wrath.up|buff.crusade.up&buff.crusade.stack>=10|cooldown.avenging_wrath.remains>45|cooldown.crusade.remains>45" );
+  cds -> add_action( "use_item,name=overwhelming_power_crystal,if=buff.avenging_wrath.up|buff.crusade.up&buff.crusade.stack>=10" );
+  cds -> add_action( "use_item,name=darkmoon_deck_voracity,if=buff.avenging_wrath.up|buff.crusade.up&buff.crusade.stack>=10" );
+  cds -> add_action( "use_item,name=macabre_sheet_music,if=buff.avenging_wrath.up|buff.crusade.up&buff.crusade.stack>=10" );
+  cds -> add_action( "use_item,name=dreadfire_vessel,if=(!talent.execution_sentence.enabled|debuff.execution_sentence.up|cooldown.execution_sentence.remains>25)" );
```
mserrano: as you've mentioned, ES doesn't affect these; let's get rid of these ES-related changes

engine/class_modules/paladin/sc_paladin_retribution.cpp
```diff
@@ -957,37 +960,33 @@ void paladin_t::generate_action_prio_list_ret()
...
+  generators -> add_action( this, "Wake of Ashes", "if=(holy_power=0|holy_power<=2&(cooldown.blade_of_justice.remains>gcd*2|(debuff.execution_sentence.up|target.time_to_die<8)|debuff.final_reckoning.up))&(!raid_event.adds.exists|raid_event.adds.in>20)&(!talent.execution_sentence.enabled|cooldown.execution_sentence.remains>15)&(!talent.final_reckoning.enabled|cooldown.final_reckoning.remains>15)&(cooldown.avenging_wrath.remains>0|cooldown.crusade.remains>0)" );
+  generators -> add_action( this, "Blade of Justice", "if=holy_power<=3&talent.execution_sentence.enabled&(debuff.execution_sentence.up|cooldown.execution_sentence.remains=0)" );
+  generators -> add_action( this, "Hammer of Wrath" );
```
mserrano, line 980: it's somewhat surprising that just dropping the clash-avoidance and waste-avoidance logic on these lines leads to an improvement. How well does this work across talent setups that don't involve ES, in particular TFoJ/DP or BoW/DP combinations?

keyboardmash: The check for BoJ not being up within 2 GCDs to use Judgment is performing lower for every talent combination I tried including those two, I suspect because the sequence of TV - BoJ - Judge just puts your generators on cooldown slower than Judge - TV - BoJ, even if BoJ is a higher priority generator.

I removed the checks for Holy Power from abilities that generate 1 Holy Power just because they're redundant, since it'll just use a spender instead of a generator above 4 Holy Power anyway. It doesn't make any actual performance difference, I can put them back in if they're there to make it read more intuitively.

engine/class_modules/paladin/sc_paladin_retribution.cpp
keyboardmash: Agreed, I'll remove them.

PR #5746, opened 2021-01-13
Added Gluttonous spike trinket logic

Alenore:


engine/player/unique_gear_shadowlands.cpp
```diff
@@ -764,11 +764,89 @@ void memory_of_past_sins( special_effect_t& effect )
...
+      auto pulse_action = create_proc_action<gluttonous_spike_pulse_t>( "gluttonous_spike_pulse", e );
+      set_tick_callback( [pulse_action]( buff_t* /* buff */, int /* current_tick */, timespan_t /* tick_time */ ) {
+        pulse_action->execute();
+      } );
+    }
+
+    void start( int stacks, double value, timespan_t duration ) override
+    {
+      // Only applies the aura if you get overhealed, according to user configuration
+      if ( rng().roll( player->sim->shadowlands_opts.gluttonous_spike_overheal_chance ) )
```
scamille: I think you could accomplish this through ```set_proc_chance( player->sim->shadowlands_opts.gluttonous_spike_overheal_chance )``` in the constructor?

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -764,11 +764,89 @@ void memory_of_past_sins( special_effect_t& effect )
...
+      {
+        add_child( pulse );
+      }
+    }
+
+    // Being overhealed by the proc applies the 344154 aura, which disables this proc while active
+    void execute() override
+    {
+      // Creates the buff and direct
+      auto buff = buff_t::find(player, "gluttonous_spike");
```
scamille: hmm finding the buff during execute() is expensive. Can you find the buff in the constructor and just store the pointer to it?

engine/player/unique_gear_shadowlands.cpp
Alenore: I suppose you meant set_chance(), it's fixed.

engine/player/unique_gear_shadowlands.cpp
Alenore: Done :)

scamille (approved):

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -764,11 +764,86 @@ void memory_of_past_sins( special_effect_t& effect )
...
+      // If the aura isn't currently running, execute the spell and triggers the buff
+      if (buff->expiration.empty())
+      {
+        proc_spell_t::execute();
+        buff->trigger();
+      }
+    }
+  };
+
+  create_proc_action<gluttonous_spike_pulse_t>( "gluttonous_spike_pulse", effect );
```
gastank: While `create_proc_action` does have safeguards to prevent multiple copies, this line isn't needed since you're also creating `gluttonous_spike_pulse_t` in `gluttonous_spike_buff_t`.

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -764,11 +764,86 @@ void memory_of_past_sins( special_effect_t& effect )
...
+      {
+        add_child( pulse );
+      }
+    }
 
+    // Being overhealed by the proc applies the 344154 aura, which disables this proc while active
+    void execute() override
+    {
+      // If the aura isn't currently running, execute the spell and triggers the buff
+      if (buff->expiration.empty())
```
gastank: If you want to check if a buff is not up, you can just do `!buff->check()`

gastank (approved):

engine/player/unique_gear_shadowlands.cpp
Alenore: That does sound much less hacky.

engine/player/unique_gear_shadowlands.cpp
Alenore: I've tried removing it before, and it isn't present, the add_child part doesn't work.
The pulse action is now created in the gluttonous_spike_t constructor, and passed to the buff through make_buff, which solves the multiple initialisation.

PR #5754, opened 2021-01-16
[Priest] localize heal actions inside parents

seanpeters86:


engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -742,43 +742,47 @@ struct unfurling_darkness_t final : public priest_spell_t
...
+    void trigger( double original_amount )
+    {
+      base_dd_min = base_dd_max = original_amount * data().effectN( 2 ).m_value();
+      execute();
+    }
+  };
 
-struct vampiric_touch_t final : public priest_spell_t
-{
+  vampiric_touch_heal_t* vampiric_touch_heal;
```
scamille: you could add the propagate_const wrapper here as well, and just initialize the heal in the initializer list instead of in the constructor body.
Same for DP.

scamille (approved):

PR #5755, opened 2021-01-17
[Priest] fix boon usage with AM and Ordnance

seanpeters86:


Saeldur:
LGTM, thanks for commiting this.

PR #5756, opened 2021-01-17
[Dockerfile] Change to release build, so march=native isn't set

Armadk:
If we want to have a compatible container, it should be built with x86-64 so that it runs on most machines.  optimized adds in -march=native so whatever the build machine cpu is is at least haswell or newer.  I had an old sandy bridge linux machine that I managed to reproduce the issue on.

Either that or a min processor age should be decided on, and march set to that instruction set.

@Bloodmallet

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,10 +40,10 @@ RUN apk --no-cache add --virtual build_dependencies \
     make && \
     if [ $NONETWORKING -eq 0 ] ; then \
         echo "Building networking version" && \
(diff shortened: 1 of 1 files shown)


scamille:
ooooh I haven't seen your PR when quickly doing 181915e872e7f75748cf9e1086cc58aa46eae911

But the content here is probably still worth implementing on top of that.

-mtune=generic seems like a sensible choice.

I am not sure if addin -march=x86-64 makes sense here. The dockerfile could be used on any other architecture (x86, arm, etc.). 
Assuming that at least the dockerhub image is tied to x86-64, it could be added to the CI job in simc-publish.

Armadk:
Yeah, I hadn't actually considered arm etc.  It's prob best to just leave it out.  So that the file can be used to build all archs.

PR #5758, opened 2021-01-18
[Soulbind] Implement Thrill Seeker killing blow stack increase

seanpeters86:
Adds a "pseudo" killing blow proc for Thrill Seeker by just checking target demises in the sim. This is then wrapped around an option to set the chance of you getting that killing blow or not with:

```
shadowlands.thrill_seeker_killing_blow_chance=0.25
```

By default I chose `0.05` to mimic a raid environment when there is a low chance you will get KBs on any adds. For DungeonSlice we alter this to be `0.25` since there is a higher chance you can get this more often. Open to suggestions on the default though :) Since it has no impact on typical sims I figured this was okay, most folks will encounter this in the DungeonSlice sim anyways.


engine/player/soulbinds.cpp
```diff
@@ -453,18 +453,33 @@ void thrill_seeker( special_effect_t& effect )
     make_repeating_event( *p->sim, eff_data->period(), [ counter_buff ] { counter_buff->trigger(); } );
   } );
 
-  auto p                  = effect.player;
-  int killing_blow_stacks = as<int>( p->find_spell( 331586 )->effectN( 1 ).base_value() );
+  auto p                     = effect.player;
+  int killing_blow_stacks    = as<int>( p->find_spell( 331586 )->effectN( 1 ).base_value() );
+  double killing_blow_chance = p->sim->shadowlands_opts.thrill_seeker_killing_blow_chance;
+  // If the user does not override the value for this we will set different defaults based on the sim here
+  // Default: 1/20 = 0.05
```
scamille, line 461: why not give the values 4 and 20 a name? num_players or something. And then just set killing_blow_chance = 1.0 / num_players;
Makes your code self-documenting and easier to understand.

Also why 1/4 for dungeon slice instead of 1/5, but still 1/20 for a raid?

seanpeters86: I think in both of these cases the reasons are because they arent directly tied to players. Dungeons should have a significantly higher chance, I just chose 4 as a conservative default. As for the 1/20 I think the same applies there, don't want to assume mythic raid, could easily have 20 dps with a 30 man heroic roster or something.

The comment here is probably doing more harm than good in this case. I think the better way to just look at it is the percents rather than how we randomly decided on these two figures :)

engine/player/soulbinds.cpp
```diff
@@ -449,12 +448,45 @@ void thrill_seeker( special_effect_t& effect )
...
+  {
+    if ( effect.player->sim->fight_style == "DungeonSlice" )
+    {
+      number_of_players = 4;
+    }
+    else
+    {
+      number_of_players = 20;
+    }
+    killing_blow_chance = 1 / number_of_players;
```
scamille:
```suggestion
    killing_blow_chance = 1.0 / number_of_players;
```

scamille (changes requested):

PR #5762, opened 2021-01-21
More Idiomatic Python in ConvertAPL.py

WillAyd:
These are simple stylistic updates to improve the readability of the Python code by reducing levels of indentation required and removing arguably unnecessary exception handling.

By default, this will provide a full traceback now to stderr if things go awry. Previously the script would supply 2 as an exit code but the default exception handling will now change that to 1. I am assuming that the actual return code is irrelevant as long as it is non-zero

### DIFF
--- a/engine/class_modules/apl/ConvertAPL.py
+++ b/engine/class_modules/apl/ConvertAPL.py
@@ -28,38 +28,32 @@ def read_apl(inputFilePath):
     commentString = ""
     aplList = []
     subaplList = ["default"]
-    try:
-        inputFile = open(inputFilePath, 'r')
-    except IOError as e:
-        print('Error opening input file with path: ' + inputFilePath)
-        print(e)
-        sys.exit(2)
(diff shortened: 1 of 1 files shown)


engine/class_modules/apl/ConvertAPL.py
```diff
@@ -119,41 +113,28 @@ def main(argv):
...
     subaplList, aplList = read_apl(inputFilePath)
-    try:
-        outputFile = open(outputFilePath, 'r')
-    except IOError as e:
-        print('Error opening output file with path: ' + outputFilePath)
-        print(e)
-        sys.exit(2)
-    else:
-        try:
-            tempFile = open('tempFile.txt', 'w')
```
WillAyd, line 130: The Python stdlib has a dedicated `tempfile` module that would be more suitable here; leaving to a separate PR

AlexanderKenny (approved):

PR #5770, opened 2021-01-24
Add SpellID into gains/loses formats

MichelPate:
Add SpellID information for gains and loses outputs

### DIFF
--- a/engine/buff/sc_buff.cpp
+++ b/engine/buff/sc_buff.cpp
@@ -2399,7 +2399,7 @@ void buff_t::aura_loss()
     {
       if ( sim->log_spell_id )
       {
-        sim->print_log( "{} loses {} ({})", *player, name_str, s_data->id() );
+        sim->print_log( "{} loses {}", *player, *this );
       }
       else
       {
--- a/engine/buff/sc_buff.cpp
+++ b/engine/buff/sc_buff.cpp
@@ -2374,14 +2374,8 @@ void buff_t::aura_gain()
     {
       if ( !player->is_sleeping() )
       {
-        if ( sim->log_spell_id )
-        {
-          sim->print_log( "{} gains {} ({}) (value={})", *player, buff_display_name, s_data->id(), current_value );
-        }
-        else
-        {
-          sim->print_log( "{} gains {} (value={})", *player, buff_display_name, current_value );
-        }
+        std::string buff_display_name = fmt::format("{} (stacks={})", *this, current_stack );
+        sim->print_log( "{} gains {} (value={})", *player, buff_display_name, current_value );
       }
     }
     else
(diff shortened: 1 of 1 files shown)


engine/buff/sc_buff.cpp
```diff
@@ -2390,7 +2397,14 @@ void buff_t::aura_loss()
   {
     if ( !player->is_sleeping() )
     {
-      sim->print_log( "{} loses {}", *player, name_str );
+      if ( sim->log_spell_id )
+      {
+        sim->print_log( "{} loses {} ({})", *player, buff_display_name, s_data->id() );
```
Melekus:
```suggestion
        sim->print_log( "{} loses {} ({})", *player, name_str, s_data->id() );
```
This should be name_str to mirror the display without log_spell_id

scamille:
It swould be easier to log just ```*this``` instead of ```name_str``` for the buff, and then you get the automatic output I put in place.

So just do:
```sim->print_log( "{} loses {}", *player, *this );```
should print 'Player xy loses Buff b' or 'Player xy loses Buff b (buff_id)'.

File: engine/buff/sc_buff.cpp [hunk 2]
```diff
@@ -2390,7 +2397,14 @@ void buff_t::aura_loss()
...
     {
-      sim->print_log( "{} loses {}", *player, name_str );
+      if ( sim->log_spell_id )
+      {
+        sim->print_log( "{} loses {}", *player, *this );
+      }
+      else
+      {
+        sim->print_log( "{} loses {}", *player, name_str );
+      }
```
scamille:
```suggestion
      sim->print_log( "{} loses {}", *player, *this );
```

engine/buff/sc_buff.cpp
```diff
@@ -2374,7 +2374,14 @@ void buff_t::aura_gain()
...
       {
-        sim->print_log( "{} gains {} (value={})", *player, buff_display_name, current_value );
+        if ( sim->log_spell_id )
+        {
+          sim->print_log( "{} gains {} ({}) (value={})", *player, buff_display_name, s_data->id(), current_value );
+        }
+        else
+        {
+          sim->print_log( "{} gains {} (value={})", *player, buff_display_name, current_value );
+        }
```
scamille:
```suggestion
        std::string buff_display_name = fmt::format("{} (stacks={})", *this, current_stack );
        sim->print_log( "{} gains {} (value={})", *player, buff_display_name, current_value );
```

This requires changing line 2371 as well.

Should give a line like:
Player X gains Buff Y (spell_id) (current_stack=N) (value=x)

scamille (changes requested):

engine/buff/sc_buff.cpp
scamille: The idea was to replace the whole block with this.

formatting *this (the buff_t object) will automatically print either 'Buff X' or 'Buff X (id)'. So there is no need for the if statement here.

PR #5776, opened 2021-02-02
[Shaman] Updated resto racial APL

Kocki0810:
Removed ascendance checking from the racial usage as resto shaman since it gives no value to dps


scamille:
I am just gonna merge this, apparently there is no active developer around for resto shamans. The changes look fine with the given reason.

PR #5801, opened 2021-02-18
[Death Knight] 9.0.5 PTR Legendary support

Armadk:


engine/class_modules/sc_death_knight.cpp
```diff
@@ -6809,6 +6823,12 @@ struct vampiric_blood_buff_t : public buff_t
     // Cooldown handled by the action
     cooldown -> duration = 0_ms;
     base_buff_duration += player -> spec.vampiric_blood_2 -> effectN( 3 ).time_value();
+    if ( player -> dbc -> ptr )
+    {
+      set_default_value_from_effect( 5 );
+      apply_affecting_aura( player -> legendary.vampiric_aura );
+      add_invalidate( CACHE_LEECH );
+    }
```
Melekus: I wouldn't invalidate leech cache if the legendary isn't equipped/the default value is still 0. It also feels a bit weird to use the buff's default value to store what is just a side-effect from a legendary.

Melekus (approved):
Looks good, could probably use some improvements/cleanup but that can wait for when the changes go live.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4107,17 +4110,28 @@ struct dancing_rune_weapon_buff_t : public buff_t
 
 struct dancing_rune_weapon_t : public death_knight_spell_t
 {
+  int bone_shield_stack_gain;
   dancing_rune_weapon_t( death_knight_t* p, const std::string& options_str ) :
     death_knight_spell_t( "dancing_rune_weapon", p, p -> spec.dancing_rune_weapon )
   {
     may_miss = may_crit = may_dodge = may_parry = harmful = false;
 
+    bone_shield_stack_gain = 0;
```
scamille:
```suggestion
    death_knight_spell_t( "dancing_rune_weapon", p, p -> spec.dancing_rune_weapon ),
    bone_shield_stack_gain( 0 )
  {
    may_miss = may_crit = may_dodge = may_parry = harmful = false;
```

engine/class_modules/sc_death_knight.cpp
```diff
@@ -6846,20 +6866,32 @@ struct vampiric_blood_buff_t : public buff_t
 
 struct vampiric_blood_t : public death_knight_spell_t
 {
+  int gorefiends_domination_energize_amount;
   vampiric_blood_t( death_knight_t* p, const std::string& options_str ) :
     death_knight_spell_t( "vampiric_blood", p, p -> spec.vampiric_blood )
   {
```
scamille:
```suggestion
  int gorefiends_domination_energize_amount;
  vampiric_blood_t( death_knight_t* p, const std::string& options_str ) :
    death_knight_spell_t( "vampiric_blood", p, p -> spec.vampiric_blood ),
    gorefiends_domination_energize_amount( 0 )
  {
```

scamille: I know that this isn't strictly necessary since all usages are inside ptr checks, but in general we try to always initialize variables, at least in class modules.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -8934,6 +8974,11 @@ double death_knight_t::composite_leech() const
     m += buffs.voracious -> data().effectN( 1 ).percent();
   }
 
+  if ( dbc -> ptr && legendary.vampiric_aura.ok() && buffs.vampiric_blood -> up() )
```
scamille:
```suggestion
  if ( dbc -> ptr && legendary.vampiric_aura.ok() && buffs.vampiric_blood -> check() )
```

engine/class_modules/sc_death_knight.cpp
scamille: The number 5 should also bei either retrieved from spell data, or otherwise add a comment.

I agree that misusing the buff value is a bit weird, but I guess it works. Until the normal buff wants to use that value :D

Melekus: The value is retrieved from spelldata, in effect5, via the use of set_default_value_from_effect(). It starts at 0 and is modified by the legendary vampiric aura data.

engine/class_modules/sc_death_knight.cpp
Melekus: Why would you remove uptime tracking here?

engine/class_modules/sc_death_knight.cpp
scamille: Right, I should have read that more carefully.

engine/class_modules/sc_death_knight.cpp
scamille: Because it is in a composite / const function which can be called an arbitrary amount of times.

PR #5847, opened 2021-03-14
Prot Warrior Charge Fix v2

Mwahi:
Was not using Charge on previous update - should be using it now.


Melekus:
The code for heroic charge is messy and misleading, I wouldn't use it in its current state

Melekus (changes requested):
Remove or modify Heroic Charge

PR #5855, opened 2021-03-18
added in soul igniter use for druid

DiggidyDale:
Added in the correct use of the soul igniter trinket so that it won't interfere with other trinkets, basically lifted from the mages


Jundarer:
This should already have been covered by the apl but thank you for pointing out it broke somewhere along the way. Fixed the condition in c999f61

PR #5864, opened 2021-03-25
[Soulbind] Volatile Solvent fixes

Armadk:
Fix affect type for beast (primary stat)
Invalidate caches for giant (physical) and elemental (magic) damage


vituscze (approved):
Looks good

PR #5901, opened 2021-04-23
Storm Elemental update

Eokira24601:
This is the culmination of work done on 'What should I do during Storm Elemental uptime?' full numbers located here: https://drive.google.com/file/d/1kBePrXOdf83npJrzDxg6NBKGl7IHNyop/view 
The gameplay change to focus Lightning Bolt spam was only a gain in specific circumstances, so all lines added are with that in mind. This does not affect Storm Elemental use when anything but EoGS is equipped. 

8642- added Fire Elemental to the precast sequence for regular opener
8659 - replaced  Flame Shock line with :  if not necrolord and not ticking. necrolords will apply FS via PW. Moved line for Flame Shock above Storm Elemental such that builds without Elemental Blast talented (and therefore pre-casting LvB) will still LvB->Flame Shock before using Storm Elemental. (8661)
8732 - line that enables the sim to only cast LvB when any one of the 'trifecta' is not currently true i.e. it will only cast LvBs when bloodlust+SEactive+EoGS equipped is not true and also when surge procs are active
8748 - added line to ensure PW buff is consumed, will delay to fish for lava surge procs but will hardcast to consume appropriately.
8750 - added line to make hardcasting LB more important than regular LvBs when EoGS equipped+SE active+BL up

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -8639,6 +8639,7 @@ void shaman_t::init_action_list_elemental()
   precombat->add_talent( this, "Stormkeeper",
                          "if=talent.stormkeeper.enabled&(raid_event.adds.count<3|raid_event.adds.in>50)",
                          "Use Stormkeeper precombat unless some adds will spawn soon." );
+  precombat->add_action( this, "Fire Elemental"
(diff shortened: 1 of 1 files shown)


Bloodmallet:
@HawkCorrigan is this still being considered? does this require an update? can it be removed?

Bloodmallet:
Closing the PR for now. If it becomes relevant again, please open it again.

PR #5913, opened 2021-04-29
Holy Paladin Stage 1

abelito75:
**This does turn on holy.** It is not completely yet if desired I can turn it back off till 100% done but personally I believe it will be helpful to have it on.
The only things left that I can see directly are Avenging Crusades's CDR aspect (its more like IT than cooldown reduction) and glimmer of light. 
Holy shocks interactions with judgment has been fixed ( moved where the dummy cast goes to after the actual damage/healing cast).
Added WoG to holy spec for Awakening resets.
Corrected LoD's CD iirc
Fixed holy shock's crit rate passive
Fixed AP calculation for holy paladin (does not include weapon AP)
Fixed Hallowed Discernment (it doesn't get modified by aoe caps. its always full damage)
There is probably a lot more but its been a hot minutes since starting on this so :(

I honestly have no clue why the spacing in sc_paladin.cpp changed. It might have been weird clang on my system but that doesn't make sense... I know it makes the PR much much harder but /shrug idk man. If you have a logical way to remove all the BS changes and fix formatting I'm all ears... sorry for not being smarter xD


Melekus (changes requested):
It would be great if you did a new PR where the clang commit is isolated from actual changes, because at the moment this is fairly hard to read.

Something nice you could also do is add a holy paladin default profile in https://github.com/simulationcraft/simc/blob/shadowlands/profiles/generators/PreRaids/PR_Generate_Paladin.simc , generate it with the generate_profiles.bat script located in the root directory, and then add Holy Paladin to the list of CI checks in https://github.com/simulationcraft/simc/blob/shadowlands/.github/workflows/main.yml#L164

abelito75:
Added the work flow requests (i believe)
Here is the isolated Clang on sc_paladin_holy.cpp and sc_paladin.cpp #5915

scamille:
I am trying to fix the problems with the CI build not working for holy. https://github.com/Bloodmallet/simc_support/pull/10 should hopefully get some more things working.

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -795,6 +813,11 @@ struct paladin_action_t : public Base
       this -> affected_by.reckoning = this -> data().affected_by( p -> spells.reckoning -> effectN( 1 ) );
       this -> affected_by.final_reckoning = this -> data().affected_by( p -> talents.final_reckoning -> effectN( 3 ) );
     }
+    if ( p->specialization() == PALADIN_HOLY )
```
scamille, line 802: There should be no need for such extra specialization checks if the talent is only enabled for holy anyway.

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -917,6 +940,11 @@ struct paladin_action_t : public Base
       am *= 1.0 + p() -> buffs.avenging_wrath -> get_damage_mod();
     }
 
+    if ( affected_by.avenging_crusader && p()->buffs.avenging_crusader->up() )
```
scamille:
```suggestion
    if ( affected_by.avenging_crusader && p()->buffs.avenging_crusader->check() )
```

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -450,16 +455,90 @@ struct lights_hammer_t : public paladin_spell_t
...
+    parse_options( options_str );
+  }
+
+  void execute() override
+  {
+    holy_power_consumer_t::execute();
 
-    cooldown = p->cooldowns.light_of_dawn;
+    // deal with awakening
+    if ( p()->talents.awakening->ok() )
```
scamille: It might make sense to move the awakening trigger into a shared function, eg. in holy_power_consumer_t, paladin_heal_t or paladin_action_t, and then just calling it from these 2 actions.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -184,6 +190,8 @@ struct avenging_wrath_t : public paladin_spell_t
 
     if ( p->talents.crusade->ok() )
       background = true;
+    if ( p->specialization() == PALADIN_HOLY && p->talents.avenging_crusader->ok() )
```
Melekus: The spec check isn't needed here

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2619,6 +2591,12 @@ double paladin_t::composite_melee_attack_power() const
 
 double paladin_t::composite_melee_attack_power_by_type( attack_power_type ap_type ) const
 {
+  // for some reason on the second pass it of AP cacluatotion
```
Melekus: This could probably use a more explanatory comment about holy's AP not scaling with weapon dps

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2632,6 +2610,11 @@ double paladin_t::composite_attack_power_multiplier() const
   if ( specialization() == PALADIN_PROTECTION )
     ap *= 1.0 + cache.mastery() * mastery.divine_bulwark->effectN( 2 ).mastery_value();
 
+  if ( specialization() == PALADIN_HOLY )
```
Melekus, line 2708: Something something comment about how Holy doesn't benefit from AP multipliers because their AP is based on their SP

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -673,6 +677,19 @@ namespace buffs {
...
+
+    double get_damage_mod() const
+    {
+      return damage_modifier;
+    }
+
+  private:
+    double damage_modifier;
+  };
+
```
Melekus: From what I see of AC so far, you don't need to use a specific struct for it.
Making it a simple buff defined in paladin_t::create_buffs_holy() and storing the damage modifier in its default value would be simpler.
You also don't need to use find_spell() for talents, and can instead use find_talent_spell( name )
Most throughput talents should already be extracted in paladin -> talents.talentname here so you can just use that.
Even if you need to evaluate a different data from it (ie. for the recharge multiplier component) you'll be able to call it with paladin -> talents.avenging_crusader -> effectN( 3 ).percent() without much effort.

I think the reason why avenging wrath has all of this behaviours is to avoid repeating multiple calculations and to make it simpler to understand which values increases healing, which increases damage, and which increases crit.

for example.
```cpp
buffs.avenging_crusader = make_buff( this, "avenging_crusader", talents.avenging_crusader )
-> set_default_value_from_effect( 1 );

File: engine/class_modules/paladin/sc_paladin.hpp (hunk 1 above)
Comment (Melekus) [reply engine/class_modules/paladin/sc_paladin.hpp:802]: There's no need for it, but the ret modifiers are also gated behind a spec check and I think it's better for readability.

File: engine/class_modules/paladin/sc_paladin.hpp [hunk 8]
```diff
@@ -917,6 +940,11 @@ struct paladin_action_t : public Base
       am *= 1.0 + p() -> buffs.avenging_wrath -> get_damage_mod();
     }
 
+    if ( affected_by.avenging_crusader && p()->buffs.avenging_crusader->up() )
+    {
+      am *= 1.0 + p() -> buffs.avenging_crusader -> get_damage_mod();
+    }
+
```
Melekus: With the changes I proposed earlier, this would turn into 
```suggestion
    if ( affected_by.avenging_crusader )
    {
      am *= 1.0 + p() -> buffs.avenging_crusader -> check_value();
    }

```

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -307,20 +313,19 @@ struct holy_shock_t : public paladin_spell_t
...
       damage->execute();
     }
     else
     {
       // heal friendly
-      heal->set_target( execute_state->target );
+      heal->set_target( target );
       heal->execute();
     }
+    paladin_spell_t::execute();
```
Melekus, line 325: You should probably comment why the subspells are executed before the main one

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -450,16 +455,90 @@ struct lights_hammer_t : public paladin_spell_t
...
+          main_buff->trigger( 1, buff_t::DEFAULT_VALUE(), -1.0, trigger_duration );
+        }
+      }
+    }
+  }
+};
+
+// Word Of Glory (Holy) =======================================================
+
+struct word_of_glory_t : public holy_power_consumer_t<paladin_heal_t>
```
Melekus: There's already a word_of_glory_t defined in sc_paladin_protection.cpp, it would be better to merge the two and place the resulting struct in sc_paladin.cpp

Melekus: Additionally, you could move the awakening proc to holy_power_consumer_t and gate it behind a bool that defaults to false and is set to true in WoG and LoD constructors. Or keep it the current way, it's your choice and I don't think one is better than the other

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -450,16 +455,90 @@ struct lights_hammer_t : public paladin_spell_t
...
+
+struct avenging_crusader_t : public paladin_spell_t
+{
+  avenging_crusader_t( paladin_t* p, const std::string& options_str )
+    : paladin_spell_t( "avenging_crusader", p, p->talents.avenging_crusader )
+  {
+    parse_options( options_str );
+
+    if ( !( p->talents.avenging_crusader->ok() ) )
+      background = true;
```
Melekus: This is automatically set during spelldata parsing

Melekus (commented):
👍

engine/class_modules/paladin/sc_paladin.hpp
abelito75: Yeah. I was just following what was already there to keep things constant!

engine/class_modules/paladin/sc_paladin.hpp
abelito75: Is there a reason to not do it like the others? Would seem weird to me that crusade/avenging wrath both use `up()` when avenging crusader uses check

abelito75: image
photo to directly show what im talking about

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: It feels weird to me to do that. Since it only is a thing for holy paladin so if it was in holy_power_consumer_t or paladin_action_t it would just needlessly run for every ret and prot sim (the check for awakening even if you pre assign a struct wide bool for it.)

I feel like it being in heal could have the same result but lead the possibility to having a singular healing cast proc awakening multiple times

engine/class_modules/paladin/sc_paladin.cpp
abelito75: Makes sense. Ret doesn't check for it with crusade so I shouldn't either

engine/class_modules/paladin/sc_paladin.cpp
abelito75: Works for me plus what I wrote doesn't make sense in the first place

engine/class_modules/paladin/sc_paladin.cpp
abelito75: 👍

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: Will do boss man

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: Good to know

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: I think the first option is probably the most logical so i'll do that

abelito75: Just to note. Before someone comes to yell at me about `p()->specialization() == PALADIN_HOLY` is not needed
Im going to leave it in there since prot paladin had it in there by default and it should keep some form of readability + the ability to blindly look at it and say `oh this is something for holy. let me not nuke it completely`

engine/class_modules/paladin/sc_paladin.hpp
abelito75: I'll blindly listen to you on this one because u cute

engine/class_modules/paladin/sc_paladin.hpp
abelito75: i dislike removing the up part but i'll test it

abelito75: Tested and works great. Love you

engine/class_modules/paladin/sc_paladin_holy.cpp
scamille: The Idea is to add a function
```
void trigger_awakening();
```
to holy_power_consumer_t, paladin_heal_t or paladin_action_t

Only in word_of_glory_t::execute and light_of_dawn_t::execute are you then actually calling the function.

So the function is not triggered everywhere, the code is just shared in some appropriate base class.

engine/class_modules/paladin/sc_paladin.hpp
scamille: As a general rule you should use check() in const functions and up()/down() in non-const functions if you want to have benefit-tracking (and also only if the benefit tracking affects a single place, otherwise the information you gain from it is somewhat meaningless).

I understand that this might not be practiced in other places as well, it's just that nobody cared to clean it up.

(The more technical explanation for paragraph 1 is that most const action functions can be called an arbitrary number of times through ready() and APL system, so benefit tracking should not be used there)

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: That makes a lot more sense. I thought you were just saying to generically put it in the parent execute.

abelito75: I kind of want to save this for stage 2 as right now its not messing anything up

scamille: Sure no problem. The danger in copy-pasting code is that it quickly gets out of sync and later you might not always remember that the 2 places are actually meant to do the same thing ;)

scamille:
Paladin CI build runs now, and there aren't any errors.

scamille (approved):

abelito75:
Thanks so much!!!

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -10,6 +10,7 @@ struct blessing_of_sacrifice_redirect_t;
 namespace buffs {
                   struct avenging_wrath_buff_t;
                   struct crusade_buff_t;
+                  struct avenging_crusader_buff_t;
```
Melekus: I think you can remove this now

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -219,23 +218,22 @@ struct holy_prism_t : public paladin_spell_t
...
 
   holy_shock_damage_t( paladin_t* p ) : paladin_spell_t( "holy_shock_damage", p, p->find_spell( 25912 ) )
   {
     background = may_crit = true;
     trigger_gcd           = 0_ms;
-
-    // this grabs the 100% base crit bonus from 20473
-    crit_chance_multiplier = p->find_class_spell( "Holy Shock" )->effectN( 1 ).base_value() / 10.0;
+    // this grabs the 30% base crit bonus from 272906
+    crit_chance_boost = p->spec.holy_shock_2->effectN( 1 ).percent();
```
Melekus, line 228: Since holy_shock_2's data and effects are already stored, you don't really have to store it here and can just call it from the composite_crit_chance() overrides. This is only necessary for values that you access with find_spell() and similar functions. (That said your code is perfectly fine, just giving you a heads up if you want to be lazy later)

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -307,20 +305,22 @@ struct holy_shock_t : public paladin_spell_t
...
     else
     {
       // heal friendly
-      heal->set_target( execute_state->target );
+      heal->set_target( target );
       heal->execute();
     }
+    // The reason the subspell is allowed to execute first is so we can consume judgment correctly. 
+    // If this is first then a dummy holy shock is fired off that consumes the judgment debuff 
+    // meaning when we attempt to do damage judgment will not be factored in even when it should be.
```
Melekus, line 322: Might be worth adding that it's consumed because of dumb blizz spelldata that makes every holy shock spell affected by judgment, instead of just affecting the damaging holy shock spell.

engine/class_modules/paladin/sc_paladin.hpp
abelito75: yup

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: I like this since it makes me feel like my code is more optimized (as in what I have)

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: did done

Melekus (approved):

PR #5916, opened 2021-04-30
Raid event name fixes

scamille:


engine/sim/sc_raid_event.cpp
```diff
@@ -244,14 +241,14 @@ struct adds_event_t final : public raid_event_t
         std::string add_name_str;
 
         if ( sim->add_waves > 1 &&
-             name_str == "Add" )  // Only add wave to secondary wave that aren't given manual names.
+             name == "adds" )  // Only add wave to secondary wave that aren't given manual names.
```
EvanMichaels: Will name be able to be populated with "adds" by default in this case? Think it would just be an empty string right? ("adds" being the default type but name would be unset?)

Wonder if we should be doing name( "adds" ) in the constructor as the default name in this case?

scamille: Right, I think I mixed up type and name in my head when looking at how the base constructor is invoked for the add event.

scamille: I changed the check so that it instead the string needs to be empty.

PR #5920, opened 2021-05-09
Holy Paladin Stage 2

abelito75:
There are some current bugs in here like AC such as it just increases the rate of recovery of all abilities by 30% not just judgment and cursader strike (this was me just being lazy and testing stuff but it seems to work.... just need to target those two spells now... maybe i don't need to idk man)
Glimmer has a nice crash when you cast divine toll :)
Glimmer also needs to force a buff/debuff cap of 8 and always remove the oldest one
iirc simc also yells about having two glimmer buffs


engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -385,13 +407,29 @@ struct holy_shock_t : public paladin_spell_t
     return rm;
   }
 
+  void handle_glimmer_targets(player_t* target)
+  {
+    // no matter what we do we check if it already exists and erase it if it does.
+    std::vector<player_t*>::iterator it = std::find( glimmer_targets.begin(), glimmer_targets.end(), target );
```
scamille:
```suggestion
    auto it = range::find( glimmer_targets, target );
```

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -385,13 +407,29 @@ struct holy_shock_t : public paladin_spell_t
...
+    if ( it != glimmer_targets.end() )
+    {
+      glimmer_targets.erase( it );
+    }
+    // You can only have 8 glimmers out at a time
+    if ( glimmer_targets.size() >= 8 )
```
scamille: Since you are adding the glimmer target after this, you would get 9 glimmers in total, I think.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1209,12 +1194,21 @@ struct divine_toll_t : public paladin_spell_t
...
     background = true;
   }
+
+  double action_multiplier() const override 
+  {
+    double am = paladin_spell_t::action_multiplier();
+    // for some god forsaken reason this ability double dips with vers :)
+    // assumption is because it takes a tick of another ability that is already scaled with vers
+    // then runs it through damage code again which applies vers again
+    am *= (1 + p()->composite_damage_versatility());
```
Melekus: Since this is (most likely) bugged behaviour, you could gate it behind a p() -> bugs check.

(Don't worry it's a bool that is enabled by default)

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: Kind of forgot you can be lazy in this god forsaken language... (ty it looks a lot cleaner!)

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: Yeah. Good catch. I remember I had just gotten that done and then got busy at work so I just said fuck it and pushed

engine/class_modules/paladin/sc_paladin.cpp
abelito75: ✔️

abelito75:
This is just basically glimmer and a bug fix. 
The only thing remaining is AC recharge rate on CS and judgment but my brain is too smoll to figure out call backs right now...
but glimmer is much higher of a prio than AC recharge rate so I kind of want to get this out

scamille (approved):

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -842,7 +842,7 @@ struct crusader_strike_t : public paladin_melee_attack_t
 
 struct word_of_glory_t : public holy_power_consumer_t<paladin_heal_t>
 {
-  word_of_glory_t( paladin_t* p, util::string_view options_str )
+  word_of_glory_t( paladin_t* p, const std::string& options_str )
```
scamille, line 845:
```suggestion
  word_of_glory_t( paladin_t* p, util::string_view options_str )
```

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -273,13 +281,37 @@ struct holy_shock_heal_t : public paladin_heal_t
...
   holy_shock_heal_t* heal;
   bool dmg;
 
-  holy_shock_t( paladin_t* p, util::string_view options_str )
+  glimmer_of_light_damage_t* glimmer_damage;
+  glimmer_of_light_heal_t* glimmer_heal;
+
+  std::vector<player_t*> glimmer_targets;
+
+  holy_shock_t( paladin_t* p, const std::string& options_str )
```
scamille, line 314:
```suggestion
  holy_shock_t( paladin_t* p, util::string_view options_str )
```

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -454,7 +554,7 @@ struct lights_hammer_t : public paladin_spell_t
 
 struct light_of_dawn_t : public holy_power_consumer_t<paladin_heal_t>
 {
-  light_of_dawn_t( paladin_t* p, util::string_view options_str )
+  light_of_dawn_t( paladin_t* p, const std::string& options_str )
```
scamille, line 558:
```suggestion
  light_of_dawn_t( paladin_t* p, util::string_view options_str )
```

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -467,28 +567,13 @@ struct light_of_dawn_t : public holy_power_consumer_t<paladin_heal_t>
...
-      }
-    }
+    trigger_awakening();
   }
 };
 
 struct avenging_crusader_t : public paladin_spell_t
 {
-  avenging_crusader_t( paladin_t* p, util::string_view options_str )
+  avenging_crusader_t( paladin_t* p, const std::string& options_str )
```
scamille, line 577:
```suggestion
  avenging_crusader_t( paladin_t* p, util::string_view options_str )
```

abelito75:
Sorry. This PR is taking a while. I've gotten a bit busy but its not forgotten about!!!

scamille:
And I thought Melekus was taking so long :) We can also merge this and you do the "AC recharge rate on CS and judgment" in a new PR, if that makes more sense.

Melekus:
We also discussed some other changes in private, I'll add them as suggestions here for clarity

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -273,13 +281,37 @@ struct holy_shock_heal_t : public paladin_heal_t
...
 {
   holy_shock_damage_t* damage;
   holy_shock_heal_t* heal;
   bool dmg;
 
-  holy_shock_t( paladin_t* p, util::string_view options_str )
+  glimmer_of_light_damage_t* glimmer_damage;
+  glimmer_of_light_heal_t* glimmer_heal;
+
+  std::vector<player_t*> glimmer_targets;
```
Melekus, line 312: Storing glimmer targets in holy_shock_t means you'll have one target list per Holy Shock line.
My suggestion here would be to:
1) Store the array/vector/queue/whatever you want it to be on the paladin_t level
2) Transfer the handle_glimmer() functions to paladin_t as trigger_glimmer() and apply_glimmer() (clearer names imho than handle_glimmer() and handle_glimmer_targets()
3) Inside apply_glimmer, make sure that: a) on glimmer application: add target to the vector and trigger the buff on that target. If the vector is full, remove the oldest target from the vector and expire its buff. b) on glimmer refresh, make sure the position of the target in the vector is updated to reflect the new application
4) Inside the glimmer target data buff's stack_change_callback, remove the target from the vector if the debuff expires naturally.

nb: I don't think stack_change_callback lets you differentiate between manually expired buffs (from apply_glimmer) and normally expired buff, so you may have to make it a specific glimmer_buff_t with an expire_override() override as this function receives the remaining duration (and stack count) on the buff at time of expiration.

abelito75:
Gonna kill this PR till I get time to do it actually

PR #5921, opened 2021-05-10
[Soulbind] Implement Pelagos Better Together

Saeldur:
Implements Better Together and adds shadowlands custom option
better_together_ally, defaulting to true, if this will proc on combat
start.


seanpeters86:
Hm. Maybe we should have the variable instead be buff uptime or something?

seanpeters86 (approved):

PR #5925, opened 2021-05-13
[Warlock] Grimoire Felguard Consistency Update

Vanishoxyact:
There is an issue where if an apl contains pet.grimoire_felguard and the Grimoire: Felguard talent is not selected, an error will ocurr.

Cannot parse expression from 'pet.grimoire_felguard.active': Cannot find pet or pet spawner 'grimoire_felguard'.

The following APL will trigger this error:
https://pastebin.com/DpfdA72t

This is due to how the name of the pet is defined in the pet definition. Changing this to match how the other warlock pets work (i.e. vilefiend) resolves the issue.

### DIFF
--- a/engine/class_modules/warlock/sc_warlock_pets.cpp
+++ b/engine/class_modules/warlock/sc_warlock_pets.cpp
@@ -632,8 +632,8 @@ void felguard_pet_t::queue_ds_felstorm()
   }
 }
 
-grimoire_felguard_pet_t::grimoire_felguard_pet_t( warlock_t* owner, const std::string& name )
-  : warlock_pet_t( owner, name, PET_SERVICE_FELGUARD, true ),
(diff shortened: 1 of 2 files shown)


vituscze:
I'm not seeing any error when I run the linked .simc file with a freshly compiled simc.

Vanishoxyact:
My apologies, you are correct. I was running into this error with a recent nightly version and assumed it wasn't fixed, as it seemed demology specific and there were no nearby changes. I still think the change is valid and makes the pet definitions more consistent, but it seems there isnt an error raised anymore.

Vanishoxyact:
Seems like this commit fixed it, https://github.com/simulationcraft/simc/commit/f7bc7856ce626c85baffa1542dce60bc7d88e47c, guess my version was older than i thought, sorry!

Azevara:
Since the code change is at least still a minor consistency update, will still go ahead and merge this with amended description.

PR #5931, opened 2021-05-16
Expression optimization options

scamille:
- Change expression optimization to work for the first N iterations. That means N iterations get the _analyze_ type expression objects, optimized after each iteration. Only after the last one, they are replaced by the normal expression objects.

- Fix some places where expressions trees where created with analyzing types, but no actual optimization was performed on it. This seems to have been the problem why in the past the whole sim->optimize_expressions wasn't performing as good as expected.
 (pool resource amount_expr and raid event player expr, and moving  target_if_expr, interrupt_if_expr, early_chain_if_expr, cancel_if_expr out of the if_expr if-check)

- Add experimental optimize_expressions_rounds option to run multiple optimize-passes. This does not seem to help much, might not really make sense in hindsight.

- Set number of expression optimization iterations to 2. This gave significant gains both on a rogue as well as a shadow priest which I tested. After that, there wasn't much more improvement.


vituscze:
I haven't looked into the optimize stuff in detail before, so I'm not sure how useful my review is going to be. But for what it's worth, I haven't spotted any obvious problems.

scamille:
Thanks for the feedback. 

This whole expression optimization is mostly Nates work, but I had started playing around with it a while back and now thought it couldn't hurt to just get this work up to date and merged.

From my very unscientific testing, there definitely is something to gain here by enabling it by default, also removing the need for custom option setting like rogues did. There is potential for fine-tuning the settings some more after doing more rigorous measuring, maybe also marking more things as constant (as I did for raid events).

Nate always said he sees even more possibilities to optimize the expressoin tree itself. I haven't really changed anything at all about the original optimization logic itself, and adding more complicated shortcuts is above my paygrade :)

natehieter:
Serge, multiple rounds of sampling+optimization is brilliant!  Also, good catch on the constants.   Given the limitations of our current logical graph optimization, I don't believe running multiple rounds of optimization on the same sampling will make any change.   The most important legs will already be bubbled up to the top.

navv1234:
This just extends the single iteration statistical sampling optimisation we have, right? Should be fine to enable by default, people have been using that for quite a while without any issues, afaik?

The only issues I seem to recall in recent years with the expression optimisation is various constant-related optimisations on more complex expressions (like variable).

scamille:
Currently I set it to optimize on the first 2 iterations, since that gave better performance than just optimizing on the first iteration. So it goes a bit farther than the previous setting with optimize_expressions=true.

As far as I can tell, marking expressions as constant is done very conservatively, there might even be more places where that could be enabled. Running sims with fixed seed/deterministic also doesn't prodcue a different output, so all should be fine.

navv1234:
Yes it's done conservatively, and we have had bugs even in that, I seem to recall having to jump through quite a few hoops to get the variable expression to do something sensible. But yeah, there may be more situations where things could be constantized with some logic.

scamille:
Ok I just did some more deterministic result comparisons just to be safe, looks all good.

scamille:
Some data to backup the number of optimizations. Ran T26_Rogue_Subtlety.simc with optimize_expressions=[0,1,2,3,4,5]. baseline is 0.

```
x  82     8.2908218     8.8924567     8.4228369     8.4622379    0.13065553
+  82     8.0088392     8.5917989     8.1662851     8.1714007    0.11522333
        -0.290837 +/- 0.0377059
        -3.43688% +/- 0.445578%
        (Student's t, pooled s = 0.123181)
*  82      7.692739     8.1399414     7.8121855     7.8433417    0.10204434
        -0.618896 +/- 0.035883
        -7.31362% +/- 0.424036%
        (Student's t, pooled s = 0.117226)
%  82     7.6827664     8.9405645     7.8259185     7.8647335    0.16786821
        -0.597504 +/- 0.0460428
        -7.06083% +/- 0.544097%
        (Student's t, pooled s = 0.150417)
#  82     7.6690339     8.0985639     7.8410282     7.8620481    0.10231799
        -0.60019 +/- 0.0359195
        -7.09257% +/- 0.424468%
        (Student's t, pooled s = 0.117345)
@  82     7.7094243     8.1172593     7.8201179     7.8490281   0.099113471
        -0.61321 +/- 0.035496
        -7.24643% +/- 0.419464%
        (Student's t, pooled s = 0.115962)
```

So about -3.5% with optimize_expressions=1, and -7% with optimize_expressions=2. No big difference after that.

PR #5934, opened 2021-05-18
Expression optimization

scamille:
More work on the expr optimization system.

Most significant part is marking some buff expressions as constant if their default_chance is 0. This somewhat bold assumption should be held together with some additional asserts.

### DIFF
--- a/engine/buff/sc_buff.cpp
+++ b/engine/buff/sc_buff.cpp
@@ -37,7 +37,7 @@ struct buff_expr_t : public expr_t
   double default_value;
 
   buff_expr_t( util::string_view n, util::string_view bn, action_t* a, buff_t* b, double default_ = 0 )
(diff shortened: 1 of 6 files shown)


navv1234 (commented):
Seems ok to me from brief glance.

One thing I was wondering is whether we could make a "default" constant function for the `fn_const_buff_expr_t`? Would clean up the code somewhat if the last parameter would only be needed when the "const check" is something else than default chance based.

scamille:
Yeah good point. I did not want to add another member to the existing buff fn class where it can be avoided, but you are right that this just adds unneeded code and is probably better merged together with a default no-op function.

scamille:
Unfortunately, I cannot. This is too high of C++ template magic for me. It does not want to work, no matter what I try.

PR #5954, opened 2021-05-26
[SV] AP/bop list merge

Thyminde:
+ testing NTA


Thyminde:
small optimizations for the nta playstyle as a whole as well, paired with Birds of Prey

PR #5963, opened 2021-06-06
Dockerfile: Add PGO, drop NoNetworking

scamille:
- Switch to clang
- Switch from LTO to LTO_THIN. The performance difference is only about 1%, and the build time is significantly improved.
- Add Profile guided optimization. Collect data from a simple T26_Raid.simc run, and rebuild. Offers about 28% performance improvement compared to current simulationcraftorg/simc image
- Drop NONETWORKING option. It is just not worth the hassle in the docker file, and since the normal version with curl works fine, there is no strong reason to keep the maintenance burden in the docker file.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -197,19 +197,6 @@ jobs:
           SIMC_ITERATIONS: 2
         run: tests/run.py ${{ matrix.spec }} -tests talent trinket covenant legendary soulbind --max-profiles-to-use 1
 
-  build-docker:
-    name: docker
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-
-      - name: Build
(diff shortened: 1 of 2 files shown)


scamille:
One problem is that this creates an outlier in the current simc CI build pipeline in terms of execution time. Unfortunately offering build options in the docker file (eg. with or without pgo) is a lot of pain, which is why I removed the networking version after consultation with bloodmallet.

Bloodmallet:
Great ideas!

scamille:
I guess with the docker build included in the nightly simc-publish CI, we could safely remove it from the standard simc CI. Not much benefit with the current tests it runs after building anyway.

scamille:
@nuoHep if you don't mind I would remove the docker build from the CI workflow, and only keep it in the Nightly workflow.

PR #5966, opened 2021-06-08
Expression optimization fixes

scamille:


scamille:
I don't fully like what I have changed here, but it does indeed fix the problems reported by @Jundarer . So I will merge this for now and refactor things some more later.

What isn't very nice now is that the and/or reduced expressions lose their token and no longer take part in possible re-arrangement of expressions when they are child to other and/or expressions.

PR #5967, opened 2021-06-09
[Death Knight] Update Rune of Sanguination to match live behavior

Armadk:


engine/class_modules/sc_death_knight.cpp
```diff
@@ -4897,10 +4897,12 @@ struct death_strike_t : public death_knight_melee_attack_t
 
     // 2020-08-23: Seems to only affect main hand death strike, not OH hit, not DRW's spell
     // No spelldata either, just "increase damage based on target's missing health"
-    // Testing shows a linear 1% damage increase for every 1% missing health
+    // Testing shows a linear 1.25% damage increase for every 1% missing health, up to 100% damage increase
```
Melekus, line 4901: Could you add the date of that testing? I know it's easy to track with git but that'd make it simpler since I added a date for my own testing two lines above

PR #5970, opened 2021-06-10
Cpp17

scamille:
Upgrade build system from C++14 to C++17


scamille:
nightly test build: https://github.com/simulationcraft/simc-publish/actions/runs/924277845

scamille:
I also did some deterministic=1 comparison with gcc10 and clang12 between C++14 and C++17 modes for T26_Raid.simc, and they gave equal results across all 4 variants.

PR #5974, opened 2021-06-13
Use charconv for integer parsing

scamille:
Should have done a PR to start with...


scamille:
Turns out this isn't possible with the targeted gcc and clang versions. I thought integer from-chars was already supported, but I didn't check properly.

PR #5978, opened 2021-06-18
Rogue Change Bonespike fractional in APL base

Vaughn1234:
To accommodate for the 9.1 legendary Deathspike. Functionality when not using it remains the same.
Made the same change for lines in the Subtlety and Assassination APL referring directly to 2.75 bonespike fractional charges.

Couldn't get the change done within the other pull request so I made a new one. Hope that's okay!


EvanMichaels:
Looks good!

PR #5979, opened 2021-06-18
[elemental] swap APL

HawkCorrigan:
swap to default to new APL


engine/class_modules/sc_shaman.cpp
```diff
@@ -8764,269 +8761,8 @@ void shaman_t::init_action_list_elemental()
...
-    single_target->add_talent( this, "Icefury", "if=talent.icefury.enabled" );
-    single_target->add_action( this, "Lightning Bolt" );
-    single_target->add_action( this, "Flame Shock", "moving=1,target_if=refreshable" );
-    single_target->add_action( this, "Flame Shock", "moving=1,if=movement.distance>6" );
-    single_target->add_action( this, "Frost Shock", "moving=1", "Frost Shock is our movement filler." );
-  }
-  else if ( options.rotation == ROTATION_NEW ) {
+  if ( options.rotation == ROTATION_STANDARD )
+  {
     action_priority_list_t* single_target    = get_action_priority_list( "single_target" );
```
Bloodmallet: just wondering in general, why don't we move `single_target` and `aoe` up to precombat and def? This way we wouldn't initialize them twice in the code.

PR #5980, opened 2021-06-18
[Gear] Remade shadowgrasp trinket into a pet spawn.

adrian-budau:
Also slightly changed the way it targets enemies since it seems to
auto-update based on player target.

### DIFF
--- a/engine/player/unique_gear_shadowlands.cpp
+++ b/engine/player/unique_gear_shadowlands.cpp
@@ -1730,8 +1730,8 @@ void shadowgrasp_totem( special_effect_t& effect )
   auto buff = buff_t::find( effect.player, "shadowgrasp_totem" );
   if ( !buff ) {
     buff = make_buff<shadowgrasp_totem_buff_t>( effect );
-    effect.custom_buff = buff;
   }
+  effect.custom_buff = buff;
 }
 
 // TODO: Implement healing?
--- a/engine/player/unique_gear_shadowlands.cpp
+++ b/engine/player/unique_gear_shadowlands.cpp
@@ -1622,48 +1622,58 @@ void bloodspattered_scale( special_effect_t& effect )
  */
 void shadowgrasp_totem( special_effect_t& effect )
 {
-  struct shadowgrasp_totem_damage_t : public generic_proc_t
+  struct shadowgrasp_totem_damage_t : public spell_t
   {
-    action_t* parent;
-
-    shadowgrasp_totem_damage_t( const special_effect_t& effect ) :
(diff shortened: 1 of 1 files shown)


EvanMichaels:
Looks nice, this was on the to-do for a while. Do you happen to have some sim comparisons to confirm the numbers look good with the changes compared to previously?

adrian-budau:
Yeah, windwalkers have Hit Combo which increases pet damage (by up to 6%).

This is before:
image

And this is after:
image

The increase is almost 6% (First few hits of the sim only get 2-5% of pet damage increase because hit combo isn't stacked yet).

EvanMichaels:
Cool, just wanted to confirm the executes were the same, that the haste increases were working, etc.

I assume changing it to a pet also implicitly (and more cleanly) fixed the issue I had the workaround for with the target multipliers?

adrian-budau:
Yeah, it benefits from pet target multipliers instead of player target multiplier (hence why I removed that logic).

adrian-budau:
Is there anything more I need to add to this PR?

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -1681,45 +1691,50 @@ void shadowgrasp_totem( special_effect_t& effect )
...
+    {}
+
+    void execute() override
+    {
+      buff->spawner.spawn();
+      buff->trigger();;
+    }
+  };
+
+  effect.execute_action = create_proc_action<shadowgrasp_totem_t>( "shadowgrasp_totem", effect );
```
gastank: You don't need to create an action just to trigger the buff. You can retain the old code with the sanity check for buff already existing and instead override `shadowgrasp_totem_buff_t::trigger()` to spawn the pet then continue with `buff_t::trigger()`

adrian-budau: Yes, sorry. Initially I wanted to make sure the pet was spawning before the buff ticks (in case they ever make the buff tick on spawn, now it doesn't), but doing it like you said works as well, I'm just not familiar with all the simc hierarchy.

adrian-budau:
Removed the extra action, fixed a typo and actually put  the npc id
image

gastank:
> 
> 
> Should I move it outside?

Yeah it was likely an oversight with the initial implementation. Sorry took a while but we were all busy fighting over gromit spawns. If you can move it out we'll get the PR merged.

adrian-budau:
Sorry I didn't notice the reply (too many rares & mounts in Korthia&Maw). Fixed it :-D

PR #5981, opened 2021-06-19
Update sc_warlock_affliction.cpp

Fatherdotter:
Preparing for Chains of Domination. 
1. Added talent requirement for refreshing Shadow Embrace
2. Implemented more optimal shard usage for Malefic Wrath users.
3. Placed restrictions on existing shard usage that should only affect Malefic Wrath users.


Fatherdotter:
Removing the changes to Shadow Embrace to allow the Malefic Wrath changes (which are still relevant in 9.0.5) to be pulled as a separate request.

PR #5995, opened 2021-06-22
Elemental Shaman APL Update

DenamPavel:
Adjusted 3 target behavior for both Necrolord and Non-NL covenants, needs a test run


HawkCorrigan:
included in #5979

PR #6002, opened 2021-06-23
[Death Knight] First pass Dk legendaries 9.1

Armadk:
Wouldn't mind a review on the Insatiable Hunger commit in particular.

### DIFF
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -2064,8 +2064,7 @@ struct ghoul_pet_t : public base_ghoul_pet_t
 
     m *= 1.0 + frenzied_monstrosity -> value();
 
-    // only buff pets, not guardians.  As only unholy ghouls are pets, use rank 2 of raise_dead to gatekeep
-    if ( dk() -> spec.raise_dead_2->ok() && dk() -> buffs.final_sentence->up() )
+    if ( dk() -> buffs.final_sentence->up() )
     {
       m *= 1.0 + dk() -> buffs.final_sentence->stack_value();
     }
--- a/engine/class_modules/sc_death_knight.cpp
+++ b/engine/class_modules/sc_death_knight.cpp
@@ -5659,6 +5659,27 @@ struct hypothermic_presence_t : public death_knight_spell_t
   }
 };
 
+// Insatiable Hunger Legendary ================================================
+struct insatiable_hunger_damage_t : public death_knight_spell_t
+{
+  double insatiable_hunger_rp_multiplier;
+  insatiable_hunger_damage_t( death_knight_t* p ) :
+    death_knight_spell_t( "insatiable_hunger_damage", p, p -> find_spell( 353720 ) ),
+    insatiable_hunger_rp_multiplier ( p -> find_spell( 353729 ) -> effectN( 1 ).percent() )
+  {
+    background = true;
+  }
+
+  double composite_aoe_multiplier( const action_state_t* state ) const override
+  {
+    double cam = death_knight_spell_t::composite_aoe_multiplier( state );
+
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_death_knight.cpp
```diff
@@ -8601,14 +8677,26 @@ void death_knight_t::create_buffs()
...
     -> add_invalidate( CACHE_PLAYER_DAMAGE_MULTIPLIER );
 
+  buffs.final_sentence = make_buff( this, "final_sentence", find_spell( 353823 ) )
+    -> set_default_value_from_effect_type( A_MOD_DAMAGE_PERCENT_DONE )
+    -> set_schools_from_effect( 2 )
+    -> add_invalidate( CACHE_PLAYER_DAMAGE_MULTIPLIER )
+    -> set_stack_change_callback( [ this ]( buff_t*, int old_stacks, int new_stacks)
+          {
+            if ( new_stacks > old_stacks )
+              replenish_rune( as<unsigned int>( find_spell( 353823 ) -> effectN( 1 ).resource( RESOURCE_RUNE ) ), gains.final_sentence );
```
Melekus: Once shackle's spreading works again, you should check whether the rune gen works when you're already at full stacks of the damage buff and move the run gen to the dot spread if necessary.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -9039,6 +9129,11 @@ double death_knight_t::composite_player_multiplier( school_e school ) const
     m *= 1.0 + buffs.frenzied_monstrosity->data().effectN( 2 ).percent();
   }
 
+  if ( buffs.final_sentence->up() )
+  {
+    m *= 1.0 + buffs.final_sentence->stack_value();
+  }
+
```
Melekus: Isn't this redundant with the buff's implementation that already makes it affect A_MOD_DAMAGE_PERCENT_DONE?

Unless you meant to put that just below in the pet multiplier, since it seems to be missing from your implementation.

edit: no it doesn't, there's no automagic for player damage buffs
That said you still have to copy this below for the pet damage mod in effect#3

engine/class_modules/sc_death_knight.cpp
```diff
@@ -6439,6 +6483,16 @@ struct swarming_mist_buff_t : public buff_t
...
   }
+  void expire_override( int expiration_stacks, timespan_t remaining_duration ) override
+  {
+    buff_t::expire_override( expiration_stacks, remaining_duration );
+    death_knight_t* p = debug_cast< death_knight_t* >( player );
+    if ( p -> legendary.insatiable_hunger.ok() )
+    {
+      insatiable_damage -> execute();
+      insatiable_hunger_rp_spent = 0;
+    }
```
Melekus: I'd reset on cast start rather than cast ends, because I'm not sure whether expire_override is called if a buff is cancelled for the end of a simulation (so you may have to reset it there too, which resetting on cast circumvents)

engine/class_modules/sc_death_knight.cpp
Armadk: good catch. Thanks.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -2950,6 +2950,7 @@ struct death_knight_action_t : public Base
         if ( p() -> legendary.final_sentence.ok() )
         {
           p() -> buffs.final_sentence -> trigger();
+          p() -> replenish_rune( as<unsigned int>( p() -> find_spell( 353823 ) -> effectN( 1 ).resource( RESOURCE_RUNE ) ), p() -> gains.final_sentence );
```
Melekus: You should cache the rune amount to be replenished somewhere to avoid calling find_spell() during runtime

Melekus (approved):
Looks good 👍

PR #6006, opened 2021-06-23
[Gear] Implement Shard of Cor

indiche:
**This still needs extensive testing as I'm still finding my way around codebase.**

Notes:

* Coldhearted is the actuall buff the player gets when he procs the shard, but the value of the buff is stored in the Shard spell data.
* Buff doesn't affect pets as it seems 
* Melee hits won't trigger the Shard

### DIFF
--- a/engine/player/unique_gear_shadowlands.cpp
+++ b/engine/player/unique_gear_shadowlands.cpp
@@ -2782,6 +2782,55 @@ void shard_of_dyz( special_effect_t& effect )
   effect.proc_flags2_ = PF2_ALL_HIT | PF2_PERIODIC_DAMAGE;
   new shard_of_dyz_cb_t( effect );
 }
+
+/**Shard of Cor
+ * id=355741 driver Rank 1
+ * id=357034 driver Rank 2 (Ominous)
+ * id=357052 driver Rank 3 (Desolate)
+ * id=357062 driver Rank 4 (Foreboding)
+ * id=357073 driver Rank 5 (Portentous)
+ * id=356364 Coldhearted buff
+ */
+void shard_of_cor( special_effect_t& effect )
+{
+  struct shard_of_cor_cb_t : public dbc_proc_callback_t
+  {
+    std::vector<int> target_list;
+
+    shard_of_cor_cb_t( const special_effect_t& e )
+      : dbc_proc_callback_t( e.player, e ),
+        target_list()
+    {
+    }
+
+    void execute( action_t* a, action_state_t* s ) override
+    {
+      if ( range::contains( target_list, s->target->actor_spawn_index ) )
+        return;
+
+      dbc_proc_callback_t::execute( a, s );
+      target_list.push_back( s->target->actor_spawn_index );
+    }
+
+    void reset() override
+    {
+      dbc_proc_callback_t::reset();
+      target_list.clear();
+    }
+  };
+
+  buff_t* buff = buff_t::find( effect.player, "coldhearted" );
+  if ( !buff )
+  {
+    buff = make_buff<damage_buff_t>( effect.player, "coldhearted", effect.player->find_spell( 356364 ) )
+               ->set_default_value( 0.0001 * effect.driver()->effectN( 1 ).average( effect.player ), 1 )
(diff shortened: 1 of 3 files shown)


engine/player/unique_gear_shadowlands.cpp
```diff
@@ -2782,6 +2782,55 @@ void shard_of_dyz( special_effect_t& effect )
...
+      dbc_proc_callback_t::reset();
+      target_list.clear();
+    }
+  };
+
+  buff_t* buff = buff_t::find( effect.player, "coldhearted" );
+  if ( !buff )
+  {
+    buff = make_buff<damage_buff_t>( effect.player, "coldhearted", effect.player->find_spell( 356364 ) )
+               ->set_default_value( 0.0001 * effect.driver()->effectN( 1 ).average( effect.player ), 1 )
```
seanpeters86, line 2826: is this needed? usually we dont need to hardcode this type of thing

indiche: i think it is since the actuall value of the buff is in the gem spell data, and not on the buff

e.g.
```
Name             : Desolate Shard of Cor (id=357052) [Passive, Hidden] 
School           : Frost
Spell Type       : Magic
Labels           : 1306
Proc Chance      : 101%
Proc Flags       : ....x... x...x... x.x..... ...x.... 
                 : Yellow Melee, Yellow Ranged, Generic Hostile Spell, Magic Hostile Spell, Periodic
Attributes       : ......xx ........ ........ ........   ........ ........ ........ ........ 
                 : ........ ........ ........ ........   ........ ........ ........ ........ 
                 : ........ ........ ........ ........   ........ ........ ........ ........ 
                 : ........ ........ ........ ........   ........ ........ ........ ........ 
                 : ........ ....x... ........ ........   ........ ........ ........ ........ 
                 : ........ ........ ........ ........   ........ ........ ........ ........ 
                 : ........ ........ ........ ........   ........ ........ ........ ......x. 
                 : ........ ........ ........ ........   
                 : Passive (6), Hidden (7)
Effects          :
#1 (id=891298)   : Apply Aura (6) | Dummy (4) | Scaling Class: -1
                   Base Value: 0 | Scaled Value: 375.0001 (coefficient=3.947369) | PvP Coefficient: 1.00000 | Target: Self (1)
Description      : |CFF20ff20Your damage is increased by ${$s1/100}.2% for $356364d after attacking an enemy you have not yet damaged.|R
```

each gem rank has it's own value in the effect, and the 0.0001 multiplier is here to convert it to percentage

the buff itself has no value on it

```
Name             : Coldhearted (id=356364) 
School           : Frost
Spell Type       : Magic
Duration         : 20 seconds
Attributes       : ........ ........ ........ ........   ........ ........ ........ ........ 
                 : ........ ........ ........ ........   ........ ........ ........ ........ 
                 : ........ ........ ........ ........   ........ ........ ........ ........ 
                 : ........ ........ ........ ........   ........ ........ ........ ........ 
                 : ........ ....x... ........ ........   ........ ........ ........ ........ 
                 : ........ ........ ........ ........   ........ ........ ........ ........ 
                 : ........ ........ ........ ........   ........ ........ ........ ........ 
                 : ........ ........ ........ ........   
Effects          :
#1 (id=890186)   : Apply Aura (6) | Modify Damage Done% (79)
                   Base Value: 0 | Scaled Value: 0 | PvP Coefficient: 1.00000 | Misc Value: 0x7f | Target: Self (1)
                   Affected School(s): All
#2 (id=890187)   : Dummy (3)
                   Base Value: 10 | Scaled Value: 10 | PvP Coefficient: 1.00000
Description      : $@spelldesc355741
Tooltip          : Damage done increased by $w1%.
```

seanpeters86 (approved):
LGTM will have @Dorovon review as well

Dorovon (approved):

seanpeters86:
ty @indiche !!

Dorovon:
Oh, one thing that I didn't notice before merging: in this case you don't need to create a damage_buff_t. A buff_t (the default for make_buff) is all that is needed in this case when the effects of the buff are manually applied.

indiche:
I was wondering about that since it seems useful only in cases where you want to access damage_buff_t specific getters and setters. Thanks for letting me know.

PR #6007, opened 2021-06-24
[Gear] Implement Shard of Bek

indiche:
I'm not sure if I'm doing it properly but it seems to be working :)

The Shard seems to activate when the percentage difference is greater than 50 e.g it will activate if the player is at 100% and the target is at 49%. If the player is at 80% and the target is at 49% it won't activate. Once it's active it won't become inactive regardless of the percentage diff as seen here:


dbc_extract3/dbc/generator.py
```diff
@@ -1297,6 +1297,8 @@ class SpellDataGenerator(DataGenerator):
          355829, 356042, 356043, 356046,
          # Banshee's Blight (Sylvanas Dagger) damage spell
          358126,
+         # Exsanguinated (Shard of Bek)
+         356372
```
indiche, line 1301: I'm not sure if this is the right place to put this. Exsanguinated (the debuff the target gets) wasn't in the spell data so I've whitelisted it and regenerated spell data and it was working after that.

I didn't push regenerated spell data since I wasn't sure if it belongs in this PR. I can push it if needed.

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -2831,6 +2831,45 @@ void shard_of_cor( special_effect_t& effect )
...
+      : dbc_proc_callback_t( e.player, e ),
+      debuff_value( 0.0001 * e.driver()->effectN( 1 ).average( e.player ) )
+    {
+    }
+
+    void execute( action_t* a, action_state_t* s ) override
+    {
+      dbc_proc_callback_t::execute( a, s );
+
+      auto health_diff = a->player->health_percentage() - s->target->health_percentage();
```
indiche, line 2861: I was thinking of adding an RNG roll to this at first but as I said in the description the debuff stays active once it is activated regardless of the current difference so it won't fall if the player gets damaged and the difference becomes smaller and RNG doesn't make sense in that scenario.

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -2831,6 +2831,45 @@ void shard_of_cor( special_effect_t& effect )
...
+      if ( health_diff > 50 )
```
Dorovon: The 50% health threshold should be pulled from spell data so that it will automatically update if Blizzard changes it.
```suggestion
    double debuff_value;
    double health_pct_threshold;

    shard_of_bek_cb_t( const special_effect_t& e )
      : dbc_proc_callback_t( e.player, e ),
      debuff_value( 0.0001 * e.driver()->effectN( 1 ).average( e.player ),
      health_pct_threshold( e.driver()->effectN( 2 ).base_value() )
    {
    }

    void execute( action_t* a, action_state_t* s ) override
    {
      dbc_proc_callback_t::execute( a, s );

      auto health_diff = a->player->health_percentage() - s->target->health_percentage();
      if ( health_diff > health_pct_threshold )
```

engine/player/unique_gear_shadowlands.cpp
Dorovon: Yeah, this seems fine. If they change it to fall off, we can add some option to reduce uptime.

dbc_extract3/dbc/generator.py
Dorovon: Yes, this is the correct place for this. The spell data will need to be updated for the implementation to work, so you can go ahead and push it.

Dorovon (changes requested):
Looks mostly good, thanks for implementing this. The only significant part I can see that should be adjusted is using spell data for the 50% health threshold instead of hardcoding it.

dbc_extract3/dbc/generator.py
indiche: Spell data is now up

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -2831,6 +2831,47 @@ void shard_of_cor( special_effect_t& effect )
...
+void shard_of_bek( special_effect_t& effect )
+{
+  struct shard_of_bek_cb_t : public dbc_proc_callback_t
+  {
+    double debuff_value;
+    double health_pct_threshold;
+
+    shard_of_bek_cb_t( const special_effect_t& e )
+      : dbc_proc_callback_t( e.player, e ),
+      debuff_value( 0.0001 * e.driver()->effectN( 1 ).average( e.player ),
```
indiche: Closing paren is missing here, I'll fix it in a moment.

SpellDataDump/spelldatadump_ptr.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/env bash
```
indiche: @Dorovon I hope you don't mind me adding this file. It's PTR version of the `SpellDataDump/spelldatadump.sh`.

Dorovon: This seems like a good idea to add, but just for consistency with the other version I think the shebang line should be the same.
```suggestion
#!/bin/sh
```

indiche: I think that shebang needs to change in `spelldatadump.sh` because `sh` doesn't support arrays. I couldn't get `spelldatadump.sh` to run on my machine, it was reporting `Syntax error: "(" unexpected` with `#!/bin/sh.` `#!/bin/env bash` or `#!/bin/bash` seems to work just fine.

Dorovon: Oh, that's interesting; I hadn't noticed problems but maybe I've just been running it with `bash` without thinking. I guess it makes sense to change them to #!/bin/bash.

navv1234: sh may or may not support arrays, depending on the actual shell it is (distribution dependant, somewhat). Using bash here is probably sensible.

Dorovon (approved):

Dorovon:
@indiche Just FYI, when generating spell data your system seemed to have enabled the execute mode for all of the *.inc and *.txt files. I reverted that to what it was previously before merging.

Thanks again for implementing this.

PR #6008, opened 2021-06-24
[Paladin] legendary updates

OrcinusDrake:


engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1331,6 +1334,8 @@ struct ashen_hallow_t : public paladin_spell_t
     hd_damage   = new hallowed_discernment_tick_t( p );
     damage_tick = new ashen_hallow_tick_t( p, hd_damage );
     heal_tick   = new ashen_hallow_heal_tick_t( p );
+    cooldown    = p -> cooldowns.ashen_hallow;
+    cooldown -> duration = p -> covenant.venthyr->cooldown();
```
Melekus: Do you really need that? this should be automagic'd and get_cooldown() will link to the same object as this action's cooldown as long as the action is called the same name as the argument of get_cooldown()

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -33,6 +33,7 @@ paladin_t::paladin_t( sim_t* sim, util::string_view name, race_e r )
 {
   active_consecration = nullptr;
   active_hallow       = nullptr;
+  active_h_hallow     = nullptr;
```
Melekus: I'd find a more explicit name because I had to dig through the code to find out what it stores. Maybe active_hallow_heal?

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -94,16 +94,35 @@ struct avengers_shield_base_t : public paladin_spell_t
...
+  avengers_shield_dr_t( paladin_t* p ):
+    avengers_shield_base_t( "avengers_shield_divine_resonance", p, "" )
+  {
+    background = true;
+  }
+
+  double action_multiplier() const override
+  {
+    // 2021-06-26 Divine resonance damage is increased by Moment of Glory, but
+    // does not consume the buff.
```
Melekus, line 114: If it's not consumed you'll have to adjust or override avenger's shield's execute code to match that. I suggest checking that the action is background=false before decrementing moment of glory

OrcinusDrake: Divine resonance and divine toll (and ringing clarity for that matter) both inherit from avengers_shield_base_t which does not have the moment of glory consumption
Avenger's shield the active ability is its own subclass of that base with the moment of glory implementation (and other "active ability" stuff)

OrcinusDrake: As a python guy I prefer that sub-classing to the way Shield of the Righteous does it with overloaded functions, but idk which is best practice.

Melekus: Makes sense, I completely overlooked that

PR #6009, opened 2021-06-24
Make Shiv proc combat potency

Vaughn1234:
According to the official 9.1 Patch Notes, Blizzard has fixed the bug that made shiv not proc combat potency.

### DIFF
--- a/engine/class_modules/sc_rogue.cpp
+++ b/engine/class_modules/sc_rogue.cpp
@@ -4125,7 +4125,7 @@ struct shiv_t : public rogue_attack_t
     }
   }
 
-  // 06/24/2021 - Now procs combat potency according to 9.1 Patch Notes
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_rogue.cpp
```diff
@@ -4125,9 +4125,9 @@ struct shiv_t : public rogue_attack_t
     }
   }
 
-  // 1/23/2021 - Does not appear to proc Combat Potency despite being an OH attack
+  // 06/24/2021 - Now procs combat potency according to 9.1 Patch Notes
```
scamille:
```suggestion
  // 2021-06-24 - Now procs combat potency according to 9.1 Patch Notes
```
https://github.com/simulationcraft/simc/wiki/CodingGuidelines#basic-rules

Vaughn1234: fixed it, thank you!

PR #6015, opened 2021-06-25
[Gear] Implement Ebonsoul Vise

indiche:


engine/player/unique_gear_shadowlands.cpp
```diff
@@ -2284,6 +2284,37 @@ void titanic_ocular_gland( special_effect_t& effect )
...
+    }
+
+    timespan_t composite_dot_duration( const action_state_t* state ) const override
+    {
+      auto target = state->target;
+      double bonus_mul = 1.0 - ( target->is_active() ? target->health_percentage() * 0.01 : 0.0 );
+
+      timespan_t new_duration = dot_duration * (1 + max_duration_bonus * bonus_mul);
+
+      return new_duration;
```
indiche, line 2311: Is this the right way to implement dot duration scaling?

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -2284,6 +2284,37 @@ void titanic_ocular_gland( special_effect_t& effect )
   } );
 }
 
+/**Ebonsoul Vise
+ * id=355327 damage debuff
+ * id=357785 crit pickup buff NYI
```
indiche: Crit pickup stuff isn't implemented since I don't know how likely is that someone will model this in the APL. The orb that gives you the buff spawns at the location where target died.

I saw that melee version of this trinket has it implemented assuming the player would pick it up, but ranged would need some kind of rng roll on this I guess.

https://github.com/simulationcraft/simc/blob/shadowlands/engine/player/unique_gear_shadowlands.cpp#L2041-L2057

Also, ranged buff has different spell id than the melee on, and I can't seem to find it in the spell data

engine/player/unique_gear_shadowlands.cpp
vituscze: Yeah, this looks fine

engine/sim/sc_sim.hpp
```diff
@@ -425,6 +425,8 @@ struct sim_t : private sc_thread_t
     timespan_t newfound_resolve_default_delay = 4_s;
     double newfound_resolve_delay_relstddev = 0.2;
     bool enable_rune_words = true;
+    /// Chance that the player will pickup Shredded Soul orb left by Ebonsoul Vise
+    double shredded_soul_pickup_chance = 1.0;
```
vituscze, line 431: There also needs to be a corresponding `add_option` call, otherwise this isn't doing much.

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -2284,6 +2284,58 @@ void titanic_ocular_gland( special_effect_t& effect )
...
+        return;
+
+      dot_t* dot     = actor->get_dot( "ebonsoul_vise", player );
+      bool picked_up = player->rng().roll( player->sim->shadowlands_opts.shredded_soul_pickup_chance );
+
+      // TODO: handle potential movement required to pick up the soul
+      if ( dot->remains() > 0_ms && picked_up )
+        buff->trigger();
+    } );
+  } );
```
vituscze: Slightly better name for the `actor` variable would be something like `target` or `enemy`, `player` vs `actor` can be a bit confusing.

vituscze (approved):

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -2284,6 +2284,58 @@ void titanic_ocular_gland( special_effect_t& effect )
...
+      double bonus_mul = 1.0 - ( target->is_active() ? target->health_percentage() * 0.01 : 0.0 );
+
+      timespan_t new_duration = dot_duration * ( 1 + max_duration_bonus * bonus_mul );
+
+      return new_duration;
+    }
+  };
+
+  player_t* player = effect.player;
+  buff_t* buff     = make_buff<stat_buff_t>( player, "shredded_soul", player->find_spell( 357785 ) )
```
Dorovon: While it probably won't ever matter because I doubt anyone will equip both at once, simc needs to use a unique buff name for each different buff. Tormentor's Rack Fragment has a similar buff with a different spell ID that is already using this name, so this should be given some suffix (e.g., "shredded_soul_ebonsoul_vise").

Dorovon (changes requested):

engine/player/unique_gear_shadowlands.cpp
indiche: Oh ok, good to know. Yeah, Tormentor's Rack Fragment is basically a melee version of this trinket. I've changed it know so it should be fine.

Dorovon (approved):

PR #6018, opened 2021-06-27
[Gear] Implement Shadowed Orb of Torment

indiche:
According to spell data, this trinket should trigger 40 seconds group cooldown, but that isn't how it behaves on PTR at the moment. I assume this will get fixed, and no pre-pull shenanigans will be possible with this trinket since it has a 40 seconds duration and 40 seconds cooldown.

### DIFF
--- a/engine/player/unique_gear_shadowlands.cpp
+++ b/engine/player/unique_gear_shadowlands.cpp
@@ -2336,6 +2336,65 @@ void ebonsoul_vise( special_effect_t& effect )
   effect.execute_action = create_proc_action<ebonsoul_vise_t>( "ebonsoul_vise", effect );
 }
 
+/**Shadowed Orb of Torment
+ * id=355321 driver
+ * id=356326 mastery buff
+ * id=356334 mastery buff value
+ */
+void shadowed_orb_of_torment( special_effect_t& effect )
+{
+  struct tormented_insight_channel_t : public proc_spell_t
+  {
+    buff_t* buff;
+
+    tormented_insight_channel_t( const special_effect_t& e, buff_t* b )
+      : proc_spell_t( "tormented_insight", e.player, e.driver(), e.item ), buff( b )
+    {
+      // Override this so it doesn't trigger self-harm portion of the trinket
+      base_td      = 0;
+      channeled    = true;
+      hasted_ticks = harmful = false;
+    }
+
+    timespan_t tick_time( const action_state_t* ) const override
+    {
+      return base_tick_time;
+    }
+
+    void execute() override
+    {
+      proc_spell_t::execute();
+
+      event_t::cancel( player->readying );
+      player->reset_auto_attacks( data().duration() );
+    }
+
+    void last_tick( dot_t* d ) override
+    {
+      bool was_channeling = player->channeling == this;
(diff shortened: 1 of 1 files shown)


engine/player/unique_gear_shadowlands.cpp
```diff
@@ -2335,6 +2335,66 @@ void ebonsoul_vise( special_effect_t& effect )
...
+    {
+      proc_spell_t::execute();
+
+      event_t::cancel( player->readying );
+      player->reset_auto_attacks( data().duration() );
+    }
+
+    void tick( dot_t* d ) override
+    {
+      buff->trigger( 1, buff_t::DEFAULT_VALUE(), 1.0, buff->buff_duration() );
```
indiche: The self-harm component of the trinket isn't implemented since I see no point in it. I can add it if it's necessary.

Dorovon: This trinket appears to actually apply the buff when the channel completes, based on the number of ticks that occurred. I think you should just be able to reference d->current_tick while in the last_tick method below to get the number of ticks that happened and trigger the buff with the appropriate full duration there.

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -2335,6 +2335,66 @@ void ebonsoul_vise( special_effect_t& effect )
...
+    }
+  };
+
+  buff_t* buff = buff_t::find( effect.player, "tormented_insight" );
+  if ( !buff )
+  {
+    double val = effect.player->find_spell( 356334 )->effectN( 1 ).average( effect.item );
+    buff       = make_buff<stat_buff_t>( effect.player, "tormented_insight", effect.player->find_spell( 356326 ) )
+               ->add_stat( STAT_MASTERY_RATING, val )
+               ->set_refresh_behavior( buff_refresh_behavior::EXTEND );
```
Dorovon: After the change I mentioned above, buff_refresh_behavior::EXTEND is no longer needed.

Dorovon (changes requested):

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -2336,6 +2336,68 @@ void ebonsoul_vise( special_effect_t& effect )
...
+    {
+      proc_spell_t::execute();
+
+      event_t::cancel( player->readying );
+      player->reset_auto_attacks( data().duration() );
+    }
+
+    void tick( dot_t* d ) override
+    {
+      // Override this so it doesn't trigger self-harm portion of the trinket
```
indiche: Is there a better way to do this?

engine/player/unique_gear_shadowlands.cpp
indiche: Right, I've missed that. It should behave correctly now.

engine/player/unique_gear_shadowlands.cpp
vituscze: I'd set the base tick damage to 0 in the constructor.

indiche: That seems a lot cleaner than overriding it with empty implementation.  Thanks :)

Dorovon (approved):

PR #6021, opened 2021-06-27
[Items] Implement Cruciform Veinripper

Saeldur:
TODO: Add unique proc behaviour, check if damage matches ingame, check
dot refreshes matches ingame behaviour.

Currently nothing shows for in game combat logs which makes testing this properly impossible.
Working and seems reasonable in sim results though, possible to merge anyway and do fixes later?


Saeldur:
While this cannot be tracked ingame properly via logs, debuff uptime puts it significantly lower than the expected uptime.
https://www.warcraftlogs.com/reports/fDHYwCmV67RPxJ9Q#fight=last&type=auras&source=17&spells=debuffs&by=target
per Surge dual wielding.

Saeldur:
As current implementation shows DW matching procrate of MH only assumed its supposed to scale tick damage from number of weapons equipped. This needs confirmation still. Previous commit achieves that.

Justw8:
> As current implementation shows DW matching procrate of MH only assumed its supposed to scale tick damage from number of weapons equipped. This needs confirmation still. Previous commit achieves that.

can confirm, proc rate didn't seem to increase but damage was exactly doubled when dual wielding today on PTR.

Saeldur:
Further testing shows procrate is approximately halved normally. If the target is impaired/cc'd/slowed/etc then the procrate returns to its original. 
Expansion options are possibly needed also to support time spent not behind the boss. (Just reduce the RPPM?)

PR #6028, opened 2021-06-29
[Gear] Implement Fine Razorwing Quill

indiche:


vituscze:
Looks like simc's unique effect automagic can handle this trinket just fine, so I just removed the empty function which was disabling the automagic.

By the way, the debuff in your PR isn't doing anything. In simc, DoTs and debuffs are two different things.

PR #6029, opened 2021-06-29
[Warlock] 9.1 trinkets

basdehaan:
used in this sim: https://www.raidbots.com/simbot/report/tcCxDe7nuNaw5fAMDszMc3
using 15 as the timing for the shadowed orb of torment works best as seen here: https://www.raidbots.com/simbot/report/dE6W4GyG1f9eFniChR9Vtr
but it also makes sense. using it at 15 before tyrant its used when vf/gfg are up and will last through the whole tyrant and most of tyrant's soul.


basdehaan:
the interference of damage trinks and buff trinkets you want active for tyrant is shown here:
without the condition: https://www.raidbots.com/simbot/report/jVGfLvcKHmm91ZVwYtCqfZ
with the condition: https://www.raidbots.com/simbot/report/3944duF9Li3UVnTXSyKLaL
it trades a ebonsoul vise use for a ruby use, but since tyrant is most of your damage that will be best in most cases.

PR #6034, opened 2021-06-29
[brm] fix shuffle extension

emallson:
the `refresh_duration` method on a buff returns the duration that it will be refreshed to.
the `refresh` method actually refreshes the buff.

(this is a confusing naming scheme)

this fixes Brewmaster's Shuffle handling to correctly extend the Shuffle buff when a Shuffle-triggering ability is used (previously, it would be applied if missing but would not extend the buff).


emallson:
@Hinalover fix from illuminati from this morning

PR #6067, opened 2021-07-06
Update mage.cpp

Rai0090:
BL Optimization.

1. Attempt to fit 2 stacked barrages while under BL
2. Fit a mini barrage between radiant sparks while under BL


Sergrand1:
I verified the gains, and it looks good. Also appreciate adding comments with the new lines.

PR #6078, opened 2021-07-10
CI: change ubuntu-clang build to Clang 12.

scamille:


nuoHep (approved):
LGTM

scamille:
Just as a note: I removed the version part from job names to make it less effort when upgrading again later. I don't think we really need that distinction there, and if we want to use multiple version it would probably be done through a matrix.

PR #6096, opened 2021-07-16
[Enemy Pause] Default Enemy + BrM updates

emallson:
Per our discussion in ThinkTank @Melekus 

- default enemy now has a 30s pause every minute (simulating a 30 on /
30 off tank cycle)
- default enemy now uses Soulrender Dormazain damage values instead of
Guardian of the First Ones
- default enemy uses Mythic values instead of Heroic. (Mythic Dormazain
< Heroic Guardian fwiw)
- Brewmaster APL now checks the time til the next pause before casting
Niuzao or dumping Purifying Brews.

(also includes changes from the `shadowlands` branch because i'd already pulled that)


emallson:
Not sure why it auto-requested priest review :thinking: possibly the changes from shadowlands. if a priest reads this you can disregard. move along, nothing to see here.

seanpeters86:
Can you merge and rebase shadowlands? Seems that's why we were tagged. Your files changed is massive :)

Hinalover (approved):

PR #6117, opened 2021-08-04
[Paladin] update Seraphim APL and swap T27 profile

seanpeters86:
https://www.raidbots.com/simbot/report/qMx4ne6mjypgH3vYaA4w4K

roughly .6% dps increase


abelito75:
The APL change looks good to me
If we are changing talents then we should swap from Awakening to Sanctified Wrath since the APL straight up doesn't fish for awakening procs and from some simming we did earlier SW is just more dps. 
Other things to look into (which I have been meaning to do) is adding bestow faith into the APL
But as long as Melekus is chill with it easy merge

Melekus:
Seems a bit weird to me to update the T27 holy profile's talent set when it's still a carbon copy of the T26 profile, maybe some more work should be done on it.

seanpeters86:
was just trying to help, should maybe merge this and make an issue with the outstanding things needed for the spec @Melekus @abelito75

abelito75:
Works for me.

PR #6120, opened 2021-08-07
Better frost set gear

FrozenSoviet:
https://www.raidbots.com/simbot/report/5FpiTQkfUFPvBDEVTAKmKw

Updates sin rogue profile to use frost set + dashing scoundrel


seanpeters86:
@FrozenSoviet is this still needed? i see there are conflicts

EvanMichaels:
I believe Mystler already updated the profiles to include the frost gems. Didn't integrate this PR since it wasn't in the correct format with bonus IDs and such. @FrozenSoviet this is helpful but review the guidelines at the top here for the rules about bonus IDs/enchants: https://github.com/simulationcraft/simc/blob/shadowlands/profiles/generators/Tier27/T27_Generate.simc

PR #6151, opened 2021-09-01
[Warlock] Precombat fixes

basdehaan:


engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -1300,16 +1300,17 @@ void warlock_t::apl_precombat()
...
   {
-    precombat->add_action( "demonbolt" );
     //tested different values, even with gfg/vf its better to summon tyrant sooner in the opener
     precombat->add_action( "variable,name=first_tyrant_time,op=set,value=10" );
     precombat->add_action( "use_item,name=shadowed_orb_of_torment" );
+    precombat->add_action( "demonbolt" );
   }
   if ( specialization() == WARLOCK_DESTRUCTION )
   {
     precombat->add_talent( this, "Soul Fire" );
```
Azevara: This Soul Fire line should be removed if moving soul fire below orb

basdehaan: Why is that? Honestly I’m not sure what that line does.

Azevara: It's another way of calling add_action for talent spells but its use is not really necessary anymore iirc.

PR #6155, opened 2021-09-06
Update T27_Generate_Warlock.simc

basdehaan:
the comparisons are in the commit messages


basdehaan:
@Azevara any idea why the necrolord profile isnt showing up on http://simulationcraft.org/reports/T27_Raid.html ?

seanpeters86:
@basdehaan you need to add it here https://github.com/simulationcraft/simc/blob/shadowlands/profiles/T27_Raid.simc#L57-L59

i.e. `T27_Warlock_Demonology_Necrolord.simc`

PR #6156, opened 2021-09-06
[Paladin] apl refactor and profile updates

seanpeters86:
Significant APL improvements for Holy Paladin, especially in AoE scenarios. Also updates the T27 profile to be more or less BiS for ST damage

https://www.raidbots.com/simbot/report/3xkrDRYES61pYi3ci111U8

APL Changes:
- default potion for holy to always be spectral intellect
- move ashen hallow priority above avenging wrath
- split out Shield of the Righteous high and low priority actions
- Add hammer of wrath logic for 2 targets
- add lights hammer conditions for aoe
- add consecration conditions for aoe
- add light of dawn logic for awakening
- add shield of the righteousness in aoe
- adjust crusader strike logic to split priority based on if charges are capped
- add arcane torrent into the apl

Other Changes:
- fix broken switch case for holy paladins using arcane torrent
- update gear set to be better suited for T27
- regen profiles


engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -648,15 +648,35 @@ void paladin_t::generate_action_prio_list_holy_dps()
   cds->add_action( "use_items,if=(buff.avenging_wrath.up)" );
   cds->add_action( "seraphim" );
 
-  priority->add_action( this, "Shield of the Righteous" );
+  priority->add_action( "restore_mana,if=mana<10000", "Assume infinite mana." );
```
abelito75: If I'm reading this correctly 
You're making it so holy paladin has infinite mana which shouldn't be in the base sim.

This should be an advanced option for those who have too many balance druids in their raid

engine/class_modules/paladin/sc_paladin_holy.cpp
```diff
@@ -648,15 +648,35 @@ void paladin_t::generate_action_prio_list_holy_dps()
...
+  priority->add_action( this, "Shield of the Righteous",
+                        "if=buff.avenging_wrath.up|buff.holy_avenger.up|!talent.awakening.enabled",
```
abelito75, line 652: When I was testing on simpler APLs awakening was always a DPS downgrade if you were playing into it IE casting LoD or WoG. 
I'd like to see some sims (which I'll do later today if my build environment isn't destroyed) of the differences of playing into Awakening vs just hitting sotr

abelito75 (changes requested):
Sorry. It took so long for me to get to this. I didn't see it while also swimming in work at work...
Looks mostly good just a few oddities

engine/class_modules/paladin/sc_paladin_holy.cpp
LuxOG: Hi, I wrote the logic behind these changes. You could definetly make this an advanced option,  however I don't know if there's anyone out there actually trying to optimize their hpal dps per unit mana. It's just a matter of adding or dropping haste.

engine/class_modules/paladin/sc_paladin_holy.cpp
LuxOG: Taking the awakening talent over Sanctified Wrath is a dps loss yes. However, if you are already running Awakening (i.e. every single situation besides doing 0 healing going for a parse in raid), it is worth it to fish for awakening up to 5 targets, unless wings or HA is up.

engine/class_modules/paladin/sc_paladin_holy.cpp
seanpeters86: removed this line, you should be able to add or remove haste to see that behavior instead

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: Sorry I haven't been able to do sims work is pounding me but
from my understanding and from older sims even on single target hitting SoTR was more dps than fishing for awakening procs when you have the talent

Which the older sims had a worse APL but I'd still like to see the fact fishing for awakenings is more dps than just saying `heck this talent imma hit sotr on cd`

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: In all reality the only time simc is amazingly accurate for healers is when you are going for that 100% ignore healing and focus on dps playstyle
You can still get critical information but in the grand scheme of things when you ignore mana you don't show a true paladin dps. It would be like ignoring mana for arcane mages

But thanks for removing it :D

engine/class_modules/paladin/sc_paladin_holy.cpp
abelito75: I haven't had time to do sims but in the grand scheme of things if you think its a dps increase 
its a dps increase till proven other wise so i'll just merge it in
Don't want to hold it up toooooooooo long

PR #6157, opened 2021-09-06
[Paladin] support time_to_hpg as Holy and fix bug with Ret

seanpeters86:
Adds support for:
- retribution paladin judgment CD (see comment below)
- holy paladin generic support (holy shock / crusader strike / hammer of wrath)
- arcane torrent generic support

### DIFF
--- a/engine/class_modules/paladin/sc_paladin.cpp
+++ b/engine/class_modules/paladin/sc_paladin.cpp
@@ -3284,6 +3284,8 @@ std::unique_ptr<expr_t> paladin_t::create_expression( util::string_view name_str
     cooldown_t* j_cd;
     cooldown_t* how_cd;
     cooldown_t* wake_cd;
+    cooldown_t* hs_cd;
+    cooldown_t* at_cd;
 
     time_to_hpg_expr_t( util::string_view n, paladin_t& p )
       : paladin_expr_t( n, p ),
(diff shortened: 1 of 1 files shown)


engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -3284,39 +3284,67 @@ std::unique_ptr<expr_t> paladin_t::create_expression( util::string_view name_str
...
+      if ( paladin.specialization() == PALADIN_RETRIBUTION )
+      {
+        if ( boj_cd->remains() < shortest_hpg_time )
+          shortest_hpg_time = boj_cd->remains();
+
+        if ( wake_cd->remains() < shortest_hpg_time )
+          shortest_hpg_time = wake_cd->remains();
+
+        if ( j_cd->remains() < shortest_hpg_time )
+          shortest_hpg_time = j_cd->remains();
```
seanpeters86, line 3331: @OrcinusDrake @Collisionc for some reason judgment was not included here, even though the CD was above and it does generate holy power for ret, so ive added that back in here. Please review and let me know if i should remove this or if it was a bug

seanpeters86: Before:
before-ret


After:
after-ret

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -3284,39 +3284,67 @@ std::unique_ptr<expr_t> paladin_t::create_expression( util::string_view name_str
...
     {
     }
 
     double evaluate() override
     {
-      if ( paladin.specialization() != PALADIN_RETRIBUTION )
+      if ( paladin.specialization() == PALADIN_PROTECTION )
       {
-        paladin.sim->errorf( "\"time_to_hpg\" only supported for Retribution" );
+        paladin.sim->errorf( "\"time_to_hpg\" not supported for Protection" );
```
scamille, line 3306: I think this whole check should be either moved to the expression constructor (throw an exception), or removed completely (by providing a fallback value for non protection paladins).

Having a sim error in a expression during runtime is strange, and it will only trigger once that code is reached, which might be hours into a profileset sim or similar.

seanpeters86: i think thats fair, i just did this to edit the code in place rather than fully refactor how it should be :) would prefer to fix the bug first and then update with a followup commit with the format/style change

mserrano: Yeah, agree, let's do that as a followup

mserrano (approved):

PR #6165, opened 2021-09-23
Update T27_Rogue_Subtlety.simc

mhunk:
Replacing phantom fire with agility to match sharpening stones.


mhunk:
adjusting on new data.

PR #6166, opened 2021-09-23
Update T27_Rogue_Subtlety.simc

mhunk:
Update for phantom fire to have an oil to proc off of.

### DIFF
--- a/profiles/Tier27/T27_Rogue_Subtlety.simc
+++ b/profiles/Tier27/T27_Rogue_Subtlety.simc
@@ -15,7 +15,7 @@ potion=phantom_fire
 flask=spectral_flask_of_power
 food=feast_of_gluttonous_hedonism
 augmentation=veiled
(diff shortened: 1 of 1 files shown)


scamille:
Please note that the files in profiles/ are auto-generated, see https://github.com/simulationcraft/simc/wiki/Participate#profiles

You would need to adjust https://github.com/simulationcraft/simc/blob/shadowlands/profiles/generators/Tier27/T27_Generate_Rogue.simc for most things in the rogue profiles and then auto-generate the profiles from that. For the default temporary enchants you changed, that would even be in the C++ side of the rogue module, here: https://github.com/simulationcraft/simc/blob/1686ce4ad59ab5e31606a3a11daca373fa751276/engine/class_modules/sc_rogue.cpp#L6637

scamille:
@simulationcraft/rogue Does anybody want to integrate this manually?

PR #6169, opened 2021-10-02
Qt 6

scamille:
With QT6.2 supporting WebEngine, it should in theory have everything we need for the Simc GUI.
The idea is to once again get dual support for QT5 and QT6 if possible and not too painful.

Tasks left to do, as far as I can tell right now:
- [x] Adjust qt regular expression usage. https://doc.qt.io/qt-6/qtcore-changes-qt6.html#the-qregularexpression-class
- [x] Retest with QT5.


scamille:
Works fine on my machine as far as I can tell. GUI feels slightly faster, but that might just be imagined. UI scaling on high DPI displays definitely seems to be working better.

https://github.com/simulationcraft/simc-publish/actions/runs/1300078905
builds both qt5 and qt6, so as far as I can tell everything is working fine now. I do plan to publish both versions as nightly, even if that has the potential to cause some confusion. Our past experiences with Qt have shown that there can always be strange bugs on different systems and graphics setup, so it should help to have a fallback to Qt5 for a while.

There is also a Win10 Arm64 cli only build now. I don't know how much usage that will have, but it can't hurt to offer it since there might finally be some increased usage of this architecture. Unfortunately, QtWebEngine does not support the architecture, thus it is cli only. That could have been done without going through qmake, but it is easier from a CI nightly perspective, and who knows maybe there will be GUI support someday.

navv1234:
Did you adjust for windows only? If so, I'll try to find some time to check the mac build process during the week.

scamille:
Yes I do not have a mac system to test it on. It would make sense to test the qt build on it eventually, but it is not really necessary. Then again it can't hurt to do it now since there might be some additional issues we need to resolve, but hopefully there aren't.
There were some slight changes in some file dialogs and standard paths which you can see in the diff, which hopefully did not break anything since I followed the guidelines in the obsolete-messages there.

Nightly build is another thing altogether again, we can just stick with qt5 there for mac.

There is no need to hurry adopting Qt6 in any case. But the only option is to try it out, and it seems to work now for windows. Overall, it was less of a pain than I imagined.

scamille:
macos nightly builds fine from this branch as well.

Ketrel:
I was curious if there was still plans to have Qt5 support as well.  (I didn't want to open an issue on that unless you'd prefer it.)

scamille:
There are no plans to continue supporting it on windows at least. It might still work (on linux as well), but I can't guarantee anything.

The whole gui part of simc is rather minimally maintained. Raidbots offers its own interface and most devs are using the command line version of simc. That also means we try to minimize the effort of maintaining the UI, likely completely scrapping QT5 support at some point in time.

If you are interested in keeping qt5 compatibility and can contribute, that of course changes things :)

Ketrel:
Ah ok, well it'd be for Windows, but I plan to get off of it as soon as I can.

I should probably learn the command line in the mean time.

PR #6179, opened 2021-10-10
Upgrade to fmt 8.0.1

scamille:


scamille:
https://github.com/fmtlib/fmt/issues/2479 is also causing issues when compiling with C++20. But that will only be fixed in 8.0.2

scamille:
@nuoHep I can't really get this upgraded to 8.0.1, so 8.0.0 it is for now.

Upgrading to 8.0.1 breaks something which might be related to our helpers in engine/util/format.hpp, but that template enable-if really is a bit out of my league, and in particular it seems somehow fmt::format is calling back into our util/format through ADL, or something like that.

scamille:
This also doesn't run with ```cd engine && make C++20=1 CXX=g++-10```, but clang++-12 builds fine. Might be related to the issue I mentioned in the previous commit, or just some compiler bug.

scamille:
@nuoHep  I found a hack to circumvent the problem, but now this could benefit even more from your review ;)

PR #6180, opened 2021-10-11
Rapidjson Upgrade

scamille:


scamille:
@navv1234 do you remember doing any other manual modifications to rapidjson? It is a bit hard to see what came with the original and what has been changed.

As far as I can tell, both html and json2 work after my changes, but it is entirely possible that I missed something.

navv1234:
Only remember doing the raw output thing.

scamille:
Ok thanks for your response.

I also just did a compare in notepad++ of json reports built with head against this, and they are more or less equal. So I think we are good here.

PR #6184, opened 2021-10-15
Try out using cmake for windows CI builds

scamille:


scamille:
(The main motivation is to remove any reliance to simc_vs2017.sln and simc_vs2019.sln and eventually kill them off)

I am quite astonished how good cmake integration in VS2019 is (and it might be even better with 2022?). And it even uses ninja to run the build.

Issue #6186, opened 2021-10-27
Add Option to reset cooldowns for pet spawners


Hinalover:
### Is your feature request related to a problem? Please describe.
Monk Fallen Order pet spawners spawn several of the same pets back-to-back and the cooldown of spells do not get reset upon despawn. So I have to manually force the cooldowns of spells to certain amounts so that they better align with how many casts are done in game.

However due to the fact that enemy summons also use pet spawners, this would need to be a boolean option.

### Describe the solution you'd like
Because this would be a boolean option, defaulting the option to false would keep the current implementation intact. However if the option is true, it would reset all spell cooldowns for the pet.

scamille:
Why would resetting the cooldowns for all pets on dismiss be a problem for enemy pets? I think it might be a good thing to have clean cooldown state when a pet gets dismissed in general.

> So I have to manually force the cooldowns of spells to certain amounts so that they better align with how many casts are done in game.

Does that mean you already have some code in place for this? Can you add a link?

scamille:
Or I guess to put it more generally: Currently we do not reset cooldowns on pet::dismiss or player::demise, as far as I can tell.
That doesn't really impact players since they usually survive throughout the whole fight, but it might affect all pets, including summonable main pets as well as the triggered ones which usually use a pet spawner implementation nowadays.

Of course pet spawner implementations usually reuse the pet objects, which is one of the advantages of using it for triggered pets.

But to begin with, an answer to the following question would be interesting: If you use a long (eg. 30s) cooldown on a main pet, dismiss and summon it again, is the cooldown reset or not?

If the cooldown is reset, we could just reset cooldowns in general on pet demise, otherwise a configurable flag by default for pets that are reused would make more sense.

Hinalover:
> Does that mean you already have some code in place for this? Can you add a link?

Ok....I'll use Fallen Monk's Fist of Fury as an example:

https://www.wowhead.com/spell=330898/fists-of-fury

Here it shows that the spell has a 24 second cooldown. When a Monk summons their Fallen Order; 4 Fallen Order lasting 8 seconds each of the Summoned Spec get staggered summoned and (on live) 4 staggered random between the other two specs for 6 seconds (PTR standardize it as 2 of each). So if a Windwalker Monk summons Fallen Order, 4 Windwalker Fallen Monks get summoned, and 4 are randomly selected from Brewmaster and Mistweaver Fallen Order.

[...]

PR #6189, opened 2021-10-30
Reset dynamic pet cooldowns on demise

scamille:
To avoid reused pets (eg pet spawner wih dynamic pets) get their pets without reset cooldown, do so on demise. See issue #6186

Remove monk fallen order pet cd adjustment, since that was the case that initiated this change.

### DIFF
--- a/engine/class_modules/monk/sc_monk_pets.cpp
+++ b/engine/class_modules/monk/sc_monk_pets.cpp
@@ -1645,14 +1645,6 @@ public:
       base_tick_time = dot_duration / 4;
       may_crit = may_miss = may_block = may_dodge = may_parry = callbacks = false;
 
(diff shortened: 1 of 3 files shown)


Hinalover:
That looks good to me.

Hinalover (approved):

scamille:
@vituscze @navv1234 Do you see any problems with this solution?

navv1234:
Looks OK the me. Mostly a question on whether you want to overload arise or demise, I suppose, but the same principle holds?

vituscze:
`arise` vs `demise` is probably going to be indistinguishable in the sim. If someone queries dead dynamic pet's cooldown it should probably say it's ready? I'm not even sure if the APL can do that. Resetting on demise might be slightly more efficient since it means we don't have to keep making recharge events for the few pets that have CDs with charges.

PR #6190, opened 2021-10-30
Try out clang-tidy CI run

scamille:


scamille:
Well I think would need to be split into some scheduled run for the complete static analysis, and something more selective for pull requests, maybe only on files changed.

I'll continue playing around :-)

PR #6191, opened 2021-10-30
Refactor dot_behavior to split DOT_REFRESH into DOT_REFRESH_DURATION …

EvanMichaels:
…and DOT_REFRESH_PANDEMIC

* Based on spell flag dumps, it appears that SX_REFRESH_EXTENDS_DURATION = 436u spell flag controls pandemic vs. standard refresh behavior for DoTs currently
* As Blizzard has started using this distinction more frequently in Shadowlands (there were currently 7 items in unique_gear_shadowlands.cpp implementing custom duration + tick calculate_dot_refresh_duration() functions)
* Split out DOT_REFRESH into DOT_REFRESH_DURATION and DOT_REFRESH_PANDEMIC and use the spell flag to set DOT_REFRESH_PANDEMIC when parsing spell data. DOT_REFRESH_DURATION appears to be default Blizzard behavior if this flag is not present
* Simplify many places in the code where this was being done manually
* Update the base action_t::calculate_dot_refresh_duration and action_t::dot_refreshable functions to be more robust for all dot_behavior types
* Initial review seems to show no major sim deviations in the T27_Raid.simc report but still needs further verification before merging


EvanMichaels:
Haven't had time to integrate this after initial discussions in Discord, but this is all still relevant without any major adjustments. Had a few other people eyeball the T27_Raid.simc reports and so far we haven't been able to notice any unintended deviations.

engine/action/sc_action.cpp
```diff
@@ -3995,14 +3998,34 @@ void action_t::do_teleport( action_state_t* state )
...
+  {
+    case dot_behavior_e::DOT_REFRESH_PANDEMIC:
+      return std::min( triggered_duration * 0.3, dot->remains() ) + triggered_duration;
+    case dot_behavior_e::DOT_REFRESH_DURATION:
+      return dot->time_to_next_tick() + triggered_duration;
+    case dot_behavior_e::DOT_EXTEND:
+      return dot->remains() + triggered_duration;
+    case dot_behavior_e::DOT_CLIP:
+    default:
+      return triggered_duration;
```
scamille, line 4020: Shouldn't this return timespan zero for DOT_NONE? Or is that already taken care of somewhere else?

engine/dbc/data_enums.hh
```diff
@@ -1283,11 +1284,17 @@ enum spell_attribute : unsigned
...
+  SX_NO_PARRY                  = 248u,
+  SX_NO_MISS                   = 249u,
+  SX_NO_BLOCK                  = 256u,
   SX_TICK_MAY_CRIT             = 265u,
   SX_DURATION_HASTED           = 273u,
+  SX_DOT_HASTED_MELEE          = 278u,
   SX_FIXED_TRAVEL_TIME         = 292u,
   SX_SCALE_ILEVEL              = 354u,
   SX_ALLOW_CLASS_ABILITY_PROCS = 416u,
+  SX_REFRESH_EXTENDS_DURATION  = 436u,
```
scamille, line 1297: Why not encode the word Pandemic into this enum as well?

EvanMichaels: Was just making it roughly match the flag text we standardized on from Blizzard, which in this case is "Periodic Refresh Extends Duration (436)"

engine/action/sc_action.cpp
EvanMichaels: Default case below in L4027 will return dot_refreshable() as false for DOT_NONE so I think this just didn't breakpoint. But I think it's probably worth changing. Wouldn't be timespan::zero though in this function, would be return dot->remains() I believe? I'll add this.

EvanMichaels: Ok should be fixed with https://github.com/simulationcraft/simc/pull/6191/commits/d1ffa15506c980ab74914772e9e927b5c24d3dba

engine/dbc/data_enums.hh
scamille: Ah ok. Well it seems there is no conversion to text output for spell attributes in spell query anyway, so it doesn't really matter. For anyone looking at it in the code base, the connection should be easy to establish.

scamille (approved):

PR #6199, opened 2021-11-05
[Gear] Fix Master Duelist's Chit trinket

### LINKED ISSUE #5930: Trinket "Master Duelist's Chit" not simming
The Trinket [Master Duelist's Chit ](https://www.wowhead.com/item=175730/master-duelists-chit?bonus=6707)'s proc is not being simmed and the proc is not shown as dmg

Full report https://www.raidbots.com/simbot/report/hU5dC6Psu7Nq3naQKCBvUs/simc
Here is an exaple of Details! image
Simc string https://pastebin.com/GMEUmiXw

ThatLing:
First SimC commit, fixes the Master Duelist's Chit trinket: https://www.wowhead.com/item=175730/master-duelists-chit / https://www.wowhead.com/item=181358/master-duelists-chit.

The trinket uses spell [Dueling Form #336219](https://www.wowhead.com/spell=336219/dueling-form), which doesn't seem to be doing much? [Dueling Form #336236](https://www.wowhead.com/spell=336236/dueling-form) appears to be calling [the spell that deals damage](https://www.wowhead.com/spell=336234/duelists-shot). In Details, the spell [Duelist's Strike](https://www.wowhead.com/spell=336222/duelists-strike) is displayed (at least for melee classes), not sure why.

I added a special effect to use the correct spell, let me know if there is a better way to do this.

Fixes #5930

### DIFF
--- a/dbc_extract3/dbc/generator.py
+++ b/dbc_extract3/dbc/generator.py
@@ -1310,8 +1310,8 @@ class SpellDataGenerator(DataGenerator):
          356305,
          # Shredded Soul (Ebonsoul Vise)
          357785,
-         # Dueling Form (Master Duelist's Chit)
-         336236,
+         # Duelist's Shot (Master Duelist's Chit)
+         336234,
          # Nerubian Ambush, Frost-Tinged Carapace Spikes (Relic of the Frozen Wastes)
          355912, 357409,
          # Volatile Detonation (Ticking Sack of Terror)
--- a/dbc_extract3/dbc/generator.py
+++ b/dbc_extract3/dbc/generator.py
@@ -1310,8 +1310,8 @@ class SpellDataGenerator(DataGenerator):
          356305,
          # Shredded Soul (Ebonsoul Vise)
          357785,
-		 # Dueling Form (Master Duelist's Chit)
-		 336236,
+         # Dueling Form (Master Duelist's Chit)
+         336236,
          # Nerubian Ambush, Frost-Tinged Carapace Spikes (Relic of the Frozen Wastes)
          355912, 357409,
          # Volatile Detonation (Ticking Sack of Terror)
--- a/dbc_extract3/dbc/generator.py
+++ b/dbc_extract3/dbc/generator.py
@@ -1310,6 +1310,8 @@ class SpellDataGenerator(DataGenerator):
          356305,
          # Shredded Soul (Ebonsoul Vise)
          357785,
+		 # Dueling Form (Master Duelist's Chit)
+		 336236,
          # Nerubian Ambush, Frost-Tinged Carapace Spikes (Relic of the Frozen Wastes)
          355912, 357409,
          # Volatile Detonation (Ticking Sack of Terror)
--- a/engine/player/unique_gear_shadowlands.cpp
+++ b/engine/player/unique_gear_shadowlands.cpp
@@ -1984,6 +1984,24 @@ void murmurs_in_the_dark( special_effect_t& effect )
   new dbc_proc_callback_t( effect.player, effect );
 }
 
+// id=336219 driver
+// id=336236 unknown use, triggered by driver
+// id=336234 damage spell
+// id=336222 melee damage spell?
+void dueling_form( special_effect_t& effect )
+{
+  struct dueling_form_t : public generic_proc_t
+  {
+    dueling_form_t( const special_effect_t& e ) : generic_proc_t( e, "dueling_form", 336236 )
+    {
+      base_dd_min = e.driver()->effectN( 1 ).min( e.item );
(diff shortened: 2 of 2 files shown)


engine/player/unique_gear_shadowlands.cpp
```diff
@@ -1984,6 +1984,24 @@ void murmurs_in_the_dark( special_effect_t& effect )
...
 
+// id=336219 driver
+// id=336236 unknown use, triggered by driver
+// id=336234 damage spell
+// id=336222 melee damage spell?
+void dueling_form( special_effect_t& effect )
+{
+  struct dueling_form_t : public generic_proc_t
+  {
+    dueling_form_t( const special_effect_t& e ) : generic_proc_t( e, "dueling_form", 336236 )
```
vituscze: Generally, you want to use the spell id of the damage spell here, so that simc correctly picks up spell attributes, spell school, etc

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -1984,6 +1984,24 @@ void murmurs_in_the_dark( special_effect_t& effect )
   new dbc_proc_callback_t( effect.player, effect );
 }
 
+// id=336219 driver
+// id=336236 unknown use, triggered by driver
+// id=336234 damage spell
+// id=336222 melee damage spell?
+void dueling_form( special_effect_t& effect )
+{
+  struct dueling_form_t : public generic_proc_t
```
vituscze: I believe the standard practice (for shadowlands items at least) is to use `proc_spell_t` base here.

vituscze (changes requested):

engine/player/unique_gear_shadowlands.cpp
ThatLing: Wowhead says [Dueling Form #336236](https://www.wowhead.com/spell=336236/dueling-form) uses [Duelist's Shot](https://www.wowhead.com/spell=336234/duelists-shot), however Details seems to display [Duelist's Strike](https://www.wowhead.com/spell=336222/duelists-strike). How do I figure out which one to pick?

engine/player/unique_gear_shadowlands.cpp
ThatLing: Updated

engine/player/unique_gear_shadowlands.cpp
vituscze: Trust the in-game spell id. If you want to be extra sure (in case Details is doing something weird), you can check the spell id in logs produced by `/combatlog`. In-game combat log is also usable if you have some addon that lets you grab ids from spell links (such as idTip).

dbc_extract3/dbc/generator.py
```diff
@@ -1310,6 +1310,8 @@ class SpellDataGenerator(DataGenerator):
          356305,
          # Shredded Soul (Ebonsoul Vise)
          357785,
+		 # Dueling Form (Master Duelist's Chit)
+		 336236,
```
Dorovon:
```suggestion
         357785,
         # Dueling Form (Master Duelist's Chit)
         336236,
```
The simc code style uses spaces for indentation instead of tabs.

engine/player/unique_gear_shadowlands.cpp
ThatLing: After some digging I found [a log](https://www.warcraftlogs.com/reports/3Pg2JLzp6Qq9jnXm#fight=3&type=damage-done&source=21&view=events) that uses the trinket. The spell ID is [Duelist's Shot](https://www.wowhead.com/spell=336234/duelists-shot), so I guess it's correct like this.

ThatLing:
I'm guessing the CI is dying because the generated DBC data isn't updated yet. I saw another PR that didn't update it so I wasn't sure. Let me know if I should include the changes.

engine/player/unique_gear_shadowlands.cpp
Dorovon: Good find, although you should still use 336234 instead of 336236 for the damage spell. Using the exact same spell ID that shows up in a combat log is important for many reasons. A few of the more visible issues that would come up with 336236 would be that the travel time of the projectile would be instant and the tooltip information will be missing in the HTML report.

I also suppose it's possible that this trinket has two separate procs, one for melee and one for ranged, but since they do the same damage I don't think it should matter much which is used. The only difference between Duelist's Strike (id=336222) and Duelist's Shot (id=336234) are the range and velocity fields.

vituscze:
We'll regenerate the spell data once this PR is merged, so just leave it as is.

engine/player/unique_gear_shadowlands.cpp
ThatLing: Changed the spell ID to 336234

PR #6201, opened 2021-11-10
Fix some incorrect hotfix outputs

Dorovon:
For some reason, Blizzard sometimes seems to push hotfix entries with HotfixType.REMOVED that reference records that do not exist. When this happens, the id will not end up in replaced_ids, which would result in the addition of an empty record.

In the current live build (9.1.5.40966), this occurs with a HotfixType.Removed entry for SpellCategory record 1999.

I'm not sure if there are other cases where this issue comes up elsewhere in dbc_extract.

### DIFF
--- a/dbc_extract3/dbc_extract.py
+++ b/dbc_extract3/dbc_extract.py
@@ -208,11 +208,12 @@ elif options.type == 'view':
             else:
                 print('{}'.format(str(record)))
 
-        for id, entry in entries.items():
-            if id in replaced_ids:
+        for id, item in entries.items():
+            entry, hotfix = item
+            if id in replaced_ids or hotfix['state'] == HotfixType.REMOVED:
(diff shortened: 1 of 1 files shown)


scamille:
Does this just have no effect right now, or would it make sense to run dbc extract with those changes so see what changes?

navv1234:
Oh oops, forgot to post something. Yeah the changes look sensible to me, without actually checking the output of the changes.

Dorovon:
> Does this just have no effect right now, or would it make sense to run dbc extract with those changes so see what changes?

I'm unsure if the current live build still has hotfixes like this, but it would have no effect on simc either way. The modified code is just for the view/json/csv options, so mainly just for people using dbc_extract for external projects. I don't think this bug is present elsewhere in dbc_extract.

PR #6206, opened 2021-11-18
Adjust monk buff and target usage to honor constness of context

scamille:
Mainly results in buff->up() replaced with check() where appropriate, and more find_target_data usage.

### DIFF
--- a/engine/class_modules/monk/sc_monk.cpp
+++ b/engine/class_modules/monk/sc_monk.cpp
@@ -114,7 +114,7 @@ public:
     return debug_cast<monk_t*>( ab::player );
   }
 
-  monk_td_t* td( player_t* t )
+  monk_td_t* get_td( player_t* t ) const
   {
     return p()->get_target_data( t );
   }
@@ -466,7 +466,7 @@ public:
 
       // Currently Bountiful Brew cannot be applied if Bonedust Brew is currently active
       // This means that RPPM will have triggered but cannot be applied.
-      if ( !td( p()->target )->debuff.bonedust_brew->up() )
+      if ( !get_td( p()->target )->debuff.bonedust_brew->up() )
       {
         p()->active_actions.bountiful_brew->execute();
         p()->proc.bountiful_brew_proc->occur();
@@ -486,7 +486,7 @@ public:
 
     p()->trigger_empowered_tiger_lightning( s, true, true );
 
-    if ( p()->bugs && td( s->target )->debuff.bonedust_brew->up() )
+    if ( p()->bugs && get_td( s->target )->debuff.bonedust_brew->up() )
       p()->bonedust_brew_assessor( s );
   }
 
@@ -624,7 +624,7 @@ struct monk_heal_t : public monk_action_t<heal_t>
     {
       player_t* t = ( execute_state ) ? execute_state->target : target;
 
-      if ( auto td = find_td( t ) )
+      if ( auto td = this->get_td( t ) ) // Use get_td since we can have a ticking dot without target-data
       {
         if ( td->dots.enveloping_mist->is_ticking() )
         {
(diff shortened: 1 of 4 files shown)


engine/class_modules/monk/sc_monk.cpp
```diff
@@ -608,22 +624,26 @@ struct monk_heal_t : public monk_action_t<heal_t>
     {
       player_t* t = ( execute_state ) ? execute_state->target : target;
 
-      if ( td( t )->dots.enveloping_mist->is_ticking() )
+      if ( auto td = find_td( t ) )
```
vituscze: I'm fairly sure you can end up in a situation where no target data has been created but a DoT is still ticking. That is, application of a DoT does not force the creation of target data for that target.

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -5440,16 +5471,20 @@ struct invoke_xuen_the_white_tiger_buff_t : public monk_buff_t<buff_t>
 
     for ( auto target : p->sim->target_non_sleeping_list )
     {
-      if ( p->get_target_data( target )->debuff.empowered_tiger_lightning->up() )
+      if ( p->find_target_data( target ) )
       {
-        double value = p->get_target_data( target )->debuff.empowered_tiger_lightning->value();
-        p->get_target_data( target )->debuff.empowered_tiger_lightning->current_value = 0;
-        if ( value > 0 )
+        auto td = p->get_target_data( target );
```
vituscze, line 5476: This seems unnecessary. Why not use the result of find_target_data above? Same issue in the code below.

engine/class_modules/monk/sc_monk.cpp
scamille: I would hope that whoever is applying the dot on the target will create the target-data through a get_target_data call.
Thus the assumption that if no target data can be found, there was never a dot applied and it can't be ticking.

Any reason why that should not be the case?

engine/class_modules/monk/sc_monk.cpp
scamille: The reason was that we can't (or shouldn't be able to) reset the current value with a const target data.
I know it looks a bit strange, but if that really is what the code should do I think it is the correct way.

vituscze: Oh right, failed to look at what's actually being done to the target data. Makes sense.

engine/class_modules/monk/sc_monk.cpp
vituscze: DoTs need not be in target data. In fact, most of them aren't. Take a look at action_t::trigger_dot, no target data in sight.

Hinalover (approved):
I'm fine with the changes. I guess my only concern is the fact that I guess I never really understood when to use "check()" and when to use "up()." I know it's related to tracking but I just didn't see the difference between the two. As for the target data part, I'm fine with making adjustments to how I am calling it.

scamille:
I always thought about renaming the buff functions at some point to remove that confusion. Problem is that this might make things even worse for everyone who has internalized what they do, and it would also be best if everyone uses the correct ones even with the existing confusing names.

engine/class_modules/monk/sc_monk.cpp
scamille: Yeah you are absolutely right. I thought they did call player_t::get_target_data at some point, but they don't.

I'll adjust this then.

scamille: I guess the other option would be to let action_t::get_dot create target data as well when initially creating the dot.
Not sure if that would be better worse performance-wise.

The main idea of using find_target_data is that you do not have to create the objects for all player<>target combinations in the first place. (Which isn't that important anymore with profilesets anyway)

vituscze: I would just leave it as is. There are plenty of actors that have dots but no debuffs to create and plenty of situations that trigger dots and don't need target data (such as dots from trinkets on debuff-light classes).

PR #6207, opened 2021-11-18
More usage of string-view based string-split

scamille:


engine/sim/sc_profileset.cpp
```diff
@@ -1205,7 +1205,7 @@ void create_options( sim_t* sim )
     auto split = util::string_split( value, "/:," );
     for ( const auto& v : split )
     {
-      sim -> profileset_output_data.push_back( v );
+      sim -> profileset_output_data.push_back( std::move( v ) );
```
nuoHep, line 1208: This move doesn't actually do anything as `v` is a const ref.
I'd drop the temp `spilt` here and do it all inside a loop: `for ( auto&& v : util::string_split( value, "/:," ) )`

scamille: 1.
```
    for ( auto&& v : util::string_split( value, "/:," ) )
    {
      sim -> profileset_output_data.push_back( std::move( v ) );
    }
```

or 2.
```
    for ( auto v : util::string_split<util::string_view>( value, "/:," ) )
    {
      sim -> profileset_output_data.emplace_back( v  );
    }
```

PR #6211, opened 2021-11-20
Assorted updates for C++17

nuoHep:
The main thing here is replacing of our custom `string_view` with the `std` one. I left the `util` alias in place to minimize the diff.
Next step if no one minds the churn could be to remove that and replace all `util::` and `fmt::` `string_view`s with `std::`.

Also some other stuff I remembered could be done with the update to c++17.


scamille:
Great to see that we can have that upgrade without actually having the additional c++20 string-view functions in the standard.

engine/util/static_map.hpp
```diff
@@ -119,24 +108,24 @@ class static_set_base {
...
     assert( is_unique(data) );
   }
 
   constexpr iterator begin() const { return cbegin(); }
-  constexpr iterator cbegin() const { return data_.begin(); }
+  constexpr iterator cbegin() const { return data_.data(); }
 
   constexpr iterator end() const { return cend(); }
-  constexpr iterator cend() const { return data_.end(); }
+  constexpr iterator cend() const { return data_.data() + N; }
```
scamille, line 122: Is there a reason to use data() and data()+N here instead of directly using the corresponding std::array functions?

scamille (approved):

engine/util/static_map.hpp
nuoHep: Yeah, different types.

`static_xxx::iterator` is a pointer. `std::array<>::iterator` is implementation defined, just has to satisfy some named requirements.
So while it [`std::array<>::iterator`] is *typically* a pointer, at least in release builds, in msvc debug builds it's a checked iterator struct.

I actually had it call the corresponding `std::array` functions initially, but had to change it (https://github.com/simulationcraft/simc/compare/212a90813c8a97d550e6799fbda4c08e073c3a44..66029fa2b3ec5140a0219d2bfb35a60efda44b0a) because of that.

PR #6212, opened 2021-11-20
[Hunter] T28 set bonus implementation

ThatLing:
https://www.wowhead.com/news/datamined-hunter-tier-set-bonuses-in-patch-9-2-beast-mastery-marksmanship-324968

Progress:
- [x] Beast Mastery
- [x] Marksmanship
- [x] Survival

### DIFF
--- a/engine/class_modules/sc_hunter.cpp
+++ b/engine/class_modules/sc_hunter.cpp
@@ -394,6 +394,7 @@ public:
     buff_t* terms_of_engagement;
     buff_t* tip_of_the_spear;
     buff_t* vipers_venom;
+    buff_t* mad_bombardier;
 
     // Legendaries
     buff_t* butchers_bone_fragments;
@@ -599,6 +600,8 @@ public:
     timespan_t pet_basic_attack_delay = 0.15_s;
     // random testing stuff
     bool stomp_triggers_wild_spirits = true;
+    bool t28_2pc                     = false;
+    bool t28_4pc                     = false;
   } options;
 
   hunter_t( sim_t* sim, util::string_view name, race_e r = RACE_NONE ) :
@@ -4838,6 +4841,15 @@ struct kill_command_t: public hunter_spell_t
         flankers_advantage.proc -> occur();
         cooldown -> reset( true );
         p() -> buffs.strength_of_the_pack -> trigger();
+
+        if ( p() -> options.t28_2pc )
+        {
+            if ( rng().roll( 0.4 ) )
+            {
+              p() -> cooldowns.wildfire_bomb -> reset( true );
+              p() -> buffs.mad_bombardier -> trigger();
+            }
+        }
       }
     }
 
@@ -5343,6 +5355,9 @@ struct wildfire_bomb_t: public hunter_spell_t
       double am = hunter_spell_t::composite_da_multiplier( s );
 
       am *= 1 + p() -> buffs.flame_infusion -> check_stack_value();
+
+      if ( p() -> options.t28_4pc )
+        am *= 1 + 0.3 + p() -> buffs.mad_bombardier -> check_value();
 
       return am;
     }
@@ -6192,6 +6207,11 @@ void hunter_t::create_buffs()
     make_buff( this, "aspect_of_the_eagle", specs.aspect_of_the_eagle )
       -> set_cooldown( 0_ms );
 
+  buffs.mad_bombardier =
+    make_buff( this, "mad_bombardier" )
+      -> set_default_value( 0.5 )
+      -> set_chance( options.t28_4pc );
+
   // Conduits
 
   buffs.brutal_projectiles =
@@ -6731,6 +6751,9 @@ void hunter_t::create_options()
   add_option( opt_obsoleted( "hunter.brutal_projectiles_on_execute" ) );
   add_option( opt_obsoleted( "hunter.memory_of_lucid_dreams_proc_chance" ) );
   add_option( opt_obsoleted( "hunter.serpenstalkers_triggers_wild_spirits" ) );
+
+  add_option( opt_bool( "hunter.t28_2pc", options.t28_2pc ) );
+  add_option( opt_bool( "hunter.t28_4pc", options.t28_4pc ) );
 }
 
 // hunter_t::create_profile =================================================
--- a/engine/class_modules/sc_hunter.cpp
+++ b/engine/class_modules/sc_hunter.cpp
@@ -594,7 +594,8 @@ public:
     events::tar_trap_aoe_t* tar_trap_aoe = nullptr;
     wildfire_infusion_e next_wi_bomb = WILDFIRE_INFUSION_SHRAPNEL;
     unsigned steady_focus_counter = 0;
-    double focus_used = 0;
+    // Focus used for Focused Trickery (363666)
+    double focus_used_FT = 0;
   } state;
 
   struct options_t {
@@ -678,6 +679,7 @@ public:
   action_t* create_action( util::string_view name, util::string_view options ) override;
   pet_t*    create_pet( util::string_view name, util::string_view type ) override;
   void      create_pets() override;
+  double resource_loss( resource_e resource_type, double amount, gain_t* g = nullptr, action_t* a = nullptr ) override;
   resource_e primary_resource() const override { return RESOURCE_FOCUS; }
   role_e    primary_role() const override { return ROLE_ATTACK; }
   stat_e    convert_hybrid_stat( stat_e s ) const override;
@@ -987,27 +989,6 @@ public:
     }
 
     return total_regen + floor( total_energize );
-  }
-
-  void consume_resource() override
-  {
-    ab::consume_resource();
-
-    if ( !p() -> specs.marksmanship_hunter -> ok() || !p() -> options.t28_4pc )
-      return;
-
-    resource_e cr = ab::current_resource();
-    if ( cr != RESOURCE_FOCUS )
-      return;
-
-    p() -> state.focus_used += cost();
-
-    // TODO: The cost is a double, can this cause rounding errors?
-    while ( p() -> state.focus_used >= 80.0 )
-    {
-      p() -> state.focus_used -= 80.0;
-      p() -> buffs.trick_shots -> trigger( 2 );
-    }
   }
 
   // action list expressions
@@ -3473,9 +3454,6 @@ struct aimed_shot_base_t: public hunter_ranged_attack_t
     base_aoe_multiplier = p -> specs.trick_shots -> effectN( 4 ).percent() +
                           p -> conduits.deadly_chain.percent();
 
-    if ( p -> options.t28_2pc )
-      base_aoe_multiplier += 0.3;
-
     trick_shots.target_count = 1 + static_cast<int>( p -> specs.trick_shots -> effectN( 1 ).base_value() );
 
     if ( p -> talents.careful_aim.ok() )
@@ -3508,6 +3486,9 @@ struct aimed_shot_base_t: public hunter_ranged_attack_t
       if ( target_health_pct > careful_aim.high || target_health_pct < careful_aim.low )
         m *= 1 + careful_aim.multiplier;
     }
+
+    if ( p() -> options.t28_2pc && trick_shots_up() )
+      m *= 1.3;
 
     return m;
   }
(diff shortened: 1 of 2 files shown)


engine/class_modules/sc_hunter.cpp
```diff
@@ -5344,6 +5384,9 @@ struct wildfire_bomb_t: public hunter_spell_t
 
       am *= 1 + p() -> buffs.flame_infusion -> check_stack_value();
 
+      if ( p() -> options.t28_4pc )
+        am *= 1 + 0.3 + p() -> buffs.mad_bombardier -> check_value();
```
nuoHep: According to spell data, this doesn't work this way. The main 2pc spell (https://ptr.wowhead.com/spell=363667/mad-bombardier) is a passive 30% increase. The buff you get from the 4pc (https://ptr.wowhead.com/spell=363805/mad-bombardier) is a separate 50% increase on top of that. Yes, it does mean the final increase is more than 80 - 95.

Also, both spells work off white lists and increase not only base bomb damage but also secondary stuff from wildfire infusion.

engine/class_modules/sc_hunter.cpp
```diff
@@ -1863,6 +1872,16 @@ struct kill_command_bm_t: public kill_command_base_t
 
     return am;
   }
+
+  double composite_crit_chance() const override
+  {
+    double cc = kill_command_base_t::composite_crit_chance();
+
+    if ( o()->options.t28_2pc )
+      cc *= 1 + (0.15 * p()->buffs.frenzy->check());
```
nuoHep: This should be `+=` as crit chance increases are typically additive.

engine/class_modules/sc_hunter.cpp
```diff
@@ -4838,6 +4867,17 @@ struct kill_command_t: public hunter_spell_t
...
         p() -> buffs.strength_of_the_pack -> trigger();
+
+        if ( p() -> options.t28_2pc )
+        {
+            if ( rng().roll( 0.4 ) )
+            {
+              p() -> cooldowns.wildfire_bomb -> reset( true );
+
+              // TODO: Does mad bombardier (4) trigger from only these resets or also when it comes off cd like normal?
+              p() -> buffs.mad_bombardier -> trigger();
```
nuoHep: Re the 4pc question, at least the wording suggests that the buff triggers only from resets.
It also kinda makes sense mechanically - KC reset has a 40% chance to trigger a buff (https://ptr.wowhead.com/spell=363805/mad-bombardier) that actually prevents charge consumption and increases damage. So triggering it only here makes sense.

engine/class_modules/sc_hunter.cpp
nuoHep: Oh, you also have to consume the buff after execute. Otherwise it will linger forever.

engine/class_modules/sc_hunter.cpp
```diff
@@ -4838,6 +4867,17 @@ struct kill_command_t: public hunter_spell_t
         flankers_advantage.proc -> occur();
         cooldown -> reset( true );
         p() -> buffs.strength_of_the_pack -> trigger();
+
+        if ( p() -> options.t28_2pc )
+        {
+            if ( rng().roll( 0.4 ) )
+            {
+              p() -> cooldowns.wildfire_bomb -> reset( true );
```
nuoHep: Judging by the wording this is wrong. It does not reset the cd but supresses triggering the cd while the 4pc buff is up. So this should be done by overriding `action_t::update_ready()` and skipping it in case the buff's up.

nuoHep: Smth like
```cpp
void update_ready( timespan_t cd_duration ) override {
  if ( p() -> buffs.mad_bombardier -> check() )
    return; // maybe sim->print_debug() smth about mad bombardier 

  hunter_spell_t::update_ready( cd_duration );
} 
```

engine/class_modules/sc_hunter.cpp
```diff
@@ -3421,6 +3472,9 @@ struct aimed_shot_base_t: public hunter_ranged_attack_t
     base_aoe_multiplier = p -> specs.trick_shots -> effectN( 4 ).percent() +
                           p -> conduits.deadly_chain.percent();
 
+    if ( p -> options.t28_2pc )
+      base_aoe_multiplier += 0.3;
```
nuoHep: According to spell data 2pc adds a damage increase on the Trick Shots buff - https://ptr.wowhead.com/spell=257622/trick-shots the second effect is new in 9.2 and while it's 0 in the spell data, experience tells me that 2pc simply updates this value. Which means it increases *all* AiS/RF damage while it's up and is not passive.

engine/class_modules/sc_hunter.cpp
```diff
@@ -983,6 +988,27 @@ struct hunter_action_t: public Base
...
+    p() -> state.focus_used += cost();
+
+    // TODO: The cost is a double, can this cause rounding errors?
+    while ( p() -> state.focus_used >= 80.0 )
+    {
+      p() -> state.focus_used -= 80.0;
+      p() -> buffs.trick_shots -> trigger( 2 );
+    }
+  }
+
```
nuoHep: Tbh, I'd not bother implementing 4pc for now considering it simply doesn't make sense with the 2 charges wording.

engine/class_modules/sc_hunter.cpp
ThatLing: Fixed

engine/class_modules/sc_hunter.cpp
ThatLing: Fixed

engine/class_modules/sc_hunter.cpp
ThatLing: Fixed

engine/class_modules/sc_hunter.cpp
ThatLing: What should I change for this?

engine/class_modules/sc_hunter.cpp
ThatLing: I'll leave it in for now

ThatLing:
@nuoHep What's the best way to handle having 2 spells with the same name? Both the 2- and 4-piece set bonus have the same name but different spell IDs. I just named it ``mad_bombardier_2`` and ``mad_bombardier_4`` for now.

nuoHep:
> @nuoHep What's the best way to handle having 2 spells with the same name? Both the 2- and 4-piece set bonus have the same name but different spell IDs. I just named it `mad_bombardier_2` and `mad_bombardier_4` for now.

You don't actually need a buff for 4pc as it's a passive effect. Only 2pc needs a buff.

So...
If we already had the spell data for it 4pc passive would be handled like any other passive spell - as an `ab::apply_affecting_aura()` call inside `hunter_action_t` constructor (look at the big list already there) and the interaction with 2pc like eg Coordinated Assault (see `hunter_action_t::affected_by`, calls to `parse_damage_affecting_aura()` etc).
We don't have spell data for now unfortunately. But we can setup it close to what we'd do if we had (so the update to spell data would be minimal too).
Here is what I was going to do:
* add manually setup Mad Bombardier buff for 2pc
* add a new `damage_affected_by` field to `hunter_action_t::affected_by` (you would need that for 4pc handling with spell data anyway)
* hook up Mad Bombardier 4pc interaction with 2pc buff inside `hunter_action_t::composite_da_multiplier()` & `hunter_action_t::composite_ta_multiplier()` (like eg Coordinated Assault, which won't change when transitioning to spell data)
* manually handle passive 4pc inside `hunter_action_t::init()` multiplying `ab::base_dd_multiplier` & `ab::base_td_multiplier`
* go through all of the spells affected by 4pc and manually set the `affected_by` field to `true`

Then when we get spell data we just replace manual `affected_by` setting with `parse_damage_affecting_aura()` and replace passive 4pc handling inside `init()` with `ab::apply_affecting_aura()`.

Also, FYI, at least for now spell data for Mad Bombardier is setup to affect only direct damage. So you should set only direct to `true` (eg `affected_by.mad_bombardier.direct = true;`)

Edit: Ughh... Sorry, I mixed up 2pc & 4pc stuff. Updated.

engine/class_modules/sc_hunter.cpp
nuoHep: Ok, then I'd move it to `hunter_t::resource_loss()`.

engine/class_modules/sc_hunter.cpp
nuoHep: Handle it inside `composite_da_multiplier()` in `rapid_fire_t::damage_t` (like the hidden BP buff) and `aimed_shot_base_t` (have to add that and Tricks should be checked with `trick_shots_up()` because of the interaction with DT, will need to test that once PTR is up though in case they played with TrS while I was away).

engine/class_modules/sc_hunter.cpp
```diff
@@ -5474,6 +5552,18 @@ struct wildfire_bomb_t: public hunter_spell_t
         wildfire_cluster -> execute();
     }
   }
+
+  void update_ready( timespan_t cd_duration ) override
+  {
+    if ( p() -> buffs.mad_bombardier_2 -> check() )
+    {
+      p() -> buffs.mad_bombardier_2 -> expire();
```
nuoHep: Move buff expiration to the end of `wildfire_bomb_t::execute()`.

nuoHep: Ughh... thinking about it we may actually need to have 2 separate buffs for 2pc and 4pc interaction with 2pc because of the way the spell is setup (both in game and in simc).

So, 2pc should be consumed inside `wildfire_bomb_t::execute()` to properly handle cd interaction, but for the purposes of 4pc interaction with 2pc the buff should actually linger until after the actual damage spell executes inside impact.
I think the easiest would be to do with 2 separate buffs tbh.

Ideally this crap should be tested in game once the PTR goes up.

ThatLing: > Ughh... thinking about it we may actually need to have 2 separate buffs for 2pc and 4pc interaction with 2pc because of the way the spell is setup (both in game and in simc).

Yep, this is what I mean. There should be 1 buff for the cd reset and 1 buff for the increased damage. That's how I have implemented it currently with ``mad_bombardier_2`` and ``mad_bombardier_4``.

Is the list you commented before still relevant in this case? If so, I will work on implementing it this way.

engine/class_modules/sc_hunter.cpp
```diff
@@ -983,6 +989,27 @@ struct hunter_action_t: public Base
     return total_regen + floor( total_energize );
   }
 
+  void consume_resource() override
+  {
+    ab::consume_resource();
+
+    if ( !p() -> specs.marksmanship_hunter -> ok() || !p() -> options.t28_4pc )
```
ThatLing: Is this the correct way to check that the actor is a MM hunter?

engine/class_modules/sc_hunter.cpp
ThatLing: Updated, please check it's implemented correct

engine/class_modules/sc_hunter.cpp
ThatLing: Updated, please check it's implemented correct

engine/class_modules/sc_hunter.cpp
nuoHep: Ok for RF. For AiS move it to `composite_da_multiplier()` please, it's a buff so it's not target specific.

engine/class_modules/sc_hunter.cpp
nuoHep: We'd typically use `specialization() == HUNTER_MARKSMANSHIP` for that. This should work too though.

engine/class_modules/sc_hunter.cpp
nuoHep: Yeah, the list is still relevant for 4pc. But there are always diff way to skin a cat, I wrote what I was going to do for it.

Re different buffs - typically I'd just name the "main" one (the one you'd see in-game) as it's named in-game and append something to our "internal" ones. In this case you could name them `_2pc` and `_4pc` as they are used this way (`_2`/`_4` is a bit obscure).

nuoHep:
@ThatLing mind if I take it off from here and finish it myself? I want to make it available on raidbots tomorrow.

ThatLing:
@nuoHep No problem, thanks for the feedback 👍

PR #6213, opened 2021-11-21
Clean string/string_view usage in class modules

scamille:
- replace most remaining const std::string& parameters with util::string_view, since they usually call other functions/constructor accepting string_view anyway.
- adjust default parameter arguments for string_view

### DIFF
--- a/engine/class_modules/sc_mage.cpp
+++ b/engine/class_modules/sc_mage.cpp
@@ -164,14 +164,14 @@ struct buff_stack_benefit_t
   const buff_t* buff;
   std::vector<benefit_t*> buff_stack_benefit;
 
-  buff_stack_benefit_t( const buff_t* _buff, const std::string& prefix ) :
+  buff_stack_benefit_t( const buff_t* _buff, std::string_view prefix ) :
     buff( _buff ),
     buff_stack_benefit()
   {
     for ( int i = 0; i <= buff->max_stack(); i++ )
     {
-      buff_stack_benefit.push_back( buff->player->get_benefit(
-        prefix + " " + buff->data().name_cstr() + " " + util::to_string( i ) ) );
+      auto benefit_name = fmt::format("{} {} {}", prefix, buff->data().name_cstr(), i);
+      buff_stack_benefit.push_back( buff->player->get_benefit( benefit_name ) );
     }
   }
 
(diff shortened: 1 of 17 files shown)


engine/class_modules/sc_mage.cpp
```diff
@@ -164,14 +164,14 @@ struct buff_stack_benefit_t
   const buff_t* buff;
   std::vector<benefit_t*> buff_stack_benefit;
 
-  buff_stack_benefit_t( const buff_t* _buff, const std::string& prefix ) :
+  buff_stack_benefit_t( const buff_t* _buff, util::string_view prefix ) :
```
vituscze: I've already swapped the mage module to `std::string_view`

engine/class_modules/sc_mage.cpp
```diff
@@ -164,14 +164,14 @@ struct buff_stack_benefit_t
...
     buff( _buff ),
     buff_stack_benefit()
   {
     for ( int i = 0; i <= buff->max_stack(); i++ )
     {
-      buff_stack_benefit.push_back( buff->player->get_benefit(
-        prefix + " " + buff->data().name_cstr() + " " + util::to_string( i ) ) );
+      auto benefit_name = fmt::format("{} {} {}", prefix, buff->data().name_cstr(), i);
```
vituscze, line 173:
```suggestion
      auto benefit_name = fmt::format( "{} {} {}", prefix, buff->data().name_cstr(), i );
```

scamille: Am I allowed to fully use clang-format in the mage module? I am always hesitant, since I can't remember which modules are using it and which aren't.

vituscze: We're not using clang-format at the moment

engine/class_modules/sc_mage.cpp
scamille: I didn't consciously look at the mage file changes in summary and back off after seeing that everything was already adjusted.

Would you like me to revert the change here in buff_stack_benefit, or should I keep it (with the formatting changes suggested below)?

vituscze: Oh, definitely keep the change. I just wanted `std` namespace instead of `util` (since that's what the mage module now uses).

scamille: Oh. I haven't thought about that, I hoped we would eventually change everything in a synced,automated fashion instead of having both worlds. But if mage module already uses that I can of course adjust.

PR #6214, opened 2021-11-21
[Shaman] Experimental Elemental T28 set

Bloodmallet:
Preliminary implementation of T28 set bonuses 
```
Fireheart (2-Set) While your Storm Elemental or Fire Elemental is active, your Lava Burst deals 20% additional damage and you gain Lava Surge every 8 sec.
Fireheart (4-Set) Casting Lava Burst extends the duration of your Storm Elemental and Fire Elemental by 1.5 sec. If your Storm Elemental or Fire Elemental is not active, spending Maelstrom has a 20% chance to reduce its remaining cooldown by 10 sec instead.
```

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -808,6 +808,7 @@ public:
   void trigger_lightning_shield( const action_state_t* state );
   void trigger_hot_hand( const action_state_t* state );
   void trigger_vesper_totem( const action_state_t* state );
+  void trigger_lava_surge( const player_t* source );
 
   // Legendary
   void trigger_legacy_of_the_frost_witch( unsigned consumed_stacks );
@@ -5689,19 +5690,7 @@ struct flame_shock_t : public shaman_spell_t
 
     if ( rng().roll( proc_chance ) )
     {
-      if ( p()->buff.lava_surge->check() )
-        p()->proc.wasted_lava_surge->occur();
-
-      p()->proc.lava_surge->occur();
-      if ( !p()->executing || p()->executing->id != 51505 )
-        p()->cooldown.lava_burst->reset( true );
-      else
-      {
-        p()->proc.surge_during_lvb->occur();
-        p()->lava_surge_during_lvb = true;
-      }
-
-      p()->buff.lava_surge->trigger();
+      p()->trigger_lava_surge( d->source );
     }
 
     // Fire Elemental passive effect (MS generation on FS tick)
@@ -8504,6 +8493,25 @@ void shaman_t::trigger_lightning_shield( const action_state_t* state )
   }
 }
 
+void shaman_t::trigger_lava_surge(const player_t* source ) {
+  if ( buff.lava_surge->check() )
+  {
+    proc.wasted_lava_surge->occur();
+  }
+
+  proc.lava_surge->occur();
+  if ( !source->executing || source->executing->id != 51505 )
+  {
+    cooldown.lava_burst->reset( true );
+  } else
+  {
+    proc.surge_during_lvb->occur();
+    lava_surge_during_lvb = true;
+  }
+
+  buff.lava_surge->trigger();
+}
+
 // shaman_t::init_buffs =====================================================
 
 void shaman_t::create_buffs()
@@ -8639,18 +8647,7 @@ void shaman_t::create_buffs()
                        // TODO: confirm if duration is affected by conduit and tier set 4pc bonus
                        ->set_duration( buff.fire_elemental->buff_duration() )
                        ->set_tick_callback( [ this ]( buff_t* b, int, timespan_t ) {
-                         if ( buff.lava_surge->check() )
-                         {
-                           proc.wasted_lava_surge->occur();
-                         }
-
-                         proc.lava_surge->occur();
-                         if ( !sim->active_player->executing || sim->active_player->executing->id != 51505 )
-                         {
-                           cooldown.lava_burst->reset( true );
-                         }
-
-                         buff.lava_surge->trigger();
+                         trigger_lava_surge( b->source );
                        } )
                        // TODO: confirm recasting elemental during uptime behaviour
                        ->set_refresh_behavior( buff_refresh_behavior::EXTEND );
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -4386,7 +4386,7 @@ struct lava_burst_overload_t : public elemental_overload_spell_t
       m *= 1.0 + p()->buff.windspeakers_lava_resurgence->data().effectN( 1 ).percent();
     }
 
-    if ( p()->sets->has_set_bonus( SHAMAN_ELEMENTAL, T28, B2 ) && p()->buff.fireheart->up() )
+    if ( p()->sets->has_set_bonus( SHAMAN_ELEMENTAL, T28, B2 ) && p()->buff.fireheart->check() )
     {
       m *= 1.0 + p()->spell.t28_2pc_ele->effectN(2).percent();
     }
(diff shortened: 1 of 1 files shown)


HawkCorrigan (approved):

engine/class_modules/sc_shaman.cpp
```diff
@@ -4374,6 +4395,11 @@ struct lava_burst_overload_t : public elemental_overload_spell_t
       m *= 1.0 + p()->buff.windspeakers_lava_resurgence->data().effectN( 1 ).percent();
     }
 
+    if ( p()->opt_sl.t28_2pc_ele && p()->buff.fireheart->up() )
```
scamille:
```suggestion
    if ( p()->opt_sl.t28_2pc_ele && p()->buff.fireheart->check() )
```

engine/class_modules/sc_shaman.cpp
```diff
@@ -4745,6 +4771,10 @@ struct lava_burst_t : public shaman_spell_t
       m *= 1.0 + p()->buff.windspeakers_lava_resurgence->data().effectN( 1 ).percent();
     }
 
+      if ( p()->opt_sl.t28_2pc_ele && p()->buff.fireheart->up())
```
scamille:
```suggestion
      if ( p()->opt_sl.t28_2pc_ele && p()->buff.fireheart->check())
```

engine/class_modules/sc_shaman.cpp
```diff
@@ -5334,6 +5377,13 @@ struct earthquake_t : public shaman_spell_t
   void execute() override
   {
     shaman_spell_t::execute();
+
+    if ( p()->opt_sl.t28_4pc_ele && !p()->is_elemental_pet_active() && rng().roll( p()->opt_sl.t28_4pc_ele_chance ) )
+    {
```
scamille, line 5380: You could add a proc_t to track this.

Bloodmallet:
Sorry for srewing up this rebase. I'll look into fixing that tomorrow. Probably in a new branch

scamille:
One nice trick to fix the git diff in github PRs is to switch the destination branch to something else and back.

engine/class_modules/sc_shaman.cpp
```diff
@@ -8536,6 +8628,25 @@ void shaman_t::create_buffs()
                             ->set_default_value_from_effect_type( A_ADD_PCT_MODIFIER, P_TICK_TIME )
                             ->apply_affecting_conduit( conduit.call_of_flame );
 
+  buff.fireheart = make_buff( this, "fireheart" )
```
scamille: Consider moving this to a custom class if you don't want it to clutter up create_buffs().

engine/class_modules/sc_shaman.cpp
Bloodmallet: Why do you suggest to use `check`? If I understand it correctly `up` would provide the benefit tracking directly?

engine/class_modules/sc_shaman.cpp
Bloodmallet: I understand, but I get the improession our shaman module is fancying creating buff like this in one big blop.

engine/class_modules/sc_shaman.cpp
Bloodmallet: done

engine/class_modules/sc_shaman.cpp
scamille: Because the function can be called multiple times without actually executing the action, depending eg. on expressions.

engine/class_modules/sc_shaman.cpp
```diff
@@ -8536,6 +8635,26 @@ void shaman_t::create_buffs()
                             ->set_default_value_from_effect_type( A_ADD_PCT_MODIFIER, P_TICK_TIME )
                             ->apply_affecting_conduit( conduit.call_of_flame );
 
+  buff.fireheart = make_buff( this, "fireheart", spell.t28_2pc_ele )
+                       // TODO: confirm if duration is affected by conduit and tier set 4pc bonus
+                       ->set_duration( buff.fire_elemental->buff_duration() )
+                       ->set_tick_callback( [ this ]( buff_t* b, int, timespan_t ) {
+                         if ( buff.lava_surge->check() )
```
navv1234: Could consider making a shaman_t accessor for handling the Lava Surge, instead of replicating code here and in flame_shock_t?

engine/class_modules/sc_shaman.cpp
```diff
@@ -8536,6 +8635,26 @@ void shaman_t::create_buffs()
...
+                         {
+                           proc.wasted_lava_surge->occur();
+                         }
+
+                         proc.lava_surge->occur();
+                         if ( !sim->active_player->executing || sim->active_player->executing->id != 51505 )
```
navv1234: This will break badly in multi-actor sims. Use "source->executing" here.

engine/class_modules/sc_shaman.cpp
```diff
@@ -8536,6 +8635,26 @@ void shaman_t::create_buffs()
                             ->set_default_value_from_effect_type( A_ADD_PCT_MODIFIER, P_TICK_TIME )
                             ->apply_affecting_conduit( conduit.call_of_flame );
 
+  buff.fireheart = make_buff( this, "fireheart", spell.t28_2pc_ele )
+                       // TODO: confirm if duration is affected by conduit and tier set 4pc bonus
+                       ->set_duration( buff.fire_elemental->buff_duration() )
```
navv1234, line 8648: Fireheart itself has a 30 second duration, this probably needs some checking if stuff is synced in game or not?

navv1234 (commented):
Some small changes in comments, otherwise seems ok to me.

engine/class_modules/sc_shaman.cpp
Bloodmallet: Done

engine/class_modules/sc_shaman.cpp
Bloodmallet: How do I access source? Where do I find that?

engine/class_modules/sc_shaman.cpp
Bloodmallet: Yes we're aware. That#s why the comment is right above this line. Once we have this on PR we need to do some testing.

engine/class_modules/sc_shaman.cpp
navv1234: The buff passed to the lambda has a source, but since you are capturing this, you could just use "executing", which would in fact be the shaman object's executing. That's probably for the best.

I'd just create a generic "trigger_lava_surge" method in the shaman module (like it has for many other kind of behavior), and all it from flame shock and this spot, since the behavior seems indentical?

Bloodmallet: Hawk found it. Added.

Bloodmallet:
Comments were accounted for. Waiting for Ci to finish. Would rebase and merge afterwards if I get your ok for the changes.

engine/class_modules/sc_shaman.cpp
```diff
@@ -8405,6 +8493,25 @@ void shaman_t::trigger_lightning_shield( const action_state_t* state )
   }
 }
 
+void shaman_t::trigger_lava_surge(const player_t* source ) {
```
navv1234: This does not need any arguments.

engine/class_modules/sc_shaman.cpp
```diff
@@ -8405,6 +8493,25 @@ void shaman_t::trigger_lightning_shield( const action_state_t* state )
...
+  if ( buff.lava_surge->check() )
+  {
+    proc.wasted_lava_surge->occur();
+  }
+
+  proc.lava_surge->occur();
+  if ( !source->executing || source->executing->id != 51505 )
```
navv1234: Replace with if ( !executing || !executing->id != 51505 )

Bloodmallet: Done. See next commit. Thank you

engine/class_modules/sc_shaman.cpp
Bloodmallet: Done. See next commit. Thank you

PR #6216, opened 2021-11-21
Monk 9.2 T28 Set Bonus

Hinalover:


engine/class_modules/monk/sc_monk.cpp
```diff
@@ -490,6 +502,28 @@ struct monk_action_t : public Base
...
+        p()->buff.primordial_power->trigger();
+  }
+
+  double composite_persistent_multiplier( const action_state_t* action_state ) const override
+  {
+    double pm = ab::composite_persistent_multiplier( action_state );
+
+    if ( p()->sets->has_set_bonus( MONK_WINDWALKER, T28, B4 ) )
+    {
+      if ( ab::data().affected_by( p()->passives.primordial_power->effectN( 1 ) ) )
```
scamille, line 520: can you move this into a bool in the affected_by struct?

PR #6220, opened 2021-11-22
Do not set `-march=native` on darwin/arm64

wfarr:
The version of `clang` in XCode 13 does not support `-march=native` for
this chipset, so we exclude it from the `optimized` target by default on
that particular architecture.

This means darwin/arm64 builds may be slightly less optimized than non-darwin/arm64
builds, but that these do work out of the box without requiring extra
flags on my 2021 MacBook Pro M1 Max.

I've only tested this out locally so far and I'm far from an expert on this particular part of the build process so I mostly just banged on stuff until `make -j 8 optimized` passed.

### DIFF
--- a/engine/Makefile
+++ b/engine/Makefile
@@ -12,6 +12,7 @@
 
 
 FLAVOR     =
+ARCH       =
 DESTDIR    =
 PREFIX    ?= ..
 
@@ -49,6 +50,7 @@ ifeq (${OS},)
   else
     OS         = UNIX
     FLAVOR     = $(shell uname)
+    ARCH       = $(shell uname -m)
   endif
 endif
 
@@ -223,11 +225,31 @@ release: $(MODULE)
 
 optimized:CPP_FLAGS += -DNDEBUG
 optimized:OPTS_INTERNAL += -ffast-math -fomit-frame-pointer
-ifneq (${FLAVOR},Darwin)
-  ifneq ($(uname -m),arm64)
+
+# MacOS's default compiler, `clang`, does not support `-march=native` on Apple
+# Silicon chips, and as of 2021-11-22 will return an error like the following
+# at compile time:
+#
+#     clang: error: the clang compiler does not support '-march=native'
+#
+# This conditional preserves the existing default `-march=native` flag on all
+# non-Apple systems, and also on Apple Intel systems, while instead setting
+# `-mcpu=apple-m1` on Apple Silicon systems to enable any M1-specific
+# optimizations that clang supports.
+#
+# In a future release, presumably when there are M2 chips, M3 chips, and so on,
+# hopefully clang will support `-march=native` on these platforms and we can
+# eliminate this block.
+ifeq (${FLAVOR},Darwin)
+  ifeq (${ARCH},arm64)
+    optimized:OPTS_INTERNAL += -mcpu=apple-m1
+  else
     optimized:OPTS_INTERNAL += -march=native
   endif
+else
+  optimized:OPTS_INTERNAL += -march=native
 endif
+
 optimized: $(MODULE)
 
 install: all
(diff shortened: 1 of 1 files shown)


wfarr:
My machine:

My machine:

```
$ system_profiler -detailLevel mini -json | jq -r '.SPHardwareDataType'
[
  {
    "_name": "hardware_overview",
    "boot_rom_version": "7429.41.5",
    "chip_type": "Apple M1 Max",
    "machine_model": "MacBookPro18,2",
    "machine_name": "MacBook Pro",
    "number_processors": "proc 10:8:2",
    "os_loader_version": "7429.41.5",
    "physical_memory": "64 GB"
  }
]

$ system_profiler -detailLevel mini -json | jq -r '.SPSoftwareDataType'
[
  {
    "_name": "os_overview",
    "kernel_version": "Darwin 21.1.0",
    "os_version": "macOS 12.0.1 (21A559)",
    "uptime": "up 19:21:5:5"
  }
]

$ clang --version
Apple clang version 13.0.0 (clang-1300.0.29.3)
Target: arm64-apple-darwin21.1.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

With this change set, I can clone a fresh copy of the repo and build the contents of `engine/` with `make -j8 optimized`. Prior to this change set, I could not build even with the April changes and suggestion of `xcrun
    --sdk macosx11.1 make UNIVERSAL=1 -j 8 optimized` due to the incompatibility with `-march=native` in clang 13.

scamille:
https://stackoverflow.com/questions/65966969/why-does-march-native-not-work-on-apple-m1
interesting, learned something here.

It would be good to add a comment to your changes in the makefile explaining why this is necessary. Other than that, looks good.

navv1234:
That's peculiar, but yeah, go ahead. Intel macs have -march=native. Commenting this in the makefile would be nice, yes.

EDIT: Could do "-mcpu=apple-m1" instead? Or, maybe that breaks on other arm64 platforms, so probably not a good idea.

wfarr:
> Could do "-mcpu=apple-m1" instead? Or, maybe that breaks on other arm64 platforms, so probably not a good idea.

I'll give this a shot. I think it _still_ blew up with `-march=native` was present at all, but I could specify the `-mcpu=apple-m1` to try and get clang to do any optimizations it knows about.

wfarr:
44c4590d1d should include both a comment better describing the underlying problem, as well as setting `-mcpu=apple-m1` for darwin/arm64 hosts.

engine/Makefile
```diff
@@ -49,6 +50,7 @@ ifeq (${OS},)
   else
     OS         = UNIX
     FLAVOR     = $(shell uname)
+    ARCH       = $(shell uname -m)
```
wfarr, line 53: It's easy to forget you need to `shell` this, and it seems likely we'll utilize this same thing down the road to flag other behavior by cpu architecture, so I've added this as a variable on all Unix systems.

navv1234:
Looks good to me now.

engine/Makefile
```diff
@@ -223,11 +225,31 @@ release: $(MODULE)
...
+# at compile time:
+#
+#     clang: error: the clang compiler does not support '-march=native'
+#
+# This conditional preserves the existing default `-march=native` flag on all
+# non-Apple systems, and also on Apple Intel systems, while instead setting
+# `-mcpu=apple-m1` on Apple Silicon systems to enable any M1-specific
+# optimizations that clang supports.
+#
+# In a future release, presumably when there are M2 chips, M3 chips, and so on,
```
scamille, line 240: I fear that even though a new release of clang might support -march=native for M1, you will have again no definition for M2 and it will break again :-)

wfarr: Thoughts and prayers in the mean time. 😂

PR #6225, opened 2021-11-25
[Warlock] Corruption for specs other than Affliction

Fatherdotter:
Implemented offspec corruption as an action. 
- Demo/destro corruption actions get a version of corruption defined in the main file with no instant damage on application, no ta_multiplier, and no legendary/talent checks.
- Aff corruption actions fall through and are executed as normal prior to this edit.


scamille:
What is the rationale for separating the action instead of making sure that the effects you mention are only enabled when they should be?

Edit: I see that warlock_t::create_action_affliction is not called for non affliction warlocks. But that just might be a bad design choice there, and with this makes matters worse by duplicating functionality.
There is no reason that actions which are accessible by all specs can not live in the warlock_affliction.cpp file. Or more generally, there is no good reason to artificially restrict action construction by specialization. The action itself should already check if you are allowed to create it based on the spell data availability, which can already be based on a talent or various other things.

cc @Azevara

Fatherdotter:
I spoke with Azevara about this, and there did not seem to be a great way to implement this with the current file layout.

The issue that we ran into was the fact that Seed of Corruption in the affliction file requires a solid reference to the corruption_t type to apply or refresh corruption.

While the corruption action is available to all specs, the issue we ran into was that we saw no clean way to define the corruption action in the main warlock file and also make it visible and available to the affliction file for an instance to be created by seed of corruption explosion.

scamille:
Just remove the specialization checks at https://github.com/simulationcraft/simc/blob/9672ae82c9ce9e322daaf1a545100871e8a826b3/engine/class_modules/warlock/sc_warlock.cpp#L1036 and you should be done. No need to move the corruption action itself to the main file.

Priest has a lot of shadow actions in the priest_shadow.cpp file, which can be used by discipline/holy.

vituscze:
It doesn't look like you need `corruption_t`, `action_t` would be sufficient. That is, the affliction module doesn't need access to the definition of `corruption_t`, you just need to create the action with something like `create_action( "corruption" )` instead of using the `corruption_t` constructor.

Fatherdotter:
How does that work with regards to Seed of Corruption in `sc_warlock_affliction.cpp` requiring a `corruption_t` instance?

https://github.com/simulationcraft/simc/blob/9672ae82c9ce9e322daaf1a545100871e8a826b3/engine/class_modules/warlock/sc_warlock_affliction.cpp#L422-L439

Edit: Embed and link code in question

scamille:
Don't change anything about corruption_t or seed_of_corruption_aoe_t. Leave it in the warlock_affliction.cpp file, no change necessary.

Just allow demo/destro warlocks to create "corruption" actions by removing the artificial constraints put into `warlock_t::creat_action`.

The spec subfiles were intended not to strictly have spec-only stuff in there, more as a way of grouping things together a bit. So don't restrict it too much.

If you really have to share things across translation units, there are also ways to achieve that - but that just isn't needed here.

Fatherdotter:
@scamille 

I guess I'm just confused about how this isn't required.

I understand in the instance of just casting corruption that it isn't necessary, but Seed of Corruption currently creates a `corruption_t` instance with several non-default variables.

We could define the `corruption_t` type in `sc_warlock.cpp` as it is defined in `sc_warlock_affliction.cpp` so there isn't a definition of the action in both `sc_warlock.cpp` and `sc_warlock_affliction.cpp` but there are several checks in the affliction version that are not necessary for demo/destro.

Do you see any way to allow Seed of Corruption to create and apply corruption to the affected actors while not having a definition in `sc_warlock_affliction.cpp` and maintaining the behavior that it has? It would still need to be a background action, no mana cost, no direct damage, and not affect the number of corruption executes in a sim.

As far as I know, just causing Seed of Corruption to create a corruption action would not have any of the above behavior.

scamille:
I thought I had a real hard time bringing across my proposed solution, so please have a look at #6229 

As far as I can tell, it resolves what your PR with the given title was intended to solve: Let Destruction and Demonology Warlock use corruption.

Fatherdotter:
@scamille 

Responded on #6229 

I am fine with considering another implementation, but I felt this was more appropriate. The only other functional implementation I had considered (and discussed with Azevara) was implementing corruption in each spec's file. We have similar precedence with regards to Shadow Bolt being defined in both `sc_warlock_affliction.cpp` and `sc_warlock_demonology.cpp` due to their different performance for each spec. I just felt that since demonology have the same effect when using corruption (other than Destruction's mastery which, as I mentioned on your PR I waited to implement until we decided how to make corruption available cross-spec) it made sense to make the action available for them in the main file while routing the more complicated performance of corruption to the affliction file (the only spec that needs it.)

scamille:
Yeah thanks for the feedback. I don't want to force my proposed solution here, the point was to just show the code to get the point across.

I personally would favor no duplication for corruption, shadow_bolt and other spell that might be used by more than 1 spec, but still are more conveniently place in a warlock_<spec>.cpp file. But your solution of duplicating the action is fine as well.

There also is the option of implementing corruption_t in warlock.cpp and still getting it to work for seed of corruption. Might require some C++ trickery, I am sure @vituscze or myself can help you with that.

As long as you guys are aware of the choices, and make an informed decision, everything is fine :-)

Fatherdotter:
No problem! I agree that I would like to move spells as high as they can, and it's entirely possible that Shadow Bolt in particular could be moved up. I would prefer to move `corruption_t`  up to the `sc_warlock.cpp` file as well but I did not find a way to do it nicely.

scamille:
Ideally, the order of the actions that can be created in player_t::create_action does not depend on the ordering. It should be more like a map lookup, with a single action returned for a given name. That is usually how most class modules have implemented it in the past, and too much logic in there usually makes life harder, especially when writing APLs.

Fatherdotter:
Of course, and this implementation does replicate that behavior. It is only possible for each corruption call to create one corruption action (and it is the desired action.) The only way that ordering matters is in the case that the spec checks are removed from the appropriate create_action calls. 

Otherwise, the current create_action calls function as you describe. You can change the order of the spec-specific create_action lines with no effect.

Azevara:
#6236 will be used (barring unforeseen issues) to handle which version of corruption to use per-spec. This will require merging in that change and a slight adjustment to the changes made to action creation in this PR.

Beyond resolving that issue, there are a few secondary items which need addressing here for the generic cross-spec version:
- There is a manual application of spec aura in Affliction's corruption, which may need to be similarly addressed for the other two specs
- Currently untested, needs verification: Sacrolash and Wrath of Consumption legendaries may interact with Corruption even when not spec'd into Affliction
- Destruction's Mastery may affect Corruption damage, requires verification

Azevara:
#6237 includes a correction to the changes from #6236, please make sure this is also included.

Fatherdotter:
@Azevara Updated to be functional with #6236 and #6237,  along with the changes we discussed

- Create_action for Corruption no longer includes a spec check
- Destro mastery added to the main Corruption definition
- Immolate set to be affected by Wrath of Consumption
- **Edit**: Also added Sacrolash to the main Corruption defition

PR #6229, opened 2021-11-26
Warlock: enable corruption for non-affliction warlocks

scamille:
More generally, allow the creation of actions located in warlock_spec.cpp files to be constructed by other specializations, assuming the action itself supports that. (eg. spell data is ok)

### DIFF
--- a/engine/class_modules/warlock/sc_warlock.cpp
+++ b/engine/class_modules/warlock/sc_warlock.cpp
@@ -1033,26 +1033,14 @@ action_t* warlock_t::create_action( util::string_view action_name, util::string_
   if ( action_name == "interrupt" )
     return new interrupt_t( action_name, this, options_str );
 
-  if ( specialization() == WARLOCK_AFFLICTION )
-  {
-    action_t* aff_action = create_action_affliction( action_name, options_str );
-    if ( aff_action )
-      return aff_action;
-  }
+  if ( action_t* aff_action = create_action_affliction( action_name, options_str ) )
+    return aff_action;
 
-  if ( specialization() == WARLOCK_DEMONOLOGY )
-  {
-    action_t* demo_action = create_action_demonology( action_name, options_str );
-    if ( demo_action )
-      return demo_action;
-  }
+  if ( action_t* demo_action = create_action_demonology( action_name, options_str ) )
+    return demo_action;
 
-  if ( specialization() == WARLOCK_DESTRUCTION )
-  {
-    action_t* destro_action = create_action_destruction( action_name, options_str );
-    if ( destro_action )
-      return destro_action;
-  }
+  if ( action_t* destro_action = create_action_destruction( action_name, options_str ) )
+    return destro_action;
 
   return player_t::create_action( action_name, options_str );
 }


Fatherdotter:
In my opinion, this is an odd choice of implementation that brings a few problems (immediate and possible) from routing a cross-spec spell through the affliction file.

While it does prevent the duplicate declaration of the corruption action, it allows specs to attempt to take any action that any spec could take (and attempts to create an action for each spec in order of aff->demo->destro) rather than checking if they are the appropriate spec before attempting to create that action.

Not only does it allow specs to take impossible actions (for example, Destruction can now use Shadow Bolt), it causes specs to utilize the action definition in the spec file above them. Demonology now takes affliction's "Shadow Bolt" action, and generates no soul shards.

This also routes a multi-spec spell (corruption) through the affliction file `sc_warlock_affliction.cpp` , which would mean that if there are any changes to be made for destro/demo to corruption they would have to be implemented in the affliction file (for example, my next commit once we decided how to implement corruption as a cross-spec spell was to add Destruction's mastery to the corruption spell. If Destro is taking the corruption action by calling the action in the affliction file, the affliction file will also need to implement Destro's mastery.)

vituscze:
> While it does prevent the duplicate declaration of the corruption action, it allows specs to attempt to take any action that any spec could take (and attempts to create an action for each spec in order of aff->demo->destro) rather than checking if they are the appropriate spec before attempting to create that action.

This is not necessarily an issue, as long as the spells are initialized with spell data from `find_specialization_spell` call (which will do the spec check for you). Take a look at the mage module. There are no restrictions like that and yet simc won't allow you to execute Fireballs as Frost.

Fatherdotter:
@vituscze I am aware :)

I explain further on why this is an issue, but to be more clear about that issue in particular the problem is that before the lines that checked for specialization prevented these issues from occurring. With the specialization checks in place, destro/demo never attempted to create the actions stored in the affliction file.

Now they do attempt to create 'affliction actions', and it created the two issues below that I listed. Sure aff can't cast Hand of Guldan with this change, but there are two cases already in this commit where issues are created. Destruction can take an action that is not possible in-game, and Demonology no longer casts the Shadow Bolt defined in their file (which means that Shadow Bolt no longer generates resources for Demonology sims.)

I didn't say that they *could* take any action of any other spec, but that they now *attempt* to. As this commit shows that can lead to undesirable behavior.

vituscze:
Right, so this PR isn't sufficient to add Corruption, but the problem is ultimately with missing checks in the destruction action and the duplicate spell.

I'm not really familiar with details of warlock spells, but I thought all specs share the same Shadow Bolt spell. Seems like it should be just one type of action.

scamille:
I wasn't aware that there are already that many duplications of spells in the different warlock modules. This solution suits the "implement action once per class", avoiding going through the duplication route that seems already to be there with shadow_bolt and potentially corruption.

Fatherdotter:
With response to both comments, I agree that it would be desirable to have actions like these defined in `sc_warlock.cpp`, and it may in fact be possible to move Shadow Bolt in particular up.

The issue I brought up in #6225 was that with the current warlock file layout I did not see a way to do this while making the needed functionality available to Affliction sims to use Seed of Corruption. The only solution I found that resolves the issue without having a duplicate definition was to move to a single warlock file like other classes use. This would make the corruption definition available to Seed of Corruption and the corruption definition available to destro/demo.

The only problem there is that, while I can't speak for Azevara, personally I do prefer the layout of multiple files for each spec. Not only does it make maintenance easier (each file only contains the definitions and checks pertinent to the spec at hand) it allows us to use inheritance to handle more complex behavior only in the files where it is needed.

While I understand the desire to avoid duplicate spells definitions, moving these definitions up and adding spec-specific code (if it is even possible with corruption here) just seems to be a move towards just having a single file for the entire class. I am not sure if that is something Azevara is interested in, but personally I would prefer not to go the route of having a single file for the entire class.

scamille:
A simple function `action_t* warlock_t::create_action_corruption` should do the trick. Declare it in warlock.hpp, implement in warlock.cpp and use it in warlock_affliction.cpp

Fatherdotter:
Of course, but that still leaves the above comment which I personally find to be an issue.

If we are going to have a spell definition with multiple actions that are specific only to a single spec within them, which means talent checks or multiple `spec()` checks then what is the purpose of breaking out the spec files?

Surely it cannot be more optimal to have every tick of incinerate checking if the spell is affected by wrath of consumption, every action checking for the appropriate mastery to apply, and every action aff/destro take checking if it is affected by things like From the Shadows.

vituscze:
The cost of these checks is practically zero compared to pets, APL, event handling overhead etc. The main focus should be on clarity and ease of maintenance.

CurtisJMC:
I wholly agree, but in my opinion having all of this code in the same file would make it more difficult to maintain the module, and less clear since there is code for multiple specs contained in each spell.

That is my opinion though, and if Azevara would prefer to refactor warlock into a single file I will happily contribute to that goal.

Fatherdotter:
I wholly agree, but in my opinion having all of this code in the same file would make it more difficult to maintain the module, and less clear since there is code for multiple specs contained in each spell.

That is my opinion though, and if Azevara would prefer to refactor warlock into a single file I will happily contribute to that goal.

scamille:
> The cost of these checks is practically zero compared to pets, APL, event handling overhead etc. The main focus should be on clarity and ease of maintenance.

That was pretty much the motivation for me to point out this solution here. Duplicating corruption_t for affliction and non-affliction was not a good thing for maintainability in my book.
Restricting spells in warlock_affliction.cpp rigorously to affliction only isn't something that I think is particularily important. For me that always was more of a guideline to put the spells with the main focus for that spec into the file. At least that is how I know it from the priest module.

If warlock maintainer think that is better to absolutely only keep affliction-only spells in warlock_affliction.cpp, then that is a valid choice.

Btw about module splitting: If I could do it again for priests, I would probably split it up into a priest_actions.cpp and priest_pets.cpp file nowadays. The split by spec was done when there was a quite large amount of holy and discipline content in there and the 3 specs did not share that much, so it was a natural decision. Nowadays there is much more shared spells between the 3 specs, at least for the remaining dps-implementations of holy and discipline, so a split alongside the spec boundary might not be the first choice anymore. I don't know the exact situation for warlocks, and how much stuff is shared or separated between the specs. Since all 3 are dps specs, I can imagine there being quite a bit of common stuff.
Keep in mind that reorganizing a module into different files can be a lot of work, and I would not recommend doing it if the current situation is good enough for you.

Fatherdotter:
For what it's worth, there is not a lot of overlap between warlock specs in terms of damaging abilities. The abilities that are shared between specs are usually not functionally similar. The only warlock (not covenant) spell I can think of that is functionally identical across specs is Drain Life, which is defined in ` sc_warlock.cpp`.

Shadow Bolt is shared between Demonology and Affliction. 
- Affliction uses it primarily as filler damage and modifies/replaces it with talents and at least one conduit. It generates no soul shards.
- Demonology uses it primarily as a shard generator, it generates soul shards and a buff using one legendary ([Balespider's Burning Core](https://www.wowhead.com/spell=337159/balespiders-burning-core))

Corruption is shared between all three specs
- Affliction uses it as a dot that is constantly maintained. For maximum level affliction warlocks it is instant cast and does damage on application. It is also modified by talents, legendaries, mastery, as well as several different ranks of the spell. It is also applied in AoE by the Seed of Corruption talent with no damage on application (seed itself does explosion damage, but the corruption instances applied do not do damage themselves.)
- Demonology/Destruction maintain the same version of the spell regardless of level, with a cast time and no damage on application. The only difference between the two is that it is likely affected by Destruction's mastery (pending in-game testing once this is resolved.)

So despite them all being DPS specs,  there are only three damaging warlock spells that are shared between specs... and two of them are drastically different for at least one specialization. The last one is defined for all three in `sc_warlock.cpp`

Azevara:
Please see #6236 for an alternative approach that I am going to suggest we utilize for now, barring any unforeseen issues. Given the uncertainty around order issues in this PR, I believe the best approach at the moment is to have actions with a matching name in the spec files supersede those found in sc_warlock.cpp.

Long-term, a complete overhaul of how we have the warlock module organized is warranted which would probably render this issue moot, but that will most likely be pushed out to 10.0 alpha/beta as a project.

Azevara:
Addendum: Any further discussion related to the Corruption action specifically should take place in #6225

PR #6230, opened 2021-11-26
[Paladin] Basic T28 Support

Woliance:
Added T28 4p for Protection
As well as basic support for retribution and holy tier sets (sc_paladin.hpp 440-450 & sc_paladin.cpp 2428--2434)


Woliance:
Note: At the moment all judgments from T28 4P for Prot is labeled as divine toll.

Woliance:
Remaking PR to add 2p

PR #6236, opened 2021-11-29
[Warlock] Update create_action logic to handle cross-spec conflicts

Azevara:


engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -1036,24 +1038,31 @@ action_t* warlock_t::create_action( util::string_view action_name, util::string_
...
 
+  return nullptr;
+}
+
+action_t* warlock_t::create_action( util::string_view action_name, util::string_view options_str )
+{
+  // The generic sc_warlock actions are checked first and then passed into the respective spec create_action
+  // create_action_[specialization] should return a more specialized action if needed (ie Corruption in Affliction)
+  // Default behavior is that if no alternate action is found, generic_action is returned
+  action_t* generic_action = create_action_warlock( action_name, options_str );
```
scamille, line 1049: ~~I don't think it is a good idea to create the generic action then if it is superseeded by a spec-specific one, just drop it silently.~~

~~I would recommend just putting the spec specific lines before the generic one.~~
Edit: Ups, didn't yet see your other PR.

PR #6238, opened 2021-11-30
Demonology set bonus for 9.2

nikibobi:
I implemented the 2-piece set bonus for demonology warlock.
I plan to test demonology on PTR and try my hand at implementing the 4-piece set bonus and I'll post progress here.


Azevara:
This PR looks to be correct based on what we know now, but we will wait until PTR is testable on Dec 2 before merging in case any additional concerns arise.

PR #6240, opened 2021-12-02
[Warlock] Removed the ability of demo Horned legendary to self-proc

Fatherdotter:
Removed the ability of the demo legendary 'Forces of the Horned Nightmare' to proc off of its own procs.


engine/class_modules/warlock/sc_warlock_demonology.cpp
```diff
@@ -250,7 +244,14 @@ struct hand_of_guldan_t : public demonology_spell_t
...
     demonology_spell_t::impact( s );
 
-    impact_spell->execute();
+    impact_spell->execute();    
+
+    if ( p()->legendary.forces_of_the_horned_nightmare.ok() &&
+         rng().roll( p()->legendary.forces_of_the_horned_nightmare->effectN( 1 ).percent() ) )
+    {
+      p()->procs.horned_nightmare->occur();
+      make_event( *sim, 400_ms, [ this, t = target ] { impact_spell->execute_on_target( t ); } );
```
vituscze, line 253: This should be `[ this, t = s->target ]`.

engine/class_modules/warlock/sc_warlock_demonology.cpp
```diff
@@ -250,7 +244,14 @@ struct hand_of_guldan_t : public demonology_spell_t
   {
     demonology_spell_t::impact( s );
 
-    impact_spell->execute();
+    impact_spell->execute();
```
vituscze, line 247: This probably needs to set a target too (or use the execute_on_target fn).

Fatherdotter: Would the target not be set by the call up to `action_t::impact()` ?

https://github.com/simulationcraft/simc/blob/d6699509e1fa07a7263ec3764dfc8ed4a26a4478/engine/action/sc_action.cpp#L3925-L3942

engine/class_modules/warlock/sc_warlock_demonology.cpp
Fatherdotter: I can understand this one, but would target not also be acceptable here?

If I understand correctly, when we make the event we are passing by value so the player's target at the time the event is created should also be the action state's target.

Is there any situation where these two variables would be different? Appreciate any advice!

Edit: For the record, I definitely agree with this change. Just trying to gain more understanding about the engine (and the situation that might cause this to arise)

vituscze: `target` and `s->target` can differ when: a) you're dealing with an AoE spell or b) when the spell has travel time and executes on multiple targets in a quick succession

vituscze: Yes, it's most likely fine here but it's a potential error waiting to happen.

engine/class_modules/warlock/sc_warlock_demonology.cpp
vituscze: That's `impact_action`, not `impact_spell`

Fatherdotter: This is the action_t::impact function, which as far as I am aware is what is called up to via 
`demonology_spell_t::impact()` (Only calls up)
->`warlock_spell_t::impact()` (Only calls up)
-> `spell_t::impact()` (Not overridden)
->`spell_base_t::impact()` (Not overridden) 
-> `action_t::impact()`

Is there some interaction along the way that would prevent this from occurring?

Edit: Corrected to type names instead of file names.

engine/class_modules/warlock/sc_warlock_demonology.cpp
Fatherdotter: Understood. Like I said I certainly agree with the change, was just trying to gain an understanding of the situation because I am a bit new to the engine.

engine/class_modules/warlock/sc_warlock_demonology.cpp
vituscze: There's nothing in the code you linked that ensures `impact_spell->target` is the same as `s->target`.

Fatherdotter: Perhaps I misunderstand
the `impact_spell` is of type `hog_impact_t`, which inherits from `demonology_spell_t`, which as we've shown above inherits up to `action_t`

The target variable of `impact_spell` does not appear to be overridden either, so it would be the reference set in `action_t`, which is set by the call at the bottom of this code to `impact_action`'s `::set_target()` function, is it not?

https://github.com/simulationcraft/simc/blob/d6699509e1fa07a7263ec3764dfc8ed4a26a4478/engine/action/sc_action.cpp#L4518-L4527

So if the impact itself has its target driven by the state passed to it (always in reference to the `state->target` in `action_t`), and the `impact_spell` the impact function creates has its target set to the state's target, So the value of `impact_spell->target` should be the same as `s->target` regardless of which `impact_spell` variable we are talking about, right?

I'm confused which target variable we believe needs to be set here. I just wonder because up until this point it has been functional and hitting the target without setting it.

Fatherdotter: Should be resolved by #6243

PR #6241, opened 2021-12-02
[Warlock] Initial implementation of affliction T28 set

Fatherdotter:
Working implementation of affliction's T28 2-piece and 4-piece set bonus.

PR is available for review, as discussed we are awaiting PTR testing to confirm behavior before merging.


Azevara:
This should be safe to merge following sanity checks once PTR is testable on Dec 2.

engine/class_modules/warlock/sc_warlock_affliction.cpp
```diff
@@ -106,10 +106,28 @@ struct shadow_bolt_t : public affliction_spell_t
...
+      {
+        // Add passive check
+        td( s->target )->debuffs_shadow_embrace->trigger();
+      }
+
+      // TODO: Check PTR to see if 4-set procs on cast finish, or on damage
+      if ( p()->sets->has_set_bonus( WARLOCK_AFFLICTION, T28, B4 ) )
+      {        
+        auto tdata = this->td( s->target );
+        bool tierDotsActive = tdata->dots_agony->is_ticking();
```
scamille: I would express this with a single boolean statement instead of (mis)using the bitwise operator.

engine/class_modules/warlock/sc_warlock_affliction.cpp
```diff
@@ -606,6 +650,19 @@ struct malefic_rapture_t : public affliction_spell_t
         p()->buffs.malefic_wrath->trigger();
         p()->procs.malefic_wrath->occur();
       }
+
+      if (p()->sets->has_set_bonus( WARLOCK_AFFLICTION, T28, B2 ) )
+      {
+        timespan_t dot_extension =  p()->sets->set( WARLOCK_AFFLICTION, T28, B2 )->effectN( 2 ).time_value() * 1000;
+        warlock_td_t* td = p()->get_target_data( target );
+
+        if ( td )
```
scamille: the pointer returned by get_target_data should always be non-null.

engine/class_modules/warlock/sc_warlock_affliction.cpp
Fatherdotter: Would this be functionally different from the existing implementation in any way? As far as I am aware the usage of 

```c++
bool val = A;
val &= B;
val &= C;
```

Is functionally equivalent to 

```c++
bool val = A && B && C;
```

With the exceptions that they do not short-circuit like a single logical statement would, and thus the minor overhead of guaranteed evaluation of both B and C in exchange for (in my opinion) clarity of the statements to be evaluated in this case. Considering I have recently had yourself and an existing simc member speak against using a similar approach in this project I am not sure the reason for this change, but welcome any advice to the contrary.

Edit: Clarified example.

engine/class_modules/warlock/sc_warlock_affliction.cpp
Fatherdotter: Considering you have, quite recently, argued against saving multiple checks per-action as being an unnecessary optimization surely a single sanity check to prevent dereferencing a null pointer is a reasonable addition to the program?

There is existing similar usage for this exact purpose in simc, and it should add negligible overhead.

vituscze: If `get_target_data` returns a nullptr, something is wrong and it'd be better to crash than to silently ignore it.

engine/class_modules/warlock/sc_warlock_affliction.cpp
vituscze: Problem here isn't functionality but meaning. `&` is a bitwise operation (and so is `&=`) but you want a logical operation. Sadly C++ doesn't have logical version of `&=` such as `&&=`, so you're left with either `val = A && B;` or `val = A; val = val && B;`.

By the way, those two are not the same when dealing with expressions that can be converted to `bool`:
```cpp
bool b = true;
b &= 2;
// b is false

bool c = true;
c = c && 2;
// c is true
```

Fatherdotter: I am aware of the bitwise nature of the assignment operators here but did not see how it was functionally different in the event that we are working purely with boolean data types. Since all of these operators are boolean in nature ( is_ticking() ) and would have no reason to become non-boolean values in the future, this interaction should not possibly arise. The only difference in functionality is the short-circuit nature being lost in the expression.

engine/class_modules/warlock/sc_warlock_affliction.cpp
Fatherdotter: If that is the decision made here, then I will happily do a separate commit to remove these checks in whole from the warlock modules (rather than removing one instance in this commit, and others in a separate commit.)

engine/class_modules/warlock/sc_warlock_affliction.cpp
vituscze: It's the same reason we don't use `+` or `*` on booleans (even though it technically works as disjunction/conjunction when all the arguments are booleans): it's just the wrong operator to use.

Fatherdotter: In the absence of a logical assignment operator for c++, for someone who prefers the clarity of having the conditions clearly spelled out there is no other functional alternative though.

I understand the concern here, and although I cannot conceive of a situation where this code fails the change can be made simply to meet conventions at the expense of clarity.

engine/class_modules/warlock/sc_warlock_affliction.cpp
scamille: I don't have anything against checking it per se, but would then change it to an assert or throw instead of silently doing nothing.

Fatherdotter: Resolved by the most recent commit. After discussing, we have elected to handle it within this commit and not take the route of an assert or throw.

engine/class_modules/warlock/sc_warlock_affliction.cpp
```diff
@@ -935,7 +934,9 @@ void warlock_t::create_buffs_affliction()
 
   buffs.malefic_wrath = make_buff( this, "malefic_wrath", find_spell( 337125 ) )->set_default_value_from_effect( 1 );
 
+  // TODO: Replace hardcoded duration with query from spell data, replace buff spell (363953) with actuall buff spell ID
   buffs.calamitous_crescendo = make_buff( this, "calamitous_crescendo", find_spell( 363953 ) )
+                                   ->set_duration(timespan_t::from_seconds(10))
```
vituscze: If you want, you can use `10_s` instead of `timespan_t::from_seconds( 10 )`

engine/class_modules/warlock/sc_warlock_affliction.cpp
```diff
@@ -935,8 +935,8 @@ void warlock_t::create_buffs_affliction()
   buffs.malefic_wrath = make_buff( this, "malefic_wrath", find_spell( 337125 ) )->set_default_value_from_effect( 1 );
 
   buffs.calamitous_crescendo = make_buff( this, "calamitous_crescendo", find_spell( 364322 ) )
-                                   ->set_chance( talents.drain_soul->ok() ? p()->sets->set(WARLOCK_AFFLICTION, T28, B4 )->effectN( 2 ).percent() 
-                                                                          : p()->sets->set(WARLOCK_AFFLICTION, T28, B4 )->effectN( 1 ).percent() );
+                                   ->set_chance( talents.drain_soul->ok() ? this->sets->set(WARLOCK_AFFLICTION, T28, B4 )->effectN( 2 ).percent()
```
vituscze: You can drop `this` and just use `sets->set(...)`

Fatherdotter: Thank you for the advice. Dropped this from a recent commit as setting the proc chance was actually a breaking change.

engine/class_modules/warlock/sc_warlock_affliction.cpp
Fatherdotter: Appreciate the advice, in the future I will try to put it to use but I do not intend to set these values.

PR #6242, opened 2021-12-02
[Warlock] Set target for Hand of Guldan impact

Fatherdotter:
Sets the target for the Hand of Guldan impact spell.


Fatherdotter:
Have received a request to convert this to `execute_on_target`

PR #6244, opened 2021-12-02
[Player] implement on-kill callbacks

gastank:
Callback triggered on all non-sleeping players on an enemy mob demise()

use `player_t::register_on_kill_callback( void(player_t*) )` to register a new callback


scamille:
What was the reason to not just use the target data on demise callback with a is_enemy() filter?

PR #6255, opened 2021-12-14
Paladin Protection DPS Optimization

xizbow1:
Increased HOW prio in AH, and only cast vanq hammer if using seraphim and DBG, or on cd otherwise.
This is the same pull request that I previously had, but I cleaned it up so that it isn't 20 commits long.


xizbow1:
nvm I messed something up this is bad and wrong

PR #6266, opened 2021-12-24
[Stats] allow actions to use custom stat objects

gastank:
`player_t::get_stats()` called with a valid action will use virtual method
`action_t::new_stats()`, which can be overriden to create a custom stat object
derived from `stats_t`.

`player_t::get_stats()` called with a null action will default to creating
a base `stats_t` object.


scamille:
What is the reason/motivation for this, if I may ask?

PR #6273, opened 2022-01-10
[Player] Add Generic Options for Precombat Buff and Cooldown States

### LINKED ISSUE #6270: Frost Mage Cold Front Legendary Prepull Implementation
### Is your feature request related to a problem? Please describe.
There does not appear to be a simple way to simulate a pull where the Cold Front legendary is precharged and ready to fire with the next cast of flurry or frostbolt.

### Describe the solution you'd like
Something similiar to what was done for Disciplinary Command or Arcane Harmony legendaries would be useful.

Disciplinary Command->
mage.prepull_dc=<0/1>

Arcane Harmony->
mage.prepull_harmony_stacks=<number>

Dorovon:
This adds a generic system to let users specify the precombat state of buffs and cooldowns for an actor.
Valid options include:
`override.precombat_state=cooldown.<name>=<remaining_cooldown_duration>`
`override.precombat_state=buff.<name>.stack=<buff_stacks>`
`override.precombat_state=buff.<name>.value=<buff_value>`
`override.precombat_state=buff.<name>.remains=<remaining_buff_duration>`

By default, this will simply `start` the cooldown with that name or `execute` the buff with that name.

Individual buffs and cooldowns can have this system disabled by setting their `allow_precombat` member to false. Buffs also have `buff_t* buff_t::set_allow_precombat( bool )` to set this.

Class modules may implement specialized behavior through overriding `player_t::add_precombat_buff_state` and `player_t::add_precombat_cooldown_state` to add support for buffs and cooldowns that will not work with the default system.

Any simulation run with this option will output the following warning due to how easy it is to misuse this system.
> Warning: The 'override.precombat_state' option may not be fully supported for all buffs and cooldowns and and may produce incorrect or misleading results.

### DIFF
--- a/engine/player/player.cpp
+++ b/engine/player/player.cpp
@@ -3033,11 +3033,37 @@ void player_t::init_finished()
   if ( !precombat_state_map.empty() )
   {
     sim->error( "Warning: The 'override.precombat_state' option may not be fully supported for all buffs and cooldowns and and may produce incorrect or misleading results." );
-    std::unordered_map<buff_t*, std::tuple<int, double, timespan_t>> precombat_buff_state;
+
+    struct buff_state_t
+    {
+      int stacks = -1;
+      double value = buff_t::DEFAULT_VALUE();
+      timespan_t duration = timespan_t::min();
+    };
+
+    std::unordered_map<buff_t*, buff_state_t> precombat_buff_state;
+    auto update_buff_state = [ this, &precombat_buff_state ] ( std::string_view buff_name, std::string_view type, std::string_view value )
+    {
+      buff_t* buff = buff_t::find( this, buff_name );
+      if ( !buff )
+      {
+        sim->error( "No buff found for 'override.precombat_state' buff expression: '{}'", buff_name );
+        return;
+      }
+
+      if ( type == "stack" )
+        precombat_buff_state[ buff ].stacks = util::to_int( value );
+      else if ( type == "value" )
+        precombat_buff_state[ buff ].value = util::to_double( value );
+      else if ( type == "remains" )
+        precombat_buff_state[ buff ].duration = timespan_t::from_seconds( util::to_double( value ) );
+      else
+        throw std::invalid_argument( fmt::format( "Invalid 'override.precombat_state' buff expression type: '{}'", type ) );
+    };
 
     for ( const auto& v : precombat_state_map )
     {
-      auto splits = util::string_split<std::string>( v.first, "." );
+      auto splits = util::string_split( v.first, "." );
 
       if ( splits.size() < 2 )
       {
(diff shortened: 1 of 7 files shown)


engine/player/player.cpp
```diff
@@ -3029,6 +3029,123 @@ void player_t::init_finished()
       }
     } );
   }
+
+  if ( !precombat_state_map.empty() )
+  {
+    sim->error( "Warning: The 'override.precombat_state' option may not be fully supported for all buffs and cooldowns and and may produce incorrect or misleading results." );
+    std::unordered_map<buff_t*, std::tuple<int, double, timespan_t>> precombat_buff_state;
```
scamille: Instead of using a tuple, wouldn't it make the code clearer and easier to read if you defined a struct for it?
```
struct buff_state_t{
int stacks;
double value;
timespan_t duration
};
std::unordered_map<buff_t*,buff_state_t> precombat_buff_state;
```

engine/player/player.cpp
```diff
@@ -3029,6 +3029,123 @@ void player_t::init_finished()
...
+
+    for ( const auto& v : precombat_state_map )
+    {
+      auto splits = util::string_split<std::string>( v.first, "." );
```
scamille: I am fairly sure you could use the std::string_view version of util::string_split here.

engine/player/player.cpp
```diff
@@ -3029,6 +3029,123 @@ void player_t::init_finished()
...
+        }
+        else if ( splits [ 2 ] == "remains" )
+        {
+          std::get<2>( precombat_buff_state[ buff ] ) = timespan_t::from_seconds( util::to_double( v.second ) );
+          continue;
+        }
+
+        // If no part of the newly created state was set, it needs to be
+        // removed to avoid triggering the buff with all default values.
+        if ( new_buff_state )
```
scamille: Adding the buff state and removing it again seems like very ugly and error prone control flow.

I see that it isn't as easy to merge the 3 statements because they parse v.second into different parts of the buff state, but it should still be possible to improve it a bit.
If you have a function
```
using buff_state_t = std::tuple<int, double, timespan_t>; // alias for your tuple, but see other comment
buff_state_t& get_buff_state(buff_t* buff)
{
  bool new_buff_state = precombat_buff_state.find( buff ) == precombat_buff_state.end();

  if ( new_buff_state )
  {
    // Initialize the values for a buff's stack, value, and duration
    // parameters to the default parameter values of buff_t::execute.
    precombat_buff_state[ buff ] = { -1, buff_t::DEFAULT_VALUE(), timespan_t::min() };
  }
  return precombat_buff_state[ buff ];
}
```

You can then do:
```

        if ( splits[ 2 ] == "stack" )
        {
          std::get<0>( get_buff_state( buff ) ) = util::to_int( v.second );
          continue;
        }
```

And you can even wrap the 3 ifs into a function so you can simple return statements instead of continue. (But that's just my taste)

engine/player/player.cpp
```diff
@@ -3029,6 +3029,123 @@ void player_t::init_finished()
...
+          sim->error( "No cooldown found for 'override.precombat_state' cooldown expression: '{}'", name );
+          continue;
+        }
+
+        timespan_t duration = timespan_t::from_seconds( util::to_double( v.second ) );
+        add_precombat_cooldown_state( cd, duration );
+      }
+      else
+      {
+        sim->error( "Invalid type '{}' for 'override.precombat_state' option.", type );
```
scamille: Might be better to throw here as well, if you are also doing that for the `!buff->allow_precombat` case further down. I would definitely treat this as invalid user input right away.

engine/player/player.cpp
```diff
@@ -3029,6 +3029,123 @@ void player_t::init_finished()
...
+        {
+          precombat_buff_state.erase( buff );
+        }
+
+        sim->error( "Invalid 'override.precombat_state' buff expression: '{}'", v.first );
+      }
+      else if ( type == "cooldown" )
+      {
+        if ( splits.size() != 2 ){
+          sim->error( "Invalid 'override.precombat_state' cooldown expression: '{}'", v.first );
```
scamille: I would also throw here, it is not something that is sometimes valid input and sometimes isn't.

engine/player/player.cpp
```diff
@@ -3029,6 +3029,123 @@ void player_t::init_finished()
...
+        }
+
+        // If no part of the newly created state was set, it needs to be
+        // removed to avoid triggering the buff with all default values.
+        if ( new_buff_state )
+        {
+          precombat_buff_state.erase( buff );
+        }
+
+        sim->error( "Invalid 'override.precombat_state' buff expression: '{}'", v.first );
```
scamille: And maybe throw here as well.

engine/player/player.cpp
```diff
@@ -3029,6 +3029,123 @@ void player_t::init_finished()
...
+        continue;
+      }
+
+      auto type = splits[ 0 ];
+      auto name = splits[ 1 ];
+
+      if ( type == "buff" )
+      {
+        if ( splits.size() != 3 ){
+          sim->error( "Invalid 'override.precombat_state' buff expression: '{}'", v.first );
```
scamille: This might be better handled by throwing an exception since you always except it to have exactly 3 splits for buffs.

engine/player/player.cpp
Dorovon: Yeah, a struct is probably a bit better. It works with default values as well, which is nice.

vituscze (approved):

PR #6278, opened 2022-01-12
[Gear] initial implementation of unity covenant legendary belt

gastank:


Dorovon:
There is a covenant field in RuneforgeLegendaryAbility.db2 added in 9.1. I haven't closely checked it, but we might be able to do this more generically if we export that.

gastank:
That should work, can get rid of the entire 'init' and just check for unity id/effect if `find_runeforge_legendary()` is looking for a power with a covenant id.

PR #6301, opened 2022-01-28
{fmt} 8.1.1.

scamille:
Upgrade fmt to version 8.1.1


nuoHep:
> Not sure if this would actually have resulted in problem when formatting the enums, but it can't hurt to explicitly cast them to a integer type as that is the desired way to format them in those places that I adjusted.

Afair {fmt} has default formatter for all enums that effectively does `static_cast<std::underlying_type_t<Enum>>()`

scamille:
> 

Hmm I did get error messages from fmt, otherwise I would not have adjusted them. If fmt has a fallback enum formatter that is not used when checking the constexpr format string, I find that a bit strange, but of course that is a possibility.

PR #6302, opened 2022-01-28
Workflows: Add C++20 builds

scamille:


scamille:
This might need #6301 merged first to succeed.

.github/workflows/main.yml
```diff
@@ -17,18 +17,22 @@ env:
...
         include:
           - cc: gcc-7
             cxx: g++-7
             packages: gcc-7 g++-7
           - cc: gcc-10
             cxx: g++-10
             packages: gcc-10 g++-10
+        exclude:
+          - cc: gcc-7
+            cppVersion: 17
```
nuoHep: 20?

scamille: Yes that was the original idea. Somewhere along the way I must have done the exact opposite.

PR #6303, opened 2022-01-29
Refactor some chart and highchart code

scamille:
Also add a GetStringView function to rapidjson stringbuffer, since the string size is known in the GetString() function, so we can use a string-view instead of c-string.

More color constexpr and use optional<color::rgb> where appropriate.

### DIFF
--- a/engine/report/color.hpp
+++ b/engine/report/color.hpp
@@ -46,15 +46,69 @@ namespace color
     std::string hex_str() const;
     operator std::string() const; // meh...
 
-    rgb& adjust(double v);
-    rgb adjust(double v) const;
-    rgb dark(double pct = 0.25) const;
-    rgb light(double pct = 0.25) const;
-    rgb opacity(double pct = 1.0) const;
 
-    rgb& operator=( util::string_view color_str );
-    rgb& operator+=( rgb other );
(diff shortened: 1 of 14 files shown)


engine/report/color.hpp
```diff
@@ -46,15 +46,69 @@ namespace color
...
+      unsigned mix_b = ( b_ + other.b_ ) / 2;
+
+      r_ = static_cast<uint8_t>( mix_r );
+      g_ = static_cast<uint8_t>( mix_g );
+      b_ = static_cast<uint8_t>( mix_b );
+
+      return *this;
+    }
+    
+    rgb operator+( rgb other ) const
```
nuoHep: Missing constexpr?

Though I'm not sure if it really makes sense to make all of these constexpr. While constexpr is nice and easy in this case I don't really see a situation where the need can arise to use is at compile time. And in non-constant evaluation context constexpr doesn't matter. But it doesn't hurt either so whatever.

nuoHep: Oh, one thing I remembered, you can slap inline on all of the named colors further below while you are at it now that we are on c++17. Just so that we don't get extra copies in case they are odr-used. It's just 8 bytes per copy, but still.

PR #6308, opened 2022-01-31
[Priest] Hungering Void should affect all pets as a Shadow Priest

seanpeters86:
Editing my previous commit by moving the pet mod to `pet_t` so pets from soulbind spells/etc will get the effect as well.


Saeldur (approved):
LGTM

PR #6309, opened 2022-02-02
[Priest] refactor how we delay the spawn of the t28 pet

seanpeters86:
With the previous code there was an issue if we consumed a Dark Thought within the 1s delay between consuming a Dark Thought and Spawning the T28 pet. Something like this:

1. Consume Dark Thought
2. Schedule T28 Spawn in 1s
3. Consume another Dark Thought before the spawn has happened

While this situation is very rare and hard to reproduce in-game it is possible with enough haste and would be easier if they ever make Dark Thought stack more. In-game if this happens it just added another pet's worth of duration to the currently spawning pet, so this PR mimics that in simc as best as we can.

Example output from debug showing how this works:
```
32.730 Player 'Base' loses Buff dark_thought
32.730 Player 'Base' consumes Dark Thought, delaying your_shadow spawn by 1 second.
33.571 Player 'Base' activated Dark Thoughts using Action mind_flay with 0.09 chance with 3 dots
33.686 Player 'Base' loses Buff dark_thought
33.686 Player 'Base' extends future your_shadow by 6.000 seconds for a new total of 12.001. Spawns in 0.044 seconds.
33.730 Player 'Base' summons your_shadow for 12.001.
33.730 New Action Execute Event: Player 'Base_your_shadow' Action torment_mind time_to_execute=0.500 (target=Fluffy_Pillow, marker=0)
34.230 Player 'Base_your_shadow' direct amount for Action torment_mind_tick: amount=1372.3690060119113 initial_amount=1372.3690060119113 weapon=0 base=0 s_mod=0.34 s_power=2273 a_mod=0 a_power=0 mult=1.7757938536941478 w_mult=0 w_slot_mod=1 bonus_da=0
```


engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -177,14 +177,24 @@ struct mind_blast_t final : public priest_spell_t
         {
           pet->adjust_duration( your_shadow_duration );
         }
+        else if ( priest().t28_4pc_summon_event &&
```
scamille: Hmm I don't like the remains check here. Please remove it but set the event pointer to null after you spawn the pet.

You can't assume that just because remains returns 0 that the event was already executed.

scamille (changes requested):

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -177,14 +177,23 @@ struct mind_blast_t final : public priest_spell_t
...
           // This is intentional due to the way the pet spawns attached to the player
-          // NOTE: This could cause 2 pets to be scheduled if you consume a Dark Thought proc while waiting for the first pet to spawn
-          // NOTE: While not ideal this is basically not possible unless Dark Thought buffs start stacking.
-          sim->print_debug( "{} consumes Dark Thought, delaying Your Shadow spawn by 1 second.", priest() );
-          make_event( *sim, timespan_t::from_seconds( 1 ), [ this ] { priest().pets.your_shadow.spawn(); } );
+          sim->print_debug( "{} consumes Dark Thought, delaying your_shadow spawn by 1 second.", priest() );
+          // Add 1ms to ensure pet is dismissed after last dot tick.
+          priest().t28_4pc_summon_duration = your_shadow_duration + timespan_t::from_millis( 1 );
+          priest().t28_4pc_summon_event    = make_event( *sim, timespan_t::from_seconds( 1 ), [ this ] {
+            priest().pets.your_shadow.spawn( priest().t28_4pc_summon_duration );
```
scamille, line 195:
```suggestion
            priest().pets.your_shadow.spawn( priest().t28_4pc_summon_duration );
            priest().t28_4pc_summon_event = nullptr;
```

navv1234: You likely want to reset the event in priest_t::reset also, just in case.

seanpeters86: i included this in the `pet_t::arise` method but i like it here better. And i added it to preist_t:reset as well in a previous commit

scamille (approved):

PR #6320, opened 2022-02-07
Refactor cmake files

scamille:
- Try to use some more target based "modern" cmake
- Try to move as much into the engine/qt cmake files that can be handled in there
- Remove unnecessary CMAKE_CURRENT_[]_DIR variables
- Move locale files to qt/ folder
- Make flat_install_folder an option
- Proper warnings for all targets and hopefully toolchains


scamille:
Well this moves the simc binary from /simc to /engine/simc , thus CI that tries to still use the old path fails.

Question is if we are ok with this breaking change of the build location so that we can move the cli into the proper cmake list.

CMakeLists.txt
```diff
@@ -31,52 +30,50 @@ if(${CMAKE_GENERATOR} STREQUAL "Ninja")
...
+if (SC_USE_FLAT_INSTALL)
+  set(SIMC_INSTALL_BIN ".")
+  set(SIMC_INSTALL_SHARED ".")
+else()
+  set(SIMC_INSTALL_BIN "bin")
+  set(SIMC_INSTALL_SHARED "share/SimulationCraft/SimulationCraft")
 endif()
 
-# output a compile_commands.json used for editor tooling
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
nuoHep, line 40: Is this the default now? There is no reason not to do this. And considering the quality of cmake docs I don't really want to check, so I'd leave this.

CMakeLists.txt
```diff
@@ -1,10 +1,5 @@
 # Projects Settings   
-cmake_minimum_required (VERSION 3.10...3.21)
-
-# allow enabling lto from the command line
-if(POLICY CMP0069)
-    cmake_policy(SET CMP0069 NEW)
-endif()
```
nuoHep, line 7: I take it this is the default for the min version we require now?

CMakeLists.txt
```diff
@@ -31,52 +30,50 @@ if(${CMAKE_GENERATOR} STREQUAL "Ninja")
...
+  set(SIMC_INSTALL_BIN "bin")
+  set(SIMC_INSTALL_SHARED "share/SimulationCraft/SimulationCraft")
 endif()
 
-# output a compile_commands.json used for editor tooling
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+# helper function to enable warnings on targets
+function(sc_enable_warnings target)
+  target_compile_options(${target} PRIVATE
+    $<$<CXX_COMPILER_ID:MSVC>:/W4>
```
nuoHep: You *really* don't want to enable /W4, it's simply obnoxious.

CMakeLists.txt
```diff
@@ -31,52 +30,50 @@ if(${CMAKE_GENERATOR} STREQUAL "Ninja")
...
+endfunction()
+
+function(sc_common_compiler_options target)
+  sc_enable_warnings(${target})
+  
+  # Require /permissive- when building with Visual Studio
+  target_compile_options(${target} PRIVATE
+    $<$<CXX_COMPILER_ID:MSVC>:/permissive->
+  )
+endfunction()
```
nuoHep, line 51: I'm not sure it's worth making a function for this considering you then call it for all targets. But even if it is there is no point in making it 2 functions.

nuoHep (commented):
> Question is if we are ok with this breaking change of the build location so that we can move the cli into the proper cmake list.

I'm not sure tbh. I don't even think the cli is not in the proper list atm as I view the cli as the main "output" of the project and so it's perfectly fine that it's in the root cmake file. It's also nice that the paths are short ($build-dir/simc).

But I don't care that much, so... At least if you do change it - update the ci in this mr so it stays green.

nuoHep:
Also, you should ask about moving the cli on discord tbh. This can break workflows for folks using cmake.

CMakeLists.txt
scamille: My understanding was that a project usually shouldn't care about setting such an option. If you as a user/ide need the compile commands you can set it during the configure step. If you don't need/want it then the project shouldn't force it on.
But we can of course also just leave it in, maybe as a cache variable so that it could still be turned off.

CMakeLists.txt
scamille: Yes, it is for 3.9+

CMakeLists.txt
scamille: Right I should probably check the output on msvc first.
So /W3 would be better?

scamille:
> Also, you should ask about moving the cli on discord tbh. This can break workflows for folks using cmake.

Yes you are absolutely right. I plan to move it back to the root anyway, just haven't got the time to continue on this. My last comment was as much for myself, after seeing why CI failed.

CMakeLists.txt
scamille: While there aren't any plans to ever export a simc engine library, I thought it couldn't hurt to properly set it as a target property.

You are right in that it is a bit verbose :-)

CMakeLists.txt
scamille: Seems level 3 is what we use in the handmade VS projects.

CMakeLists.txt
nuoHep: Well... kind of. Theoretically sure, a user can pass in `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` when invoking CMake if they need it.
Realistically though, considering the amount of tooling that can make use of it (or even *requires* it) currently (e.g. everything clang based requires it, vscode cpp plugin works miles better with it) and no downsides to having it on by default you may as well always generate it and make life easier for everyone.

PR #6347, opened 2022-02-16
[Paladin] Prot T28 4pc bonus update

OrcinusDrake:
Now has a chance to proc on any damage taken rather than guaranteed on blocks. Also implemented the icd.


metacursor:
Ok made a change so that it only procs on blockable attacks. In-game it procs on some sources of damage but not on others: like it doesn't tend to proc on aoe abilities. Simc doesn't make that distinction. So this change makes the 4pc perform more realistically in sims despite not following exactly what it does in game.

OrcinusDrake:
Ok so made a change to the T28 4pc to only proc on blockable damage. In-game it says it procs on any damage, but it tends not to proc on aoe damage for example. Simc doesn't really distinguish between these different types of damage, so this change makes it perform more realistically despite not following exactly what it does in game.
Sims also tend to use a magic dot which would greatly inflate the value of the 4pc, but such dots are pretty rare in-game. You could just use a different enemy profile, but this is an alternative solution.

PR #6361, opened 2022-02-17
[Gear] Bells of the Endless Feast

indiche:
Trinket seems to be gaining between 5-7 stacks per proc. There is also this line "Effects that cause your target to move generate bonus stacks." in the tooltip, but I'm unsure how to test it and if it's even relevant for the sim.

### DIFF
--- a/engine/player/unique_gear_shadowlands.cpp
+++ b/engine/player/unique_gear_shadowlands.cpp
@@ -3427,6 +3427,56 @@ void elegy_of_the_eternals( special_effect_t& effect )
   std::rotate( vec->rbegin(), vec->rbegin() + 1, vec->rend() );
 }
 
+// id=367336 - Brood of the Endless Feast	(driver - damage on effect 1)
+// id=368585 - Scent of Souls (target debuff)
+// id=368587 - Rabid Devourer Chomp (physical damage at max stacks)
+void bells_of_the_endless_feast( special_effect_t& effect )
+{
+  struct rabid_devourer_chomp_t : generic_proc_t
+  {
+    rabid_devourer_chomp_t( const special_effect_t& effect )
+      : generic_proc_t( effect, "rabid_devourer_chomp", effect.player->find_spell( 368587 ) )
+    {
+      base_dd_min = base_dd_max = effect.driver()->effectN( 1 ).average( effect.item );
+    }
+  };
+
+  struct brood_of_the_endless_feast_cb_t : dbc_proc_callback_t
+  {
+    action_t* damage;
+
+    brood_of_the_endless_feast_cb_t( const special_effect_t& effect )
+      : dbc_proc_callback_t( effect.player, effect ),
+        damage( create_proc_action<rabid_devourer_chomp_t>( "rabid_devourer_chomp", effect ) )
+    {
+    }
+
+    void execute( action_t* a, action_state_t* s ) override
+    {
+      dbc_proc_callback_t::execute( a, s );
+
+      actor_target_data_t* td = a->player->get_target_data( s->target );
+
+      // NOTE: Damage triggers on the next tick of the debuff after it reaches max stacks, but I'm not sure if it's
+      // worth modeling that properly, so we instead trigger it as soon as it reaches max stacks.
+      if ( td->debuff.scent_of_souls->at_max_stacks() )
+      {
+        td->debuff.scent_of_souls->expire();
+        damage->execute_on_target( s->target );
+      }
+      else
+      {
+        // NOTE: I couldn't find it anywhere in the spell data, but from testing, the target seems to be gaining between
+        // 5 and 7 stacks of the debuff on each proc.
+        double stacks = rng().range( 5, 7 );
+        td->debuff.scent_of_souls->trigger( stacks );
+      }
+    }
+  };
+
+  new brood_of_the_endless_feast_cb_t( effect );
+}
+
 // Weapons
 
 // id=331011 driver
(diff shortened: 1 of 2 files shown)


engine/player/unique_gear_shadowlands.cpp
```diff
@@ -4990,6 +5038,19 @@ void register_target_data_initializers( sim_t& sim )
...
   } );
 
+  // Bells of the Endless Feast
+  sim.register_target_data_initializer( []( actor_target_data_t* td ) {
+    if ( unique_gear::find_special_effect( td->source, 367336 ) )
+    {
+      assert( !td->debuff.scent_of_souls );
+
+      td->debuff.scent_of_souls = make_buff<buff_t>( *td, "scent_of_souls", td->source->find_spell( 368585 ) );
+      td->debuff.scent_of_souls->reset();
```
nuoHep, line 5108: Looking at the spell data the buff has a dummy periodic effect. Current implementation doesn't need it so you can `set_period( 0_ms )` here to remove a pointless tick event every 1s.

There is also a possibility that the damage is driven by these tick events. As in the effect simply triggers the debuff, and the debuff checks its stacks on each tick and executes the damage if it's at max stacks.
Dunno if it's the case and if we should "properly" model that even if it is.

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -3422,6 +3422,53 @@ void elegy_of_the_eternals( special_effect_t& effect )
...
+      if ( td->debuff.scent_of_souls->at_max_stacks() )
+      {
+        td->debuff.scent_of_souls->expire();
+        damage->set_target( s->target );
+        damage->execute();
+      }
+      else
+      {
+        double stacks = rng().range( 5, 7 );
+        td->debuff.scent_of_souls->execute( stacks );
```
nuoHep:
```suggestion
        td->debuff.scent_of_souls->trigger( stacks );
```
Or a comment why we are going straight to `execute()`.

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -3422,6 +3422,53 @@ void elegy_of_the_eternals( special_effect_t& effect )
...
+
+      if ( td->debuff.scent_of_souls->at_max_stacks() )
+      {
+        td->debuff.scent_of_souls->expire();
+        damage->set_target( s->target );
+        damage->execute();
+      }
+      else
+      {
+        double stacks = rng().range( 5, 7 );
```
nuoHep, line 3471: I'd add these as named constants, but that's up to you.
*Definitely* add a comment why these vales & rng as there is nothing in the spells indicating that.

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -3422,6 +3422,53 @@ void elegy_of_the_eternals( special_effect_t& effect )
...
+    {
+      dbc_proc_callback_t::execute( a, s );
+
+      actor_target_data_t* td = a->player->get_target_data( s->target );
+
+      if ( td->debuff.scent_of_souls->at_max_stacks() )
+      {
+        td->debuff.scent_of_souls->expire();
+        damage->set_target( s->target );
+        damage->execute();
```
nuoHep:
```suggestion
        damage->execute_on_target( s->target );
```

engine/player/unique_gear_shadowlands.cpp
indiche: It seemed that it triggers on the tick rather than immediately after reaching max stacks. The only scenario where this is relevant is if you would proc the trinket directly after getting the max stacks. The current model would reset the debuff and award you with new stacks. 

With the actual in-game behavior, there is a possibility that you are still on max stacks since the debuff hasn't ticked yet by the time you proc the trinket again, meaning the debuff hasn't reset yet, and you would lose new stacks. 

I'm not really sure we need to model it this way because of it.

nuoHep: Hm, in that case I'd model the in-game behavior tbh. Considering it should be easy.

Proc action should still be created in the special effect initialization function (because actions have to be created before the simulation starts) but it can be dropped from the custom `dbc_proc_callback_t` with only debuff triggering left there.

Then in the target data initializer just find the proc action and slap a tick callback on the debuff triggering it. Smth like
```cpp
auto damage = td->source->find_action( "rabid_devourer_chomp" );
assert( damage );
[...]
td->debuff.scent_of_souls->set_tick_callback( [ damage ]( buff_t* b, int, timespan_t ){
    if ( b->at_max_stacks() ) {
      damage->execute_on_target( b->player );
      b->expire( 1_ms );
    }
  } );
```
`1_ms` in the expire is needed because we can't directly expire buffs from callbacks, so there will be a 1ms window where you can lose stacks, but eh.

I *think* this should work and properly model in-game behaviour.

indiche: @nuoHep hmm, I've implemented it as suggested, but I've noticed some issues with the `at_max_stacks` method. Once the debuff reaches max stacks method continues returning true, even tho debuff was expired. I can see the stack fall off, but the damage portion is triggered on each tick never the less.




As you can see, the uptime on 100 stacks is 5%, but the damage is triggered over 600 times.

nuoHep: Hm, interesting.

Can you please rebase it once more so we can merge this as is. I'll try to look at it later why this happens.

indiche: I've rebased it. I also saw that you whitelisted the damage spell id and removed the commit with the generated spell data since it already contains everything.

nuoHep:
Thanks

PR #6371, opened 2022-02-18
[Shaman] childishforlife/shaman_surge_of_power_se_st

Childishforlife:
Use surge of power procs on lightning bolts instead of LvB's as 100% uptime is already achieved


engine/class_modules/sc_shaman.cpp
```diff
@@ -9908,6 +9909,7 @@ void shaman_t::init_action_list_elemental()
         "60&spell_targets.chain_lightning>=2&(!runeforge.echoes_of_great_sundering.equipped|buff.echoes_of_great_"
         "sundering.up)|spell_targets.chain_lightning<2&buff.wind_gust.stack>=18&(!runeforge.echoes_of_great_sundering."
         "equipped|buff.echoes_of_great_sundering.up)&maelstrom>=60" );
+    se_single_target->add_action( this, "Lightning Bolt", "if=buff.surge_of_power.up");
```
Bloodmallet: 1. can be accidentally munched by Flame Shock/Primordial Wave lines above
2. quite likely to consume Echoing Shock if taken

Are you fine with the above?

Childishforlife: Probably not ideal, I will change it around a little so it can't be munched/consumed that way and see if its an improvement!

Bloodmallet (approved):

Bloodmallet (approved):

Bloodmallet:
Thank you for providing Surge of Power support.

PR #6409, opened 2022-02-23
Cthuulu destro apl optimizations

Cthuulu:
Avoid shard capping
Avoid using ritual proc instantly if using madness and low on soul shards
End of fight optimization, prefer spending shards and last conflag charge
Time to die change to avoid wasting uptime of trinkets

### DIFF
--- a/profiles/Tier28/T28_Warlock_Demonology.simc
+++ b/profiles/Tier28/T28_Warlock_Demonology.simc
@@ -45,6 +45,7 @@ actions+=/variable,name=buff_sync_cd,op=set,value=variable.next_tyrant_cd,if=!va
 actions+=/variable,name=buff_sync_cd,op=set,value=cooldown.decimating_bolt.remains_expected,if=variable.use_bolt_timings
 actions+=/call_action_list,name=trinkets
 actions+=/call_action_list,name=ogcd,if=(!variable.use_bolt_timings&pet.demonic_tyrant.active)|(variable.use_bolt_timings&buff.shard_of_annihilation.up&(!talent.power_siphon.enabled|buff.power_siphon.up))
+actions+=/implosion,if=target.time_to_die<2*gcd
 actions+=/call_action_list,name=opener,if=time<variable.first_tyrant_time
 actions+=/interrupt,if=target.debuff.casting.react
 actions+=/doom,if=refreshable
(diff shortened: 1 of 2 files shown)


Azevara:
actions.cds+=/use_items,if=pet.infernal.active|target.time_to_die<21
This change is most likely neutral given its small impact, but as it's human-understandable it should be fine to keep
https://www.raidbots.com/simbot/report/u5BTb9tHD4CHKMmu4RjB5d

Azevara:
Destruction End of Fight combined changes is ~0.3% increase
https://www.raidbots.com/simbot/report/4ibvD1Qh5T1gFEMzFsE4to
The gain appears to come strictly from the Chaos Bolt line (not shown in the above link). Similar to previous comment, the conflagration line is simple enough for human play to keep as well.

Azevara:
actions+=/chaos_bolt,if=active_enemies<3&(pet.infernal.active|pet.blasphemy.active)&soul_shard>=4.6
While this line (placed above Cataclysm) is executed in the standard sim ~3 times on average, it is mostly a duplicate of a similar line with a slightly lower priority, and the two versions have negligible DPS difference with no change in Soul Shard overflow.
https://www.raidbots.com/simbot/report/9zR75CPje7VHZj25dJkFGa

Azevara:
actions+=/chaos_bolt,if=(pet.infernal.active|buff.rain_of_chaos.remains>cast_time)&!(buff.ritual_of_ruin.up&buff.madness_of_the_azjaqir.remains>3.7&soul_shard<1.5)
The latter half of this expression always evaluates to true, so it's unnecessary. The result is this change has no noticeable effect.
https://www.raidbots.com/simbot/report/oQE6orukct9MxBSNKk7poi

Azevara:
Altogether this has a small increase due to end of fight gains, with some additional expressions that might be catching edge cases not tested with the current T28 profile. A sample sim of this comparison using most of the above changes can be found at this link: https://www.raidbots.com/simbot/report/p2YqEzaHhkPRuiT5xNDaWs

That being said, changes to the files in this PR will not actually end up being utilized, as the changes need to be made in the sc_warlock_destruction.cpp and sc_warlock_demonology.cpp files. I will be closing this PR and implementing the changes from the raidbots link in this comment (as well as the end of fight Demo implosion line) as a separate PR.

PR #6410, opened 2022-02-24
[Cooldown] Refactor out target-specific cooldowns into target_specifi…

EvanMichaels:
…c_cooldown_t

* Refactor into dedicated wrapper class so that they can be used elsewhere--such as actions, buffs, or as standalone objects within modules
* Tested as having the same results results with Dream Delver in HAC/DungeonSlice simulations

### DIFF
--- a/engine/sim/cooldown.cpp
+++ b/engine/sim/cooldown.cpp
@@ -777,9 +777,16 @@ void sc_format_to( const cooldown_t& cooldown, fmt::format_context::iterator out
 }
 
 target_specific_cooldown_t::target_specific_cooldown_t( player_t& p, cooldown_t& base_cd ) :
+  name_str( base_cd.name() ),
   player( &p ),
-  base_cooldown( &base_cd ),
-  name_str( base_cooldown->name() )
+  base_duration( base_cd.duration )
+{
+}
+
+target_specific_cooldown_t::target_specific_cooldown_t( util::string_view name, player_t& p, timespan_t duration ) :
+  name_str( name ),
+  player( &p ),
+  base_duration( duration )
 {
 }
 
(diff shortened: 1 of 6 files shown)


engine/sim/cooldown.cpp
```diff
@@ -775,3 +775,43 @@ void sc_format_to( const cooldown_t& cooldown, fmt::format_context::iterator out
...
+    tcd.spawn_index = spawn_index;
+  }
+
+  return tcd.cooldown;
+}
+
+void target_specific_cooldown_t::reset()
+{
+  target_cooldowns.clear();
+}
```
Dorovon: I don't think clearing the vector on reset is necessary. If I remember correctly, this would happen at start of each iteration, meaning that for simulations with many iterations this vector would have to be rebuilt every time. This probably doesn't slow things down too much, but the main purpose of keeping pointers to the cooldowns in a vector like this is to allow fast access by minimizing the number of times `player_t::get_cooldown` is called.

engine/sim/cooldown.hpp
```diff
@@ -121,3 +122,29 @@ struct cooldown_t
 private:
   void adjust_remaining_duration( double delta ); // Modify the remaining duration of an ongoing cooldown.
 };
+
+struct target_specific_cooldown_t
+{
+  player_t* const player;
+  cooldown_t* const base_cooldown;
```
Dorovon: I think just storing the duration and name here would be better than actually referencing the regular cooldown. It's possible someone could want to use this without creating a base cooldown.

engine/sim/cooldown.cpp
EvanMichaels: I was kinda torn on this myself. The reason I plugged it in in the end was that it did seem like it would avoid a bunch of useless calls to cooldown->reset as I assumed the spawn_index values would be initially mismatched from the previous iteration. However I think you're probably right that resetting the vector would likely auto_dispose the cooldown_t* objects themselves on the player (since they aren't referenced anywhere else) and we'd have to re-create them all.

Not against removing this, just kinda seemed "technically correct" but it probably isn't strictly needed.

engine/sim/cooldown.hpp
EvanMichaels: My thought process here for the time being is that I wasn't totally sure if it worked with RPPM objects yet (it is assigned to that one RPPM spell but I did not have time to test if it actually functions) and I figured we may need additional aspects of the cooldown data. But just storing the duration could be sufficient for now.

engine/sim/cooldown.cpp
Dorovon: That is true that there are some wasted reset calls happening, since all of the cooldowns in the player's cooldown list already run cooldown_t::reset_init when the player is reset.

I suspect that rebuilding the vector during each iteration will be more computationally expensive than those reset calls, but I'm not entirely sure, because there is a decent amount of stuff happening in cooldown_t::reset. I guess another option would be to reset the spawn indices to some negative value and to not reset the cooldowns if the old index was equal to that value, but that does add further complication to the code that might not be needed.

PR #6416, opened 2022-02-26
[Warrior] start work on T28

seanpeters86:
Current issues:
- something with the rage trigger is causing issues. If I hard code the rage value do be an integer things seem okay, so the conversion between rage and stacks for the tracking buff needs to be fixed/refactored. 

Tried adding a cast to the rage like so:
`p()->buff.seeing_red_tracking->trigger( as<int>( rage ) );`

but that just causes this:
`Assertion failed: f == static_cast<From>( t ), file B:\simc\engine\util\generic.hpp, line 534 `

Aside from the above issue things work as expected on a debug iteration.


Armadk:
Moving the expire out of the stack change seems to take care of the asserts triggering.  Tested it in game as well, and during execute you can have multiple back to back 40 rage executes.  When it happens it can trigger the buff to stack twice.

PR #6427, opened 2022-02-27
Bonestorm RP Fix

TheAsaSpades:
Bonestorm was only having its base runic power cost (10) added to Insatiable Hunger, this fix adds the correct amount of runic power spent on Bonestorm to Insatiable Hunger


engine/class_modules/apl/apl_death_knight.cpp
```diff
@@ -108,32 +108,33 @@ void blood( player_t* p )
   default_->add_action( "raise_dead" );
   default_->add_action( "blooddrinker,if=!buff.dancing_rune_weapon.up&(!covenant.night_fae|buff.deaths_due.remains>7)" );
   default_->add_action( "blood_boil,if=charges>=2&(covenant.kyrian|buff.dancing_rune_weapon.up)" );
-  default_->add_action( "raise_dead" );
+//  default_->add_action( "raise_dead" );
```
nyterage: Why remove raise dead here? its not on a GCD, and is just a free dps increase? Doesn't make much sense to remove it.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -7793,7 +7793,7 @@ double death_knight_t::resource_loss( resource_e resource_type, double amount, g
     double base_rp_cost = actual_amount;
 
     // If an action is linked, fetch its base cost. Exclude Bonestorm from this otherwise it uses the base cost for Insatiable Hunger instead of the actual rp spent.
-    if ( action && action->id != 194844 )
+    if ( action && specialization() != DEATH_KNIGHT_BLOOD )
```
Armadk: Stepping through the RP spending spells, it looks like only Bonestorm, and Death Strike do this.  

I tried with death coil while wearing deadliest coil, and I was getting the base cost, rather than the actual cost.  Interestingly,  when I first implemented this death strike was giving the base cost, rather than actual.  I don't have data for bonestorm in my notes from that time.

engine/class_modules/apl/apl_death_knight.cpp
TheAsaSpades: Cause it's used 2 lines before, and you want to use it before blooddrinker so you don't interrupt blooddrinker.

engine/class_modules/sc_death_knight.cpp
Armadk: Re-tested this today after reporting the issue with Death Strike, and it now correctly grants a value of 90 (as if 45 rp was spent) when ossuary is up.  Only Bonestorm should need the adjustment.  Bonestorm on current build gives 200 value (100 rp) with a full rp bar.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -7792,10 +7792,12 @@ double death_knight_t::resource_loss( resource_e resource_type, double amount, g
...
     double base_rp_cost = actual_amount;
 
-    // If an action is linked, fetch its base cost
-    if ( action )
+    // If an action is linked, fetch its base cost. Exclude Blood from this otherwise it uses the base cost for
+    // Insatiable Hunger instead of the actual rp spent for Bonestorm and Death Strike.
+    if ( action && specialization() != DEATH_KNIGHT_BLOOD )
       base_rp_cost = action -> base_costs[ RESOURCE_RUNIC_POWER ];
 
+
```
Armadk:
```suggestion

```

Armadk (changes requested):

Armadk (approved):

PR #6435, opened 2022-03-02
[Paladin] Holy T28 4pc Implementation

xizbow1:
Not implementing the 2pc because the bonus LODs from it do not proc awakening and thus are never a dps gain. Had help from abelito75.


xizbow1:
I mangled this pull request a lot and did a bunch of stupid stuff, will reopen in a few hours with less stupidity.

PR #6440, opened 2022-03-05
[shaman] Implement 2022-03-04 hotfixes for Elemental Shaman

wfarr:
Changes can be found here: https://us.forums.blizzard.com/en/wow/t/elemental-shaman-class-tuning-march-8/1195446


Bloodmallet (approved):
Under the assumption that people are simulating their mythic information **now** rather than later, I agree to hotifx Storm Elemental Wing Gust max stacks.

Resetting Wind Gust stacks on recast does not affect people yet. Therefore I consider this good to add now.

PR #6496, opened 2022-03-27
Adding 2p and 4p to Vengeance DH

Woliance:


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -1040,6 +1043,29 @@ struct soul_fragment_t
...
     }
 
+    if ( is_type( soul_fragment::LESSER ) && dh->sets->has_set_bonus( DEMON_HUNTER_VENGEANCE, T28, B4))
+    {
+
+     timespan_t duration =  (dh->set_bonuses.t28_vengeance_4pc->effectN( 1 ).time_value()) ;
+
+      if ( dh->cooldown.immolation_aura->is_ready() && !dh->cooldown.fel_devastation->is_ready() )
+      {
+        dh->cooldown.immolation_aura->adjust( -duration );
```
EvanMichaels, line 1053: Is this (and the next check) not backwards?  This won't really do anything since it's adjusting IA's cooldown only when it is already ready.

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -1040,6 +1043,29 @@ struct soul_fragment_t
       dh->resource_gain( RESOURCE_FURY, dh->spec.blind_faith_fury->effectN( 1 ).resource( RESOURCE_FURY ), dh->gain.blind_faith );
     }
 
+    if ( is_type( soul_fragment::LESSER ) && dh->sets->has_set_bonus( DEMON_HUNTER_VENGEANCE, T28, B4))
```
EvanMichaels: This check is fine (the result will be the same) but as a minor point it may make it easier to search if it was consistent with the other checks in the module which compare against p()->set_bonuses.t28_vengeance_4pc->ok()

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -2625,6 +2651,11 @@ struct immolation_aura_t : public demon_hunter_spell_t
           p()->spawn_soul_fragment( soul_fragment::LESSER, 1 );
           p()->proc.soul_fragment_from_fallout->occur();
         }
+        if ( p()->sets->has_set_bonus( DEMON_HUNTER_VENGEANCE, T28, B2 ) && rng().roll( (p()->set_bonuses.t28_vengeance_2pc->effectN( 1 ).base_value()) ) )
```
EvanMichaels: Same as the note above, maybe more advantageous in this case because it already requires set_bonuses.t28_vengeance_2pc spell data to be ok() for this line to function.

Also, I believe the rng().roll() here is using the incorrect value, should probably be using .percent() rather than .base_value(). Alternative, you can just use set_bonuses.t28_vengeance_2pc->proc_chance() since it is on the spell data for the proc.

> Name             : Burning Hunger (id=364454) [Spell Family (107), Passive, Hidden] 
> Proc Chance      : 10%

EvanMichaels (changes requested):
Thanks for the PR, just a few comments you may want to look at!

Woliance:
I fixed the inverted checks, and the incorrect proc chance, I couldn't get p()-> to work on 1046, which is why it was in that format to begin with, Im unsure why ^^

EvanMichaels (approved):

PR #6501, opened 2022-03-28
Use CPack to create a release package on windows

scamille:
Tries to mirror the current nightly to some degree (Version number separator is different), and also better structures windeployqt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,23 @@ function(sc_common_compiler_options target)
   )
 endfunction()
 
+
+### Git Hash ###
+# Get the current working branch
+execute_process(
+    COMMAND git rev-parse --abbrev-ref HEAD
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    OUTPUT_VARIABLE GIT_BRANCH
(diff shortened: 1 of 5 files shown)


scamille:
Example nightly build: https://github.com/simulationcraft/simc-publish/actions/runs/2060128165#artifacts

navv1234:
Do we include the various licenses and README.md from the root directory into the package somewhere else? I think those should be there as well?

scamille:
Yes all the things that get added through the cmake install directives get added to the package. I am already using that when building the windows nightly.

I think those files I explicitly added for CPack would only be relevant if you used it with a .deb or Apple bundle generator.

What is new is only the automated packaging into a 7z file.

Please download and have a look at the build artefact from the run I linked.

navv1234:
Yeah so I see. Well, seems good to me .. wonder if I should do something about the Macos build as well...

scamille:
I had that thought as well. Not sure if it's worth it compares to the qmake solution, but worth a try.

I do not have access to any MacOS system, but I could play with it through our CI setup.

navv1234:
I should be able to play around it much easier locally, so shouldn't be an issue.

PR #6531, opened 2022-04-20
[Death Knight]  Fix melee background abilities that do not have not a proc flags

Armadk:


Armadk:
Had a report that Phial of Putrefaction wasn't providing any dps beyond what stats gave for Frost DK in sims.  Found via bisect that 8b0c3aee6eb25e89b was the commit that changed the behavior.

From what I can see, no Melee abilities for DK are flagged with not a proc (105), however, via log analysis I can confirm that all these abilities do proc the trinket.  For frost in particular, obliterate and frost strike were found to proc the trinket.

Armadk:
@Jayezi I'm not sure if any other class falls into this.  Phial is yellow melee only, so there is far fewer abilities that can proc it.  DK spell's do generally have not a proc set on them, but for whatever reason, the melee ones do not.

Jayezi:
Most things slip by fine without worrying about the not a proc flag since there's also checks for if the ability is a background ability along with it, but the dk abilities mentioned seem to all be implemented as background abilities. Seems like if it's necessary to have some of them background that are actually acting like casted then it'd be safest to set each one as not a proc individually so that other abilities derived from that base melee action that are actually background don't erroneously get considered to be triggers.

Jayezi:
cb0d8928f5f27d0e2a18e701f702d679449abfd6 Actually decided to try this instead which should fix anything else like this, Phial was proccing like it should again off the mentioned abilities.

PR #6548, opened 2022-05-08
Parse coefficients from trigger data

gastank:
A common motif found in recent spells, particularly damage procs from items & effects
is to have the damage coefficient determined by the drivers, rather than the damage
spell itself, with a dummy trigger on the damage spell referring back to the driver.

If ALL of the following are true:
  1) the special_effect_t is an offensive spell action or an attack action
  2) the proc action has 0 damage after parsing the spell data
  3) the proc action has a trigger effect referencing the driver
  4) the driver has a trigger effect referencing the proc action
then the coefficient of the effect in 4) will be used for the damage of action 2).

### DIFF
--- a/engine/player/unique_gear_helper.hpp
+++ b/engine/player/unique_gear_helper.hpp
@@ -336,7 +336,47 @@ struct proc_action_t : public T_ACTION
     if ( effect )
     {
       override_data( *effect );
+
+      // If ALL of the following are true:
+      //   1) the special_effect_t is an offensive spell action or an attack action
+      //   2) the proc action has 0 damage after parsing the spell data
+      //   3) the proc action has a trigger effect referencing the driver
+      //   4) the driver has a trigger effect referencing the proc action
+      // then the coefficient of the effect in 4) will be used for the damage of action 2)
+
+      bool is_dot = effect->duration() > 0_ms && effect->tick_time() > 0_ms;
+      auto type   = effect->action_type();
+
+      if ( ( type == SPECIAL_EFFECT_ACTION_SPELL || type == SPECIAL_EFFECT_ACTION_ATTACK ) &&
+           ( ( is_dot && this->base_td == 0 ) || ( this->base_dd_min == 0 && this->base_dd_max == 0 ) ) )
+      {
+        for ( const auto& eff : this->data().effects() )  // go thru the driver's effects
+        {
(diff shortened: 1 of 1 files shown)


navv1234:
Shouldn't you make sure here that you are actually handling a damaging or healing effect on the triggered spell?

gastank:
`special_effect_t::action_type()` checks the triggered spell and returns if it has a damaging effect type. will have to either replicate some of what it does for `SPECIAL_EFFECT_ACTION_CUSTOM` or just leave it out with assumption that this should only apply this automagic parsed effects.

engine/player/unique_gear_helper.hpp
```diff
@@ -336,7 +336,48 @@ struct proc_action_t : public T_ACTION
...
+
+      if ( ( type == SPECIAL_EFFECT_ACTION_SPELL || type == SPECIAL_EFFECT_ACTION_ATTACK ||
+             type == SPECIAL_EFFECT_ACTION_CUSTOM ) &&
+           ( ( is_dot && this->base_td == 0 ) || ( !is_dot && this->base_dd_min == 0 && this->base_dd_max == 0 ) ) )
+      {
+        for ( const auto& eff : this->data().effects() )  // go thru the driver's effects
+        {
+          auto trg = eff.trigger();
+
+          if ( trg->ok() )  // trigger'd damage spell found
```
scamille, line 358: You could use init statements in the if conditional here and the one below at line 362 to make this a bit denser (but maybe instead spell out  the variable names some more? But that is just my personal preference not to shorten things that much :D)
E.g: 
```cpp
if ( const auto& trigger_spell = eff.trigger(); trigger_spell->ok() )
```

engine/player/unique_gear_helper.hpp
```diff
@@ -336,7 +336,48 @@ struct proc_action_t : public T_ACTION
...
+                if ( is_dot )
+                  this->base_td = coeff;
+                else
+                  this->base_dd_min = this->base_dd_max = coeff;
+
+                break;
+              }
+            }
+
+            break;
```
scamille, line 375: I assume this break here is made as a conscious decision? If you don't find any match in the first trigger spell, you don't continue looking in other trigger spells (assuming there even is something like that in real spell data? probably not).
Maybe add that as a small comment to the code.

PR #6552, opened 2022-05-10
[Core] Add support for Antumbra Weapon Swapping

Saeldur:
Allows for Antumbra "Weapon Swapping" - Note: This does not actually ever swap weapons at any point and is merely a stopgap measure to allow sims to be ran until Blizzard stops this from being possible.

Adds global action, tentatively called "antumbra_swap". This takes a full 1.5s GCD to perform and cannot be done during other actions, nor is it hasted.
Action will "swap weapons", if swapping is enabled, by providing a temporary stat buff as defined by a series of player variables.
This doesn't protect against swapping on or off of the weapon at bad times,
Swapping off during active buffs will cancel them all, do not swap off during Singularity Supreme.

Example non-optimized Swap APL, using shadow priest as an example:
```
#Swap off Antumbra when it goes on CD if no buffs are active
actions+=/antumbra_swap,if=buff.singularity_supreme_lockout.up&!buff.power_infusion.up&!buff.voidform.up&!pet.fiend.active&!buff.singularity_supreme.up&!buff.swap_stat_compensation.up&!buff.bloodlust.up
#Swap onto Antumbra approximately 30s before CDS
actions+=/antumbra_swap,if=buff.swap_stat_compensation.up&!buff.singularity_supreme_lockout.up&(cooldown.power_infusion.remains<=30&cooldown.void_eruption.remains<=30&!((time>80&time<100)&((fight_remains+time)>=330&time<=200|(fight_remains+time)<=250&(fight_remains+time)>=200))|fight_remains<=40)
```

Adds 6 player variables, all prefixed "antumbra."

antumbra.swap - Boolean, setting this to 1 will allow sims containing an
antumbra_swap action to "switch" your weapons.

Series of variables to contain the stat difference between your chosen
weapon swap and antumbra, as if you had antumbra and its offhand
equipped. Likely to be an increase in secondary stats (Positive for a
gain in stats by changing off of antumbra, negative for a loss)

antumbra.int_diff - Double
antumbra.haste_diff - Double
antumbra.mastery_diff - Double
antumbra.crit_diff - Double
antumbra.vers_diff - Double
antumbra.stam_diff - Double

Example Player Options
```
antumbra.swap=1
antumbra.haste_diff=14
antumbra.mastery_diff=88
```

Adds a buff tentatively called "swap_stat_compensation", alternative or better names would be good.
This buff provides the stat difference between the two weapon swaps and disables the Antumbra procs, as if you had weapon swapped it properly. This is an infinite duration buff and Antumbra cannot trigger while this is active.

Important  things for APL 
buff.singularity_supreme_lockout - Antumbras ICD. Don't put the weapon back on if it's still on CD
buff.swap_stat_compensation - the buff provided while you have swapped off of Antumbra.
buff.singularity_supreme - Singularity supreme, the big vers proc. To reiterate, the action does not prevent against swapping while this is up, you'll need to not do that in APL.
buff.singularity_supreme_counter - The stacking vers proc before singularity supreme. Also not guarded against but the item wont be on lockout yet.

### DIFF
--- a/engine/player/player.cpp
+++ b/engine/player/player.cpp
@@ -11571,6 +11571,8 @@ void player_t::copy_from( player_t* source )
   temporary_enchant_str = source->temporary_enchant_str;
 
   external_buffs = source->external_buffs;
+
+  antumbra = source->antumbra;
 }
 
 void player_t::create_options()
--- a/engine/player/player.cpp
+++ b/engine/player/player.cpp
@@ -11733,6 +11733,15 @@ void player_t::create_options()
   add_option( opt_bool( "infinite_runic", resources.infinite_resource[ RESOURCE_RUNIC_POWER ] ) );
   add_option( opt_bool( "infinite_astral_power", resources.infinite_resource[ RESOURCE_ASTRAL_POWER ] ) );
 
+  // Rygelon Dagger / Antumbra
+  add_option( opt_bool( "antumbra.swap", antumbra.swap ) );
+  add_option( opt_float( "antumbra.int_diff", antumbra.int_diff ) );
+  add_option( opt_float( "antumbra.crit_diff", antumbra.crit_diff ) );
+  add_option( opt_float( "antumbra.haste_diff", antumbra.haste_diff ) );
+  add_option( opt_float( "antumbra.mastery_diff", antumbra.mastery_diff ) );
+  add_option( opt_float( "antumbra.vers_diff", antumbra.vers_diff ) );
+  add_option( opt_float( "antumbra.stam_diff", antumbra.stam_diff ) );
+
   // Resources
   add_option( opt_func( "initial_resource", parse_initial_resource ) );
 
--- a/engine/player/player.hpp
+++ b/engine/player/player.hpp
@@ -660,6 +660,17 @@ struct player_t : public actor_t
     const spell_data_t* brush_it_off;
   } racials;
 
+  struct antumbra_t
+  {
+    bool swap           = false;
+    double int_diff     = 0.0;
+    double haste_diff   = 0.0;
+    double crit_diff    = 0.0;
+    double mastery_diff = 0.0;
+    double vers_diff    = 0.0;
+    double stam_diff    = 0.0;
+  } antumbra; // 9.2 Rygelon Dagger
+
   struct passives_t
   {
     double amplification_1;
--- a/engine/player/unique_gear_shadowlands.cpp
+++ b/engine/player/unique_gear_shadowlands.cpp
@@ -4324,35 +4324,130 @@ void cruciform_veinripper(special_effect_t& effect)
   new dbc_proc_callback_t(effect.player, effect);
 }
 
+struct singularity_supreme_t : public stat_buff_t
+{
+  stat_buff_t* supremacy_buff;
+  buff_t* lockout;
+  stat_buff_t* swap_stat_compensation;
+
+  singularity_supreme_t( player_t* p, const item_t* i )
+    : stat_buff_t( p, "singularity_supreme_counter", p->find_spell( 368845 ), i )
+  {
+    lockout = buff_t::find( p, "singularity_supreme_lockout" );
+    if ( !lockout )
+    {
+      lockout = make_buff( p, "singularity_supreme_lockout", p->find_spell( 368865 ) )
+                    ->set_quiet( true )
+                    ->set_can_cancel( false );
+    }
+
+    supremacy_buff = dynamic_cast<stat_buff_t*>( buff_t::find( p, "singularity_supreme" ) );
+    if ( !supremacy_buff )
+    {
+      supremacy_buff = make_buff<stat_buff_t>( p, "singularity_supreme", p->find_spell( 368863 ), i );
+      supremacy_buff->set_can_cancel( false );
+    }
+
+    swap_stat_compensation = dynamic_cast<stat_buff_t*>( buff_t::find( p, "swap_stat_compensation" ) );
+
+    if ( !swap_stat_compensation )
+    {
+      swap_stat_compensation = make_buff<stat_buff_t>( p, "swap_stat_compensation" );
+      swap_stat_compensation->set_name_reporting( "antumbra_swapped_with_other_weapon" );
+      swap_stat_compensation->set_duration( 0_s )->set_can_cancel( false );
+    }
+
+    if ( p->antumbra.swap )
+    {
+      swap_stat_compensation->stats.clear();
+      if ( p->antumbra.int_diff != 0.0 )
+        swap_stat_compensation->add_stat( STAT_INTELLECT, p->antumbra.int_diff );
+      if ( p->antumbra.crit_diff != 0.0 )
+        swap_stat_compensation->add_stat( STAT_CRIT_RATING, p->antumbra.crit_diff );
+      if ( p->antumbra.haste_diff != 0.0 )
+        swap_stat_compensation->add_stat( STAT_HASTE_RATING, p->antumbra.haste_diff );
+      if ( p->antumbra.vers_diff != 0.0 )
+        swap_stat_compensation->add_stat( STAT_VERSATILITY_RATING, p->antumbra.vers_diff );
+      if ( p->antumbra.mastery_diff != 0.0 )
+        swap_stat_compensation->add_stat( STAT_MASTERY_RATING, p->antumbra.mastery_diff );
+      if ( p->antumbra.stam_diff != 0.0 )
+        swap_stat_compensation->add_stat( STAT_STAMINA, p->antumbra.stam_diff );
+    }
+
+    set_expire_at_max_stack( true );
+    set_stack_change_callback( [ this ]( buff_t* b, int, int ) {
+      if ( b->at_max_stacks() )
+      {
+        lockout->trigger();
+        supremacy_buff->trigger();
+      }
+    } );
+  }
+};
+
 void singularity_supreme( special_effect_t& effect )
 {
-  auto lockout = make_buff( effect.player, "singularity_supreme_lockout", effect.player->find_spell( 368865 ) )
-    ->set_quiet( true );
-
-  auto buff =
-      make_buff<stat_buff_t>( effect.player, "singularity_supreme", effect.player->find_spell( 368863 ), effect.item );
-
   // despite spell data proc flags, logs seem to show it only procs on damage spell casts
   effect.proc_flags2_ = PF2_CAST_DAMAGE;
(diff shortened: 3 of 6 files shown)


engine/player/unique_gear_shadowlands.cpp
```diff
@@ -4280,80 +4273,178 @@ void yasahm_the_riftbreaker( special_effect_t& effect )
...
+  else if ( proc_option > 0.0 )
+  {
     effect.rppm_modifier_ *= proc_option;
   }
 
-  new dbc_proc_callback_t(effect.player, effect);
+  new dbc_proc_callback_t( effect.player, effect );
 }
 
+struct singularity_supreme_t : public stat_buff_t
```
Saeldur, line 4327: Because I clangformatted and made a mess, here's the start of the singularity supreme buff struct

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -4280,80 +4273,178 @@ void yasahm_the_riftbreaker( special_effect_t& effect )
...
   effect.player->callbacks.register_callback_trigger_function(
       effect.driver()->id(), dbc_proc_callback_t::trigger_fn_type::CONDITION,
-      [ lockout ]( const dbc_proc_callback_t*, action_t*, action_state_t* ) {
-        return !lockout->check();
+      [ singularity_buff ]( const dbc_proc_callback_t*, action_t*, action_state_t* ) {
+        return !singularity_buff->swap_stat_compensation->check() && !singularity_buff->lockout->check();
       } );
 }
 
+// Action to trigger the chain break for the Chains of Domination trinket
```
Saeldur: Because I clangformatted and made a mess, here's the start of the swap action

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -5985,196 +6088,198 @@ void register_hotfixes()
...
+  unique_gear::register_special_effect( 364087, items::DISABLED_EFFECT );  // Cypher effect Cosmic Boom
 }
 
 action_t* create_action( player_t* player, util::string_view name, util::string_view options )
 {
-  if ( util::str_compare_ci( name, "break_chains_of_domination" ) ) return new items::break_chains_of_domination_t( player, options );
-
+  if ( util::str_compare_ci( name, "break_chains_of_domination" ) )
+    return new items::break_chains_of_domination_t( player, options );
+  else if ( util::str_compare_ci( name, "antumbra_swap" ) )
```
Saeldur, line 6281: and add the new action to create_action

engine/player/unique_gear_shadowlands.cpp
Saeldur: Noticed comment was wrong. Fixed comment,  but it's still in the right place.
Fixed comment for now:
https://github.com/simulationcraft/simc/blob/a2b7467b50ed603c03c044c12e038b883a5e2884/engine/player/unique_gear_shadowlands.cpp#L4397

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -4280,80 +4273,172 @@ void yasahm_the_riftbreaker( special_effect_t& effect )
...
+    {
+      supremacy_buff = make_buff<stat_buff_t>( p, "singularity_supreme", p->find_spell( 368863 ), i );
+      supremacy_buff->set_can_cancel( false );
+    }
+
+    swap_stat_compensation = dynamic_cast<stat_buff_t*>( buff_t::find( p, "swap_stat_compensation" ) );
+
+    if ( !swap_stat_compensation )
+    {
+      swap_stat_compensation = make_buff<stat_buff_t>( p, "swap_stat_compensation" );
```
Dorovon, line 4355: Because this will show up in reports, I think the current name is confusing to someone who does not know what it is already. This should have a more descriptive name like "antumbra_swapped_with_other_weapon". If you'd rather have a shorter string name in the code itself, you could set the buff's name_str_reporting to the more descriptive name instead.

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -4280,80 +4273,172 @@ void yasahm_the_riftbreaker( special_effect_t& effect )
...
+// Action to weapon swap Antumbra
+struct antumbra_swap_t : public action_t
+{
+  singularity_supreme_t* singularity_buff;
+
+  antumbra_swap_t( player_t* p, util::string_view opt ) : action_t( ACTION_OTHER, "antumbra_swap_t", p )
+  {
+    parse_options( opt );
+    trigger_gcd           = 1500_ms;
+    gcd_type              = gcd_haste_type ::NONE;
```
Dorovon:
```suggestion
    gcd_type              = gcd_haste_type::NONE;
```

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -4280,80 +4273,172 @@ void yasahm_the_riftbreaker( special_effect_t& effect )
...
+    }
+
+    supremacy_buff = dynamic_cast<stat_buff_t*>( buff_t::find( p, "singularity_supreme" ) );
+    if ( !supremacy_buff )
+    {
+      supremacy_buff = make_buff<stat_buff_t>( p, "singularity_supreme", p->find_spell( 368863 ), i );
+      supremacy_buff->set_can_cancel( false );
+    }
+
+    swap_stat_compensation = dynamic_cast<stat_buff_t*>( buff_t::find( p, "swap_stat_compensation" ) );
```
Dorovon, line 4351: I don't think there is any need to find or create this buff when the option to enable swapping is not set.

engine/player/player.cpp
```diff
@@ -11733,6 +11733,14 @@ void player_t::create_options()
   add_option( opt_bool( "infinite_runic", resources.infinite_resource[ RESOURCE_RUNIC_POWER ] ) );
   add_option( opt_bool( "infinite_astral_power", resources.infinite_resource[ RESOURCE_ASTRAL_POWER ] ) );
 
+  // Rygelon Dagger / Antumbra
+  add_option( opt_bool( "antumbra.swap", antumbra.swap ) );
+  add_option( opt_float( "antumbra.int_diff", antumbra.int_diff ) );
+  add_option( opt_float( "antumbra.crit_diff", antumbra.crit_diff ) );
+  add_option( opt_float( "antumbra.haste_diff", antumbra.haste_diff ) );
+  add_option( opt_float( "antumbra.mastery_diff", antumbra.mastery_diff ) );
+  add_option( opt_float( "antumbra.vers_diff", antumbra.vers_diff ) );
```
Dorovon, line 11744:
```suggestion
  // Rygelon Dagger / Antumbra
  add_option( opt_bool( "antumbra.swap", antumbra.swap ) );
  add_option( opt_float( "antumbra.int_diff", antumbra.int_diff ) );
  add_option( opt_float( "antumbra.crit_diff", antumbra.crit_diff ) );
  add_option( opt_float( "antumbra.haste_diff", antumbra.haste_diff ) );
  add_option( opt_float( "antumbra.mastery_diff", antumbra.mastery_diff ) );
  add_option( opt_float( "antumbra.vers_diff", antumbra.vers_diff ) );
  add_option( opt_float( "antumbra.stam_diff", antumbra.stam_diff ) );
```
Stamina affects DPS for some classes, so it wouldn't hurt to have that too.

engine/player/player.hpp
```diff
@@ -660,6 +660,17 @@ struct player_t : public actor_t
...
+  struct antumbra_t
+  {
+    bool swap           = false;
+    double int_diff     = 0.0;
+    double haste_diff   = 0.0;
+    double crit_diff    = 0.0;
+    double mastery_diff = 0.0;
+    double vers_diff    = 0.0;
+
+  } antumbra;
```
Dorovon:
```suggestion
  struct antumbra_t
  {
    bool swap           = false;
    double int_diff     = 0.0;
    double haste_diff   = 0.0;
    double crit_diff    = 0.0;
    double mastery_diff = 0.0;
    double vers_diff    = 0.0;
    double stam_diff    = 0.0;
  } antumbra; // 9.2 Rygelon Dagger
```

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -4280,80 +4273,172 @@ void yasahm_the_riftbreaker( special_effect_t& effect )
...
-        return !lockout->check();
+      [ singularity_buff ]( const dbc_proc_callback_t*, action_t*, action_state_t* ) {
+        return !singularity_buff->swap_stat_compensation->check() && !singularity_buff->lockout->check();
       } );
 }
 
+// Action to weapon swap Antumbra
+struct antumbra_swap_t : public action_t
+{
+  singularity_supreme_t* singularity_buff;
```
Dorovon, line 4409: What you have here is fine, but personally I think the code would be cleaner if you instead stored the swap_stat_compensation buff here as a buff_t*, and had a stack change callback defined for that buff to handle canceling the other buffs. That would allow all of the code except for this action to be in the `void singularity_supreme( special_effect_t& effect )` block and would not require a dynamic cast in this action.

engine/player/unique_gear_shadowlands.cpp
```diff
@@ -4280,80 +4273,172 @@ void yasahm_the_riftbreaker( special_effect_t& effect )
...
+
+  antumbra_swap_t( player_t* p, util::string_view opt ) : action_t( ACTION_OTHER, "antumbra_swap_t", p )
```
Dorovon:
```suggestion
  antumbra_swap_t( player_t* p, util::string_view opt ) : action_t( ACTION_OTHER, "antumbra_swap", p )
```

Dorovon (commented):
Overall, this seems mostly fine. I dislike that the user has to manually put in the stat difference quite a bit, but it's probably okay to allow for now until someone either adds proper complete weapon swapping or adjusts this to take an option like `antumbra_swap_weapon=<item_string>`.

Also, I think it might be a good idea to redo this PR without the clang formatting just so that the git history will be clear when it is merged.

Saeldur:
> Overall, this seems mostly fine. I dislike that the user has to manually put in the stat difference quite a bit, but it's probably okay to allow for now until someone either adds proper complete weapon swapping or adjusts this to take an option like `antumbra_swap_weapon=<item_string>`.
> 
> Also, I think it might be a good idea to redo this PR without the clang formatting just so that the git history will be clear when it is merged.

Allowing for an antumbra_swap_weapon=<item_string> and parsing it would be better but this was easier than reading how dbc works for me, that stuff is a nightmare.

And yeah I'll redo the PR after I sort my way through everything here.

Dorovon (approved):

gastank (approved):

PR #6553, opened 2022-05-13
Switch to C++20

scamille:
Adjust minimum compiler versions and set C++ standard to use to C++20.

Adjust CI workflows to new compiler minimums and maximum available versions.

Drop Visual Studio 2017 build files.


scamille:
This is really just a proposal, and I would be happy to get feedback on it.
The PR content is mostly about laying the technical groundwork and ensuring through CI the new setup is tested as good as possible.

* Is it too soon to switch to C++20?
* Are the selected minimum compilers okay (see engine/config.hpp)?
* Is switching to C++20 with the listed minimum compilers causing too many problems with not all features being supported?
* Or do you think it is overdue to switch to C++20?

navv1234:
I'll just repeat it here, I don't really understand why we are actively wanting to limit the number of compilers supported in simc, if there's really no specific need for having minimum C++ version increased?

PR #6561, opened 2022-05-21
[Warlock] gain buff on Scouring Tithe refresh

Facenomore23:
Implementation of Kyrian legendary interaction with languishing soul detritus when Scouring Tithe is refreshed (dot ticking on target while it gets reapplied).

When Scouring Tithe is on the target and get refreshed, the player gain the buff from Unity Kyrian legendary. It can be reproduced in M+ under effect of Urh and during Rygelon encounter.

This fix should not affect at all any sims if the duration of the cooldown of Scouring Tithe is not changed ( override.spell_data="spell.312321.cooldown=0" ), but allows to sim whether it is better to let the buff drop or not during a fight or keep the infinite cooldown of ST.

### DIFF
--- a/engine/class_modules/warlock/sc_warlock.cpp
+++ b/engine/class_modules/warlock/sc_warlock.cpp
@@ -355,6 +355,17 @@ struct scouring_tithe_t : public warlock_spell_t
 
     LSD_alt_value = p->find_spell( 360953 )->effectN( 2 ).percent();
   }
+  
+  void impact( action_state_t* s ) override
+  {
+    warlock_spell_t::impact( s );
+
+    if ( td( s->target )->dots_scouring_tithe->remains() > p()->find_spell( 312321 )->duration() &&
+         p()->legendary.languishing_soul_detritus.ok() )
+    {
+      p()->buffs.languishing_soul_detritus->trigger( 1, LSD_alt_value );
+    }
+  }
 
   void last_tick( dot_t* d ) override
   {


Azevara:
As of this point, most if not all bugs which allowed this behavior in-game have been fixed, though the behavior is still relevant with CDR effects such as S3 M+ (Urh). Since that will still be possible for some time, this PR is fine conceptually. Some small adjustments needed.

engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -355,6 +355,17 @@ struct scouring_tithe_t : public warlock_spell_t
 
     LSD_alt_value = p->find_spell( 360953 )->effectN( 2 ).percent();
   }
+  
+  void impact( action_state_t* s ) override
+  {
+    warlock_spell_t::impact( s );
+
+    if ( td( s->target )->dots_scouring_tithe->remains() > p()->find_spell( 312321 )->duration() &&
```
Azevara, line 363: Please confirm whether or not this is triggering LSD on the initial cast with this code (unsure if the remains() check is happening after dot has been applied due this being placed after impact() call)

Also, find_spell should not be used during combat.

Facenomore23:
I will close the PR as the feature with Urh is not possible anymore. No ST reset if cast on active dot + no LSD if cast on active ST.

engine/class_modules/warlock/sc_warlock.cpp
Facenomore23: So the cast was not triggered on initial cast, as the remaining time was equal to the spell duration. With that check it was only triggered when refreshed.

I tried a lot of things to check if the cast was running on the target at the end of the cast, and if yes, give LSD buff. But the check was always done after the end of cast so I could not check it that way.

Maybe you have an idea how it could have been made more cleanly...

PR #6563, opened 2022-05-24
[Player] cache find_spell lookups

seanpeters86:
This is roughly a 33% increase in speed when testing locally for shadow priest. I added a debug line when the cache is hit and it seems like almost all of these are for items or special effects. There were over 300,000 cache hits in a single iteration when testing locally.

### DIFF
--- a/engine/player/player.cpp
+++ b/engine/player/player.cpp
@@ -10250,9 +10250,15 @@ const spell_data_t* player_t::find_spell( unsigned int id ) const
 {
   if ( id )
   {
+    auto spell_cache_entry = spell_cache.find( id );
+    if ( spell_cache_entry != spell_cache.end() ) {
+      return spell_cache_entry->second;
+    }
+
     auto spell = dbc::find_spell( this, id );
     if ( spell->id() && as<int>( spell->level() ) <= true_level )
     {
+      spell_cache[id] = spell;
       return spell;
     }
   }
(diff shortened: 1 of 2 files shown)


scamille:
My gut feeling says that we might want to rather fix the underlying code that calls find_spell so often instead of potentially increasing memory usage so significantly.

On the other hand, a 33% increase does sound lovely. How many iterations and thread did you use when getting this number?

seanpeters86:
@scamille the root cause of most of these issues was fixed here: https://github.com/simulationcraft/simc/commit/7df2bff49b4d0a23eab02c0ab26ae6fdc8efce04

so really the question is do we leave this in as insurance in case it happens again?

Was talking with @lgkern and @Saeldur about this and i think it would make sense to refactor this into something that simply alerts on change if you have added a significant amount of find_spells compared to previous or something.

Dorovon:
One option might be to add a counter in debug builds with an assert (or maybe just a warning) that goes off if some excessive number of lookups occur.

navv1234:
I don't mind having a base level cache for this stuff at least, but it's just a bit sad that we have gone from having a relatively fast (O(1)) access back in the day to having an issue with the id-based find_spell() performance-wise.

scamille:
> One option might be to add a counter in debug builds with an assert (or maybe just a warning) that goes off if some excessive number of lookups occur.

I like the idea. I think we could just generally warn on any find_spell after combat begin on iteration>0. Sure there might be some false positives for things with lazy initialization, but it would be interesting to see how bad it is.

lgkern:
Yea, that was one idea we bounced around, to have a way for it to fail tests when find_spell is being called during iterations, that would make it easy to detect, at least after a couple of callings.

PR #6573, opened 2022-06-06
[Monk] Add BDB Expressions

JeremyMonk:


scamille:
What is bdb? Why the abbreviation, is it that long and troublesome to spell out in full?

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -327,16 +344,159 @@ struct monk_action_t : public Base
       p()->cooldown.bonedust_brew->adjust( timespan_t::from_seconds( time_reduction ), true );
   }
 
+   int get_bdb_zone()
```
scamille: First, this function should be const, and the ones deriving from it as well.

Second: Is it really best to return an int instead of a enum that naturally restricts the possible zones that can be returned?

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -327,16 +344,159 @@ struct monk_action_t : public Base
...
+  {
+    std::vector<player_t*> target_list = p()->sim->target_non_sleeping_list.data();
+
+    int N           = static_cast<int>( target_list.size() );
```
scamille: You seem to have very inconsistent format for local variables. Why not stick to snake_case?

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -327,16 +344,159 @@ struct monk_action_t : public Base
...
+      auto sef_multiplier = ( 1 + p()->spec.storm_earth_and_fire->effectN( 1 ).percent() ) * 3;
+      flAmp *= sef_multiplier;
+      flAmp *= ( 2 * CO_bonus + 1 ) / 3;
+    }
+
+    flAmp = flAmp * ( 1 + ( .5 * .4 * ( static_cast<double>( bdb_targets ) / N ) * ( 1 + BMH_bonus ) ) );
+
+    auto energy_regen_per_second = p()->resource_regen_per_second( RESOURCE_ENERGY );
+
+    // TP->SCK
```
scamille, line 461: It would be best if you can move the following 3 rdps calculations into some shared function that you call to get the rdps values. Should make your code much smaller and easier to understand.

JeremyMonk:
Thanks Scamille, this is derived from Tostad0ra's scholarly article here

https://colab.research.google.com/drive/1IlNnwzigBG_xa0VdXhiofvuy-mgJAhGa?usp=sharing

and I tried to implement it quickly so that an APL can be derived from it.
I will clean it up some more after work.

PR #6580, opened 2022-06-14
[Elemental Shaman] Updated T28 Profile for Vulpera.simc

git-fabio:
turns out using bag of tricks on CD is a dps loss over using it as a low prio filler with the standard build for ST
I've fiddled around with it with SE but it doesnt seem to make that big of a change
https://www.raidbots.com/simbot/report/nKPNYCXMw7BMhot9JpbsAK
https://www.raidbots.com/simbot/report/agXTiqZvkNkPKBiLTnLG74


Bloodmallet:
Thank you for the suggestions. The actual change needed to happen in a different file. I did that in above commit.

PR #6581, opened 2022-06-14
rename 'pvp_crit' option to 'pvp_mode', make sim use pvp ilvl when enabled.

Zurigan:
This extends the existing pvp flag to also use the pvp ilvl of any pvp gear when the flag is enabled. GUI tooltip/description also updated to match new functionality. 

Bonus ilvl for pvp items appears to be held in the used bonus_type value_1 ... I couldn't find any documentation to back this up though! 

Tested as working on current (SL S3) Honor and Conquest gear.

### DIFF
--- a/engine/action/action.cpp
+++ b/engine/action/action.cpp
@@ -1190,7 +1190,7 @@ double action_t::total_crit_bonus( const action_state_t* state ) const
   double crit_multiplier_buffed = crit_multiplier * composite_player_critical_multiplier( state );
 
   double base_crit_bonus = crit_bonus;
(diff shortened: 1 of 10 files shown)


Dorovon:
The changes to the simulator look good. I think this reduces the major PvP differences that aren't implemented down to just PvP multipliers on spells and PvP talents.

I'm not too familiar with the GUI code, so I'm not sure if anything needs to be adjusted with respect to localization.

navv1234:
IIRC QT translation tool (qt linguist?) should be run, but I think it'll work without it, honestly.

PR #6587, opened 2022-06-25
[Destruction] Bugfix for Infernals

Saeldur:
Infernals do not take any actions for ~1.2s after being summoned, this is visible in logs. This implements that delay.

e.g.

https://www.warcraftlogs.com/reports/rY6AWmk8fR4LKHTt#fight=1&type=casts&source=16&translate=true&view=events

summon infernal at 0.02.217
infernal awakening 0.03.203
infernal melee at 0.04.491


Azevara (approved):
This appears roughly correct and is only an issue for the Infernal pet based on my testing (Demonology pets do not have a delay issue after summoning). Will also appropriately affect Blasphemy behavior so this change appears to have no issues.

Issue #6536, opened 2022-04-24
Destruction Warlock: Soul Fire applies Immolate on cast, not on impact.


Hekili:
### Describe the bug

Current sim behavior has Soul Fire applying Immolate on impact rather than on cast.  This leads to the profile opening with Soul Fire -> Immolate in single target, when the Immolate is not needed.

### Expected behavior

Soul Fire should apply Immolate on cast and not wait for the projectile to land.

### To Reproduce

Please include any steps necessary to recreate the problem, or include logs/data to show what is occurring in game if there is an issue with how something is modeled.

Simulate a character with the Soul Fire talent.

https://www.raidbots.com/simbot/report/9HoMkKXdeLm1LoLXkw17Z7/simc

![image](https://user-images.githubusercontent.com/9438134/164950801-430b9928-4eba-4902-98ab-0f5653ca82b0.png)

In the opener, Immolate immediately follows Soul Fire.  In reality, the conditions for that Immolate entry would not fire because Immolate's duration would be > 3 following SF.

### Additional information
* SimulationCraft Version (required)
* [ff3026eda9](https://github.com/simulationcraft/simc/commits/ff3026eda9)

* If SimulationCraft is crashing, please include your **dxdiag** file - Instructions here - 
  https://us.battle.net/support/en/article/Obtaining-System-Files

N/A

* **simc_gui.simc** file or contents
  This file is saved in the same folder as SimulationCraft after each simulation. Make sure to run the 
  simulation that is giving you issues, then include the file here by either uploading to github, 
  copy/pasting the contents, or using http://pastebin.com/

PR #6588, opened 2022-06-25
[Warlock] Soul Fire applies Immolate on cast success, not impact

Azevara:


Azevara:
Should resolve #6536  and render #6584 moot.

PR #6608, opened 2022-07-21
[Priest] Living Shadow dragonflight version

seanpeters86:
This is the new implementation for [Living Shadow](https://www.wowhead.com/beta/spell=373381/living-shadow). This is currently our Tier 28 4pc set in shadowlands, however in Dragonflight it only shares a name and a model as the actual spells are quite different.

Living Shadow now "replicates" your casts by casting its version of that cast. I've created an enum to store these different casts [here](https://github.com/simulationcraft/simc/pull/6608/files#diff-d81f48f18c48acbff561af80cd3e4dd75d281aae4f61b84a1c73ccb741373157R564-R571). The basic idea with the implementation is that every spell will be coded manually as to what corresponding spell it triggers, with the default being `NONE`. 

Some TODOs I'm not sure about:
- Should the trigger be called in `impact()` or `execute()`?
- Is there a way to set this property on each spell without having to prepend it with `priest_t`? i.e. `living_shadow_action = living_shadow_action::SHADOW_SPIKE_VOLLEY;` instead of `living_shadow_action = priest_t::living_shadow_action::SHADOW_SPIKE_VOLLEY;`
- The channel abilities seem to be borked, they just cast once per channel rather than starting their own channels
- Need to make sure that we can _interrupt_ spells with this method. I.e. the Player casts Mind Flay -> Your Shadow casts Shadow Spike Volley -> Player interrupts cast to use Mind Blast -> Your Shadow should interrupt its cast to cast a Shadow Spike. Not sure if this needs special handling or not

Some logs:
- [Shadow Spike](https://www.warcraftlogs.com/reports/L6dFMQwJNqC3rZX7/#fight=2&type=damage-done&source=1&ability=376914&view=events)
- [Shadow Spike Volley](https://www.warcraftlogs.com/reports/L6dFMQwJNqC3rZX7/#fight=2&type=damage-done&source=1&ability=376891&view=events)
- [Shadow Sear](https://www.warcraftlogs.com/reports/L6dFMQwJNqC3rZX7/#fight=2&type=damage-done&source=1&ability=373387&view=events)
- [Shadow Sear + Shadow Nova](https://www.warcraftlogs.com/reports/BJH23kfV9ZaNjzCd/#fight=6&type=damage-done&view=events)


engine/class_modules/priest/sc_priest.hpp
```diff
@@ -558,6 +561,15 @@ struct priest_t final : public player_t
     const spell_data_t* boon_of_the_ascended;
   } covenant;
 
+  enum living_shadow_action
```
scamille: 1. If you don't want to prepend priest_t:: all the time for the enum values just move the whole thing out of priest _t class ;)
2. Prefer using a enum class. You can also simplify the enum values since they are bound to its scope,.a call might eg become `living_shadow_action::spike`

engine/class_modules/priest/sc_priest_pets.cpp
```diff
@@ -1155,6 +1264,43 @@ void priest_t::trigger_shadowflame_prism( player_t* target )
...
+{
+  priest_pet_t* pet = debug_cast<priest_pet_t*>( pets.your_shadow.active_pet() );
+
+  if ( pet )
+  {
+    switch ( action )
+    {
+      case living_shadow_action::SHADOW_SPIKE:
+        assert( pet->shadow_spike );
+        pet->shadow_spike->set_target( target );
```
scamille: Create a function to select the correct action_t* and then have the set_target and execute lines a single time.

Saeldur: Good shout, added this.
Neatens it up and makes it easier in the future if they add anything.

engine/class_modules/priest/sc_priest.hpp
seanpeters86: Wasn't sure if that first point would cause issues lol, but seems to work fine :) As for the simpler enum values I think I prefer the more verbose ones so its clear which action it triggers 🤷🏻

scamille: I'd still recommend using a class enum so they don't leak out of scope :-)

engine/class_modules/priest/sc_priest_pets.cpp
```diff
@@ -1155,6 +1284,51 @@ void priest_t::trigger_shadowflame_prism( player_t* target )
...
+}
+
+void priest_t::trigger_living_shadow_action( player_t* target, living_shadow_action action )
+{
+  auto pet_action = get_living_shadow_action( action );
+
+  if ( pet_action )
+  {
+    pet_action->set_target( target );
+    pet_action->execute();
```
scamille, line 1316: The main issue as fasr as I can see is that you call execute() here instead of schedule_execute(). That means you are directly executing the main part of the pet action right here and now (unless there is some travel time), instead of properly scheduling it through the pets action system.

That might definitely explain why you have trouble with channeling actions and the like.

engine/class_modules/priest/sc_priest.hpp
seanpeters86: I did make it an `enum class` insetad already, unless im misunderstanding what you mean? @scamille

engine/class_modules/priest/sc_priest_pets.cpp
seanpeters86: @scamille Right now it is working as expected while using execute, as we manually manage cancelling channels by overriding `last_tick` in priest_spell. We need to do this anyways because once the player stops channeling mind flay we need to also cancel the Living Shadows channel. Because of this i dont think we have any issues with `execute` but i could be missing something.

With that being said are there advantages to using the pet action system that we would want to swap this to schedule execute instead?

scamille: Maybe @navv1234 can better assess the advantages or disadvantages of each approach - I am quite rusty when it comes to player scheduling and all that.

seanpeters86:
After doing a good amount of testing in game the current implementation is near identical to how simc is performing. Going to merge this and can always take future suggestions from @scamille and/or @navv1234 if either have suggestions! Thanks for the help everyone

PR #6632, opened 2022-08-04
Updated Ignore Pain cap to 30% of max health.

TheAsaSpades:


engine/class_modules/sc_warrior.cpp
```diff
@@ -6032,12 +6039,12 @@ struct ignore_pain_t : public warrior_spell_t
 
     double previous_ip = p() -> buff.ignore_pain -> current_value;
 
-    // IP is capped at 1.3 times the amount of the current IP
-    double ip_cap_ratio = 1.3;
+    // IP is capped to 30% of max health
```
Armadk, line 6035: Is this value in DF as well? or is DF still using the old method?  If it is using the new method, we should clean up the commented out free_revenge_should_trigger, and I'll apply this to the DF branch as well.

TheAsaSpades: It looks like it's 30% max in DF as well. I'll clean up the comments and repush.

Armadk (approved):

PR #6649, opened 2022-08-15
[Gear] Return functionality of cruciform_veinripper_proc_rate

JeremyMonk:
Tanks have been using this option to mean "time spent behind boss," and not literally changing the Proc rate. This is how the variable used to work prior to the latest commit.

This returns the functionality of that variable to the state that the tank Discords are using it and makes it clearer. Perhaps the variable name should be changed as well?

### DIFF
--- a/engine/player/unique_gear_shadowlands.cpp
+++ b/engine/player/unique_gear_shadowlands.cpp
@@ -4285,79 +4285,66 @@ void yasahm_the_riftbreaker( special_effect_t& effect )
 // TODO: Add proc restrictions to match the weapons or expansion options.
 void cruciform_veinripper(special_effect_t& effect)
 {
-  struct cruciform_veinripper_cb_t : public dbc_proc_callback_t
-  {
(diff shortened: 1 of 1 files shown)


emallson (approved):
this looks fine to me. restores old functionality of the option. technically the option should probably get renamed but I can also just update the wiki to explain exactly what it is for. that avoids the class discords having to update pins etc. especially for an item that is going away soon:tm:

EvanMichaels:
> this looks fine to me. restores old functionality of the option. technically the option should probably get renamed but I can also just update the wiki to explain exactly what it is for. that avoids the class discords having to update pins etc. especially for an item that is going away soon™️

I already checked in a version of this with a newly renamed option, see: https://github.com/simulationcraft/simc/commit/84a43a1349b18a9bb551bf89975a33740614a101

emallson:
I was just about to comment on that 👍

JeremyMonk:
I meant to close this following Koji's last commit

PR #6672, opened 2022-08-25
[Paladin] Multiple changes

Woliance:
Adding HoW, AW, MoG, Seal of.. Yeah most of them ^-^


engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -213,11 +207,16 @@ struct avenging_wrath_t : public paladin_spell_t
   {
     paladin_spell_t::execute();
 
-    p()->buffs.avenging_wrath->trigger();
+    if ( p()->talents.avenging_wrath->ok() )
```
mserrano: this `if` shouldn't be necessary; the `background` above should do it

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1029,6 +1047,9 @@ void judgment_t::impact( action_state_t* s )
 {
   if ( result_is_hit( s->result ) )
   {
+      if ( p()->talents.judgment->ok() )
+      td( s->target )->debuff.judgment->trigger();
```
mserrano:
```suggestion
      if ( p()->talents.judgment->ok() )
        td( s->target )->debuff.judgment->trigger();
```

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1586,7 +1612,7 @@ struct blessing_of_the_seasons_t : public paladin_spell_t
 struct hammer_of_wrath_t : public paladin_melee_attack_t
 {
   hammer_of_wrath_t( paladin_t* p, util::string_view options_str )
-    : paladin_melee_attack_t( "hammer_of_wrath", p, p->find_class_spell( "Hammer of Wrath" ) )
+    : paladin_melee_attack_t( "hammer_of_wrath", p, p->find_talent_spell( talent_tree::CLASS,"Hammer of Wrath" ) )
```
mserrano:
```suggestion
    : paladin_melee_attack_t( "hammer_of_wrath", p, p->find_talent_spell( talent_tree::CLASS, "Hammer of Wrath" ) )
```

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -3052,6 +3115,8 @@ double paladin_t::resource_loss( resource_e resource_type, double amount, gain_t
   double result     = player_t::resource_loss( resource_type, amount, source, action );
   if ( resource_type == RESOURCE_HOLY_POWER && result > 0 && legendary.of_dusk_and_dawn->ok() &&
        resources.current[ RESOURCE_HOLY_POWER ] == legendary.of_dusk_and_dawn->effectN( 2 ).base_value() )
+  if ( resource_type == RESOURCE_HOLY_POWER && result > 0 && talents.of_dusk_and_dawn->ok() &&
+       resources.current[ RESOURCE_HOLY_POWER ] == talents.of_dusk_and_dawn->effectN( 2 ).base_value() )
```
mserrano: this seems wrong - did you mean to delete the legendary check above, or to duplicate the whole block?

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -937,9 +946,18 @@ struct paladin_action_t : public Base
...
     if ( affected_by.blessing_of_dawn && p() -> buffs.blessing_of_dawn -> up() )
     {
-      am *= 1.0 + p() -> legendary.of_dusk_and_dawn -> effectN ( 1 ).trigger() -> effectN ( 1 ).percent();
+      am *= 1.0 + p() -> talents.of_dusk_and_dawn -> effectN ( 1 ).trigger() -> effectN ( 1 ).percent();
+    }
+
+    if ( affected_by.divine_purpose && p()->buffs.blessing_of_dawn->up() && p()->talents.seal_of_order->ok() )
+    {
+      am *= 1.0 + p()->talents.seal_of_order->effectN( 1 ).percent();
     }
```
mserrano, line 961: we can refactor this to only need to check `buffs.blessing_of_dawn -> up` once, which is marginally better

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -48,17 +48,21 @@ struct avengers_shield_base_t : public paladin_spell_t
...
     {
       aoe = as<int>( p -> azerite.soaring_shield.spell() -> effectN( 2 ).base_value() );
     }
 
     // Soaring Shield hits +2 targets
+    if ( p->talents.soaring_shield->ok() )
+    {
     aoe += as<int>( p -> talents.soaring_shield -> effectN( 1 ).base_value() );
+    }
+  }
```
mserrano: fix the indentation here

engine/class_modules/paladin/sc_paladin.cpp
Woliance: The idea was to duplicate the whole block in order to keep it working during prepatch, could've done an OR check but decided against it in favour of being able to delete it easier once the legendary gets disabled.

mserrano: Right now it doesn't duplicate the whole block, just the condition, which is equivalent to `&&`-ing these conditions - you'll need to duplicate the `trigger` line as well

Woliance: Ah I didnt catch that, pushed a fix for that now :)

mserrano (approved):

PR #6688, opened 2022-08-30
Docker image build option to build smaller images

Horaddrim:
Adding a simple build argument to run the PTR data exclusion shell script only when a build argument called PTR_DATA is set to anything other than 1. 
Using an inverse logic in the if so you have to explicitly set the variable in order to exclude PTR related data making the current behaviour unaltered to everybody not using the flag.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,7 @@ FROM alpine:latest AS build
 
 ARG THREADS=1
 ARG APIKEY=''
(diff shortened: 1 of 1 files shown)


Horaddrim:
@Bloodmallet if you think I should add anything else, please let me know :shipit:

Bloodmallet (approved):
Looks good. Thank you for the contribution.

Issue #6452, opened 2022-03-08
Dungeon Slice should allow disabling of class buffs


diab64:
### Describe the bug
First posted under raidbots here: https://github.com/seriallos/raidbots-issues/issues/324

Dungeon Slice sims with Bloodlust, Arcane Intellect, Power Word: Fortitude, and Battle Shout active, with no way for them to be unchecked. According to a post in the above-linked issue, this was originally because of the fact that if those classes were not in the group, groups would use scrolls to supplement those buffs. As this is no longer the case, you should be able to sim Dungeon Slice without those buffs in effect.

### Expected behavior
Expect to be able to check and uncheck Bloodlust, Arcane Intellect, Power Word: Fortitude, and Battle Shout from being used in the simulation, based on the group comp.

### To Reproduce
Run a simulation with Dungeon Slice and see results for active buffs: Bloodlust, Arcane Intellect, Power Word: Fortitude, Battle Shout

### Additional information
I am using Raidbots, but I was told that Raidbots is based on SimulationCraft, thus Raidbots cannot update without SimulationCraft updating.

![raidbotsbug](https://user-images.githubusercontent.com/44482861/157292306-455829cd-a348-46bc-b3b3-1c7c0cf5cdb2.JPG)

JontiH:
I made a PR for this, seemed simple enough
https://github.com/simulationcraft/simc/pull/6708

PR #6708, opened 2022-09-10
get rid of overrides on dslice due to lack of scrolls

JontiH:
this is a fix for 
https://github.com/simulationcraft/simc/issues/6452

this is my first contrib to this repo so let me know if theres anything I am missing!

### DIFF
--- a/engine/sim/sim.cpp
+++ b/engine/sim/sim.cpp
@@ -2256,7 +2256,9 @@ void sim_t::init_fight_style()
       //Based on the Hero Dungeon setup
       max_time = timespan_t::from_seconds( 360.0 );
 
-      optimal_raid = 1;
+      optimal_raid = 0;
+      // it's reasonable to assume there will be a lust in the group
+      overrides.bloodlust = 1;
 
       shadowlands_opts.enable_rune_words = false;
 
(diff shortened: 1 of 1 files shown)


engine/sim/sim.cpp
```diff
@@ -2255,13 +2255,8 @@ void sim_t::init_fight_style()
...
       max_time = timespan_t::from_seconds( 360.0 );
-      //Disables all raidbuffs, except those provided by scrolls or the character itself.
-      optimal_raid = 0;
-      overrides.arcane_intellect = 1;
-      overrides.battle_shout = 1;
-      overrides.power_word_fortitude = 1;
-      overrides.bloodlust = 1;
-      overrides.windfury_totem = 0;
+
+      optimal_raid = 1;
```
cremor: Why change this to 1? Will this not again enable **all** raid buffs?

JontiH: It will have them on by default, but you can toggle them off. I guess maybe for Dslice we should turn bloodlust on and everything else off?

Before they were not toggleable, but I also thought having nothing on by default would be worse than having everything.

jharris-tc:
any thoughts on this?

Issue #6683, opened 2022-08-28
Docker build for Dragonflight branch failling


Horaddrim:
### Describe the bug
Trying to build the Docker image for the Dragonflight branch is currently failing due to an error on using old spells with the new version of the engine.
From what I could gather from my currently superficial understanding of the code is that by using the "T26_raid.simc" file in the LLVM profilling run in the build, you ran into an error related to the Arcane mage talents as seen in the screenshot below:
![image](https://user-images.githubusercontent.com/25853444/187083750-a8e16d9c-df2c-4297-a7a5-c182f5c4b848.png)

I think the fact it's reporting for the Arcane mage is unrelated to the mage simulation itself, probably just the first thing to break.

If anyone can provide me any guidance on what is nice to have on this profilling run, I can try to adapt the simc file itself to not use any of the old spells but still generate a good profilling, but this is probably better to get done by more experienced members of the community 😅 

### Expected behavior
The Docker image build succesfully.

### To Reproduce
Command used: `docker build -t simc-dragonflight .`

### Additional information
* SimulationCraft [commit](https://github.com/Horaddrim/simc-dragonflight/commit/5389ac2d45247c7df81587fe7b963d0e5fc847ce)
[...]

PR #6709, opened 2022-09-11
added dummy profiles to get docker to work

JontiH:
As discussed here:

https://github.com/simulationcraft/simc/issues/6683

The empty profile doesn't work for building the docker image.

I created a folder of super basic dummy profiles for each spec in order to get the dockerfile to build properly.

Hopefully this will suffice until things are all ported and proper APLs are built, and furthermore the dummy APLs should be pretty agnostic across expansions, so they can be used in the time between other expansions


jharris-tc:
@Horaddrim can you merge this?

Bloodmallet (changes requested):
Please add Devastation Evoker, too.

I think empty profile could be useful for something. I'm undecided whether these should be generated as generator-profiles too. This way they would contain an APL.
Oh and while the goal of "these will work between expansions" is a good one, it won't be fulfilled because between expansions whole specs usually break in SimulationCraft because spelldata is shifted around so much.
Therefore this reason might not apply.

jharris-tc:
> Please add Devastation Evoker, too.
> 
> I think empty profile could be useful for something. I'm undecided whether these should be generated as generator-profiles too. This way they would contain an APL. Oh and while the goal of "these will work between expansions" is a good one, it won't be fulfilled because between expansions whole specs usually break in SimulationCraft because spelldata is shifted around so much. Therefore this reason might not apply.

well the whole idea with these dummy profiles is that they that they literally don't do anything. 

so for example, look at the VDH dummy

demonhunter="T28_Demon_Hunter_Vengeance"
source=default
spec=vengeance
level=60
race=blood_elf
role=tank
position=front
actions=use_items
main_hand=drustlords_greataxe,id=178713,bonus_id=6807/1498/6646,enchant=sinful_revelation


All it does is uses items, and beyond having a weapon equipped (I just picked one), it doesn't have any gear. So even if spelldata is shifted around, this should theoretically still build.

if I'm being honest, maybe the better path longterm is to add the ability to turn profiling off when building the Dockerfile. The engine itself should still be able to build even if APLs aren't working, because I would still be able to make my own custom APL and run it. But for now I think this is a good stopgap to provide a docker image people can use to play around with.

Does that make sense @Bloodmallet

JontiH:
> Please add Devastation Evoker, too.
> 
> I think empty profile could be useful for something. I'm undecided whether these should be generated as generator-profiles too. This way they would contain an APL. Oh and while the goal of "these will work between expansions" is a good one, it won't be fulfilled because between expansions whole specs usually break in SimulationCraft because spelldata is shifted around so much. Therefore this reason might not apply.

> Please add Devastation Evoker, too.
> 
> I think empty profile could be useful for something. I'm undecided whether these should be generated as generator-profiles too. This way they would contain an APL. Oh and while the goal of "these will work between expansions" is a good one, it won't be fulfilled because between expansions whole specs usually break in SimulationCraft because spelldata is shifted around so much. Therefore this reason might not apply.

well the whole idea with these dummy profiles is that they that they literally don't do anything. 

so for example, look at the VDH dummy

```
demonhunter="T28_Demon_Hunter_Vengeance"
source=default
spec=vengeance
level=60
race=blood_elf
role=tank
position=front
actions=use_items
main_hand=drustlords_greataxe,id=178713,bonus_id=6807/1498/6646,enchant=sinful_revelation
```

All it does is uses items, and beyond having a weapon equipped (I just picked one), it doesn't have any gear. So even if spelldata is shifted around, this should build. I have built it myself, and run a sim with a custom (shitty) apl for a hunter.

if I'm being honest, maybe the better path longterm is to add the ability to turn profiling off when building the Dockerfile. The engine itself should still be able to build even if APLs aren't working, because I would still be able to make my own custom APL and run it. But for now I think this is a good stopgap to provide a docker image people can use to play around with.

Does that make sense @Bloodmallet

PR #6748, opened 2022-09-28
[Priest] break out PL actions by trigger spell

seanpeters86:
issue: https://github.com/WarcraftPriests/df-shadow-priest/issues/8

This PR changes how PL is built such that you can clearly see which spell triggered what portion of that damage.
psychic_link_breakdown

Testing before/after damage is still the same, this is just a reporting improvement.


Saeldur (approved):
It's not pretty but it works I suppose.

PR #6809, opened 2022-10-06
[Priest] Pet spells need their own buff apply code

seanpeters86:
- pet spells were not getting auto buffed by things like voidform, twist of fate, or shadowform
- clean up old living shadow actions
- inescapable shadows only works with mindbender, removed switch case
- inescapable shadows was dealing damage twice from new spelldata

Still need to check:
- other buff effects that work with pets?
- potentially add apply_debuff code as well
- Is there a way to use the same functions that priest_spell_t already has? (guessing no since we need to maintain our own list of buffs for pets/base action is different)


Saeldur (approved):
LGTM

PR #6878, opened 2022-10-15
fix NPEs in brewmaster sim

emallson:
1. Tranquil Spirit sample data wasn't initialized
2. Touch of Karma is only initialized for WW, not Brewmaster.


emallson:
Hina covered this in a separate commit

PR #6914, opened 2022-10-18
Add cancelaura_lava_surge_t action

zab-w:


vituscze:
There's already `cancel_buff` which does the same thing: https://github.com/simulationcraft/simc/blob/8a6ae4a1c1cc3db5e43ed9a59a0e3b2b2cafe75b/engine/player/player.cpp#L9479-L9542

zab-w:
I was sure that something like this must be there.

PR #6955, opened 2022-10-28
[Shaman] Implement "Lightning Turret APL"

SchmidNiklas:
Implementing "Lightning Turret" APL as suggested by Lucenia. This will most likely NOT be an improvement at level 70!

Will break with several talent combinations but I tried to keep the functionality intact for the previous build.

We observe a significant increase in Single Target DPS using this APL.

"Lightning Turret" APL: https://www.raidbots.com/simbot/report/1SWnyC3e7zpodM1f3Mg49P

Current APL: https://www.raidbots.com/simbot/report/hFMsBvb5DzSUaWrQAsRa8E
I added a line to cast Fae Transfusion for a somewhat fair comparison.

Edit:
Requested comparison with same talents and legendaries:
https://www.raidbots.com/simbot/report/cToL1gfwbnY7ESFR7CigLv

It might be enough/better to make the current APL drop Lava Burst in single target looking at this sim. The performance uplift is significantly lower (though the proposed APL doesn't auto attack).

### DIFF
--- a/engine/class_modules/sc_druid.cpp
+++ b/engine/class_modules/sc_druid.cpp
@@ -12042,6 +12042,10 @@ std::unique_ptr<expr_t> druid_t::create_expression( std::string_view name_str )
         return make_fn_expr( name_str, [ this ]() { return eclipse_handler.state == IN_LUNAR; } );
       else if ( util::str_compare_ci( splits[ 1 ], "in_both" ) )
         return make_fn_expr( name_str, [ this ]() { return eclipse_handler.state == IN_BOTH; } );
+      else if ( util::str_compare_ci( splits[ 1 ], "starfire_counter" ) )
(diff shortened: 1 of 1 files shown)


SchmidNiklas:
Since this APL will break many builds it's probably better to keep this as a draft for now.

Bloodmallet:
For a better comparison: please add a simulation of the "current apl simulation" but using
- Lightning Rod
- Primal Elementalist
- Elemental Blast
instead of
- 2x Echo Chamber
- Mountains Will Fall

SchmidNiklas:
Requested comparison with same talents:
https://www.raidbots.com/simbot/report/vpLYhF9Q9TRu58wzDkQ1uk

It might be enough/better to make the current APL drop Lava Burst in single target looking at this sim. The performance uplift is significantly lower (though the proposed APL doesn't auto attack).

PR #6995, opened 2022-11-10
[ele] default apl precombat and single target changes

Snow998:
*added stormkeeper, fire elemental, storm elemental, lava burst, and icefury to the precombat actions 
*placed liquid magma totem higher up on the single target action list to apply flame shock 
*added maelstrom and talent requirements to stormkeeper, for the purpose of the double SoP combo 
*added totemic recall, with time constraints as it is currently bugged 
*added master of the elements requirement to Stormkeeper empowered lightning bolt, without SoP talented 
*added earth shock maelstrom and talent requirements, for the purpose of resource pooling for the double SoP Stormkeeper combo

https://www.raidbots.com/simbot/report/urFX6qsZ1ZQop6mugxXCo2


engine/class_modules/sc_shaman.cpp
```diff
@@ -11372,6 +11372,13 @@ void shaman_t::init_action_list_elemental()
     precombat->add_action( "snapshot_stats", "Snapshot raid buffed stats before combat begins and pre-potting is done." );
     precombat->add_action( this, "Flametongue Weapon", "if=talent.improved_flametongue_weapon.enabled", "Ensure weapon enchant is applied." );
     precombat->add_action( "potion" );
+    precombat->add_action( "stormkeeper,if=!talent.surge_of_power.enabled|spell_targets.chain_lightning>=3" );
+    precombat->add_action( "fire_elemental" );
+    precombat->add_action( "storm_elemental" );
+    precombat->add_action( "lava_burst,if=spell_targets.chain_lightning<3" );
+    precombat->add_action( "icefury,if=spell_targets.chain_lightning>=3&talent.electrified_shocks.enabled" );
+    precombat->add_action( "lava_burst,if=spell_targets.chain_lightning>=3&talent.master_of_the_elements.enabled" );
+    precombat->add_action( "icefury,if=spell_targets.chain_lightning>=3" );
```
Bloodmallet: SimulationCraft doesn't know "time" during `precombat`. Therefore it does not protect you against casting as many damage spells as your apl can fit. Please update your conditionals to prevent multiple damage casts.

Bloodmallet (changes requested):

engine/class_modules/sc_shaman.cpp
Bloodmallet: Hawk let me know that the first damage applying spell will end the precombat phase for simc. I didn't know that. Still a small cleanup here would be good.

PR #7016, opened 2022-11-15
[Warlock] Destruction AOE

Facenomore23:
Update on Destruction AOE.

TODO : 
- Number of targets rules with havoc active and different variables (inferno/eradication/AR/ AoD ??).
- Number of target for cleave/aoe 
- Trinkets


Facenomore23:
Havoc management, RoF/CB rules and headaches.

Still to do : Trinkets ! We see the light !

PR #7049, opened 2022-11-25
[Paladin] Protection - Adding a bunch of stuff, including APL and T29 Profile

Woliance:
I am so so sorry for the big PR <3


engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1039,12 +1065,65 @@ struct crusader_strike_t : public paladin_melee_attack_t
...
+      periodic( new light_of_the_titans_hot_t( p ) )
+  {
+    parse_options( options_str );
+    tick_zero = false;
+    impact_action = periodic;
+  }
+
+  void impact( action_state_t* s ) override
+  {
+    paladin_t* p = debug_cast<paladin_t*>( player );
```
Armadk, line 1102: Does the hot keep the boost for the entire duration, even if the dot falls off while the hot is still active?  I can help with getting this working, but I just need to understand the behavior to model it.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -3025,8 +3160,12 @@ double paladin_t::composite_attribute_multiplier( attribute_e attr ) const
     if ( passives.aegis_of_light -> ok() )
       m *= 1.0 + passives.aegis_of_light -> effectN( 1 ).percent();
 
+    // This literally never gets triggered. Apparently, invalidating the Stamina cache doesn't recalculate Stamina?
     if ( buffs.redoubt->up() )
       m *= 1.0 + buffs.redoubt->stack_value();
+
+    if ( buffs.sentinel->up() )
```
Armadk, line 3164: Sadly we can't just boost stam here, as sentinel is a % health buff, not stam buff.  There is a number of talents that do similar.  Last Stand, Desperate Prayer.  You may need to check specific health behavior around when it is proc'd vs when it drops for if it does heal you or not.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -3356,6 +3485,18 @@ double paladin_t::composite_parry_rating() const
   return p;
 }
 
+double paladin_t::composite_parry() const
+{
+  double p = player_t::composite_parry();
+
+  if ( buffs.strength_in_adversity->up())
+  {
+    p += buffs.strength_in_adversity->value()*buffs.strength_in_adversity->stack();
```
Armadk, line 3491: minor, but you could use ->stack_value() here instead.

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -1305,6 +1370,21 @@ struct holy_power_consumer_t : public Base
     if ( this -> affected_by.the_magistrates_judgment && ab::p() -> buffs.the_magistrates_judgment -> up() )
       c += ab::p() -> buffs.the_magistrates_judgment -> value();
 
+    if ( ab::p()->buffs.seal_of_clarity->up() )
+    {
+      if ( this->affected_by.seal_of_clarity )
+      {
+        c += ab::p()->spells.seal_of_clarity_buff->effectN( 1 ).base_value();
+      }
+      // 2022-25-11 Seal of Clarity's Buff has no entry as of now, so hardcoding until it's there
```
Armadk: I see you do have the right spell id for the buff, but to import it, it needs to be added to the generator, and new spell data pulled in.  DM me if you need help with how to do this.  If there is other buff spells missing we can add them all at the same time.

engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -1518,6 +1608,15 @@ struct holy_power_consumer_t : public Base
     {
       p -> buffs.sealed_verdict -> trigger();
     }
+
+    if ( p->buffs.seal_of_clarity->up() )
+      p->buffs.seal_of_clarity->expire();
+
+    auto foo = p->talents.seal_of_clarity->effectN( 1 ).percent();
```
Armadk: we should be able to setup the roll and everything in the buff once it gets imported

engine/class_modules/paladin/sc_paladin_protection.cpp
```diff
@@ -546,6 +611,110 @@ struct judgment_prot_t : public judgment_t
...
+  double stacks;
+  redoubt_buff_t( paladin_t* p, util::string_view name, const spell_data_t* s )
+    : buff_t( p, name, s ), health_change( data().effectN( 1 ).percent() )
+  {
+    add_invalidate( CACHE_STRENGTH );
+    add_invalidate( CACHE_STAMINA );
+    set_cooldown( timespan_t::zero() );
+    set_default_value( p->talents.redoubt->effectN( 1 ).trigger()->effectN(1).percent() );
+  }
+  void increment(int stacks, double value, timespan_t duration) override
```
Armadk, line 626: was this entire section just trying to get the stam portion to work?

PR #7062, opened 2022-11-28
[Ele Shaman] Incorporate use of MotE

Whaitiri:
- when not using SoP, cast lava burst before using a SK-buffed lightning bolt to buff it with MotE
- if we have enough Maelstrom to spend, cast Lava Burst to gain MotE if it would not cap Maelstrom
- With EotE, if MotE is up and we have one LvB charge with no Surge, use Icefury FS or LB before spending the LvB charge (very minor increase)

These changes show a 608~ dps increase on the T29 profile according to raidbots:
[Standard APL](https://www.raidbots.com/simbot/report/hFAA2sUDN5ucJg9NnNJyW4) 84,500 DPS
[Custom APL from this PR](https://www.raidbots.com/simbot/report/gbMqgoA954gyXaTc9koAx2) 85,108 DPS

The DPS increase is much lower with Surge of Power (~50dps), and higher on Lightning Bolt builds that don't use Surge of Power (approaching 1000).


Whaitiri:
Here is a similar talent tree setup, switching the right side for the left to test Stormkeeper

[Standard APL](https://www.raidbots.com/simbot/report/aDKyU91RSFWKfLTsjJTsdE) 74,327 DPS
[Custom APL from this PR](https://www.raidbots.com/simbot/report/jKZagJXfrCizFp9iUAL8rp) 75,165 DPS

836 dps difference

PR #7091, opened 2022-12-02
[Gear] Primal Ritual Shell initial implementation

jacobnlsn:
I'm going to be honest with you, I'm jumping into SimC for the first time, and it's a bit overwhelming.  This was an attempt, I made some assumptions that are probably incorrect and am looking to learn from this.

WoWHead link to trinket: https://www.wowhead.com/item=200563/primal-ritual-shell?bonus=7935#comments
WoWHead link to the one of four option I'm trying to implement: https://www.wowhead.com/spell=390899/wind-turtles-blessing

This is the trinkets default option when it drops, and probably the easiest to create the sim for.

I made some assumptions, and will probably get corrected, or this PR will get ignored as garbage, but thought I'd go through some of them.

- I have seen manual_stats_added get set to false in a lot of other places, no idea what this actually does.
- effect.driver()->effectN( 1 ) this is grabbing the effect at index 0 and triggering it? In this case it's grabbing how much master rating the buff should give based on the items ilvl?
- How often this effect procs, where is that calculated, this is still mostly a black box to me, trying to start to crack it open


jacobnlsn:
Huge thanks to @gastank and @nyterage for getting me this far!

PR #7095, opened 2022-12-03
[Paladin] Fix Sentinel

Fluttershy-git:
Sentinel was never used in the APL, because it's a new spell. Fixed Sentinel (and other) talents to no longer grant 20% Damage/Heal when not being talented into Avenging Wrath. Also fixed Sentinel to grant this when talented into Avenging Wrath.


Fluttershy-git:
Woli looked into Sentinel doing way lower damage than it maybe should. Appears Sanctified Wrath didn't give Holy Power with Sentinel.

Fluttershy-git:
Removed some bugs that were fixed on launch

PR #7096, opened 2022-12-03
[Vengeance] add Soul Carver action

ToppleTheNun:
image

@EvanMichaels - can't request you for review.

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -405,7 +405,7 @@ public:
       player_talent_t charred_flesh;              // NYI
 
       player_talent_t soulcrush;                  // NYI
-      player_talent_t soul_carver;                // NYI
+      player_talent_t soul_carver;
       player_talent_t last_resort;                // NYI
       player_talent_t fodder_to_the_flame;        // NYI
       player_talent_t elysian_decree;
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -4723,6 +4723,49 @@ struct vengeful_retreat_t : public demon_hunter_spell_t
...
+      may_miss = may_parry = may_dodge = false;  // TOCHECK
+    }
+  };
+
+  soul_carver_oh_t* oh;
+
+  soul_carver_t( demon_hunter_t* p, util::string_view options_str )
+    : demon_hunter_attack_t( "soul_carver", p, p->talent.vengeance.soul_carver, options_str )
+  {
+    oh        = new soul_carver_oh_t( p );
```
EvanMichaels: I would do this as
p->get_background_action<soul_carver_oh_t>( "soul_carver_oh" );

And change the signature above to:
  soul_carver_oh_t( util::string_view name, demon_hunter_t* p )
      : demon_hunter_attack_t( name, p ... )
      
Additionally, you could assign this as 
impact_action = p->get_background_action ... 

That way you don't need to do the manual execute below.

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -4723,6 +4723,49 @@ struct vengeful_retreat_t : public demon_hunter_spell_t
...
+    oh->schedule_execute();
+
+    p()->spawn_soul_fragment( soul_fragment::LESSER, data().effectN( 3 ).base_value() );
+  }
+
+  void tick( dot_t* d ) override
+  {
+    demon_hunter_attack_t::tick( d );
+
+    p()->spawn_soul_fragment( soul_fragment::LESSER );
```
EvanMichaels: Technically this value comes from spell data, so instead of always spawning one, I would do data().effectN( 4 ).base_value(). Since they could change it at any point.

EvanMichaels (changes requested):

PR #7064, opened 2022-11-28
[Vengeance] initial pass at passive talents

ToppleTheNun:

PR #7118, opened 2022-12-08
[Vengeance] implement Void Reaver, Frailty, Soulcrush

ToppleTheNun:
dependent on #7064 or a replacement for it, up to you, @EvanMichaels.

**Before**
beforesoulcrush
**After**
image
**Bonus: 5 target cleave**
image

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -6758,9 +6758,9 @@ void demon_hunter_t::target_mitigation( school_e school, result_amount_type dt, 
       s->result_amount *= 1.0 + spec.fiery_brand_dr->effectN( 1 ).percent();
     }
 
-    if ( td->debuffs.frailty->check() && talent.vengeance.void_reaver )
+    if ( td->debuffs.frailty->check() && talent.vengeance.void_reaver->ok() )
     {
-      s->result_amount *= 1.0 + spec.frailty_debuff->effectN( 3 ).percent() * td->debuffs.frailty->stack();
+      s->result_amount *= 1.0 + spec.frailty_debuff->effectN( 3 ).percent() * td->debuffs.frailty->check();
     }
   }
 }
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -367,7 +367,7 @@ public:
       player_talent_t shear_fury;
       player_talent_t fracture;                   // NYI
       player_talent_t calcified_spikes;           // NYI
-      player_talent_t roaring_fire;
+      player_talent_t roaring_fire;               // NYI
       player_talent_t sigil_of_silence;           // NYI
       player_talent_t retaliation;                // NYI
       player_talent_t fel_flame_fortification;    // NYI
@@ -384,7 +384,7 @@ public:
       player_talent_t void_reaver;
       player_talent_t fallout;                    // NYI
       player_talent_t ruinous_bulwark;            // NYI
-      player_talent_t volatile_flameblood;
+      player_talent_t volatile_flameblood;        // NYI
       player_talent_t revel_in_pain;              // NYI
 
       player_talent_t soul_furnace;               // NYI
@@ -1345,13 +1345,8 @@ public:
 
     ab::apply_affecting_aura( p->talent.vengeance.perfectly_balanced_glaive );
     ab::apply_affecting_aura( p->talent.vengeance.meteoric_strikes );
-    ab::apply_affecting_aura( p->talent.vengeance.shear_fury );
-    ab::apply_affecting_aura( p->talent.vengeance.roaring_fire );
-    ab::apply_affecting_aura( p->talent.vengeance.agonizing_flames );
-    ab::apply_affecting_aura( p->talent.vengeance.extended_spikes );
     ab::apply_affecting_aura( p->talent.vengeance.burning_blood );
     ab::apply_affecting_aura( p->talent.vengeance.ruinous_bulwark );
-    ab::apply_affecting_aura( p->talent.vengeance.volatile_flameblood );
     ab::apply_affecting_aura( p->talent.vengeance.chains_of_anger );
     ab::apply_affecting_aura( p->talent.vengeance.stoke_the_flames );
     ab::apply_affecting_aura( p->talent.vengeance.down_in_flames );
@@ -1482,7 +1477,7 @@ public:
 
     if ( affected_by.frailty )
     {
-      m *= 1.0 + p()->spec.frailty_debuff->effectN( 4 ).percent() * td( target )->debuffs.frailty->stack();
+      m *= 1.0 + p()->spec.frailty_debuff->effectN( 4 ).percent() * td( target )->debuffs.frailty->check();
     }
 
     return m;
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -5889,8 +5992,9 @@ void demon_hunter_t::init_spells()
   spec.soulrend_debuff = talent.havoc.soulrend->ok() ? find_spell( 390181 ) : spell_data_t::not_found();
   spec.tactical_retreat_buff = talent.havoc.tactical_retreat->ok() ? find_spell( 389890 ) : spell_data_t::not_found();
   spec.unbound_chaos_buff = talent.havoc.unbound_chaos->ok() ? find_spell( 347462 ) : spell_data_t::not_found();
-  
+
   spec.fiery_brand_dr = talent.vengeance.fiery_brand->ok() ? find_spell( 207744 ) : spell_data_t::not_found();
+  spec.frailty_debuff = talent.vengeance.frailty->ok() ? find_spell( 247456 ) : spell_data_t::not_found();
```
EvanMichaels, line 5997: You may need the debuff to be found without frailty talented as isn't it possible to proc it while skipping that talent technically via other talents?

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -6654,9 +6758,9 @@ void demon_hunter_t::target_mitigation( school_e school, result_amount_type dt,
       s->result_amount *= 1.0 + spec.fiery_brand_dr->effectN( 1 ).percent();
     }
 
-    if ( td->debuffs.void_reaver )
+    if ( td->debuffs.frailty->up() && talent.vengeance.void_reaver )
```
EvanMichaels: This should probably be check() rather than up() due to not wanting to have another source of benefit tracking. (Think it uses up() in the damage reduction calculations.)

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -1463,6 +1480,11 @@ class demon_hunter_action_t : public Base
       m *= 1.0 + td( target )->debuffs.burning_wound->check_value();
     }
 
+    if ( affected_by.frailty )
+    {
+      m *= 1.0 + p()->spec.frailty_debuff->effectN( 4 ).percent() * td( target )->debuffs.frailty->stack();
```
EvanMichaels: Should use check() here rather than stack()

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -1342,6 +1343,19 @@ class demon_hunter_action_t : public Base
     ab::apply_affecting_aura( p->talent.havoc.any_means_necessary );
     ab::apply_affecting_aura( p->talent.havoc.dancing_with_fate );
 
+    ab::apply_affecting_aura( p->talent.vengeance.perfectly_balanced_glaive );
+    ab::apply_affecting_aura( p->talent.vengeance.meteoric_strikes );
+    ab::apply_affecting_aura( p->talent.vengeance.shear_fury );
```
EvanMichaels: Unfortunately since this is done as Spell Effect 2 indirection rather than a mod aura, you'll have to add this manually into the Shear energize_amount in the shear_t() constructor.

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -1342,6 +1343,19 @@ class demon_hunter_action_t : public Base
...
+    ab::apply_affecting_aura( p->talent.vengeance.roaring_fire );
```
EvanMichaels: Roaring Fire will have to be done manually since it's conditional.

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -1342,6 +1343,19 @@ class demon_hunter_action_t : public Base
...
+    ab::apply_affecting_aura( p->talent.vengeance.agonizing_flames );
```
EvanMichaels: This is already hooked up in the Immolation Aura buff.

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -1342,6 +1343,19 @@ class demon_hunter_action_t : public Base
...
+    ab::apply_affecting_aura( p->talent.vengeance.extended_spikes );
```
EvanMichaels: Will need to be added to the Demon Spikes buff rather than the ability.

engine/class_modules/sc_demon_hunter.cpp
ToppleTheNun: If you tried to build a tree without it, it would be the 8th point: https://www.wowhead.com/talent-calc/demon-hunter/vengeance/DAAAERFBUBA

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -1342,6 +1343,19 @@ class demon_hunter_action_t : public Base
...
+    ab::apply_affecting_aura( p->talent.vengeance.burning_blood );
+    ab::apply_affecting_aura( p->talent.vengeance.ruinous_bulwark );
+    ab::apply_affecting_aura( p->talent.vengeance.volatile_flameblood );
```
EvanMichaels: Volatile Flameblood will need to be implemented manually as a gain in the Immolation Aura impact (is it per hit? or per execute?)

Look at Isolated Prey handling in Fel Rush as an example.

EvanMichaels (changes requested):
Added comments for everything that stood out to me. Hope that's helpful!

engine/class_modules/sc_demon_hunter.cpp
EvanMichaels: Got it, sounds good.

engine/class_modules/sc_demon_hunter.cpp
ToppleTheNun: Even though it's setup as an `APPLY_AURA` on the spell data?

engine/class_modules/sc_demon_hunter.cpp
ToppleTheNun: I'll pull that out into a separate PR then

engine/class_modules/sc_demon_hunter.cpp
EvanMichaels:
```
 #1 (id=1028507)  : Apply Aura (6) | Modify Healing% Based on Target Health% (354)
                    Base Value: 50 | Scaled Value: 50 | PvP Coefficient: 1.00000 | Misc Value: 2 | Target: Self (1)
                    Affected Spells: Fel Devastation (212106), Fel Devastation (333120)
```

This isn't an aura type we support manually applying. Since it would dynamically scale based on your health. Think I would implement it as an action_multiplier() override on fel_devastation_heal_t probably.

ToppleTheNun: I'll pull that out into a separate PR, good to know!

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -6654,9 +6758,9 @@ void demon_hunter_t::target_mitigation( school_e school, result_amount_type dt,
       s->result_amount *= 1.0 + spec.fiery_brand_dr->effectN( 1 ).percent();
     }
 
-    if ( td->debuffs.void_reaver )
+    if ( td->debuffs.frailty->check() && talent.vengeance.void_reaver )
```
EvanMichaels: talent.vengeance.void_reaver->ok()

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -6654,9 +6758,9 @@ void demon_hunter_t::target_mitigation( school_e school, result_amount_type dt,
...
     {
-      s->result_amount *= 1.0 + td->debuffs.void_reaver->stack_value();
+      s->result_amount *= 1.0 + spec.frailty_debuff->effectN( 3 ).percent() * td->debuffs.frailty->stack();
```
EvanMichaels: This one can also be
> ... * td->debuffs.frailty->check()
as stack() has benefit tracking as well

PR #7119, opened 2022-12-08
[Monk] BoC BoF applies to initial hit only

renanthera:


Hinalover:
Closing due to similarity to #7132

PR #7132, opened 2022-12-09
[Monk] RF RJW EK & DfB

renanthera:
- RF ICD race condition
- EK GCD
- DfB occurring once per target, instead of twice
- RJW target cap


renanthera:
Was doing some local testing and merged in #7119, so that's now duplicated by this pr.

JeremyMonk (changes requested):
Please change line L2036 to use the shared RJW as Windwalkers also use this spell.

**reduced_aoe_targets = p->shared.rushing_jade_wind->effectN( 1 ).base_value();**

PR #7138, opened 2022-12-10
[Vengeance] only register painbringer buff if talented

ToppleTheNun:
@EvanMichaels

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -5773,7 +5773,6 @@ void demon_hunter_t::init_spells()
   spec.soul_cleave            = find_specialization_spell( "Soul Cleave" );
   spec.soul_cleave_2          = find_rank_spell( "Soul Cleave", "Rank 2" );
   spec.riposte                = find_specialization_spell( "Riposte" );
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -5999,6 +5998,7 @@ void demon_hunter_t::init_spells()
   spec.fiery_brand_debuff = talent.vengeance.fiery_brand->ok() ? find_spell( 207744 ) : spell_data_t::not_found();
   spec.fiery_brand_dot_damage = talent.vengeance.fiery_brand->ok() ? find_spell( 207771 ) : spell_data_t::not_found();
   spec.frailty_debuff = talent.vengeance.frailty->ok() ? find_spell( 247456 ) : spell_data_t::not_found();
+  spec.painbringer_buff = talent.vengeance.painbringer->ok() ? find_spell( 212988, DEMON_HUNTER_VENGEANCE ) : spell_data_t::not_found();
```
EvanMichaels: Technically you don't need the DEMON_HUNTER_VENGEANCE check here as the optional parameter if you're doing a talent check. Won't make any difference here but just for reference.

PR #7143, opened 2022-12-10
[Vengeance] update APL

ToppleTheNun:
update APL based on theorycrafting happening in Fel Hammer and Bear Troupe Discords.


ToppleTheNun:
@EvanMichaels

PR #7145, opened 2022-12-11
[Vengeance] implement Soul Furnace

ToppleTheNun:
@EvanMichaels

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -223,6 +223,8 @@ public:
     buff_t* demon_spikes;
     buff_t* painbringer;
     absorb_buff_t* soul_barrier;
+    buff_t* soul_furnace_damage_amp;
+    buff_t* soul_furnace_stack;
 
     // Set Bonuses
     damage_buff_t* t29_havoc_4pc;
@@ -388,7 +390,7 @@ public:
       player_talent_t volatile_flameblood;        // NYI
       player_talent_t revel_in_pain;              // NYI
 
-      player_talent_t soul_furnace;               // NYI
+      player_talent_t soul_furnace;
       player_talent_t painbringer;
       player_talent_t darkglare_boon;             // NYI
       player_talent_t fiery_demise;
@@ -517,6 +519,8 @@ public:
     const spell_data_t* soul_cleave_2;
     const spell_data_t* thick_skin;
     const spell_data_t* painbringer_buff;
+    const spell_data_t* soul_furnace_damage_amp;
+    const spell_data_t* soul_furnace_stack;
   } spec;
 
   // Set Bonus effects
@@ -3178,6 +3182,13 @@ struct spirit_bomb_t : public demon_hunter_spell_t
       reduced_aoe_targets = p->talent.vengeance.spirit_bomb->effectN( 2 ).base_value();
     }
 
+    void execute() override
+    {
+      demon_hunter_spell_t::execute();
+
+      p()->buff.soul_furnace_damage_amp->expire();
+    }
+
     void impact(action_state_t* s) override
     {
       demon_hunter_spell_t::impact(s);
@@ -3186,6 +3197,18 @@ struct spirit_bomb_t : public demon_hunter_spell_t
       {
         td(s->target)->debuffs.frailty->trigger();
       }
+    }
+
+    double composite_da_multiplier( const action_state_t* s ) const override
+    {
+      double m = demon_hunter_spell_t::composite_da_multiplier( s );
+
+      if ( p()->buff.soul_furnace_damage_amp->up() )
+      {
+        m *= 1.0 + p()->buff.soul_furnace_damage_amp->check_value();
+      }
+
+      return m;
     }
   };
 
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -4575,6 +4578,12 @@ struct soul_cleave_t : public demon_hunter_attack_t
         m *= 1.0 + p()->talent.vengeance.focused_cleave->effectN( 1 ).percent();
       }
 
+      if ( p()->buff.soul_furnace->up() )
+      {
+        m *= 1.0 + p()->buff.soul_furnace->check_value();
+        p()->buff.soul_furnace->decrement();
```
EvanMichaels: Decrement probably needs to be put in the ability execute() after the base execute() call. This function could be called multiple times potentially.

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -6925,6 +6938,14 @@ unsigned demon_hunter_t::consume_soul_fragments( soul_fragment type, bool heal,
   {
     it->consume( heal );
     souls_consumed++;
+    if (talent.vengeance.soul_furnace) {
+      soul_furnace_soul_fragments_consumed++;
```
EvanMichaels: In-game this is implemented as a buff (391166) and it may be easier to implement this as a stacking buff checking ->at_max_stacks()

ToppleTheNun:
Ran this through 40k iterations:
| Talent String | Name | # of Iterations | DPS ST |
| ---- | ---- | ---- | ---- |
| BUkAAAAAAAAAAAAAAAAAAAAAAAAgIIJREARCQSJBJR7AhkEJhDEtk0SSSSoEJBAAAAA | No Soul Furnace | 21363 | 15023 |
| BUkAAAAAAAAAAAAAAAAAAAAAAAAgIIJREARCQSJBJR7ARiEJhEtk0SSSSoEJBAAAAA | Soul Furnace | 21261 | 15370 |

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -3178,6 +3182,13 @@ struct spirit_bomb_t : public demon_hunter_spell_t
       reduced_aoe_targets = p->talent.vengeance.spirit_bomb->effectN( 2 ).base_value();
     }
 
+    void execute() override
+    {
+      spell_base_t::execute();
```
EvanMichaels: This should be demon_hunter_spell_t::execute(); in this case

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -4566,6 +4589,13 @@ struct soul_cleave_t : public demon_hunter_attack_t
       }
     }
 
+    void execute() override
+    {
+      attack_t::execute();
```
EvanMichaels: This should be demon_hunter_attack_t::execute();

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -6925,6 +6965,13 @@ unsigned demon_hunter_t::consume_soul_fragments( soul_fragment type, bool heal,
   {
     it->consume( heal );
     souls_consumed++;
+    if (talent.vengeance.soul_furnace) {
```
EvanMichaels: Should be talent.vengeance.soul_furnace->ok()

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -3187,6 +3198,18 @@ struct spirit_bomb_t : public demon_hunter_spell_t
         td(s->target)->debuffs.frailty->trigger();
       }
     }
+
+    double composite_da_multiplier( const action_state_t* s ) const override
+    {
+      double m = demon_hunter_action_t::composite_da_multiplier( s );
```
EvanMichaels: I notice here this call is also slightly wrong. Outside of funny cases, generally, we just want to call the direct parent class, whichever is referenced in the constructor. 

```
    spirit_bomb_damage_t( util::string_view name, demon_hunter_t* p )
      : demon_hunter_spell_t( name, p, p->find_spell( 247455 ) )
```

In this case all the entry calls should just be demon_hunter_spell_t::

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -4566,6 +4589,13 @@ struct soul_cleave_t : public demon_hunter_attack_t
...
+    void execute() override
+    {
+      demon_hunter_attack_t::execute();
+
+      p()->buff.soul_furnace_damage_amp->expire();
+    }
+
     double composite_da_multiplier( const action_state_t* s ) const override
     {
       double m = demon_hunter_action_t::composite_da_multiplier( s );
```
EvanMichaels: Same issue here actually. I just missed it in the previous review for Focused Cleave.

engine/class_modules/sc_demon_hunter.cpp
ToppleTheNun: This was just me using the "generate override" and not noticing it was referencing the top level class instead of the parent.

PR #7149, opened 2022-12-11
Add precombat cast for Sigil of Flame

hammundo:
Generates more fury than Immolation aura, wastes 1 less tick that could be used to build ragefire (if taken) and starts the fight with the DoT. 

Both spells can be cast before combat begins due to the detonation time of Sigil of Flame.


hammundo (commented):
Running this against the current APL results in the precast of immolation aura being skipped for some reason.

DPS should however be slightly higher once this is resolved.

hammundo:
It appears that time is being incremented as soon as any action is taken when time=0.
This is why the 'second precast' for Immolation aura is then being skipped.

Is there any way to cast multiple spells during the precombat window when time technically is still 0?

nyterage:
> 

No, only 1 damage action can be executed precombat, even if there is a delay in the damage being done.

EvanMichaels:
Fixed this with https://github.com/simulationcraft/simc/commit/dfc412a69d6e7c251288f3f4079641392e0ad63a and https://github.com/simulationcraft/simc/commit/4e2fb5efb99f02c9aabe0fb74d9928cd0d0ee296 by extending the support for precombat actions as well as some Havoc-specific logic in the module.

PR #7150, opened 2022-12-11
[Vengeance] implement Darkglare Boon

ToppleTheNun:
@EvanMichaels

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -2389,7 +2389,7 @@ struct fel_devastation_t : public demon_hunter_spell_t
       double maximum_fury_refund = p()->talent.vengeance.darkglare_boon->effectN( 4 ).base_value();
       double fury_refund         = rng().range( minimum_fury_refund, maximum_fury_refund );
 
-      p()->cooldown.fel_devastation->adjust( cdr_reduction );
+      p()->cooldown.fel_devastation->adjust( -cdr_reduction );
       p()->resource_gain( RESOURCE_FURY, fury_refund, p()->gain.darkglare_boon );
     }
   }
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -358,24 +358,24 @@ public:
 
     struct vengeance_talents_t
     {
-      player_talent_t fel_devastation;            // NYI
+      player_talent_t fel_devastation;
 
-      player_talent_t frailty;                    // NYI
-      player_talent_t fiery_brand;                // NYI
+      player_talent_t frailty;
+      player_talent_t fiery_brand;
 
       player_talent_t perfectly_balanced_glaive;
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -2371,6 +2371,25 @@ struct fel_devastation_t : public demon_hunter_spell_t
...
+    {
+      // CDR reduction and Fury refund are separate rolls per Realz
+      timespan_t minimum_cdr_reduction = timespan_t::from_seconds(p()->talent.vengeance.darkglare_boon->effectN(1).base_value());
+      timespan_t maximum_cdr_reduction = timespan_t::from_seconds(p()->talent.vengeance.darkglare_boon->effectN(2).base_value());
+      timespan_t cdr_reduction = rng().range(minimum_cdr_reduction, maximum_cdr_reduction);
+      double minimum_fury_refund = p()->talent.vengeance.darkglare_boon->effectN(3).base_value();
+      double maximum_fury_refund = p()->talent.vengeance.darkglare_boon->effectN(4).base_value();
+      double fury_refund = rng().range(minimum_fury_refund, maximum_fury_refund);
+
+      p()->cooldown.fel_devastation->adjust(cdr_reduction);
```
EvanMichaels: Isn't the adjustment value coming from spell data the percentage of the total cooldown reduced? Or is the tooltip wrong? Either way, this value should be negative though?

ToppleTheNun: That is correct, not a flat value; I'll fix it.

ToppleTheNun: Just a note that a flat value ends up working out atm due to Fel Dev CD being 1min by default.

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -2372,6 +2372,28 @@ struct fel_devastation_t : public demon_hunter_spell_t
...
+      timespan_t minimum_cdr_reduction =
+          timespan_t::from_seconds( p()->talent.vengeance.darkglare_boon->effectN( 1 ).percent() * base_cooldown );
+      timespan_t maximum_cdr_reduction =
+          timespan_t::from_seconds( p()->talent.vengeance.darkglare_boon->effectN( 2 ).percent() * base_cooldown );
+      timespan_t cdr_reduction   = rng().range( minimum_cdr_reduction, maximum_cdr_reduction );
+      double minimum_fury_refund = p()->talent.vengeance.darkglare_boon->effectN( 3 ).base_value();
+      double maximum_fury_refund = p()->talent.vengeance.darkglare_boon->effectN( 4 ).base_value();
+      double fury_refund         = rng().range( minimum_fury_refund, maximum_fury_refund );
+
+      p()->cooldown.fel_devastation->adjust( cdr_reduction );
```
EvanMichaels: This value needs to be passed in as -cdr_reduction since right now this would add to the cooldown.

PR #7156, opened 2022-12-12
[Marksmanship] Add Puzzlebox usage

Ghosteld:
Algeth'ar Puzzle Box wasn't used optimally with the current Trinket APL.
- Added precombat usage.
- Added specific line to use Puzzlebox before trueshot.


Azortharion:
MM DPS increase.

PR #7157, opened 2022-12-12
Comitting A Feral Druid profile

Tomslack:
Comitting an initial Feral Druid profile, one with Survival Instincts one without


Khazakdk:
@Xanzara I see you pushed some feral profiles as well, could you work with this PR to determine if it is still needed?

PR #7159, opened 2022-12-12
[Vengeance] implement T29 tier set

ToppleTheNun:
@EvanMichaels

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -3216,7 +3216,7 @@ struct spirit_bomb_t : public demon_hunter_spell_t
 
     void copy_state(const action_state_t* s) override
     {
-      action_state_t::initialize();
+      action_state_t::copy_state(s);
       t29_vengeance_4pc_proc = debug_cast<const spirit_bomb_state_t*>(s)->t29_vengeance_4pc_proc;
     }
 
@@ -4686,7 +4686,7 @@ struct soul_cleave_t : public demon_hunter_attack_t
 
     void copy_state(const action_state_t* s) override
     {
-      action_state_t::initialize();
+      action_state_t::copy_state(s);
       t29_vengeance_4pc_proc = debug_cast<const soul_cleave_state_t*>(s)->t29_vengeance_4pc_proc;
     }
 
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -81,6 +81,9 @@ public:
 
     // Vengeance
     buff_t* frailty;
+
+    // Set Bonuses
+    buff_t* t29_vengeance_4pc;
   } debuffs;
 
   demon_hunter_td_t( player_t* target, demon_hunter_t& p );
@@ -529,8 +532,8 @@ public:
   {
     const spell_data_t* t29_havoc_2pc;
     const spell_data_t* t29_havoc_4pc;
-    const spell_data_t* t29_vengeance_2pc;          // NYI
-    const spell_data_t* t29_vengeance_4pc;          // NYI
+    const spell_data_t* t29_vengeance_2pc;
+    const spell_data_t* t29_vengeance_4pc;
   } set_bonuses;
 
   // Mastery Spells
@@ -638,7 +641,7 @@ public:
     proc_t* soul_fragment_from_hunger;
 
     // Set Bonuses
-
+    proc_t* soul_fragment_from_t29_2pc;
   } proc;
 
   // RPPM objects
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -4590,10 +4650,20 @@ struct soul_cleave_t : public demon_hunter_attack_t
...
+      {
+        td( s->target )->debuffs.t29_vengeance_4pc->trigger();
+      }
     }
 
     void execute() override
     {
+      if ( p()->set_bonuses.t29_vengeance_4pc->ok() )
+      {
+        t29_proc = rng().roll( p()->set_bonuses.t29_vengeance_4pc->effectN( 2 ).percent() );
```
EvanMichaels: This mechanic is a bit strange and while this may work it may not work for all methods of triggering (such as scheduled executes.)

It may be more correct to override demon_hunter_attack_t::snapshot_internal() and then add to the 
```
if ( p()->set_bonuses.t29_vengeance_4pc->ok() && flags & STATE_MUL_DA 
     && rng().roll( p()->set_bonuses.t29_vengeance_4pc->effectN( 2 ).percent() ) )
    state->da_multiplier *= p()->set_bonuses.t29_vengeance_4pc->effectN( 1 ).percent();
```

This will snapshot it into the state that gets passed around and used in any delayed executes and for all the impacts.

ToppleTheNun: I'll give that a shot!

ToppleTheNun: Should the trigger of the debuff go in the snapshot as well?

EvanMichaels: No, that would be before there are any targets. I guess that's tricky... could make a custom action state perhaps. Seems overkill but may actually be the cleanest way. I'm not totally sure. Let me look at a few options.

ToppleTheNun: Since there are two spells affected by it, should they both use the same custom action state type?

ToppleTheNun:
Tried out the custom action state approach for the procs and it did feel more clean.

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -4640,6 +4768,10 @@ struct soul_cleave_t : public demon_hunter_attack_t
 
   void execute() override
   {
+    soul_cleave_state_t* soul_cleave_state = debug_cast<soul_cleave_state_t*>(execute_action->get_state());
+    soul_cleave_state->t29_vengeance_4pc_proc = rng().roll( p()->set_bonuses.t29_vengeance_4pc->effectN( 2 ).percent() );
+    execute_action->snapshot_state(soul_cleave_state, result_amount_type::DMG_DIRECT);
```
EvanMichaels: This will actually leak memory since the state isn't assigned anywhere due to this just being a normal execute. There's ways you could add this into the execute_state but actually think the cleanest way to handle this (and could do the same thing for Soul Bomb) is just inject the t29_vengeance_4pc_proc set/rng into the snapshot_state() function itself.

```
void snapshot_state( action_state_t* state, result_amount_type rt ) override
 {
     demon_hunter_attack_t::snapshot_state( state, rt );

     if( p()->set_bonuses.t29_vengeance_4pc->ok() && rng().roll( p()->set_bonuses.t29_vengeance_4pc->effectN( 2 ).percent() )
       debug_cast<soul_cleave_state_t*>(state)->t29_vengeance_4pc_proc = true
}
```

Or something along those lines.

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -3175,6 +3178,34 @@ struct pick_up_fragment_t : public demon_hunter_spell_t
...
+
+    void initialize() override
+    {
+      action_state_t::initialize();
+      t29_vengeance_4pc_proc = false;
+    }
+
+    void copy_state(const action_state_t* s) override
+    {
+      action_state_t::initialize();
```
EvanMichaels: Should this call `action_state_t::copy_state( s )`;?

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -4563,16 +4648,49 @@ struct shear_t : public demon_hunter_attack_t
...
+  {
+    bool t29_vengeance_4pc_proc;
+
+    soul_cleave_state_t(action_t* a, player_t* target): action_state_t(a, target), t29_vengeance_4pc_proc(false)
+    {
+    }
+
+    void initialize() override
+    {
+      action_state_t::initialize();
```
EvanMichaels, line 4693: Same as above note here

PR #7166, opened 2022-12-13
[Mage] Add T29 Fire profile

chrisholder:


vituscze:
You should also specify flask for the profile (see frost and arcane). I've recently swapped the default flask to the boring vers flask, since defaulting to one of the kiss-curse flasks didn't seem great.

PR #7183, opened 2022-12-17
[Vengeance] add darkglare_boon_cdr_roll expression

ToppleTheNun:
intended to help ease the development of the Fiery Demise and Down in Flames combination APL, as we're currently operating under the assumption that you would only go into the FD+DiF combination APL if you have a high roll on Darkglare Boon's CDR refund.

@EvanMichaels

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -191,6 +191,7 @@ public:
   double ragefire_accumulator;
   unsigned ragefire_crit_accumulator;
   double shattered_destiny_accumulator;
+  double darkglare_boon_cdr_roll;
 
   event_t* exit_melee_event;  // Event to disable melee abilities mid-VR.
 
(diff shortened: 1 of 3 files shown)


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -5730,6 +5735,22 @@ std::unique_ptr<expr_t> demon_hunter_t::create_expression( util::string_view nam
       return this->cooldown.eye_beam->create_expression( "remains" );
     }
   }
+  else if ( util::str_compare_ci( name_str, "dh.darkglare_boon_cdr_roll" ) )
+  {
```
EvanMichaels, line 5739: This could probably just be simplified down to
`return make_mem_fn_expr( name_str, *this, &demon_hunter_t::darkglare_boon_cdr_roll );`

vituscze:
The `dh.` prefix seems unnecessary here. The expression is already only usable in DH APLs, so there's no ambiguity.

ToppleTheNun:
> The `dh.` prefix seems unnecessary here. The expression is already only usable in DH APLs, so there's no ambiguity.

I was trying to follow the example that druid has with its expressions, but I do agree. I was unsure if there were conventions around custom expressions that I needed to follow so I just went with what I saw there.

vituscze:
I think the reason druids did it that way is to match the names of the player-scope options (where the `druid.` prefix makes sense, partly because of Raidbots). That's not the case here as far as I can tell.

PR #7184, opened 2022-12-17
[Vengeance] use spell_targets instead of target

ToppleTheNun:
@EvanMichaels


ToppleTheNun:
Closing in favor of #7192.

PR #7192, opened 2022-12-18
[Vengeance] update Soul Cleave conditions

ToppleTheNun:
Soul Sigils was sometimes generating a Soul Fragment and the previous APL would cause us to sit and Throw Glaive forever instead of just pressing Soul Cleave to get rid of it.

Without change: https://www.raidbots.com/simbot/report/pCLiB3WsV1Ly2bSNy36ZX2
image

With change: https://www.raidbots.com/simbot/report/3M4FKVtbR5ySPngV2p2Z8x
image


ToppleTheNun:
@EvanMichaels

PR #7194, opened 2022-12-19
[Vengeance] change fracture conditions

ToppleTheNun:
only fracture during ED and SC ramps if below 5 stacks of Frailty.


ToppleTheNun:
@EvanMichaels

PR #7202, opened 2022-12-21
[Paladin] Updating Prot APL and Consumeables

Woliance:
Additionally; making Avenging wrath and sentinel be interchangeable in the APL for ease of use.


engine/class_modules/paladin/sc_paladin.hpp
```diff
@@ -1601,6 +1603,50 @@ struct holy_power_consumer_t : public Base
   }
 };
 
+// Avenging Wrath ===========================================================
```
Armadk: Was this not working being in the base sc_paladin.cpp file?  Generally declaration goes in hpp, and definition goes in cpp.

Woliance: Yeah, It didn't work in the sc_paladin.cpp
the change in sc_paladin_protection.cpp wouldnt work unless it was moved
If theres a better way to solve it, Id be happy to do that instead.

sc_paladin_protection.cpp has no active link to sc_paladin.cpp as far as I can understand.

PR #7215, opened 2022-12-29
[Vengeance] add max Fiery Brand duration

ToppleTheNun:
Charred Flesh can only extend Fiery Brand 9 times
maximum due to the interaction with Agonizing Flames, which leads to a maximum Fiery Brand duration of
14.5s.

also use `add_child` instead of `stats=stats`
based on a recommendation in the simc Discord.

@EvanMichaels


ToppleTheNun:
@EvanMichaels I'm still working on this based on our Discord discussion, I'll re-ping you when it's good to go.

PR #7223, opened 2022-12-30
LvBm Ascendance buff check

Eokira24601:
Should prevent time wasted from starting casts it can't finish due to Asc buff running out


Snoogens101:
This is good, can confirm

Bloodmallet (approved):

PR #7226, opened 2023-01-01
[Vengeance] whitelist Soul Fragments buff

ToppleTheNun:
it can be useful for reviewing actions taken
to know how many Soul Fragments stacks were
available at any given time.


ToppleTheNun:
@EvanMichaels as this now also has sc_demon_hunter.cpp changes.

PR #7229, opened 2023-01-03
[Monk] Fix FoF Damage Check

JeremyMonk:
Aura 18 appears to have been merged into spell effect 6 of FoF and is no longer used


JeremyMonk:
@Hinalover

PR #7241, opened 2023-01-08
[Hunter] Update Hunter APL with Temporary Weapon Enchants

howlsome:
Hello!

I've added in the following temporary weapon enchants (at level 70 and 60) for:

- Beastmastery
- Markmanship
- Survival (renamed the strings for clarity)

Please be gentle, this is my first simc PR and I'm keen to learn to contribute to this more in the future.

### DIFF
--- a/engine/class_modules/apl/apl_hunter.cpp
+++ b/engine/class_modules/apl/apl_hunter.cpp
@@ -39,11 +39,15 @@ std::string rune( const player_t* p )
 
 std::string temporary_enchant( const player_t* p )
 {
-  std::string lvl70_temp_enchant = ( p -> specialization() == HUNTER_SURVIVAL ) ? "main_hand:primal_whetstone_3" : "main_hand:completely_safe_rockets_3";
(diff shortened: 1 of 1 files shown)


Pewtro:
The temporary enchants are already supported for all three specs, what is this intended to do?

howlsome:
> The temporary enchants are already supported for all three specs, what is this intended to do?

Mind showing me where it's supported. When I was looking through the APL and so on, I found that only survival was listed in this file hence I thought maybe put it in then.

Pewtro:
The two strings you are replacing are ternaries that change depending on if they're Survival or not, and that adds the necessary support since that means MM/BM will use Completely Safe Rockets and Survival will use Whetstone

howlsome:
Hmm. Okay, I guess this is not needed then.

Do you think a comment line for clarity would help there for future purpose?

Pewtro:
Not really, ternaries are pretty rudimentary syntax, and it's easily confirmable by running a sim as either BM, MM or SV that the enchants are working.

PR #7242, opened 2023-01-09
[Warrior/Fury] Handle trinkets better

oBusk:
I noticed that Trinket combinations was being weird. On-use damage trinkets were being used irregardless of recklessness burn phases, and stat trinkets were not used in Recklessness which means they'll be less efficient.

I found and copied a lot of what @JeremyMonk did for Windwalker in #7074 with some success. Tried lot's of combinations of trinkets and compared to the default APL and saw a good increase.


seanpeters86:
closed due to stale/merge conflicts. feel free to re-open if still relevant

PR #7245, opened 2023-01-09
Run CI tests on gcc built binaries

scamille:
As requested by @navv1234 , we want to have tests run for both gcc and clang, in particular so that both a libstdc++ and libc++ build is covered.

Also enable all jobs without branch restrictions, bringing checks back to pull requests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -131,9 +131,8 @@ jobs:
             tests
           key: ubuntu-clang-for_run-${{ github.sha }}-cpp-${{ matrix.cppVersion }}
 
-
   ubuntu-run:
-    name: ubuntu-clang-cpp-${{ matrix.cppVersion }}-${{ matrix.type }}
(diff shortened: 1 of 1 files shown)


nuoHep:
> in particular so that both a libstdc++ and libc++ build is covered

I will only note that both clang and gcc builds use the system c++ standard library, which is libstdc++

navv1234:
Curious, wonder why gcc/libstdc++ didn't trip on the github side to that particular std::sort misuse, when all gcc/libstdc++ boxes I tried did.

nuoHep:
I guess because CI tests don't have this new dungeonroute stuff?

From the convo on Discord (if I looked it up properly re commit https://github.com/simulationcraft/simc/commit/1a7c9af43ca01b74bafaad6f78f745b24b196bfa) it seems like it was segfaulting only with specific dungeonroutes. Which, looking at the commit, somewhat makes sense.

navv1234:
Would have expected the code to run with dungeonslice as well, but perhaps not then.

PR #7249, opened 2023-01-10
[Warlock] Add support for python generated APL in all specs

Denis101:
Allow all Warlock specs to be generated from `.simc` files using same python script used elsewhere.


Azevara:
This does not seem to be adding anything that isn't already in the codebase, the python generator can already be called via command line.

PR #7255, opened 2023-01-11
Update T29_Druid_Guardian to have proper DF temp enchant and pot

JordanSavesBear:
Changing the potion and temp enchant to dragonflight stuff.


seanpeters86:
looks like this was already fixed. feel free to open again if not

PR #7263, opened 2023-01-13
Do not use lld for gcc build

scamille:
This might speed up CI ~build~ *dependency setup* times. Although I am not sure why we use lld in the first place for gcc builds...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,13 +49,12 @@ jobs:
       - name: Install deps
         run: |
           sudo apt-get update
-          sudo apt-get install -y libcurl4-openssl-dev ninja-build lld ccache ${{ matrix.packages }}
+          sudo apt-get install -y libcurl4-openssl-dev ninja-build ccache ${{ matrix.packages }}
 
       - name: Configure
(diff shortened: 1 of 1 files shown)


scamille:
@nuoHep do you remember if there was a specific reason to use the lld linker instead of the default system one?

nuoHep:
Sorry for the delay, missed the mail notification.

The reason was that lld is faster. So afair initial idea was that "installing" lld and linking with it was faster than linking with whatever the system linker is (which is afaik bfd on ubuntu). May not be true now.

On the other hand both ubuntu 20.04 and 22.04 have all of the clang/llvm suite preinstalled, including lld. So the long CI dep setup time is only a problem for the gcc7 build which uses ubuntu 18.04 runner. And the latter will be fully unsupported by 2023-04-01 - https://github.com/actions/runner-images/issues/6002

scamille:
Thanks for the info. I guess it might be sensible to just revert this PR then, and let the additional setup time whenever the ubuntu 18.04 runner is unsupported and we have to change whatever the lowest available gcc version is. (Or just require a newer gcc version for SimC, but last time I asked that wasn't desired).

PR #7283, opened 2023-01-21
[Warlock] Soul Fire

Daztey:
Adding buff.backdraft.up to the Soul Fire line should be about a 0.5% increase over the current profile


Azevara:
This edit will not be used by simulations as a default APL. The file that would require editing is
https://github.com/simulationcraft/simc/blob/dragonflight/engine/class_modules/apl/warlock.cpp

Daztey:
Closing this as Vocox made an updated version after we had some testing last night

PR #7288, opened 2023-01-22
[Warlock] SF optimisation in ST and cleave

Facenomore23:
Also prophylactic soulkeeper in st, as it's dps neutral currently


Facenomore23:
Also updated destruction profile with new talents

Facenomore23:
Race was till orc, but orc is bad for destruction. I changed to human

PR #7297, opened 2023-01-23
[Warlock] Update default flasks to vers

Azevara:


Azevara:
Will replace #7296 . Versatility is being used as default for all specs to be generally applicable to various profiles.

0xjc:
@Azevara hi, I think Elemental Chaos should at least be the default for Destro - it sims about 0.3% higher than vers [for me](https://www.raidbots.com/simbot/report/icjQZ1qFLPfyTf919dGXoo) - this should generalize to most characters because Destro gains a higher benefit than usual from the crit damage effect, and the spec also doesn't like vers very much.

PR #7306, opened 2023-01-24
Blood DK Trinket Fix

xNeptune:
An attempt to fix the scenarios where a player may have a single on use buff trinket and another that doesn't fall into either category of being a buff trinket or damage trinket; Causing the buff trinket to never proc as the second trinket would lock conditions from being met.


engine/class_modules/apl/apl_death_knight.cpp
```diff
@@ -160,8 +160,8 @@ void blood( player_t* p )
...
   trinkets->add_action( "use_item,slot=trinket2,if=!variable.trinket_2_buffs" );
-  trinkets->add_action( "use_item,slot=trinket1,if=variable.trinket_1_buffs&(buff.dancing_rune_weapon.up|!talent.dancing_rune_weapon|cooldown.dancing_rune_weapon.remains>20)&(trinket.2.cooldown.remains|variable.trinket_2_buffs)" );
-  trinkets->add_action( "use_item,slot=trinket2,if=variable.trinket_2_buffs&(buff.dancing_rune_weapon.up|!talent.dancing_rune_weapon|cooldown.dancing_rune_weapon.remains>20)&(trinket.1.cooldown.remains|variable.trinket_1_buffs)" );
+  trinkets->add_action( "use_item,slot=trinket1,if=variable.trinket_1_buffs&(buff.dancing_rune_weapon.up|!talent.dancing_rune_weapon|cooldown.dancing_rune_weapon.remains>20)&(trinket.2.cooldown.remains|!trinket.2.has_cooldown|variable.trinket_2_buffs)" );
+  trinkets->add_action( "use_item,slot=trinket2,if=variable.trinket_2_buffs&(buff.dancing_rune_weapon.up|!talent.dancing_rune_weapon|cooldown.dancing_rune_weapon.remains>20)&(trinket.1.cooldown.remains|!trinket.1.has_cooldown|variable.trinket_1_buffs)" );
```
nyterage: This should fix a few trinkets. but, will still have issues with Ruby Whelp Shell and Whispering Incarnate Icon (as they both do have cooldowns for some god forsaken reason)

nyterage: To remedy this, would recommend adding 2 new precombat variables, which are checked here as well. those being

`  precombat->add_action( "variable,name=trinket_1_exclude,value=trinket.1.is.ruby_whelp_shell|trinket.1.is.whispering_incarnate_icon" );`
  
`  precombat->add_action( "variable,name=trinket_2_exclude,value=trinket.2.is.ruby_whelp_shell|trinket.2.is.whispering_incarnate_icon" );`
  
  and calling `|variable.trinket_x_exclude` within the lines as well to properly use them

nyterage (changes requested):

engine/class_modules/apl/apl_death_knight.cpp
xNeptune: So, I've since adjusted the apl as you suggested with adding in the precombat checks for trinket exclusion on both RWS & WII. Also adding in the trinket exclusion check for the trinket action list. I appreciate the review, definitely learned from it.

nyterage (approved):

PR #7318, opened 2023-01-28
Dark theme for Qt application

Reksotiv:
d1
d2

### DIFF
--- a/qt/sc_OptionsTab.cpp
+++ b/qt/sc_OptionsTab.cpp
@@ -15,6 +15,30 @@
 
 #include <QtCore/QDateTime>
 
+#ifdef Q_OS_WIN
+  #include <dwmapi.h>
+  #pragma comment( lib, "Dwmapi.lib" )  // fixes error LNK2019: unresolved external symbol __imp__DwmExtendFrameIntoClientArea
+
+  enum : WORD
+  {
+    DwmwaUseImmersiveDarkMode           = 20,
+    DwmwaUseImmersiveDarkModeBefore20h1 = 19
+  };
+
+  bool setDarkBorderToWindow( HWND hwnd, bool dark )
+  {
+    const BOOL darkBorder = dark ? TRUE : FALSE;
+    const bool ok =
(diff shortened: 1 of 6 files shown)


scamille:
Looks great, thanks for your contribution. Seems there are still people interested in the UI, even with all these web services available for SimC nowadays :-)

qt/sc_OptionsTab.cpp
```diff
@@ -15,6 +15,30 @@
 
 #include <QtCore/QDateTime>
 
+#ifdef Q_OS_WIN
```
scamille, line 18: Can you please explain (and write down as a comment in the code) why all this special DWMA code is necessary for setting a dark border to the window? It looks like the code is copied from https://github.com/qt/qtbase/blob/cde904b38ab6ef1ddcc9a0cbcae766cf42d58d81/src/plugins/platforms/windows/qwindowswindow.cpp#L3177 - Is the Qt version we use for SimC just too old? Or any other reason why this isn't accessible somehow through a public API?

scamille:
I tried it out, works fine on windows :-)

scamille (approved):

qt/sc_OptionsTab.cpp
Reksotiv: This code makes the title bar dark when window is inactive.
t
The only way to make it is through Qt. Create a qt.conf file in the directory with .exe file and write in it
```
[Platforms] 
WindowsArguments = darkmode=2
```
1 for permanent dark mode, 2 for automatic switching depending on Windows theme.
We could just make darkmode=2 and the application will redraw itself depending on the Windows theme. But some users want to use a dark application with a light Windows theme and here you have to do everything manually. If the user wants to use the dark theme of the application, and there is a light theme in Windows, then the title bar will be light.

scamille: Ok thanks for the explanation.

I am fine with just tightly coupling the border to the theme selected in SimC. I just wanted to critically understand the necessity for it, since introducing such platform dependent code always has some maintenance burden associated with it.

PR #7325, opened 2023-01-31
Change rime base duration from 5 to 4 seconds

ItsJustMeChris:
It appears to me that darkmoon rime was changed from 4 to 5 seconds when someone was adding a different trinket, and I believe it's an error based on the tooltip for the trinket. 

```
Deal 3845 Frost damage and apply Awakening Rime to your target for 4-12s. If the target dies while Awakening Rime is active, it explodes for 10572 Frost damage split between enemies in a 10 yard radius. Duration is determined by the top-most card of the deck.
```

https://www.wowhead.com/spell=386624/awakening-rime

### DIFF
--- a/engine/player/unique_gear_dragonflight.cpp
+++ b/engine/player/unique_gear_dragonflight.cpp
@@ -894,7 +894,7 @@ void darkmoon_deck_rime( special_effect_t& effect )
 
     timespan_t get_duration( size_t index ) const
     {
-      return 5_s + timespan_t::from_seconds( index );
+      return 4_s + timespan_t::from_seconds( index );
     }
 
     void impact( action_state_t* s ) override


ItsJustMeChris:
[WoWCombatLog-013123_134008.txt](https://github.com/simulationcraft/simc/files/10550220/WoWCombatLog-013123_134008.txt)
Here's the combat log, 

timestamps to see 
`1/31 13:45:24.005` `1/31 13:45:28.071`

gastank:
Which card does it skip an extra second? Looks like the ace card gives 12s, and the 2 card gives 4s, so somewhere in the middle it skips a second.

ItsJustMeChris:
It looks like 8 is 10 and im not noticing any 9 card, so I'm lead to believe that it might just skip for 10 to 12 for an A.. I'll log some more testing on a dummy and see if i can bring any more light to it

seanpeters86:
closing for now, feel free to reopen later

PR #7336, opened 2023-02-05
[Monk] Update T29 Profile gear and talents

renanthera:


renanthera:
It appears that there is some issue with Void Elf at 5+ targets.

Second best it is. My condolences.

Issue #5047, opened 2020-02-12
UI crash when starting on second monitor


W0D1N:
Seems since latest Windows updates or since i got my new pc i get this weird issue with simc:
When i open Simc on my second monitor Simc crashes (program is not responding).
If i open Simc on my main monitor it works ok. Seems like a weird second screen bug or incompatibility? 

SimulationCraft Version (required):
Nightly Build 830-02 x64

PC Specs:
AMD Threadripper 3960X
ASUS Zenith II Extreme
G.Skill TridentZ Neo 64GB 3600 C16
Samsung EVO Plus 970 1TB
Nvidia 2080 Super with latest drivers

Monitor 1 (main): Asus ROG PG278QR @ 144hz
Monitor 2: (second): Asus VG248QE @ 144hz
Windows 10 1909 18363.657

jazz836062:
I am getting this as well.  I pulled this from Reliability Details:

Problem signature
Problem Event Name:	AppHangB1
Application Name:	SimulationCraft.exe
Application Version:	0.0.0.0
Application Timestamp:	5e431e64
Hang Signature:	58ba
Hang Type:	134217728
OS Version:	10.0.18363.2.0.0.256.48
Locale ID:	1033
Additional Hang Signature 1:	58baca757fa647a708f618c7797d55bb
Additional Hang Signature 2:	fa79
Additional Hang Signature 3:	fa79f0d1beb6bd7fbd7cd6887d54c8bc
Additional Hang Signature 4:	58ba
Additional Hang Signature 5:	58baca757fa647a708f618c7797d55bb
Additional Hang Signature 6:	fa79
Additional Hang Signature 7:	fa79f0d1beb6bd7fbd7cd6887d54c8bc

I hope that helps.  Very strange error.

navv1234:
We don't do anything specific with a second monitor with the GUI program. I suspect this may be an issue with Qt instead. Does it crash if you open it on the first monitor and drag it to the second one?

jazz836062:
It does not crash for me when I drag it to the second screen.  Only becomes non-responsive when it tries to start on the second monitor.

scamille:
Can you please try the latest nightly from http://downloads.simulationcraft.org/?C=M;O=D and also try to get updated graphic drivers if there are any available.

I am not sure if the QT version we use for nightly builds has changed since February, but there might have been some fixes for it.

jazz836062:
[...]

PR #7353, opened 2023-02-14
Fixed restoring app position for multiscreen

Reksotiv:
For multiple screens we need to use [QWidget::restoreGeometry()](https://doc.qt.io/qt-6/qwidget.html#restoreGeometry)
Source code: https://codebrowser.dev/qt5/qtbase/src/widgets/kernel/qwidget.cpp.html#7215
and QWidget::setGeometry() for comparison https://codebrowser.dev/qt5/qtbase/src/widgets/kernel/qwidget.cpp.html#7005

### DIFF
--- a/qt/MainWindow.cpp
+++ b/qt/MainWindow.cpp
@@ -101,24 +101,7 @@ void SC_MainWindow::loadHistory()
                               tr( "We have reset your configuration settings due to major changes to the GUI" ) );
   }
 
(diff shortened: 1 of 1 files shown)


scamille:
When a user closes the UI on a second monitor, this will still lead to the UI being opened again on the second monitor, correct?

And that is exactly an issue that leads to unrecoverable crashes for users in #5047 if they can't force the UI to open on their primary monitor.

Besides that, sure using restoreGeometry instead of setGeometry is better, thanks for the input.

Reksotiv:
Works fine on two monitors, unlike setGeometry. It would be nice to test on more monitors

Issue #7363, opened 2023-02-20
'gcd' in variable actions evaluates to 0 and is misused (i.e., Frost DK APL)


Hekili:
### Describe the bug
Variables are very often used to simplify/reuse criteria in action lists, but the `gcd` expression appears to evaluate to `0` since it is a member of the `variable` action rather than the ability being tested.

### Expected behavior
I *think* folks are expecting variables to function like the old aliases(?) that were rarely used to replace criteria in the APL.  My assumption is that they are expecting `gcd` in a variable to evaluate to `gcd.max` or possibly `gcd.remains` -- though there are other places where variables divide by gcd (`%gcd`) so both `gcd` and `gcd.remains` are risks for dividing by zero.

Either:
a) APLs should be using `gcd.max` (or whatever the intended value is) instead of `gcd`, or
b) `gcd` in variable actions should evaluate to `gcd.max` (or other desired value) instead of zero.

### To Reproduce

Frost DK priority has `rp_buffs` variable that uses `3*gcd` in time comparisons:
> `variable,name=rp_buffs,value=talent.unleashed_frenzy&(buff.unleashed_frenzy.remains<gcd*3|buff.unleashed_frenzy.stack<3)|talent.icy_talons&(buff.icy_talons.remains<gcd*3|buff.icy_talons.stack<3)`

https://www.raidbots.com/simbot/report/n4mD4czULgU7ErBUhzD6dZ/simc 

Sorry for using a L60 character; I didn't have a 70 available but it would not impact this issue.  Here's an entry in the above sim that references the above variable (see Hokulea actor):

\# | n | Action + Conditions
-|-|:-
H | 1.80 | frost_strike,if=active_enemies=1&variable.rp_buffs&!talent.breath_of_sindragosa&talent.obliteration&!buff.pillar_of_frost.up

It's used less than twice on average, but clearly `rp_buffs` evaluates to true at some point.  I made two edits to the priority in the copy labeled OblitElseZeroAndRPBuffs.

1. I changed `rp_buffs` so it doesn't rely on stack counts for UF or IT; it will only be 1/true based on the buffs' remaining times.

> `variable,name=rp_buffs,value=talent.unleashed_frenzy&(buff.unleashed_frenzy.remains<gcd*3)|talent.icy_talons&(buff.icy_talons.remains<gcd*3)`

2. I added an entry for Howling Blast contingent only on `rp_buffs` being 1/true.

> `howling_blast,if=variable.rp_buffs`

[...]

PR #7366, opened 2023-02-22
[Warlock] Fix usage of gcd expression

Azevara:


Azevara:
Should resolve Warlock portion of #7363

PR #7379, opened 2023-03-02
[Priest] Manually create PTW dot

taherbert:
1. PTW did not have a ticking event on the spell itself, so manually creating the dot.

2. Remove spirit shell from APL as it is no longer an available spell (and was triggering errors)

### DIFF
--- a/engine/class_modules/priest/sc_priest_discipline.cpp
+++ b/engine/class_modules/priest/sc_priest_discipline.cpp
@@ -158,10 +158,7 @@ struct purge_the_wicked_t final : public priest_spell_t
     purge_the_wicked_dot_t( priest_t& p )
       : priest_spell_t( "purge_the_wicked", p, p.talents.discipline.purge_the_wicked->effectN( 2 ).trigger() )
     {
-      may_crit = true;
-      // tick_zero = false;
-      energize_type = action_energize::NONE;  // disable resource generation from spell data
-      background    = true;
+      background = true;
 
       apply_affecting_aura( priest().talents.throes_of_pain );
     }
@@ -181,7 +178,6 @@ struct purge_the_wicked_t final : public priest_spell_t
     : priest_spell_t( "purge_the_wicked", p, p.talents.discipline.purge_the_wicked )
   {
     parse_options( options_str );
-    may_crit       = true;
     tick_zero      = false;
     execute_action = new purge_the_wicked_dot_t( p );
   }
--- a/engine/class_modules/priest/sc_priest_discipline.cpp
+++ b/engine/class_modules/priest/sc_priest_discipline.cpp
@@ -151,25 +151,39 @@ struct power_word_solace_t final : public priest_spell_t
 // Purge the wicked
 struct purge_the_wicked_t final : public priest_spell_t
 {
+  struct purge_the_wicked_dot_t final : public priest_spell_t
+  {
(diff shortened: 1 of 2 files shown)


engine/class_modules/priest/sc_priest_discipline.cpp
```diff
@@ -151,25 +151,39 @@ struct power_word_solace_t final : public priest_spell_t
...
-
-    // Throes of Pain: Spell Direct and Periodic 3%/5% gain
-    apply_affecting_aura( priest().talents.throes_of_pain );
-  }
+    // Manually create the dot effect because "ticking" is not present on
+    // primary spell
+    purge_the_wicked_dot_t( priest_t& p )
+      : priest_spell_t( "purge_the_wicked", p, p.talents.discipline.purge_the_wicked->effectN( 2 ).trigger() )
+    {
+      may_crit = true;
```
seanpeters86: the spell data has `Periodic Can Crit` so we shouldnt need to set this manually. Would suggest removing and seeing if this is added automatically

engine/class_modules/priest/sc_priest_discipline.cpp
```diff
@@ -151,25 +151,39 @@ struct power_word_solace_t final : public priest_spell_t
...
+      // tick_zero = false;
```
seanpeters86: lets remove this entirely instead of adding it back as a comment

engine/class_modules/priest/sc_priest_discipline.cpp
```diff
@@ -151,25 +151,39 @@ struct power_word_solace_t final : public priest_spell_t
...
+      energize_type = action_energize::NONE;  // disable resource generation from spell data
```
seanpeters86: AFAIK looking at the spell data this is not needed

engine/class_modules/priest/sc_priest_discipline.cpp
```diff
@@ -151,25 +151,39 @@ struct power_word_solace_t final : public priest_spell_t
...
+        trigger_power_of_the_dark_side();
+      }
     }
+  };
+
+  purge_the_wicked_t( priest_t& p, util::string_view options_str )
+    : priest_spell_t( "purge_the_wicked", p, p.talents.discipline.purge_the_wicked )
+  {
+    parse_options( options_str );
+    may_crit       = true;
```
seanpeters86: i would assume we dont need this similar to the above

seanpeters86 (changes requested):

engine/class_modules/priest/sc_priest_discipline.cpp
taherbert: Resolved

engine/class_modules/priest/sc_priest_discipline.cpp
taherbert: Removed

engine/class_modules/priest/sc_priest_discipline.cpp
taherbert: Removed

engine/class_modules/priest/sc_priest_discipline.cpp
taherbert: Removed

seanpeters86 (approved):

PR #7380, opened 2023-03-02
[Priest] Add Dark Indulgence talent

taherbert:
Apply dark indulgence to add an extra charge to mind blast for discipline.

Needed to wrap the current reset() logic in a conditional to only apply to shadow spec.

+ some small formatting fixes


seanpeters86 (approved):
LGTM

PR #7383, opened 2023-03-03
[Outlaw] Grieftorch apl update

Zac-oihfdwrrtuinvbcp:
Use grief with bte + gs (only in st) debuffs up and off gcd.

Impact: about 0.5% st for standards HO builds, +1% st for gs builds, about 1.5% in slice for standard HO builds

### DIFF
--- a/engine/class_modules/apl/apl_rogue.cpp
+++ b/engine/class_modules/apl/apl_rogue.cpp
@@ -576,7 +576,7 @@ void outlaw_df( player_t* p )
   cds->add_action( "berserking" );
   cds->add_action( "fireblood" );
   cds->add_action( "ancestral_call" );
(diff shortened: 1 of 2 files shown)


engine/class_modules/apl/apl_rogue.cpp
```diff
@@ -576,7 +576,7 @@ void outlaw_df( player_t* p )
   cds->add_action( "berserking" );
   cds->add_action( "fireblood" );
   cds->add_action( "ancestral_call" );
-  cds->add_action( "use_item,name=manic_grieftorch,if=!stealthed.all&!buff.adrenaline_rush.up|fight_remains<5", "Default conditions for usable items." );
+  cds->add_action( "use_item,name=manic_grieftorch,use_off_gcd=1,if=gcd.remains>gcd.max-0.1,if=!stealthed.all&debuff.between_the_eyes.up&(!talent.ghostly_strike|debuff.ghostly_strike.up|spell_targets.blade_flurry>2)|fight_remains<=5", "Default conditions for usable items." );
```
EvanMichaels: There are two if= statements here, which seems not intentional? One of those isn't going to be used I don't think?

EvanMichaels (changes requested):

PR #7388, opened 2023-03-04
[Death Knight] Cleanup code for 10.0.7

nyterage:


engine/class_modules/sc_death_knight.cpp
```diff
@@ -4105,7 +4077,7 @@ struct army_of_the_dead_t : public death_knight_spell_t
 
     if ( is_precombat )
     {
-      double MIN_TIME = player -> base_gcd.total_seconds(); // the player's base unhasted gcd: 1.5s
+      double MIN_TIME = 1.5; // the player's base unhasted gcd: 1.5s
```
Khazakdk: We should continue to rely on the engine for this value instead of hard coding it.  While its unlikely the base gcd changes anytime soon, magic numbers should be avoided whenever possible.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4790,64 +4740,56 @@ struct unholy_pact_buff_t : public buff_t
...
+    set_stack_change_callback( [ this, p ] ( buff_t*, int, int new_ ) 
+    {
+      if ( new_ && p -> talent.unholy.ghoulish_frenzy.ok() )
+      {
+        p -> buffs.ghoulish_frenzy -> trigger();
+        debug_cast<pets::ghoul_pet_t*>( p -> pets.ghoul_pet ) -> ghoulish_frenzy -> trigger();
+      }
+      else
+      {
+        p -> buffs.ghoulish_frenzy -> expire();
```
Khazakdk: This block should be guarded by a check for the ghoulish frenzy talent so sims that don't have GF aren't making repeated cancel calls for a buff that doesn't get used.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5273,17 +5157,14 @@ struct death_coil_damage_t : public death_knight_spell_t
   {
     double m = death_knight_spell_t::composite_da_multiplier( state );
 
-    if ( p() -> talent.unholy.improved_death_coil.ok() )
-    {
-      m *= 1.0 + p() -> talent.unholy.improved_death_coil -> effectN( 1 ).percent();
-    }
+    m *= 1.0 + p() -> talent.unholy.improved_death_coil -> effectN( 1 ).percent();
```
Khazakdk: Did the talent check get lost here?

nyterage: This one ends up unnecessary, if the talent isnt present, it returns 0. Granted, for defensive coding, might be useful to have it there as a failsafe.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5959,19 +5813,24 @@ struct festering_wound_t : public death_knight_spell_t
           p()->gains.replenishing_wounds, this );
     }
   }
+
+  double composite_da_multiplier( const action_state_t* state ) const override
+  {
+    double m = death_knight_spell_t::composite_da_multiplier( state );
+
+    m *= 1.0 + p() -> talent.unholy.bursting_sores -> effectN( 1 ).percent();
+    m *= 1.0 + p() -> talent.unholy.improved_festering_strike -> effectN( 1 ).percent();
```
Khazakdk: Should these be in the constructor since these multipliers are always active?

engine/class_modules/sc_death_knight.cpp
```diff
@@ -5993,34 +5852,30 @@ struct festering_strike_t : public death_knight_melee_attack_t
...
+    double m = death_knight_melee_attack_t::composite_da_multiplier( state );
 
-      return m;
+    if ( p() -> talent.unholy.reaping.ok() && target -> health_percentage() < p() -> talent.unholy.reaping -> effectN( 2 ).base_value() )
+    {
+      m *= 1.0 + p() -> talent.unholy.reaping -> effectN( 1 ).percent();
     }
-  };
+     
+    m *= 1.0 + p() -> talent.unholy.improved_festering_strike -> effectN( 1 ).percent();
```
Khazakdk: Since this multiplier is always active if the talent is active should this be in the constructor?

engine/class_modules/sc_death_knight.cpp
```diff
@@ -6144,25 +6001,26 @@ struct frostwyrms_fury_t : public death_knight_spell_t
...
-      if ( p() -> buffs.shattering_blade -> up() )
-      {
-          m *= 1.0 + p() -> talent.frost.shattering_blade -> effectN( 1 ).percent();
-      }
+    if ( p() -> buffs.shattering_blade -> up() )
+    {
+      m *= 1.0 + p() -> talent.frost.shattering_blade -> effectN( 1 ).percent();
+    }
+
+    m *= 1.0 + p() -> talent.frost.improved_frost_strike -> effectN( 1 ).percent();
```
Khazakdk: This appears to be missing the talent `ok` check

PR #7390, opened 2023-03-05
[Priest] Add embrace shadow

taherbert:
Extends duration of shadow covenant

### DIFF
--- a/engine/class_modules/priest/sc_priest.hpp
+++ b/engine/class_modules/priest/sc_priest.hpp
@@ -110,6 +110,7 @@ public:
     propagate_const<buff_t*> power_of_the_dark_side;
     propagate_const<buff_t*> sins_of_the_many;
     propagate_const<buff_t*> shadow_covenant;
+    propagate_const<buff_t*> embrace_shadow;
 
     // Holy
     propagate_const<buff_t*> apotheosis;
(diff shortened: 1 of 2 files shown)


engine/class_modules/priest/sc_priest.hpp
```diff
@@ -735,6 +738,7 @@ struct priest_action_t : public Base, public parse_buff_effects_t<priest_td_t>
     // 280398 applies the buff to the correct spells, but does not contain the correct buff value (12% instead of 40%)
     // So, override to use our provided default_value (40%) instead
     parse_buff_effects( p().buffs.sins_of_the_many, false, true );
+    parse_buff_effects( p().buffs.embrace_shadow );
```
seanpeters86, line 741: can you add a comment here for what automagic is doing here

engine/class_modules/priest/sc_priest_discipline.cpp
```diff
@@ -279,7 +279,9 @@ void priest_t::create_buffs_discipline()
 
   buffs.shadow_covenant =
       make_buff( this, "shadow_covenant", talents.discipline.shadow_covenant->effectN( 4 ).trigger() )
-          ->set_trigger_spell( talents.discipline.shadow_covenant );
+          ->set_trigger_spell( talents.discipline.shadow_covenant )
+          ->modify_duration( talents.discipline.shadow_covenant->duration() +
```
seanpeters86, line 283: reading the code i believe `modify_duration` _ADDS_ the input duration to base. So here you want to just add the extra amount

seanpeters86 (changes requested):

PR #7395, opened 2023-03-06
[Priest] Update penance tickrate

taherbert:
Dynamically set penance's `base_tick_rate` and `dot_duration` to account for the talents castigation and harsh discipline.

There seemed to be some rounding error when creating a `timespan_t`, so changed `dot_duration` to be the tick rate times the number of ticks. It will only ever be a few milliseconds off.


engine/class_modules/priest/sc_priest.cpp
```diff
@@ -141,6 +141,11 @@ struct mind_blast_t final : public priest_spell_t
 
       priest().buffs.coalescing_shadows->expire();
     }
+
+    if ( priest().talents.discipline.harsh_discipline.enabled() )
+    {
+      priest().buffs.harsh_discipline->increment();
```
seanpeters86: I would use `trigger()` here instead, functionally they are the same in this case.

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -538,6 +543,10 @@ struct smite_t final : public priest_spell_t
       sim->print_debug( "{} adjusted cooldown of Chastise, by {}, without Apotheosis.", priest(),
                         cooldown_base_reduction );
     }
+    if ( priest().talents.discipline.harsh_discipline.enabled() )
+    {
+      priest().buffs.harsh_discipline->increment();
```
seanpeters86: I would use `trigger()` here instead, functionally they are the same in this case.

seanpeters86 (approved):

PR #7406, opened 2023-03-10
[Paladin] Initialize Divine Wrath

Fluttershy-git:
Currently, Paladin PTR sims end in a SegFault because of uninitialized Divine Wrath


mserrano:
oops, thanks

PR #7409, opened 2023-03-11
[Priest] Add expiation (discipline)

taherbert:
Adds the talent Expiation. I initially attempted to put this spell in discipline's cpp, but was getting the following error:

image

You can see the basic implementation structure here: https://github.com/simulationcraft/simc/compare/dragonflight...taherbert:simc:expiation

Co-locating expiation with the spells that utilize it (mind blast and SW:D) resolved the issue.

* Damage tested and confirmed
* Expiation placed as a child spell of mind blast and SW:D
* Damage amp applied via `apply_affecting_aura`
* Methodology based on Painbreaker Psalm, with minor modifications
* Using `tick_damage_over_time()` from shadow.cpp

Some formatting elsewhere in the file was adjusted by the c++ auto formatter. Let me know if problematic and I can disable (or adjust to a standard formatter if there is one).


taherbert:
@seanpeters86 getting some weird build errors, but don't see any code conflicts

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -144,6 +175,12 @@ struct mind_blast_t final : public priest_spell_t
       }
 
       priest().buffs.coalescing_shadows->expire();
+      if ( priest().talents.discipline.expiation.enabled() )
+      {
+        impact_action = new expiation_t( priest() );
```
seanpeters86: instead of creating the spell here I would prefer adding this to the constructor, should be a bit faster that way as well. We do this for things like `vampiric_touch_t` if you want to see a few examples.

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -1024,6 +1061,12 @@ struct shadow_word_death_t final : public priest_spell_t
         priest_td_t& td = get_td( s->target );
         td.buffs.death_and_madness_debuff->trigger();
       }
+      if ( priest().talents.discipline.expiation.enabled() )
+      {
+        impact_action = new expiation_t( priest() );
```
seanpeters86: same as the comment on mind blast

engine/class_modules/priest/sc_priest.cpp
taherbert: Revised to follow same methodology as vampiric touch, thanks!

engine/class_modules/priest/sc_priest.cpp
taherbert: Same as above

seanpeters86 (approved):

PR #7418, opened 2023-03-15
Implement a strawman version of Ret Aura

mserrano:
This is more or less the thing lightly mentioned in Discord, though I haven't touched the core global buffs before so figure it's worth getting feedback.

The approach here is that every player will, once a second, roll for whether ret aura procced that second or not - and we can customize the roll probability. 15% was completely arbitrarily chosen; we could pick a real estimated value to approximate what logs estimate the uptime to be - or we could just not have this proc at all by default, since (unlike the personal Retribution Aura from 10.0.0) it seems unlikely that it will meaningfully affect character optimization choices and may be less misleading than assuming a high uptime.

What do y'all think? Do we bother with this?

### DIFF
--- a/engine/player/player.cpp
+++ b/engine/player/player.cpp
@@ -3891,6 +3891,25 @@ void player_t::create_buffs()
         ->set_duration( sim->max_time * 3 )
         ->set_chance( as<double>( sim->overrides.windfury_totem ) );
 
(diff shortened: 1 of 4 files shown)


mserrano:
They do seem to have added some form of debuff decay here - I'm not yet sure how it works

Fluttershy-git:
I think with the removal of Retribution Aura, we can close this PR, since it's no longer needed?

PR #7442, opened 2023-03-20
[Havoc] Expose T30 fury tracking to the APL

hotbees:
For the T30 2pc, exposes the following values to the APL:
- The total amount of fury required to be spent to proc Seething Fury (`seething_fury_threshold`)
- The current fury spent towards proccing Seething Fury (`seething_fury_spent`)
- The remaining fury until proccing Seething Fury (`seething_fury_deficit`)

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -5957,16 +5957,15 @@ std::unique_ptr<expr_t> demon_hunter_t::create_expression( util::string_view nam
   }
   else if ( util::str_compare_ci( name_str, "seething_fury_deficit" ) )
   {
-    return make_fn_expr( "seething_fury_deficit", [ this ] {
-      if ( this->set_bonuses.t30_havoc_2pc->ok() )
-      {
+    if ( this->set_bonuses.t30_havoc_2pc->ok() )
+    {
+      return make_fn_expr( "seething_fury_deficit", [ this ] {
         return this->set_bonuses.t30_havoc_2pc->effectN( 1 ).base_value() - this->set_bonuses.t30_havoc_2pc_fury_tracker;
-      }
-      else
-      {
-        return 0.0;
-      }
-    } );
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -5945,6 +5945,22 @@ std::unique_ptr<expr_t> demon_hunter_t::create_expression( util::string_view nam
...
     });
   }
+  else if ( util::str_compare_ci( name_str, "seething_fury_threshold" ) )
+  {
+    return expr_t::create_constant( "seething_fury_threshold",
+                                    this->set_bonuses.t30_havoc_2pc->effectN( 1 ).base_value() );
+  } 
+  else if ( util::str_compare_ci( name_str, "seething_fury_spent" ) )
+  {
+    return make_fn_expr( "seething_fury_spent",
```
EvanMichaels: Can this just be a make_mem_fn_expr?

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -5945,6 +5945,22 @@ std::unique_ptr<expr_t> demon_hunter_t::create_expression( util::string_view nam
...
+                         [ this ] { return this->set_bonuses.t30_havoc_2pc_fury_tracker; } );
+  }
+  else if ( util::str_compare_ci( name_str, "seething_fury_deficit" ) )
+  {
+    return make_fn_expr( "seething_fury_deficit", [ this ] {
```
EvanMichaels: Wonder if this is worth returning a constant zero if not running T30?

EvanMichaels (changes requested):

engine/class_modules/sc_demon_hunter.cpp
hotbees: I don't know if it's necessarily worth it but it seems like a sane thing to do so I've made the change

PR #7452, opened 2023-03-23
[Priest] refactor AM to support upcoming PTR changes

seanpeters86:
Ancient Madness is moving towards being a passlist of spells found in Dark Ascension of Voidform. To make this work we remove the unique buff we have for Ancient Madness and work that effect into DA/VF respectively so that we can use parse_buff_effects to handle the passlist for us.

## PTR
Sim before the changes: <https://www.raidbots.com/simbot/report/krudZWtfkPsjZkkv5whbAz>
image

Sim with these changes locally:
image

## Live
Sim before the changes: <https://www.raidbots.com/simbot/report/sas8yqTdEbcXY8ctdnFTtc>
image

Sim with these changes locally:
image


engine/class_modules/priest/sc_priest.hpp
```diff
@@ -786,15 +785,20 @@ struct priest_action_t : public Base, public parse_buff_effects_t<priest_td_t>
...
       parse_buff_effects( p().buffs.mind_melt,
                           p().talents.shadow.mind_melt );  // Mind Blast instant cast and Crit increase
-
       parse_buff_effects( p().buffs.deathspeaker );
+
+      if ( p().talents.shadow.ancient_madness.enabled() )
+      {
+        parse_buff_effects( p().buffs.dark_ascension, 0b0001U, true, true );  // Skip Effect 1
+        parse_buff_effects( p().buffs.voidform, 0b0011U, true, true );        // Skip Effects 1 and 2
+      }
```
seanpeters86, line 794: wrapped in the talent check so debug output looks a bit better

engine/class_modules/priest/sc_priest.hpp
```diff
@@ -776,7 +775,7 @@ struct priest_action_t : public Base, public parse_buff_effects_t<priest_td_t>
   {
     // using S = const spell_data_t*;
 
-    parse_buff_effects( p().buffs.voidform );
+    parse_buff_effects( p().buffs.voidform, 0x4U, false, false );  // Skip 3 for Ancient Madness
```
seanpeters86, line 778: We use the masks to apply different rules for the different effects. We effectively parse buff effects twice for DA and VF:
1. Parse the original spell for the damage multiplier. Dont use stacks for this and mask off the effect for AM
2. Parse the original spell _again_ for the ancient madness passlist. Mask off the damage multiplier so it isnt applied twice and use stacks and the default value for this to get the crit increase from the spell.

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -2227,34 +2221,55 @@ namespace buffs
...
+      // When Ancient Madness expires, make sure we do not expire Voidform
+      set_stack_change_callback( [ this ]( buff_t* b, int old, int cur ) {
+        if ( cur == 1 )
+        {
+          set_reverse( false );
+          set_max_stack( 1 );
+          set_default_value( 0 );
+        }
+      } );
+    }
```
seanpeters86: This is a slight hack, without this when the buff tries to "remove" the crit portion it would also expire Voidform. We use this to say, actually stop giving the crit bonus and let the buff finish out.

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -2227,34 +2221,55 @@ namespace buffs
...
-    bool r = base_t::trigger( stacks, value, chance, duration );
-
+    // Reset this back
     if ( priest().talents.shadow.ancient_madness.enabled() )
     {
-      priest().buffs.ancient_madness->trigger();
+      set_reverse( true );
+      set_max_stack( 20 );
+      set_default_value( am_value );
     }
```
seanpeters86, line 2266: This is required so that on subsequent casts we aren't left with the old values

PR #7468, opened 2023-03-27
[Priest] Holy priest update

taherbert:
Fairly large update that adds all of the relevant Holy updates. A few changes to note that should be reviewed carefully, as they could impact other specs:

* Added a missing `override` on `summon_fiend_t's` `impact()`
* Divine image added to shared spells: SW:P, SW:D, Smite
* Added specialization conditionals to `init_background_actions` to declutter report
* Added specialization and PTR conditionals to `parse_buff_effects` to declutter report


engine/class_modules/priest/sc_priest.cpp
```diff
@@ -529,26 +529,36 @@ struct power_word_fortitude_t final : public priest_spell_t
...
+    {
+      child_searing_light = priest().background_actions.searing_light;
+    }
   }
 
   double composite_da_multiplier( const action_state_t* s ) const override
   {
-    double d = priest_spell_t::composite_da_multiplier( s );
+    double d              = priest_spell_t::composite_da_multiplier( s );
+    const priest_td_t* td = priest().find_target_data( s->target );
```
seanpeters86: this line is somewhat "expensive" at least relatively. Would highly suggest instead wrapping it below with `priest().talents.holy.searing_light.enabled()` and then declaring it and checking if the dot is ticking. That way you aren't finding target data EVERY time smite is called even if you do not have the talent

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -604,32 +620,48 @@ struct smite_t final : public priest_spell_t
...
+        {
+          chastise_cdr *= ( 1 + priest().talents.holy.apotheosis->effectN( 1 ).percent() );
+        }
+        if ( priest().talents.holy.light_of_the_naaru.enabled() )
+        {
+          chastise_cdr *= ( 1 + priest().talents.holy.light_of_the_naaru->effectN( 1 ).percent() );
+        }
+        sim->print_debug( "{} adjusted cooldown of Chastise, by {}, with light_of_the_naaru: {}, apotheosis: {}",
+                          priest(), chastise_cdr, priest().talents.holy.light_of_the_naaru.enabled(),
+                          ( priest().talents.holy.apotheosis.enabled() && priest().buffs.apotheosis->up() ) );
```
seanpeters86: you most certainly want to use `check()` here instead of `up()`. I think using up here would skew benefit calcs and you just need check since its a print statement

engine/class_modules/priest/sc_priest_holy.cpp
```diff
@@ -12,130 +12,359 @@ namespace priestspace
...
-      return 0;
-    return priest_spell_t::cost();
+    priest_spell_t::execute();
+
+    priest().buffs.apotheosis->trigger();
+    sim->print_debug( "starting apotheosis. " );
+    priest().cooldowns.holy_word_chastise->reset( false );
+    priest().cooldowns.holy_word_serenity->reset( false, -1 );
+    priest().cooldowns.holy_word_sanctify->reset( false, -1 );
+    sim->print_debug( "apotheosis reset holy word cooldowns." );
```
seanpeters86: any reason in particular for these debug messages?

engine/class_modules/priest/sc_priest_holy.cpp
```diff
@@ -12,130 +12,359 @@ namespace priestspace
...
     parse_options( options_str );
     harmful = false;
   }
+
+  void execute() override
+  {
+    priest_spell_t::execute();
+
+    priest().buffs.divine_word->trigger();
+    sim->print_debug( "starting divine_word." );
```
seanpeters86: this print statement feels superfluous

engine/class_modules/priest/sc_priest_holy.cpp
```diff
@@ -12,130 +12,359 @@ namespace priestspace
...
   {
     parse_options( options_str );
     harmful = false;
   }
 
+  void execute() override
+  {
+    priest_spell_t::execute();
+
+    sim->print_debug( "{} starting empyreal_blaze. ", priest() );
```
seanpeters86: this print statement feels superfluous

engine/class_modules/priest/sc_priest_holy.cpp
```diff
@@ -12,130 +12,359 @@ namespace priestspace
...
+    if ( priest().talents.holy.divine_word.enabled() && priest().buffs.divine_word->check() )
+    {
+      priest().buffs.divine_word->expire();
+      priest().buffs.divine_favor_chastise->trigger();
+    }
+    if ( priest().talents.holy.divine_image.enabled() )
+    {
+      priest().buffs.divine_image->trigger();
+      // Activating cast also immediately executes searing light
+      sim->print_debug( "searing_light triggered by holy_word_chastise" );
```
seanpeters86: im assuming this might be done for tracking purposes? not sure why you need this print statement though. Perhaps you could pass in the action name when triggering searing light and then do something with it to make the reporting clear?

engine/class_modules/priest/sc_priest_holy.cpp
```diff
@@ -12,130 +12,359 @@ namespace priestspace
...
+    }
+  }
+
+  void impact( action_state_t* s ) override
+  {
+    priest_spell_t::impact( s );
+    sim->print_debug( "divine_image_buff: {}", priest().buffs.divine_image->up() );
+    if ( child_searing_light && priest().buffs.divine_image->up() )
+    {
+      sim->print_debug( "searing_light triggered by holy_word_chastise" );
```
seanpeters86: same as above

engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -855,6 +859,15 @@ struct shadow_word_pain_t final : public priest_spell_t
       }
 
       priest().refresh_insidious_ire_buff( s );
+
+      if ( child_searing_light && priest().buffs.divine_image->up() )
+      {
+        sim->print_debug( "searing_light triggered by shadow_word_pain" );
```
seanpeters86: same as above, i think instead of making a bunch of print statement telling you what procced searing light storing that inside of a proc or even a separate action per spell (like we do with psychic link) would feel better than adding a bunch of debug statements

seanpeters86 (changes requested):

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -604,32 +620,48 @@ struct smite_t final : public priest_spell_t
...
+
+        priest().cooldowns.holy_word_chastise->adjust( -chastise_cdr );
+      }
+      if ( child_searing_light && priest().buffs.divine_image->up() )
+      {
+        sim->print_debug( "searing_light triggered by smite" );
```
scamille:
```suggestion
        sim->print_debug( "{} searing_light triggered by smite", priest() );
```

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -1033,6 +1071,15 @@ struct shadow_word_death_t final : public priest_spell_t
         child_expiation->target = s->target;
         child_expiation->execute();
       }
+
+      if ( child_searing_light && priest().buffs.divine_image->up() )
+      {
+        sim->print_debug( "searing_light triggered by shadow_word_death" );
```
scamille:
```suggestion
        sim->print_debug( "{} searing_light triggered by shadow_word_death", priest() );
```

engine/class_modules/priest/sc_priest.cpp
taherbert: Resolved

engine/class_modules/priest/sc_priest_holy.cpp
```diff
@@ -162,54 +409,49 @@ action_t* priest_t::create_action_holy( util::string_view name, util::string_vie
 
   if ( name == "holy_fire" )
   {
-    return new holy_fire_t( *this, options_str );
+    return new holy_fire_t( *this );
```
scamille: Hmm removing the options_str for holy_fire does not seem like a good idea. For the version that is not a background action you do want to have this passed to the action and parsed.

A clean solution is having 2 constructors, one that takes the options_str and creates a non-background action, and another that does not take the options_str and creates the background action. You can delegate one of the constructors to the other.

engine/class_modules/priest/sc_priest.cpp
taherbert: Changed to `check()` here, as well as for divine image.

engine/class_modules/priest/sc_priest_holy.cpp
taherbert: Primarily for my testing, removed.

engine/class_modules/priest/sc_priest_holy.cpp
taherbert: Removed

engine/class_modules/priest/sc_priest_holy.cpp
taherbert: Removed

engine/class_modules/priest/sc_priest_holy.cpp
taherbert: Added a proc to track divine_image instead, removed these debug statements.

engine/class_modules/priest/sc_priest_holy.cpp
taherbert: Resolved above

engine/class_modules/priest/sc_priest_shadow.cpp
taherbert: Resolved above

engine/class_modules/priest/sc_priest_holy.cpp
taherbert: Added this back in, and passed an empty string to `background_actions.holy_fire = new actions::spells::holy_fire_t( *this, "" );`

engine/class_modules/priest/sc_priest_holy.cpp
```diff
@@ -164,52 +407,47 @@ action_t* priest_t::create_action_holy( util::string_view name, util::string_vie
...
 
   return nullptr;
 }
 
-expr_t* priest_t::create_expression_holy( action_t*, util::string_view /*name_str*/ )
+void priest_t::init_background_actions_holy()
 {
-  return nullptr;
+  background_actions.holy_fire     = new actions::spells::holy_fire_t( *this, "" );
+  background_actions.searing_light = new actions::spells::searing_light_t( *this );
```
scamille: You could only create the searing light action if ( priest().talents.holy.divine_image.enabled() ) here, and then you don't have to do the check in the actions where you are using it.

The actual advantage is that this way the action doesn't get created in the first place if you don't have the talent.

engine/class_modules/priest/sc_priest_holy.cpp
```diff
@@ -164,52 +407,47 @@ action_t* priest_t::create_action_holy( util::string_view name, util::string_vie
...
   }
 
   return nullptr;
 }
 
-expr_t* priest_t::create_expression_holy( action_t*, util::string_view /*name_str*/ )
+void priest_t::init_background_actions_holy()
 {
-  return nullptr;
+  background_actions.holy_fire     = new actions::spells::holy_fire_t( *this, "" );
```
scamille: Same thing for holy fire.
```suggestion
  if ( talents.holy.divine_word.enabled() )
  {
  background_actions.holy_fire     = new actions::spells::holy_fire_t( *this, "" );
  background_actions.holy_fire -> background = true;
  }
```

In smite_t you can then do
```
smite_t(...) : child_holy_fire(priest().background_actions.holy_fire) ...
```
instead of first initializing it to nullptr and then setting it in the ctor body.

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -2157,10 +2209,19 @@ void priest_t::init_rng()
...
-  background_actions.echoing_void = new actions::spells::echoing_void_t( *this );
-
-  init_background_actions_shadow();
-  init_background_actions_discipline();
+  if ( specialization() == PRIEST_SHADOW )
+  {
+    background_actions.echoing_void = new actions::spells::echoing_void_t( *this );
+    init_background_actions_shadow();
+  }
+  else if ( specialization() == PRIEST_DISCIPLINE )
```
scamille: I am not the biggest fan of having to use specialization() checks everywhere in initialization. If you correctly conditionalize things on spell data being enabled (see my comments on background actions holy fire and searing light), you don't need artificial specialization() checks. This makes things also just work out if you have a background action that is available for both discipline and holy, let's say based on a shared talent, without having yo create yet another condition branch.

Same applies for buff and debuff effects.

That said, I know that they can make life easier and a lot of other class modules do it that way just as well. I don't really care that much, but thought it worth mentioning.

@seanpeters86 what is your opinion on this?

scamille (approved):
Looks great overall! Nice to see someone investing the time to get the non-shadow module working again.

engine/class_modules/priest/sc_priest_holy.cpp
taherbert: Awesome insight, revised!

engine/class_modules/priest/sc_priest.cpp
seanpeters86: I generally agree, I think the check here probably isnt needed @taherbert ?

engine/class_modules/priest/sc_priest.hpp
```diff
@@ -577,6 +613,9 @@ struct priest_t final : public player_t
 
     // Last tick of Mind Sear does not get buffed by Gathering Shadows
     bool gathering_shadows_bug = true;
+
+    // Time in seconds between prayer of mending bounces
+    double prayer_of_mending_bounce_rate = 1.5;
```
seanpeters86: just a note we'll want to add this option into the wiki post-merge

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -2349,6 +2406,7 @@ void priest_t::create_options()
   add_option( opt_bool( "priest.gathering_shadows_bug", options.gathering_shadows_bug ) );
   add_option( opt_bool( "priest.t30_multiple_bender", options.t30_multiple_bender ) );
   add_option( opt_bool( "priest.t30_yshaarj", options.t30_yshaarj ) );
+  add_option( opt_float( "priest.prayer_of_mending_bounce_rate", options.prayer_of_mending_bounce_rate ) );
```
seanpeters86: We probably want to add at least a min for this option since you are passing it directly into the buff duration, potentially a max as well?

seanpeters86 (changes requested):

engine/class_modules/priest/sc_priest_shadow.cpp
seanpeters86: @taherbert i still see this debug statement here btw, is that an error?

engine/class_modules/priest/sc_priest.cpp
```diff
@@ -1033,6 +1065,15 @@ struct shadow_word_death_t final : public priest_spell_t
         child_expiation->target = s->target;
         child_expiation->execute();
       }
+
+      if ( child_searing_light && priest().buffs.divine_image->check() )
+      {
+        sim->print_debug( "{} searing_light triggered by shadow_word_death", priest() );
```
seanpeters86: is this print still needed?

engine/class_modules/priest/sc_priest.cpp
taherbert: Agree, check here not needed. I think we may still need the check around `parse_buff_effects` to de-clutter the report, though.

engine/class_modules/priest/sc_priest.cpp
taherbert: Removed (and one more above)

engine/class_modules/priest/sc_priest.cpp
taherbert: Added both min and max, with comments describing

seanpeters86 (approved):

PR #7482, opened 2023-03-30
Feature/dre proc change

Bloodmallet:
Deeply Rooted Elements procs differently now.

Based on [this log](https://www.warcraftlogs.com/reports/wD6XTY8Ap3GMa1BL#fight=last&type=damage-done&source=3) it looks like the proc chance increments by 1% each time it tries to proc. While the first two attempts are missed. Which means the attempts have a chance like:
- 0%
- 0%
- 1%
- 2%
- 3%
- ...

The implementation simply substracts 2% and let's the buff stack 1%for each stack of the artificial buff.


Bloodmallet:
Merged after a brief discussion on discord.

PR #7488, opened 2023-04-02
[Vengeance] reorder Fracture implementation

ToppleTheNun:
order of events matters for how Fracture damage
is calculated with T30 4pc.


ToppleTheNun:
@EvanMichaels

PR #7489, opened 2023-04-02
[Vengeance] use new effect on Fiery Brand debuff

ToppleTheNun:
it adds Fracture to the list of spells affected by the Fiery Brand debuff for Fiery Demise.


ToppleTheNun:
@EvanMichaels

PR #7499, opened 2023-04-04
[Priest] twilight equilibrium bugfix

taherbert:
should use execute() instead of impact(). This avoids ticks of dots counting toward the buff.


vituscze:
Is there any reason the code is using `this->execute_state->action` instead of just `this`?

taherbert:
> Is there any reason the code is using `this->execute_state->action` instead of just `this`?

Just my own ignorance, `this` works perfectly! TIL. Updated.

seanpeters86 (approved):

PR #7533, opened 2023-04-07
[Feral] Updated talent string

jordanritzie19:


nyterage:
needs to be updated in the generator at [/profiles/generators/Tier29/](https://github.com/simulationcraft/simc/blob/dragonflight/profiles/generators/Tier29/T29_Generate_Druid.simc)

Otherwise will revert to the old talent string the next time profiles are regenerated.

jordanritzie19:
Sent another pr updating in the generator instead, thanks.

PR #7559, opened 2023-04-11
[Paladin] Added support for Executioner's Will talent in Retribution Paladin APL

vincent-clipet:
Executioner's Will is a talent extending the duration of Execution Sentence from 8s to 12s.
Just added a check for it in the APL so that sims don't try to cast a 12s Execution Sentence _(which might do 0 damage if the target dies too fast)_ when `target.time_to_die` is between **8s** and **12s**.
The impact should be very minimal (the current APL has around `(4s x 2 casts per minute) / 60s` = 13.3% chance to cast an ES that won't do any damage at the end of a single-target fight; a bit more on fight styles with adds)


vincent-clipet:
@mserrano sorry for the ping, Bolas told me to contact you for this kind of things
https://discord.com/channels/115288062094868481/401132846342864896/1103032328286126111

mserrano (approved):

PR #7580, opened 2023-04-21
[Havoc] 10.1 Accelerated Blade and Serrated Glaive changes

hotbees:
In 10.1 PTR build 49222:
* Accelerating Blade has been renamed to Accelerated Blade
* Serrated Glaive now affects Chaos Strike + Annihilation

I added a fallback for renaming Accelerating Blade and I moved Serrated Glaive to be a generic affected_by check similar to Essence Break. I also added separated benefit tracking via dummy procs similar to Essence Break.


hotbees:
@EvanMichaels

PR #7592, opened 2023-04-22
[Gear] Dominance correct spellID for shaman

Dsune0:
192241 was used, should be 192249 for storm ele


Dsune0:
@HawkCorrigan please correct if the ID is wrong or merge if there is no issue

PR #7593, opened 2023-04-23
[shaman] Fix Arrogance Trinket for Ele

HawkCorrigan:
Storm Elemental spell id was typoed and I set both Elemental casts to harmful as well now.

Works fine in local sims now.

Kind of sad that we cannot precast the Elementals with that change but should be fine for now.


Bloodmallet (approved):

Saeldur:
Tested - Working correctly for both Storm Elemental and Fire Elemental.

I mentioned in discord but if you want the precombat functionality back you can use something similar what druid has here: https://github.com/simulationcraft/simc/blob/dragonflight/engine/class_modules/sc_druid.cpp#L9938-L9976
Loop through precombat actionlist in `shaman_t::init_finished` and coerce the actions to storm/fire elementals like druid does for wrath and set them to be harmful = false.

PR #7597, opened 2023-04-27
[Mage] Add support for Tier 30 fire set bonus

Biosins:


vituscze:
I'm going to close this since @Dorovon updated the APL earlier. But thanks for the PR!

PR #7610, opened 2023-04-29
[Havoc] 10.1 Transition Updates

hotbees:
* Removes fallback for dealing with different names between live and ptr
* Renames Accelerating Blade to Accelerated Blade for 10.1
* Adds reporting rename for new essence break damage type SCHOOL_CHROMASTRIKE to SCHOOL_CHAOS for consistency with chromatic rename.


hotbees:
@EvanMichaels

PR #7621, opened 2023-05-01
Create T30_Mage_Fire.simc

Degenhours:


nyterage:
This has been done improperly for the way we generate these profiles. The base profile should be placed in the T30_Generate_Mage generator file, located here https://github.com/simulationcraft/simc/blob/dragonflight/profiles/generators/Tier30/T30_Generate_Mage.simc . Once its been placed there, the generate_profiles script in the main directory should be run to generate the file in question.

PR #7635, opened 2023-05-04
[Gear] Add option to controll Spoil of Netharus' buff on pull

Shurtugal:
Using out-of-combat spells, you can set the Spoils of Neltharus trinket to whatever stat you want prepull. This is an attempt at mirroring this, by adding an option that forces the trinket to a specific stat if used at time=0. It's a bit janky, I would have preferred to set the buff_list to have the chosen buff in first instead of this hack, but I couldn't find how.


Shurtugal:
Here is a compressed html file of a sim showcasing the use of this option, with all 4 stats as well as random (default) use. The new_T30 line is a slightly different APL that doesn't use the trinket right on pull, so it gets very similar results to random.

[spoils.zip](https://github.com/simulationcraft/simc/files/11402525/spoils.zip)

As you can see, if used correctly this option can result in a decent dps gain (and probably even more on other specs than frost mage), so it feels important to me that this feature gets added to simc.

engine/player/unique_gear_dragonflight.cpp
```diff
@@ -1815,6 +1815,18 @@ void spoils_of_neltharus( special_effect_t& effect )
       proc_spell_t::execute();
 
       buff_list.front().first->expire();
+
+      if ( sim->current_time() == timespan_t::zero() )
```
gastank: You can now check the `is_precombat` bool in the action to see if it's being executed in the precombat apl. However, this entire section is unnecessary. Instead of individual string comparison, you can edit the init_buff lambda in the constructor here https://github.com/simulationcraft/simc/blob/dragonflight/engine/player/unique_gear_dragonflight.cpp#L1784 to `emplace_front` instead if the option string matches `n`.

You would then move shuffle line here https://github.com/simulationcraft/simc/blob/dragonflight/engine/player/unique_gear_dragonflight.cpp#L1803 to `void init_finished()` and only shuffle if `!is_precombat`

Shurtugal: I tried doing what you suggested. The `emplace` part works, I even added a `print_debug` to make sure the `buff_list` is in the correct order after the 4 `init_buff` lines. I moved the shuffle to `void init_finished() override`, and then commented it out for testing (the problem of `is_precombat` is that the trinket can't be used in precombat currently as it's flagged as harmful, but that's not the main issue right now).
But the buff obtained is random, as when I use the trinket only on pull each of the 4 stats is obtained 25% of the time, indicating the callback is executed. This is still true when the trinket is used as the first action of the sim, with nothing in precombat. But for some reason, when I turn on `debug=1`, the trinket always give the chosen buff in the log. Is the `buff_list` initialized only in the first iteration and not at the beginning of each?
I I'm pretty lost on what's happening here, and would gladly take any help.

I can commit my non-functional work if it helps resolving the issue. We can also move this discussion on Discord, I'm on the simc server.

Shurtugal:
I fixed everything as discussed, and also added an event in `void reset()` to properly trigger the counter buff on pull: previously, there was no counter buff until the first proc, as nothing was triggering it. As a result, an APL trying to use the trinket with a specific stat wasn't able to do so if the fight started with that buff active, and had to wait for it to swap and then swap back to that stat.

PR #7644, opened 2023-05-05
[Warrior] Update T30 Protection profile

mooman219:
Talked to Nom, this seems like a good starting point for a single target T30 profile.


Armadk:
Looks like we were both working on it at the same time.  From what I could tell, it was just some talent changes and the gem changes?

mooman219:
I tested several talent batches and landed on this one for ST. I made a spreadsheet for tier and LS uptime and it looked like we should tool are talents around that, which worked out to a large DPS increase. I can link that if you feel it's interesting.

The gems and enchantments are alright, this is roughly as good as it gets but I attempted to keep extra haste because it's a higher quality of life if swapping was within margin of error. I didn't sim the full range or embellishment options for the two shadowflame patches because it was kind of a pain, so I'm not 100% sure this ideal, but it's probably close enough.

This picture is a little out of date but is roughly what I landed on. 
image-1

PR #7663, opened 2023-05-06
[Warrior] Remove redundant APL  entry

mooman219:
The following line catches this one either way.

How is warrior's APL generated? Does the cpp generate the simc, or the other way around? It looks like the simc generates the cpp here <https://github.com/simulationcraft/simc/tree/dragonflight/engine/class_modules/apl>, but not for warrior.


Armadk:
Warrior is currently generated inside the classes/sc_warrior.cpp file, however many classes have moved apl generation to outside the core file, and into the one under the apl/ directory.  There is currently an open PR for moving the warrior APL to this file.  Just waiting on final approval from Archi and it should be fine to be merged in.

PR #7702, opened 2023-05-10
[Warrior] Update Protection profile

mooman219:
Bonus IDs should be correct.

This is after re-simming the item and bugfix changes on May 9th


mooman219:
cc @Armadk If you have a chance to take a look

PR #7718, opened 2023-05-15
[Profiles] T30 Regenerate

hotbees:
Forgot to do this when changing the default havoc phial in #7715. Also includes arcane & balance apl changes that happened to be updated since last regeneration.


Bloodmallet:
Looks like in the meantime some regenerations happened. I'm closing this one. Thank you for providing it, though!

PR #7756, opened 2023-05-31
[Death Knight] Experimental pet inheritance delay 

Khazakdk:


nyterage:
Since i am changing a core pet calculation, verifying every profile remains static (outside of the expected DK changes)

Heres a before and After for the pet ap_to_ap conversion calculation. 

https://imgur.com/a/eoPPNYC

PR #7760, opened 2023-06-01
[Monk] Execute Bountiful Brew action on proc

justindriggers:
`create_proc_callback` searches for an effect that matches the spell ID of an existing action. Bountiful Brew's effect has no spell ID, which results in it never being executed.

We can provide the action as an override, but we additionally need to set it to the effect's `execute_action`.

Bountiful Brew was previously not set as `background = true`, but I'm not sure why. Executing the proc'd action without this results in an error.


justindriggers:
Handled by https://github.com/simulationcraft/simc/pull/7762

PR #7767, opened 2023-06-07
[WARLOCK] Changes to the APL -  target.time_to_die.

Mlklock:
Chang TTD in favor of Target.Time_To_Die  to avoid issues with Iced Phial of Corrupting Rage

Exemple of Before/After 

https://www.raidbots.com/simbot/report/uAmi7mdNwt3dHqjnfgSVwR

https://www.raidbots.com/simbot/report/boKrETD55TYMtSpzhfQ7Yz

### DIFF
--- a/engine/class_modules/apl/warlock.cpp
+++ b/engine/class_modules/apl/warlock.cpp
@@ -110,7 +110,7 @@ void affliction( player_t* p )
   aoe->add_action( "seed_of_corruption,if=talent.sow_the_seeds" );
   aoe->add_action( "malefic_rapture" );
   aoe->add_action( "drain_life,if=(buff.soul_rot.up|!talent.soul_rot)&buff.inevitable_demise.stack>10" );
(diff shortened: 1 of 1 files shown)


Saeldur:
LGTM and massively fixes sim time issues Demo is facing rn

Azevara:
`fight_remains` or `interpolated_fight_remains` should be used rather than `target.time_to_die` if the intention is to still have the expression measure time until the simulation iteration ends. I recognize these are not currently searchable in the wiki; `fight_remains` is present in several other APLs and should be sufficient for this fix

Saeldur:
~Some should still be `target.time_to_die` unless you wish to completely break support for sims with add events.~
Actually upon closer inspection should be `fight_remains` - the baseline Demo apl doesn't really distinguish raid events at all and would need extending with support for raid events anyway, especially to do with tyrant, dogs, and implosion. Better to not have it rip resources on pulls ending for the most part unless it can be made more aware.

PR #7773, opened 2023-06-12
[Warlock] Implement Destro & Affliction 10.1.5 Changes

Saeldur:
Implement New Crashing Chaos for Destro

Should support Both Live and PTR.

Note: This is actually bugged on 10.1.5 PTR and is only applying the R1 always (R2 does nothing), I implemented it here as if it wasn't bugged though, since it should get fixed.

Confirmed the ROF does gain a Persistent Multiplier for the entire ROF. No other funny business like with madness.

https://www.warcraftlogs.com/reports/GaLpd8qc1hHYDbzT/#fight=last&type=damage&source=1&target=2.1&ability=42223&view=events

Test of ROF -> Infernal ROF ROF
Can see the normal ROF continue without being buffed.


Saeldur:
Affliction functions for both 10.1 and 10.1.5
- Implements Xavius' Gambit (Auto implemented by apply affecting)
- Handles not triggering Malefic Affliction/Dread Touch
- Handles new Doomblossom trigger and disables old

Azevara:
The functionality of the code appears fine but for the sake of future reviews please separate individual talent updates into separate PRs.

There are formatting/style/comments issues but that is not currently blocking anything, so I will merge.

PR #7799, opened 2023-06-16
[APL] Demonology Warlock APL Change - Power Siphon use inside Tyrant …

Mlklock:
…Setups.

https://www.raidbots.com/simbot/report/sKA3uRxf2MaGKhY77RMUVc


nyterage:
just a minor heads up

`use_pet_stat_update_delay` is a player scope option. putting it in the header as you did in the sim just causes it to ignore the option. 

needs to be placed within the player scope (somewhere after the first profiles name, but, before the next profile)

heres a sample sim of this change with that being placed properly. 
https://www.raidbots.com/simbot/report/nSpH4RHqiW65saggbyS13E

Mlklock:
Ops my bad and thanks for it

PR #7800, opened 2023-06-16
[Warlock] Add precombat timer for Power Siphon

SirPownzalot:
Adding precombat timer for Power Siphon in the same style as some rogue abilities currently have.
The current sim uses Power Siphon at 0s, but in reality we use it at -10s (this has implications mainly for our Fel Covenant build, but is still relevant to all other Power Siphon builds).

This isn't perfect because I am not adjusting the Inner Demons imp spawn cycle at all, but is better than what we currently use.


seanpeters86:
closed due to stale/merge conflicts. feel free to re-open if still relevant

PR #7823, opened 2023-06-23
Make Surv APL use Fury of the Eagle in ST when Djaruun is not equipped

SchmidNiklas:
The current Survival APL does not use Fury of the Eagle in single target if Djaruun is not equipped since the expression cooldown.elder_flame_408821.remains>40 always evaluates to false. This PR attempts to fix that by adding a conditional for when Djaruun is not equipped to allow the APL to use FotE in single target scenarios.

https://www.raidbots.com/simbot/report/jtA8cT2sV8AYJany9hKCKS Comparing the APL change on the T30 profile (adjusted talents and equipped crafted weapon in place of Djaruun)


Jayezi:
Can you also add this to the [text apl](https://github.com/simulationcraft/simc/blob/6f0dc74d2f7959679a14640cda321a2814693a14/engine/class_modules/apl/hunter/sv.txt), we have a workflow to edit those and convert them all to the generated cpp that would lose the change if it isn't mirrored there.

SchmidNiklas:
Done :+1:

PR #7837, opened 2023-06-27
[Hunter] Prioritise Rapid Fire a bit more if you won't cap Aimed Shot charges

ZingusWoW:
It's a gain to Prioritise Rapid Fire over Aimed Shot if you won't cap Aimed Shot charges. Seems to be slightly gear dependent as it's bigger for my personal sim.

Personal Sim: https://www.raidbots.com/simbot/report/ksyP6FjBeBzPkYZRdFreVw
T30 MM Sim (Using new best talents): https://www.raidbots.com/simbot/report/jvZcjUkYAy4bTC4U74GqKW

### DIFF
--- a/engine/class_modules/apl/apl_hunter.cpp
+++ b/engine/class_modules/apl/apl_hunter.cpp
@@ -180,7 +180,7 @@ void marksmanship( player_t* p )
   st->add_action( "stampede" );
   st->add_action( "death_chakram" );
   st->add_action( "wailing_arrow,if=active_enemies>1" );
-  st->add_action( "rapid_fire,if=talent.surging_shots" );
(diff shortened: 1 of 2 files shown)


Pewtro:
Shouldn't this be looking at aimed shot cast time instead of the GCD? so it would be something like action.aimed_shot.cast_time+cast_time rather than gcd+cast_time (of RF)

ZingusWoW:
> Shouldn't this be looking at aimed shot cast time instead of the GCD? so it would be something like action.aimed_shot.cast_time+cast_time rather than gcd+cast_time (of RF)

Yep makes more sense to do that, I've changed both to do that and pushed it back up

PR #7863, opened 2023-07-02
[Warlock] Add Expression to Track Expiring IGB

Saeldur:
Expose number of expiring Imp Gang Bosses to APL.
Lets you check how many IGB are on last cast or second to last cast.

This is needed to be able to do APL work (reasonably fast, local sims are too slow for demonology sadly) to check if caring about IGB expiry provides a significant damage gain towards Implosion ST for 10.1.5.

### DIFF
--- a/engine/class_modules/warlock/sc_warlock.cpp
+++ b/engine/class_modules/warlock/sc_warlock.cpp
@@ -2092,6 +2092,22 @@ std::unique_ptr<expr_t> warlock_t::create_pet_expression( util::string_view name
       } );
     } );
   }
+  else if ( name_str == "last_cast_igb_imps" )
+  {
+    return make_fn_expr( "last_cast_igb_imps", [ this ]() {
+      return warlock_pet_list.wild_imps.n_active_pets( []( const pets::demonology::wild_imp_pet_t* pet ) {
+        return pet->resources.current[ RESOURCE_ENERGY ] < 32 && pet->buffs.imp_gang_boss->check();
+      } );
+    } );
+  }
+  else if ( name_str == "two_cast_igb_imps" )
+  {
+    return make_fn_expr( "two_cast_igb_imps", [ this ]() {
+      return warlock_pet_list.wild_imps.n_active_pets( []( const pets::demonology::wild_imp_pet_t* pet ) {
+        return pet->resources.current[ RESOURCE_ENERGY ] < 48 && pet->buffs.imp_gang_boss->check();
+      } );
+    } );
+  }
   else if ( name_str == "igb_ratio" )
   {
     return make_fn_expr( "igb_ratio", [ this ]() {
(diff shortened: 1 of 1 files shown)


Azevara:
Is there an example of WeakAura or addon that currently exists and is used in game that can track this behavior accurately?

Saeldur:
Malacath sent me https://wago.io/RT5SwAJwa originally, I don't believe it specifically shows if you're going to lose an Imp gang boss next (it does some other calculations) but the reasons for adding this is to attempt to evaluate if it's actually worth caring about in any reasonable metric. 
E.g. if the increase provided by caring about this is large.

Other alternatives would be like how that weakaura works and exposing the number of imp gang firebolts left, and how many total imp firebolts are left. 
Neither of which are in this, but would potentially be good although I don't believe having to do the calculation that precisely is necessary.

In my eyes a combination of both is likely to produce the best results. E.g. only use implosion when other aspects result in a gain by checking aspects like firebolts, current core economy state, and only do the implosion if you're about to lose an imp gang boss which will be one of the main factors for imploding.
This may actually be how the weakaura triggers to tell you to implode (I haven't checked how it functions)

Mlklock:
> https://wago.io/RT5SwAJwa

Just to add to your own commentary, it won't directly tell you are going to loose a IGB, but if you look at the bar it have a color code in the bar for Wild Imps vs Imp Gang Boss.

The Blue Bar with pink background are specifically Imp Gang Bosses, so for exemple, https://media.wago.io/screenshots/RT5SwAJwa/6375785852f0995dd5e7578e.jpg 

shows you have 2 Regular Wild Imps with 2 cast remaining, 1 IGB with 2 casts remaining, 1 Wild imp with 4 casts remaining, 2 IGB's with 4 cast remaining and  1 Wild Imps with 6 cast remaining  and 1 IGB with 6 cast remaining. 

----------------------------

The WA also does some calculation about if Implosion is worth in single target.

https://imgur.com/y6aSBpz 

In this image exemple (the exemple is in the WA page), he calculates that implosion is worth an extra 25% spellpower 

I think his method to calculate the worth of a Implosion is very interesting and might be worth to check for the sake of simcraft.

Rhaki:
I was invited to take a look at this conversation. I am the creator of the linked weak aura. If you want I'm available to talk about how to calculate the gain to cast or not the implosion. Several internal calculations are made, if you want I can expose them. The latest version on wago is outdated (there are a couple of things that I tweaked but didn't push, sadly I didn't have time to test everything definitively before making it public)

Azevara:
Going to commit this PR with the following caveats:
- The weakaura is an available tracking mechanism that provides state info that player can use, so that meets the minimum criteria for an acceptable expression
- However, calculations being done by the weakaura to recommend Implosion may still not be 100% replicable in the APL given the complexity of the strings that would be written
- With that in mind, custom expressions to replicate those later calculations will most likely not be included in the module

PR #7875, opened 2023-07-05
[Warlock] Aff BiS Profile Update

abousumer:
https://www.raidbots.com/simbot/report/wafJRGMsAJB3hCioVcDthv


Fluttershy-git:
That profile has two main stat gems (One in head, one in neck)

abousumer:
Oh missed that, thank you. It's neutral with that fixed

PR #7894, opened 2023-07-10
[Paladin] Update Talents to 10.1.5 compatible, change Ret Ring to Fre…

Fluttershy-git:
…ehold, regen Profiles


cremor:
I think this is not needed any more now that #7904 has been merged, right?

Fluttershy-git:
> I think this is not needed any more now that #7904 has been merged, right?

https://github.com/simulationcraft/simc/commit/f338ef2a0ce39afbce9645b7731869d599089d80 
But yes, discussed this quickly with @mserrano , this PR isn't needed anymore

PR #7904, opened 2023-07-11
[Paladin] Retribution Updates

keyboardmash:
Updated APL including adding PI support and AoE improvements.
Updated default profile for 10.1.5 and added an alternative base CS profile.


cremor:
@keyboardmash @mserrano What about the file [apl_paladin.cpp](https://github.com/simulationcraft/simc/blob/dragonflight/engine/class_modules/apl/apl_paladin.cpp), does it not need to be updated/regenerated?
The APL changes are not visible in the regenerated profiles, so I assume this is the missing link.

**edit:** The file has been updated, thanks.

PR #7934, opened 2023-07-15
[Havoc] Remove redundant trinket lines in APL

hotbees:


hotbees:
The initial commit removed the top set of generic trinket lines because that's how it was in our working APL but previous APLs had generics lower in priority and it probably makes more sense to have them down there, so I moved the default item use logic to the bottom of the priority

PR #7938, opened 2023-07-16
[Arcane] Low Target AOE Updates and Touch optimization for T30

ArcanePorom:


Sergrand1:
Hey Porom, appreciate the work here. I can't approve this PR as it is. As part of your commit, you reordered a few action lists, and because of that, it's very difficult for someone to tell what changed when looking at the diff. For very large changes that include a refactor that would be fine, but for changes like this please only include the functional changes to the apl.

ArcanePorom:
Yeah, I don't blame you. I'll make sure to double check the commit before
putting in a PR next time, and get to resubmitting this one whenever I can
get back to it. Thanks!

PR #7944, opened 2023-07-17
[Arcane] Low Target AOE Updates and Touch optimization for T30

ArcanePorom:


Sergrand1:
Appreciate the work here. This is good to be merged.

PR #7968, opened 2023-07-25
Update Live Spell Data

JeremyMonk:
Live data July 25th (wow build 10.1.5.50585)


Pewtro:
This is adding files it shouldn't be. 

You should also probably look to change branches from this mistweaver branch that is being merged in repeatedly for a cleaner git history.

It also looks to be a duplicate of existing PRs https://github.com/simulationcraft/simc/pull/7958

PR #8002, opened 2023-08-05
[Mage] Fire - Apl support for disabling Bloodlust

Degenhours:
Current APL will ignore override.bloodlust=0 and lust at the end of the fight. 
Bloodlust enabled: https://www.raidbots.com/simbot/report/cMco6AfU31MdH2bDJ8x8DN
Bloodlust disabled: https://www.raidbots.com/simbot/report/1JnZbUbiGqe3cCHCWorcur


vituscze:
What's the firestarter bit doing there? That doesn't seem related.

PR #8008, opened 2023-08-09
Fix qmake macos QT 6.5 build

scamille:


navv1234:
Nightlies seem to compile with Qt 6.5, so I'll close this as completed.

PR #8012, opened 2023-08-12
[Arcane] Conjure mana gem and Call to Chaos 

ArcanePorom:
Also Nelth trinket optimization


Sergrand1:
Good to merge. Thanks for contributing.

PR #8034, opened 2023-09-06
[Warlock] fix shard spending with PS

Facenomore23:
Fixing an oversight where shards were spend outside of PS without reason.

5min : https://www.raidbots.com/simbot/report/6vctAAq7gUVmi5SfYQhU5n

5:10min, 0 variance https://www.raidbots.com/simbot/report/rYiydFN1c1DMT6an8m7MLM
(0 var comparison as we might lose a cast of SDG if the fight is shorter than 4:40)

### DIFF
--- a/engine/class_modules/apl/warlock.cpp
+++ b/engine/class_modules/apl/warlock.cpp
@@ -78,7 +78,7 @@ void affliction( player_t* p )
   
   default_->add_action( "malefic_rapture,if=talent.tormented_crescendo&buff.tormented_crescendo.react&!debuff.dread_touch.react" );
   default_->add_action( "malefic_rapture,if=talent.tormented_crescendo&buff.tormented_crescendo.stack=2" );
-  default_->add_action( "actions+=/malefic_rapture,if=variable.cd_dots_up|variable.vt_ps_up&soul_shard>1" );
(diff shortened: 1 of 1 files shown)


engine/class_modules/apl/warlock.cpp
```diff
@@ -78,7 +78,7 @@ void affliction( player_t* p )
   
   default_->add_action( "malefic_rapture,if=talent.tormented_crescendo&buff.tormented_crescendo.react&!debuff.dread_touch.react" );
   default_->add_action( "malefic_rapture,if=talent.tormented_crescendo&buff.tormented_crescendo.stack=2" );
-  default_->add_action( "malefic_rapture,if=variable.cd_dots_up|variable.vt_up&soul_shard>1" );
+  default_->add_action( "actions+=/malefic_rapture,if=variable.cd_dots_up|variable.vt_ps_up&soul_shard>1" );
```
seanpeters86: this syntax is invalid, you need something like this

```suggestion
  default_->add_action( "malefic_rapture,if=variable.cd_dots_up|variable.vt_ps_up&soul_shard>1" );
```

Facenomore23:
Thanks @seanpeters86, I went too fast on the copypaste

seanpeters86 (approved):
lgtm

PR #8047, opened 2023-09-09
Add wowhead ptr-2 item data source

weswigham:
Since the `(Beta)` wowhead isn't in use this time around, and instead we have the `ptr-2` instance of wowhead for 10.2.


Bloodmallet:
Sounds good. I'm not familiar with that portion of the code, but adding the additional endpoint for precise ptr tooltips makes sense. We'll need to somehow set this to the correct one each time the ptr flip-flops around.

PR #8051, opened 2023-09-10
[Outlaw] Initial Outlaw T31 set effect & 10.2 PTR talent changes

weswigham:
Included Outlaw PTR work items:
- [x] T31 set
- [x] Roll the Bones
- [x] Restless Blades
- [x] Between the Eyes
- [x] Swift Slasher
- [x] Deft Maneuvers
- [x] Killing Spree
- [x] Ghostly Strike
- [x] Improved Adrenaline Rush
- [x] Underhanded Upper Hand
- [x] Crackshot
- [x] Sanity check that everything in the updated base class tree is still working as intended with PTR data (MFD?)


engine/class_modules/sc_rogue.cpp
```diff
@@ -6588,6 +6697,21 @@ struct adrenaline_rush_t : public buff_t
       int cp_gain = as<int>( rogue->spec.improved_adrenaline_rush_energize->effectN( 1 ).resource() );
       rogue->resource_gain( RESOURCE_COMBO_POINT, cp_gain, rogue->gains.improved_adrenaline_rush );
     }
+
+    // "Pause" blade flurry
+    if ( rogue->talent.outlaw.underhanded_upper_hand->ok() && rogue->buffs.blade_flurry->check() )
+    {
+      auto current_duration = rogue->buffs.blade_flurry->remains();
+      rogue->buffs.blade_flurry->extend_duration( rogue, (current_duration * 100.0) - current_duration );
```
weswigham: Is this a janky way to "pause" a buff? Absolutely. But it basically matches what's done by the game, which sets a `modRate` of 100 to make them count down slower, while still displaying the same number on these buffs while they're "paused". The "paused" buffs can, in fact, still expire - just takes awhile.

EvanMichaels:
Discussed a bit with Solo and Zac in Ravenholdt. I think most of this is good, although I have some concerns about some of it. (Killing Spree, as an example, I think could be done in a bit of a better way.) But this is a pretty large PR to review all at once so I may not be able to review and comment on everything immediately. May be easier to reach out on Discord about this.

PR #8083, opened 2023-09-13
[Feral] Implement Saber Jaws

jordanritzie19:


jordanritzie19:
This is my first implementation in a class module, not 100% sure if done properly.

engine/class_modules/sc_druid.cpp
```diff
@@ -3744,7 +3745,7 @@ struct ferocious_bite_t : public cat_finisher_t
   double composite_da_multiplier( const action_state_t* s ) const override
   {
     auto dam = cat_finisher_t::composite_da_multiplier( s );
-    auto energy_mul = is_free() ? 2.0 : 1.0 + excess_energy / max_excess_energy;
+    auto energy_mul = is_free() ? 2.0 : 1.0 + (excess_energy / max_excess_energy * (1+p()->talent.saber_jaws->effectN( 1 ).percent()));
```
Saeldur: Implementation seems correct but the formatting of this line should be fixed.

PR #8090, opened 2023-09-15
[Feral] T31 Class Set Implementation part 3

jordanritzie19:
found an error calling for trigger, should thereotically be fixed now


Saeldur:
From a brief check `parse_buff_effects` does not currently handle Modify Pet Damage Done% or Modify Guardian Damage Done% so these will have to be done manually via overriding `composite_player_pet_damage_multiplier`.

PR #8092, opened 2023-09-16
Support setting buff time modifiers

weswigham:
This  is extracted from #8051, this is support for setting time modifiers on buffs, as used be things like the buff freeze on the jailer's stun in shadowlands, or, more relevantly, the buff pause from the Underhanded Upper Hand talent for outlaw rogue on the 10.2 PTR.


engine/buff/buff.cpp
```diff
@@ -627,6 +628,7 @@ buff_t::buff_t( sim_t* sim, player_t* target, player_t* source, util::string_vie
     buff_duration_multiplier( 1.0 ),
     default_chance( 1.0 ),
     manual_chance( -1.0 ),
+    time_modifier( 0.0 ),
```
vituscze: Tracking the modifier rather than multiplier is very unusual as far as simc goes. Presumably, the time mod stuff stacks multiplicatively (though that might be hard to test right now). But even if Blizz never stacks these, it might still be useful in simc.

engine/buff/buff.cpp
```diff
@@ -829,6 +831,76 @@ buff_t* buff_t::set_duration_multiplier( double multiplier )
   return this;
 }
 
+buff_t* buff_t::set_time_modifier( double modifier )
+{
+  if ( s_data->flags( spell_attribute::SX_IGNORE_FOR_MOD_TIME_RATE ) )
```
vituscze: This check seems like something that should be done when parsing the spell data rather than when calling this function. For highly specific effects like this, I'd expect that explicitly calling this function does what I tell it regardless of spell data.

engine/buff/buff.cpp
```diff
@@ -829,6 +831,76 @@ buff_t* buff_t::set_duration_multiplier( double multiplier )
...
+
+      if ( player )
+        sim->print_log( "{} increases {} expiry by a factor of {}. New expiration: {}", *player, *this, ratio,
+                      exp->occurs() );
+    }
+  }
+  // Speeding up the clock, expiry moved closer
+  else
+  {
+    auto old_expirations = new std::vector<event_t*>();
```
vituscze: This leaks memory. Is this vector even necessary? Seems like you could just change the elements of the vector rather than create a new one.

engine/buff/buff.cpp
```diff
@@ -829,6 +831,76 @@ buff_t* buff_t::set_duration_multiplier( double multiplier )
...
+                      exp->occurs() );
+    }
+    old_expirations->erase( old_expirations->begin(), old_expirations->end() );
+
+  }
+
+  return this;
+}
+
+buff_t* buff_t::pause()
```
vituscze: I'm not convinced these functions should be part of the public interface of buff_t. Especially when pause() doesn't really pause anything, just slows it down by A TON.

engine/buff/buff.cpp
```diff
@@ -2068,6 +2148,8 @@ void buff_t::extend_duration( player_t* p, timespan_t extra_seconds )
 
   assert( expiration.size() == 1 );
 
+  extra_seconds = extra_seconds * get_time_modifier_multiplier();
```
vituscze: Didn't have time to check if the time mult is used in all the places it should be, so this comment is just a note for my future self to do so.

vituscze (changes requested):

engine/buff/buff.cpp
```diff
@@ -1516,6 +1588,10 @@ buff_t* buff_t::apply_affecting_effect( const spelleffect_data_t& effect )
         apply_percent_modifier( effect );
         break;
 
+      case A_MOD_TIME_RATE:
+        set_time_modifier( effect.base_value() );
+        break;
+
```
Dorovon: This is not the right place to automatically parse this aura type. A_MOD_TIME_RATE is global while `spell_data_t::affected_by( const spelleffect_data_t* effect )` is for checking if the spell's class flags match the ones in the effect.

If supporting this aura type is needed, it would make more sense to add a virtual function like `player_t::composite_time_rate_multiplier` and something that can be called to adjust buffs when changing it similar to how `action_t::recharge_rate_multiplier` is used with `cooldown_t::adjust_recharge_multiplier` to adjust cooldowns.

For now, it looks like the Rogue changes only need A_MOD_TIME_RATE_BY_SPELL_LABEL (and it only needs to apply to the buff duration), so it probably makes more sense to leave this out for now until it is needed since fully supporting the aura type would also require adjusting many other parts of the actor (attack speed, movement speed, cooldown recharge rate, etc.) and Blizzard has recently avoided using A_MOD_TIME_RATE in relevant content outside of occasionally pausing an encounter briefly with a -99% aura.

engine/buff/buff.cpp
Dorovon: If you want to keep the attribute check, The normal way to do that would be to add a `bool ignore_mod_time_rate` to `buff_t` and set that based on the presence of the attribute; that way developers can easily override it for specific buffs as needed and you would check for that flag here instead of checking the spell data.

engine/buff/buff.cpp
```diff
@@ -829,6 +831,76 @@ buff_t* buff_t::set_duration_multiplier( double multiplier )
   return this;
 }
 
+buff_t* buff_t::set_time_modifier( double modifier )
```
Dorovon: While it's not needed for this Rogue case, but you should add a comment here or somewhere else to let people know that this does not apply the modifications to buff tick times and existing tick events that would appear in game with modify time rate effects.

Dorovon (changes requested):

engine/buff/buff.cpp
vituscze: I think this is a fairly important point that I didn't catch. If this code was in the rogue module I don't think the lack of tick time modification is really an issue, but in simc core it should work correctly with everything buff_t supports.

engine/buff/buff.cpp
weswigham: It's definitely not - it's only used in the bare minimum number of places for it to affect real-time buff durations. Modifying tickrate and showing the apparent duration (as modified by the time rate) are... painful. The knock-on implications of accounting for time modifier rate are kinda profound in downstream code, especially with the difference between apparent duration and real duration. For example, for pandemic refreshing a buff - that relies on the apparent duration, as modified by the time modifier, not the real-time duration. But the real-time duration is what's useful for things like scheduling simulated events and APL logic. Thankfully apparent duration is not currently needed to sim anything, since it's only applied to things that don't tick (anymore).

engine/buff/buff.cpp
weswigham: Correct! But in the verbage of talents like underhanded upper hand, the text says the buff duration "does not decrease", when in reality what's happening is that this style of time modifier is being applied so long as the other buff is active. It seems likely this is how buff pausing effects will be implemented in the future (via this fake pause via 100x slowdown), since UI addons and the like handle it better than a true pause, thanks to already using the time modifier for _accelerated_ cooldown scenarios presented by some bosses since legion, so a function for that behavior seems useful. At least it was useful for me, since I had to call it a fair bit because the aura-applying infrastructure for buffs doesn't seem to handle _un_-applying auras too well, so seems unsuitable for auras applying temporary changes to other auras (like what the serverside scripts for underhanded upper hand seem to do with its' background spells).

Basically I just found it useful and think it'll probably be useful in the future, too. If not in the base buff type, I'd probably add it into the rogue buff base, because it seems cleaner, intent-wise, than calling `set_time_modifier` directly a bunch, and rogue needs it on at least 3 buffs.

engine/buff/buff.cpp
weswigham: I was looking into doing tick time modification, but had two (three) issues:

0. I am horribly unfamiliar with all this, and absolutely need direction to take this much farther.

1. It's basically impossible to test weather the behavior matches game behavior, since the modifier isn't applied to anything that ticks right now (slice and dice's healing was moved into a separate buff to head off this issue (see spell 426605 which now holds the heal separately from the main buff and has no duration of its own, plus a fixed 2s period), adrenaline rush just modifies stats, blade flurry just procs on damage).

2. If we assume the time modifier speeds up/slows down ticks like it does the buff's expiry, we need to calculate tick times using the "apparent" buff duration and not the real-time buff duration (and then scale those into the real-time domain using the time modifier), and dealing with real-time vs apparent buff durations _at all_ seems like a massive hassle for something that doesn't seem to matter just yet. Minimally, I have no idea what other parts of the sim should be looking at real-time vs apparent-time durations. 
    * Not dealing with it technically means slice and dice's pandemic refresh when paused is bugged in the PTR rogue sim I've been working on (since it should probably pandemic at 1200 real-seconds left while paused and not 12s - this is the in-game behavior, the pandemic duration is definitely based on the apparent duration), but that'll never happen in any real sim, since the sim'll never intentionally cast slice with anything close that much real-time left (why would you, massive waste of resources).

weswigham: > It's basically impossible to test weather the behavior matches game behavior, since the modifier isn't applied to anything that ticks right now

I've since learned the old https://www.wowhead.com/spell=320919/echoes-of-elisande anima power actually uses this modifier (as a positive number). I vaguely remember playing a mage in torghast with it, so I do think it multiplies the tickrate of all hots and dots. And cooldowns. And other stuff, for the full applies-to-everything effect. And, iirc, weird stuff like environmental damage ticks, too.

I still don't have a good grasp of everything that'd need to change to properly support it, though, especially what needs the real-time vs apparent-time distinction.

engine/buff/buff.cpp
weswigham: I can do stacking, but, if how [echos of elisande](https://www.wowhead.com/spell=320919/echoes-of-elisande) stacked is any indication, it'll be additive and not multiplicative (though that was multiple stacks of an aura effecting an effect value, rather than multiple effect values combining). But I'm not sure we can know which until we have a live example. Though I'm thinking it's probably be additive, since I'm *pretty sure* time warp (with a base value of 0), didn't break people out of jailer stun (value of -99), but I can't say that for sure.

vituscze: Echoes of Elisande doesn't really tell us anything. 2 stacks of a +10% damage buff give you +20% damage, even though damage buffs stack multiplicatively. You'd need to find two different effects to combine (not one effect with multiple stacks) to tell them apart. I'm not sure how being additive/multiplicative has anything to do with Time Warp breaking people out of Jailer stun. If it's additive then you have `0 + (-99) = -99` mod, if it's multiplicative you have `(1 + 0) * (1 + (-99)/100) = 0.01` multiplier; both of which don't indicate you should be broken out. I suppose you could go to Gul'dan, use the extra action button and check the time mod by using one of the buff API functions.

engine/buff/buff.cpp
vituscze: I don't mind the idea in principle, it's just that right now, the only example we have is the new rogue stuff. If you make a shortcut in the rogue module that's perfectly fine. If they reuse it somewhere else, that's when I think we should be considering making it part of the buff_t interface.

I guess the bigger issue for me is that this function conveys the wrong idea to devs not familiar with the rogue rework, especially since we sometimes use buff_t for stuff that doesn't have a corresponding buff in game, where you might want to actually pause a buff rather than slow it down by a lot. If this way of "pausing" buffs becomes established as a normal thing they use, then I don't think there would be any risk of confusion.

engine/buff/buff.cpp
weswigham: So the magistrix's [time stop](https://www.wowhead.com/spell=208944/time-stop) allowed me to witness a modRate of 10000 applied to my paused buffs (two 100x slows multiplied together). While [time dilation](https://www.wowhead.com/spell=210339/time-dilation) saw my paused buff rates become 76.923080444336, which is equivalent to a 100x slowdown and a 30% speedup being multiplied together, so multiplicative stacking it is. Good to know that while it seems like individual effects seem to be limited to 100x slower, though, the combined modifier can still exceed that.

engine/buff/buff.cpp
```diff
@@ -620,13 +620,16 @@ buff_t::buff_t( sim_t* sim, player_t* target, player_t* source, util::string_vie
...
+    ignore_mod_time_modifier( false ),
     reverse_stack_reduction( 1 ),
     current_value(),
     current_stack(),
     base_buff_duration( timespan_t::min() ),
     buff_duration_multiplier( 1.0 ),
     default_chance( 1.0 ),
     manual_chance( -1.0 ),
+    base_time_modifier_duration_multiplier( 1.0 ),
+    dynamic_time_modifier_duration_multiplier( 1.0 ),
```
vituscze: Using both "modifier" and "multiplier" in the same name is highly unusual. We don't call damage multipliers "damage modifier multiplier." I do wonder if there is a good use case for the base multiplier, these sorts of effects are always dynamic.

engine/buff/buff.cpp
```diff
@@ -620,13 +620,16 @@ buff_t::buff_t( sim_t* sim, player_t* target, player_t* source, util::string_vie
     is_fallback(),
     requires_invalidation(),
     expire_at_max_stack(),
+    ignore_mod_time_modifier( false ),
```
vituscze: This uses "modifier" twice. Same as above, I'd call it something like `ignore_time_multiplier`.

engine/buff/buff.cpp
```diff
@@ -829,6 +835,80 @@ buff_t* buff_t::set_duration_multiplier( double multiplier )
...
+  if ( ignore_mod_time_modifier )
+    return this;
+  modifier = std::max( modifier, -99.0 ); // Limit slow down to 100x slower
+
+  auto mul = 100.0 / ( 100 + modifier );
+
+  base_time_modifier_duration_multiplier = base_time_modifier_duration_multiplier * mul;
+
+  return this;
+}
```
vituscze: This is a fairly large departure from how multipliers are handled in simc. If the underlying thing you're changing is a multiplier, it should be accessed as a multiplier, not a modifier. I realize that you can't just use `1 + time_mod.percent()` (since you need a reciprocal of that), but this is more of a question of whether to store `mult` or `1 / mult` (see haste%).

vituscze (changes requested):

engine/buff/buff.cpp
weswigham: > I do wonder if there is a good use case for the base multiplier

Mmmm, Time warp has an effect that would constantly set it to `1` on its' buff (it's modified by an torghast anima power). It's one of only two non-boss abilities that have time modifier stuff on them, so... 🤷 @EvanMichaels  said a split like this was normal and expected, and is why things like `recharge_multiplier` are split into base and dynamic parts in `cooldown_t`, so the base part can be set by spell data loading, while the dynamic part can be dynamically modified by transient effects.

engine/buff/buff.cpp
weswigham: > This is a fairly large departure from how multipliers are handled in simc

> Tracking the modifier rather than multiplier is very unusual as far as simc goes.

Surely either the modifier from the effect or the multiplier used in the calculations as a result is "correct" to track, and not some derived 3rd thing.

This absolutely is not a datum like other multipliers used in simc, from what I can tell, because the `%`-equivalent of the modifier isn't a straight multiplier on durations, what with how values less than zero are handled, especially with the apparent hard cap at -99% per-effect. If you'd like, I could add a

```c++
  double time_multiplier() const
  { return ( 100.0 / ( 100.0 + std::max( _base_value, -99.0 ) ) ); }
```

to the `spell_data` type itself, and push the multiplier calculation there, if doing it on-use like this seems somehow lacking...

vituscze: > This absolutely is not a datum like other multipliers used in simc, from what I can tell, because the %-equivalent of the modifier isn't a straight multiplier on durations

Haste and cooldown recharge effects behave the same way. If you have a +50% cooldown recharge rate, you need to transform it into 1/1.5 to use as a multiplier. Same with haste.

> especially with the apparent hard cap at -99% per-effect

Same cap exists for damage modifiers (if you have -100% damage taken, the game caps it at -99%). The main point here is that I'm not convinced we need to depart from the standard multiplier "interface." Why not just do this at the point of use and only make sure the user doesn't try to make the multiplier 0?

engine/buff/buff.cpp
vituscze: I suppose there's no harm in having a base mult.

engine/buff/buff.cpp
weswigham: > Why not just do this at the point of use and only make sure the user doesn't try to make the multiplier 0?

While individual effects seem to be capped at -99% (based on how the UHUH buffs (which have a -100% effect value but only apply a 100x slowdown, not an infinite one) actually scale time and are reported by other APIs), the aggregate multiplier is able to be much higher - a UHUH-paused buff in an elisande boss slow field is observed at a 10000x slowdown. So yeah, the multiplier obviously has to be positive and nonzero, but the modifier itself also has to be capped when you create the multiplier from it to emulate how stacking appears to work.

weswigham: > Haste and cooldown recharge effects behave the same way.

Ignoring the seemingly unique per-effect capping behavior:

* `haste` is stored as the raw multiplier, and every application of an effect does the `haste *= 1.0 / (1.0 + effect.percent())` transform. Which seems like a footgun, but alright (maybe we should build that transform into the spell data type, so you can just do `effect.multiplier()`!).
 
* Cooldown recharge, meanwhile, _is_ stored as a reciprocal multiplier. So every effect application does `base_recharge_multiplier *= 1 + effect.percent()`. Except for `architects_ingenuity_t` which stores the multiplier instead.

The two seem to have picked different options here - looking to them for guidance, both the multiplier _and_ the reciprocal multiplier seem fine. But there's definitely a preference for the multiplier and not the reciprocal - `haste` is used far more than cooldown recharge, and cooldown recharge isn't fully consistent anyway.

The only difference currently here from how `haste` behaves would seem to be that the `apply_modifier` spell data loading function does the transform from modifier to multiplier (whichever it is), rather than the use-site? Which isn't actually uncommon, since the local function inside spell data loading, `apply_percent_effect_modifier` also does that? (And tbh, the apply function I have here *could* be a local function inside spell data loading - maybe it should be.)

vituscze (approved):

PR #8099, opened 2023-09-21
[Mage] Removed ignite bug

rootkit1337tv:
Removed the part of the mage module that was factoring out the ignite amp from bonus crit damage. Blizzard has fixed this and  will release it with the next PTR build


vituscze:
I will remove it when the bugfix is actually available and can be tested. Until then, sim with `bugs=0`.

vituscze:
Bug wasn't entirely fixed, so I'm gonna close this one.

PR #8137, opened 2023-10-01
[Havoc] Initial Dash of Chaos Implementation

Remnxnt:
First attempt at dipping my toes in please let me know if you see anything out of line, thanks.


vituscze:
Have you checked that this actually works? A lot of the conditions seem nonfunctional (you're checking whether talent's spell data pointer isn't `nullptr` rather than checking if the talent is enabled, for example).

Remnxnt:
It does appear to be functional. Though there are some things that I couldn't quite figure out, such as the dash_of_chaos action is executing twice for every hit for example. This doesn't have an effect on the DPS contribution of the ability but does impact the DPE statistics. As for talent conditions you'd have to point out which one's specifically would be non-functional.

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -5198,6 +5214,13 @@ struct fel_rush_t : public demon_hunter_attack_t
     // Fel Rush and VR shared a 1 second GCD when one or the other is triggered
     if ( p()->cooldown.movement_shared->down() )
       return false;
+   
+    // Not usable while Dash of Chaos buff is present and is locked out for 1 second upon buff expiration
+    if ( p()->talent.havoc.dash_of_chaos && p()->buff.dash_of_chaos && p()->buff.dash_of_chaos->check())
```
ToppleTheNun: `p()->talent.havoc.dash_of_chaos` needs to be `p()->talent.havoc.dash_of_chaos->ok()`

engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -5164,6 +5169,17 @@ struct fel_rush_t : public demon_hunter_attack_t
 
     p()->buff.unbound_chaos->expire();
 
+    if (p()->talent.havoc.dash_of_chaos->ok())
+    {
+      p()->buff.dash_of_chaos->trigger();
+    }
+
+    // Dash of chaos incurs 1 second lockout on Felrush
+    if ( p()->talent.havoc.dash_of_chaos && !p()->buff.dash_of_chaos->up() )
```
ToppleTheNun: `p()->talent.havoc.dash_of_chaos` needs to be `p()->talent.havoc.dash_of_chaos->ok()`

ToppleTheNun (commented):
These are what were meant by checking if the talent is selected rather than `nullptr`.

Remnxnt:
I did not intend to perform the second commit

engine/class_modules/sc_demon_hunter.cpp
Remnxnt: this line has been removed in most recent commit

engine/class_modules/sc_demon_hunter.cpp
Remnxnt: This has been fixed in most recent commit

Remnxnt:
TODO: 

Add a method to cast dash_of_chaos while debuff_out_of_range is present, I suspect this is causing dash of chaos to be executed slower (0.7-0.9s) than is possible in game (0.5s)

Resolve issue of execute statistics (exactly double the amount of real casts)

PR #8150, opened 2023-10-03
[Havoc] Soulrend -> Soulscar Rename

Remnxnt:
Renames Soulrend to Soulscar


Remnxnt:
No longer needed, changes incorporated in commit 0d055ff

PR #8151, opened 2023-10-03
[Havoc] Add talent "Champion of the Glaive"

Remnxnt:


Remnxnt:
No longer needed, changes incorporated in commit f0a282f

PR #8155, opened 2023-10-04
[Hunter] change BM 4pc to better model currently bugged behavior

JontiH:
this should model behavior described in https://github.com/SimCMinMax/WoW-BugTracker/issues/1135

I'm not 100% sure if active_pets() returns the earliest pet or latest pet first, so might need to tweak this

### DIFF
--- a/engine/class_modules/sc_hunter.cpp
+++ b/engine/class_modules/sc_hunter.cpp
@@ -3821,12 +3821,15 @@ struct multishot_bm_t: public hunter_ranged_attack_t
 
       if ( p() -> tier_set.t31_bm_4pc.ok() ) {
 
-        if ( p() -> bugs ){
-          if (!( p() -> pets.dire_beast.active_pets().empty)){
-            p() -> pets.dire_beast.active_pets().last() -> buffs.beast_cleave -> trigger();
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_hunter.cpp
```diff
@@ -5531,9 +5542,19 @@ struct kill_command_t: public hunter_spell_t
       pet -> active.kill_command -> execute_on_target( target );
 
     if ( p() -> tier_set.t31_bm_4pc.ok() )
```
Pewtro, line 5536: you should add brackets on this if statement so it's less ambiguous what happens there and end it on li 5557. Your compiler locally should also complain about it

jharris-tc: good spot, I missed that

JontiH: good spot, I missed that.

jharris-tc:
seems like it works, dire beast kc damage goes down by about 56%

JontiH:
seems like it works, dire beast kc damage goes down by about 56%

JontiH:
can someone merge this?

PR #8158, opened 2023-10-06
[Warlock] Destruction 2p can crit

Facenomore23:
)-:<


Azevara:
Closing as today's PTR notes indicate this should not occur.

PR #8168, opened 2023-10-08
Use matching lld version and adjust tcb span tuple struct definitions

scamille:
The hope is that building with clang-7 gets faster.


scamille:
Hmm after doing some research, I think the main cause of `ccache` misses is that `-ffile-prefix-map` is not available < clang10, leading to agent-dependent generated code, resulting in cache misses.

There isn't really much that can be done to prevent that.

PR #8169, opened 2023-10-08
[Paladin] Add Ret T31

Fluttershy-git:


engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -1467,13 +1468,26 @@ struct divine_toll_t : public paladin_spell_t
 
     if ( result_is_hit( s->result ) )
     {
+      // T31 only procs on the first valid target, others are not affected, even if valid
+      if ( !t31HasProcced && p()->sets->has_set_bonus( PALADIN_RETRIBUTION, T31, B2 ) &&
```
mserrano, line 1472: You should be able to check whether the current impact is for the primary target via looking at `s->chain_target`

engine/class_modules/paladin/sc_paladin_retribution.cpp
```diff
@@ -324,7 +324,14 @@ struct blade_of_justice_t : public paladin_melee_attack_t
 
       if ( p->talents.jurisdiction->ok() )
       {
-        base_multiplier *= 1.0 + p->talents.jurisdiction->effectN( 4 ).percent();
+        // Jurisdiction doesn't increase Expurgation's damage in-game
+        // base_multiplier *= 1.0 + p->talents.jurisdiction->effectN( 4 ).percent();
```
mserrano: Interesting; it's supposed to according to the spell data. Can you gate this behind a `bugs` check?

engine/class_modules/paladin/sc_paladin_retribution.cpp
```diff
@@ -390,14 +397,27 @@ struct blade_of_justice_t : public paladin_melee_attack_t
 
     if ( p()->talents.expurgation->ok() )
     {
-      expurgation->set_target( state->target );
-      expurgation->execute();
+      p()->active.expurgation->target = state->target;
```
mserrano, line 391: If we're no longer going to use the direct child expurgation instance, can we not initialize it for each blade_of_justice_t initialization?

engine/class_modules/paladin/sc_paladin.cpp
mserrano: Ah - but you can't quite do that here because it'll proc on the "first" expurgation target even if it's not the primary target? Or am I misunderstanding?

Fluttershy-git: Correct. Divine Toll shoots up to 5 Judgments and if any of those Judgments hits a target with an Expurgation ticking, it procs the Tier. Even if your current target (chain_target 0) doesn't have Expurgation ticking

engine/class_modules/paladin/sc_paladin_retribution.cpp
Fluttershy-git: Spell data increases direct damage, but Expurgation has no direct damage, only ticking damage. Jurisdiction increases Spell Direct Amount instead of Spell Periodic Amount (See Avenging Wrath effect 2 (and 11))
Sure, can do that

engine/class_modules/paladin/sc_paladin_retribution.cpp
Fluttershy-git: Fixed in https://github.com/simulationcraft/simc/pull/8169/commits/8f40d6bcddea9e7cb62dfe72e2c43c29b4d07553

engine/class_modules/paladin/sc_paladin_retribution.cpp
Fluttershy-git: Bugs-Flag added in https://github.com/simulationcraft/simc/pull/8169/commits/5bf5ff3a2ab6555cc25622a5b07ad4805379ea3e

engine/class_modules/paladin/sc_paladin.cpp
Fluttershy-git: To clarify further:
Judgment only procs Tier on the main target, even if Blessed Champion is talented and a cleave hits another target with Expurgation
Divine Resonance is exactly the same as a normal Judgment
Divine Toll's initial cast will proc the Tier on the first target it hits that has an Expurgation debuff ticking (including travel time etc. But I don't think it's really needed to simulate this)

mserrano (approved):

PR #8193, opened 2023-10-13
[Gear] Thorncaller Claw initial implementation

hotbees:


nyterage:
LGTM. 

For the target swap on death mechanic, depending on how its setup, you should be able to register an on demise callback to trigger the effect on a new target. not sure how exactly it works in game atm, but, if it applies to a new target with the same duration it had before death, you might need to adjust the damage action to be executed on debuff tick, rather than its own action. That would simplify the setup so you could just trigger on a new target in the target_non_sleeping list with the proper duration. 

if you need a reference for the on_demise_callback, could check out Coiled Serpent Idol, at L6071.

engine/player/unique_gear_dragonflight.cpp
```diff
@@ -7440,6 +7440,91 @@ void dreambinder_loom_of_the_great_cycle( special_effect_t& effect )
...
+      } );
+
+  auto thorn_spirit = create_proc_action<thorn_spirit_t>( "thorn_spirit", effect );
+  thorn_spirit->add_child( thorn_burst_damage );
+
+  // 2023-10-14: When Thorn Spirit spreads to a new target on demise it is applied with full duration.
+  // Can spread to a target with an existing Thorn Spirit even if other targets are in range, refreshes normally.
+  range::for_each( effect.player->sim->actor_list, [ effect, thorn_spirit ]( player_t* target ) {
+    target->register_on_demise_callback( effect.player, [ effect, thorn_spirit ]( player_t* t ) {
+      thorn_spirit->target_cache.is_valid = false;
```
nyterage: should probably check that the debuff exists on the current target before running any of this just for safety 

something like `effect.player->get_target_data( t )->debuff.thorn_spirit->check()` should do the trick

engine/player/unique_gear_dragonflight.cpp
```diff
@@ -7440,6 +7440,91 @@ void dreambinder_loom_of_the_great_cycle( special_effect_t& effect )
...
+      std::vector<player_t*> targets      = thorn_spirit->target_list();
+      if ( targets.size() != 0 )
+      {
+        // Choose a random new target to spread to
+        player_t* new_target =
+            targets[ static_cast<int>( effect.player->rng().range( 0, static_cast<double>( targets.size() ) ) ) ];
```
nyterage: for simc specifically, we should use `debug_cast` in these situations, rather than a `static_cast`, while its converted to a `static_cast` during compilation with the release config, helps for debugging if something goes wrong.

nyterage (changes requested):

nyterage (approved):

PR #8224, opened 2023-10-19
[Gear] Witherbark's Branch initial implementation

JontiH:


JontiH:
I based this on macabre sheet music. Ideally I could add a parameter to change duration of the time spent collecting the ball, but to start its just static

engine/player/unique_gear.cpp
```diff
@@ -1828,6 +1830,24 @@ void item::infallible_tracking_charm( special_effect_t& effect )
   new dbc_proc_callback_t( effect.item, effect );
 }
 
+void item::witherbarks_branch( special_effect_t& effect )
+{
+  auto data_spell = effect.player->find_spell( 429257 );
+  auto stat_buff  = make_buff<stat_buff_t>( effect.player, "aqueous_dowsing", effect.player->find_spell( 429257 ) )
+    ->add_stat( STAT_MASTERY_RATING, data_spell->effectN( 1 ).average( effect.item ) );
```
nyterage: Looks like you've got the wrong spell id here for the buff, should be `429262`

Also should be named `aqueous_enrichment`

with that fixed, you dont need the `add_stat` function all all, since the value, the stat it gives, etc, will be parsed from the spell data.

doing this, will automatically parse most of the buff effects, such as the 10s duration, the 3 stack cap, and what stat it gives (in case it ever changes in the future). 

final note, dont need to store the data spell, since its just the driver, can call that spell with `effect.driver()` wherever you need to call one of its effects.

nyterage (changes requested):

engine/player/unique_gear.cpp
```diff
@@ -1828,6 +1830,24 @@ void item::infallible_tracking_charm( special_effect_t& effect )
...
+
+  effect.custom_buff = make_buff( effect.player, "aqueous_dowsing_driver", effect.driver() )
+    ->set_quiet( true )
+    // assumes an average of 1 seconds per ball
+    // TODO: make this confirgurable?
+    ->set_cooldown( timespan_t::from_seconds(1) )
+    ->set_tick_callback( [ stat_buff ]( buff_t*, int, timespan_t ) {
+      stat_buff->trigger();
+    } );
```
nyterage, line 1843: for this effect, if youre attempting to make it trigger a stack every 1 second, should use `set_period( 1_s )`, rather than `set_cooldown`. right now, theres no periodic effect on the driver, so, the tick callback is likely never called.

nyterage (changes requested):

JontiH:
nevermind, it got implemented!

Issue #8256, opened 2023-10-25
[Warrior] Colossus Smash, Executioner's Precision, and Skullsplitter debuffs not properly applying in multi-target


atarr3:
### Describe the bug
Current sim behavior is not correctly applying certain debuffs from single target abilities in multi-target scenarios when Sweeping Strikes is up. This issue is evident from the reports, which show no applications of Skullsplitter or Executioner's Precision debuffs on secondary targets and an incorrect number of Colossus Smash debuff applications (the number of applications equals the number of executes). Here's an example for a 5 minute two target sim with `ptr=1`:

#### Primary Target
![image](https://github.com/simulationcraft/simc/assets/7365692/c33bd6f6-7f38-4b98-b931-2ab1d2ac5399)

#### Secondary Target
![image](https://github.com/simulationcraft/simc/assets/7365692/758c6a6b-f53a-402f-806a-c6454985ddbb)

Additionally, there's ~10k DPS loss in damage dealt to the primary target compared to a single target sim for the same character, yet the only rotational difference is that half a GCD is used every 30s to cast Sweeping Strikes. I believe the DPS loss is due to Colossus Smash not being applied to the primary target with every cast.

The picture above also shows that Skullsplitter is somehow refreshed 4 times on the primary target, even though this debuff can only be applied once every 21 seconds with a 10 second duration. The same issue is observed in debug logs, which offers more information:
![image](https://github.com/simulationcraft/simc/assets/7365692/48097e81-c6ac-4120-824d-ca3bd253e3ec)

### Expected behavior
When Sweeping Strikes is up and one of the corresponding abilities is cast, the debuff should be applied to the primary target and one other secondary target. Since Sweeping Strikes has 50% uptime, the debuff uptime on the secondary target should be around 50% of the uptime on the primary target.

### To Reproduce
Run a two target sim for warrior with the appropriate talents selected (Colossus Smash, Skullsplitter, Executioner's Precision) and `ptr=1`. [Here](https://pastebin.com/urQtsRTK) is the simc profile used to create the pictures above. Note that the issue is not exclusive to PTR sims, this flag is just used to show the Skullsplitter debuff behavior. The issues with Colossus Smash and Executioner's Precision are present in Live sims as well.

### Additional information
* SimulationCraft 1015-01 (wow build level 51851)

* [Debug log](https://github.com/simulationcraft/simc/files/13145300/2targets_BUGGED.txt)

PR #8257, opened 2023-10-25
[Warrior] Multitarget debuff application fixes for CS, EP, and SS

atarr3:
Current sim behavior is not correctly applying certain debuffs from single target abilities in multi-target scenarios when Sweeping Strikes is up, referenced in https://github.com/simulationcraft/simc/issues/8256.

The issue is that these debuffs are triggered in function overrides of `execute()` using `execute_state->target` to specify the target on which to apply the debuff. But this points to the primary target of the player rather than the target hit by the attack. I've moved the associated `trigger()` and `expire()` calls into function overrides for `impact()`, which seems to have fixed the issue. Note there are other debuffs I haven't touched that are almost certainly bugged as well (Exploiter, for example, but this is an outdated ability).

I also made some cosmetic changes to the new Skullsplitter debuff behavior, removing redundant calls to `is_ticking()` before calling `adjust()`, since `adjust()` already checks if the associated DoT is ticking.

### DIFF
--- a/engine/class_modules/sc_warrior.cpp
+++ b/engine/class_modules/sc_warrior.cpp
@@ -2476,7 +2476,7 @@ struct mortal_strike_unhinged_t : public warrior_attack_t
 
     warrior_td_t* td = this->td( execute_state->target );
     td->debuffs_exploiter->expire();
-    td->debuffs_executioners_precision->expire();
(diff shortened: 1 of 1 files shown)


Armadk:
All looks good.  You are right, there likely is a lot of these.  I know Archi had noticed slam the other day as well and fixed that one.

PR #8274, opened 2023-10-29
[gear] Fyrakk's Tainted Rageheart

renanthera:
Shadowflame Lash Target Scaling:
<https://www.warcraftlogs.com/reports/anMtvy9RwbjYTp3h#type=damage-done&view=events&ability=425701&boss=-3&difficulty=0>

Tainted Heart Target Scaling:
<https://www.warcraftlogs.com/reports/bhpgXYPaAMRwdzDH#fight=1&source=3&type=damage-done&ability=425461&view=events>

Spell data regen will be required after merge, but I opted to remove it for easier conflict resolution.


engine/player/unique_gear_dragonflight.cpp
```diff
@@ -7113,6 +7113,167 @@ void gift_of_ursine_vengeance( special_effect_t& effect )
...
+// 425461 Enemy Damage
+// 422750 On-Use Debuff
+// 425571 On-Use Absorb
+// 425703 Self Damage (On-Use)
+// 425701 Enemy Damage (On-Use)
+void fyrakks_tainted_rageheart( special_effect_t& effect )
+{
+  struct tainted_heart_t : public proc_spell_t
+  {
+    struct self_damage_t : public generic_aoe_proc_t
```
nyterage, line 7134: Should probably just be generic_proc_t here, so you dont have to edit the aoe settings later.

engine/player/unique_gear_dragonflight.cpp
```diff
@@ -7113,6 +7113,167 @@ void gift_of_ursine_vengeance( special_effect_t& effect )
...
+    {
+      self_damage->execute();
+      enemy_damage->execute();
+      proc_spell_t::execute();
+    }
+  };
+
+  struct shadowflame_rage_t : public proc_spell_t
+  {
+    struct self_damage_t : public generic_aoe_proc_t
```
nyterage, line 7184: should also be `generic_proc_t` so you dont have to override the aoe settings

engine/player/unique_gear_dragonflight.cpp
```diff
@@ -7113,6 +7113,167 @@ void gift_of_ursine_vengeance( special_effect_t& effect )
...
+      }
+    };
+
+    struct enemy_damage_t : public generic_aoe_proc_t
+    {
+      enemy_damage_t( const special_effect_t& e )
+        : generic_aoe_proc_t( e, "shadowflame_lash_enemy", e.player->find_spell( 425701 ) )
+      {
+        background          = true;
+        aoe                 = -1;
```
nyterage: generic_aoe_proc_t already sets aoe to -1, so this is redundant

engine/player/unique_gear_dragonflight.cpp
```diff
@@ -7113,6 +7113,167 @@ void gift_of_ursine_vengeance( special_effect_t& effect )
...
+        stats->type = stats_e::STATS_NEUTRAL;
+        callbacks   = false;  // TODO: confirm if this triggers any proc flags.
+        base_dd_min = base_dd_max = e.player->find_spell( 422652 )->effectN( 2 ).average( e.item );
+      }
+    };
+
+    struct enemy_damage_t : public generic_aoe_proc_t
+    {
+      enemy_damage_t( const special_effect_t& e )
+        : generic_aoe_proc_t( e, "shadowflame_lash_enemy", e.player->find_spell( 425701 ) )
```
nyterage:
```suggestion
        : generic_aoe_proc_t( e, "shadowflame_lash_enemy", e.player->find_spell( 425701 ), true )
```

setting the bool true here will handle the 15% increase per target up to 5 that trinkets often do, with all its behavior pre-set.

engine/player/unique_gear_dragonflight.cpp
```diff
@@ -9829,6 +9990,7 @@ void register_special_effects()
   register_special_effect( 422479, items::infernal_signet_brand );
   register_special_effect( 429228, items::rezans_gleaming_eye );
   register_special_effect( 421990, items::gift_of_ursine_vengeance );
+  register_special_effect( 422750, items::fyrakks_tainted_rageheart );
```
nyterage, line 9992: Should also disable the equip driver below around line 10083 to be sure simc never auto parses it and messes with the implementation if theres any spell data changes.

engine/player/unique_gear_dragonflight.cpp
renanthera: This is intended, as the damage is reduced by avoidance.

engine/player/unique_gear_dragonflight.cpp
renanthera: This is intended, as the damage is reduced by avoidance.

engine/player/unique_gear_dragonflight.cpp
renanthera: I'm not sure what you mean by equip driver. Is this line incorrect?

nyterage: so the trinket has 2 drivers, one is the equip effect, that does AoE damage constantly, id `422652`. 

usually when we custom write these types of effects, we disable that driver so that nothing funny happens with spell data changes. look at line 10083 for some examples.

PR #8318, opened 2023-11-06
[Shaman] Update to Elemental 10.2 st APL

MadBubble:
- Restricted Flame shock for refreshing if Pwave is coming
- Limited non SK SoP Lightning bolts to lightning builds
- Limited high prio IF&FrS to lightning builds
- Blocked LvBm usage on 2t for T31 4p
- Added LvB in DRE (with respect to FB depletion) line above spenders
- Added LvB above all spenders with T31 4p (only exception is EB+MWF)
- Changed "buff spenders with MotE" to "just buff smth" for fire builds
- Added updated st spender section that attempts to buff EB with MotE and ElS
- Removed weird 2t flux-related line (DRE fishing is not restricted to 1t anymore so no need)
- Added "ElS > hardcast lvb" on 2t
- Expanded fishing for DRE to 2t (no idea why it was restricted tbh)
- 3rd EB line in case everything goes to shit for fire builds
- Expanded "filling" with FrS to Els for nonlightning builds before you default to Lightining Bolt (whole bottom part with need a cleanup later on, a lot of redundant stuff left here after we moved LvB line up)

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -11232,7 +11232,7 @@ void shaman_t::init_action_list_elemental()
     single_target->add_action(
         "primordial_wave,target_if=min:dot.flame_shock.remains,if=!buff.primordial_wave.up&!buff.splintered_elements.up",
         "Use Primordial Wave as much as possible without wasting buffs." );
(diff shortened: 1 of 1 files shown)


Bloodmallet (approved):
Checked out locally. Simulation was looking good.

PR #8332, opened 2023-11-06
Add the ability to override the 'has_use_buff' property for trinkets

JontiH:
This change should let trinkets specify manually whether they should be picked up by the expression 'has_use_buff' in the APL.

Sometimes, items are written where they don't actually have a stat buff, so they aren't picked up.

I also added this to mirror of tomorrows and witherbarks branch, which had this issue

to manually add the has_use_buff, just add
```e.set_override_has_use_buff(1);```
to the item.

1 = has_use_buff is true, 0= has_use_buff is false

### DIFF
--- a/engine/class_modules/sc_enemy.cpp
+++ b/engine/class_modules/sc_enemy.cpp
@@ -1515,6 +1515,16 @@ void enemy_t::add_tank_heal_raid_event( tank_dummy_e tank_dummy )
     throw std::invalid_argument( "Cooldown lower than cooldown standard deviation." );
   }
 
+  if ( raid_event->cooldown_min == timespan_t::zero() )
+    raid_event->cooldown_min = raid_event->cooldown * 0.5;
+  if ( raid_event->cooldown_max == timespan_t::zero() )
+    raid_event->cooldown_max = raid_event->cooldown * 1.5;
+
+  if ( raid_event->duration_min == timespan_t::zero() )
+    raid_event->duration_min = raid_event->duration * 0.5;
+  if ( raid_event->duration_max == timespan_t::zero() )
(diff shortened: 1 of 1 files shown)


nyterage:
looks like something like this would work, but, would prefer a method to actually pass in the `use_buff` to this, allowing us to tell simc what the buff is. 

This would allow us to then pass it in to other auto parsing things, such as `trinket.x.proc.any_dps.duration` to actually return the duration of the buff provided.

JontiH:
> looks like something like this would work, but, would prefer a method to actually pass in the `use_buff` to this, allowing us to tell simc what the buff is.
> 
> This would allow us to then pass it in to other auto parsing things, such as `trinket.x.proc.any_dps.duration` to actually return the duration of the buff provided.


okay, so do you think override_has_use_buff should take a buff object? I guess the functionality of turning off has_use_buff is kind of irelevant.

should it be buff_t, or stat_buff_t? I can't figure out if they are interchangable

Jayezi:
Unless you really feel like working on it I think that stuff is out of scope of the original problem of just getting has_use_buff to return correctly. Setting effect.stat for trinkets with custom or disabled buffs should get the expression to pick it up.

JontiH:
> Unless you really feel like working on it I think that stuff is out of scope of the original problem of just getting has_use_buff to return correctly. Setting effect.stat for trinkets with custom or disabled buffs should get the expression to pick it up.

I can add that then.

Do you think this feature is useless going forward?

Jayezi:
not entirely but would probably need to be done in a way other than simply setting a buff at initialization, like for these trinkets that create multiple buffs and choose one depending on your highest stat etc

PR #8334, opened 2023-11-07
[Brewmaster] Add stagger level override

justindriggers:
Brewmaster has a handful of talents that modify damage or stats based on current stagger level (light, moderate, heavy). Sometimes it's desirable to run a simple sim to answer a question like "at what stagger level does X outperform Y?"

### DIFF
--- a/engine/class_modules/monk/sc_monk.cpp
+++ b/engine/class_modules/monk/sc_monk.cpp
@@ -9721,6 +9721,7 @@ namespace monk
     add_option( opt_int( "monk.chi_burst_healing_targets", user_options.chi_burst_healing_targets, 0, 30 ) );
     add_option( opt_int( "monk.motc_override", user_options.motc_override, 0, 5 ) );
     add_option( opt_float( "monk.squirm_frequency", user_options.squirm_frequency, 0, 30 ) );
+    add_option( opt_int( "monk.stagger_level_override", user_options.stagger_level_override, 0, 3 ) );
   }
 
   // monk_t::copy_from =========================================================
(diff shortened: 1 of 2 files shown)


renanthera:
This only allows for the uptime of a single stagger buff for that actor at all times. I would not consider this a useful improvement, as that is not something that occurs in game in any real context.

My current desire is to improve fluffy pillow (the default enemy actor) modeling so that the resultant stagger behaviour is more in line with real gameplay, and would absolutely be interested in changes along those lines, but I don't think that this approach helps toward improving sim accuracy.

justindriggers:
I agree that the ultimate goal should be to improve enemy modeling, but I fear that it's a much bigger undertaking than anyone is willing to take on. This approach is intended to simply answer targeted questions until a more thorough solution exists. Since simc is explicitly not a tank modeling program, it seems more useful to me to assume environmental conditions (such as damage intake) are being met in order to model DPS under such conditions.

It seems wasteful to thoroughly model enemies capable of keeping you at heavy stagger, just to determine that talent X can't keep up with Y under those conditions anyway, when we could easily _assume_ 100% heavy stagger uptime to answer the question quickly.

renanthera:
Tank modeling is an incredibly important aspect of simulating Brewmaster in any context. This fundamentally disconnects two aspects of stagger that are both VITALLY important for accurate sim results.

Consider the following sim:
https://www.raidbots.com/simbot/report/8F2vQcD84ogQEf1d3wh4TF/simc

In this case, the ONLY talent I swapped was Quick Sip for Healing Elixir. You'll note that there is a 2.6% dps loss, which many players would consider to be very, very significant, from a change that many players would consider to be incredibly minor. Not-QS/SS turns into a pretty substantial dps gain, when compared to the impact of many talent points.

Unlinking Stagger color from the pool will cause the simulator to produce results that are entirely impossible to match in game under a very large number of input conditions.

This sort of feature results in two different outcomes:
- Without Improved Niuzao:
* (a) Similar to either incredibly high damage intake, which you don't have in game (permanent red stagger).
* (b) Similar to having no effects that reduce your stagger pool with a constant damage intake.

- With Improved Niuzao:
* (c) You could trivially produce a situation where the effects provided by HT/ToN are not equivalent to the effects provided by Improved Niuzao, as the effects are "disconnected" by this change.

(a) doesn't really exist in game.
(b) is degenerate gameplay and absolutely not worth modeling with a player variable. If someone wishes to remove those talents and remove PB from the APL, they may do so, but an override can and likely will get misused.
(c) results in invalid state that doesn't even attempt to model in game behaviour.

My primary sticking point is (c), but they are certainly all relevant concerns. A higher level approach of accurately modeling damage intake accurately benefits all tanks, as they also have many effects, although closer in class to HT/ToN than Improved Niuzao in most cases, also resolves these concerns.

justindriggers:
With regard to (a) and (b), conceptually this type of override is intended to answer questions about moment-to-moment gameplay, e.g. given heavy stagger, is it still more beneficial to utilize BoC with TP over BoF in single-target? Though I suppose we already have stagger conditionals that can model these questions via the APL, if only it were more trivial to configure the sim’s damage intake profile.

I will concede to (c) producing inconsistent results, as well as the likelihood of users abusing the flag to produce inflated results that they feel validate otherwise degenerate play styles. I’ll keep this one in my back pocket to answer such targeted questions locally.

PR #8381, opened 2023-11-11
[Warlock] Affliction STS feature

Facenomore23:
STS feature is still ingame.

@Azevara  I guess with this you don't have to change backend. But we need to monitor if it get fixed at some point, and revert this apl change.

Last time checked : 11.11.2023

https://mimiron.raidbots.com/simbot/report/xkn5hLnNBWsx2FoYSe685p

After feature implement : https://mimiron.raidbots.com/simbot/report/eKQeGi2hWvfrcGGxpyHRN9


Azevara:
I had to tweak the backend behavior in #8383 , I don't think it will end up changing what you've requested here but please confirm so I know it's okay to merge.

PR #8394, opened 2023-11-12
[Mage] Create T31_Mage_Arcane Profile

ArcanePorom:


vituscze:
The profile needs to go here: https://github.com/simulationcraft/simc/blob/dragonflight/profiles/generators/Tier31/T31_Generate_Mage.simc

PR #8446, opened 2023-11-17
[APL] Removes duplicate conditional within Thunderous Roar APL

hrothmar:
**Summary:**
This change removes a duplicate "debuff.colossus_smash.up" which was present within the Thunderous Roar entry for Arm's single target APL. 
- Before: `|debuff.colossus_smash.up|debuff.colossus_smash.up|`
- After: `|debuff.colossus_smash.up|`

**Impact:**
This should have no perceivable change to performance of the spec, and solely exists to clean up an extraneous entry.


nyterage:
Heads up on this change, should be done in the `apl_warrior.cpp` file in `/engine/class_modules/apl/`. 

Doing it in the generated profile itself will only stick until the next time the profiles are regenerated.

hrothmar:
@nyterage thanks for the heads up and context that this was a generated profile. I have gone ahead and updated the core `.cpp` file within the engine as you recommended.

PR #8518, opened 2023-12-12
[Windwalker] Typo fix

Hurstilol:


vituscze:
Can we have normal PR/commit names, please?

PR #8519, opened 2023-12-12
Update warlock.cpp

Mlklock:
Fixing an issue this line had with Pip's.

PR #8521, opened 2023-12-12
[Warlock APL] Demonology on use Cantrip Update

Mlklock:
Redoing https://github.com/simulationcraft/simc/pull/8519 PR. 

Updates a line who uses the cantrip-effects on staffs like Iridal, Nymue for Demo. 

With how the line currently works it would conflict with users of Pip's Emerald Friendship badge shown in this sim https://www.raidbots.com/simbot/report/kSeofo6h9K6xmKjhCCjGth


Azevara:
Iridal does not share trinket cd so this should be fine.

PR #8550, opened 2023-12-25
[core] enable_all_item_effects

renanthera:
Work in progress implementation that enables all item effects registered via `register_special_effect`.

Unimplemented:
- Providing some behaviour to limit enabling all item effects to a specific expansion or other subset of effects.

Current issues with some potential solutions:
Many special effects use the pattern `effect.item->player` instead of directly using `effect.player`. This causes segfaults as `effect.item` is uninitialized due to the source item being unknowable and very likely incorrect.
- Provide a default item. Possibly fails if other information is extracted from the item.
- Update all implementations with `effect.player`.

Additionally, effects may depend on `item` through `effect.weapon_proc` being true. None have been discovered at this point.
- Provide a default item. An implementation for providing the mainhand was added, but is not guaranteed to work, as it could be the offhand that causes this to occur.

### DIFF
--- a/engine/action/dbc_proc_callback.cpp
+++ b/engine/action/dbc_proc_callback.cpp
@@ -169,7 +169,8 @@ void dbc_proc_callback_t::trigger( action_t* a, action_state_t* state )
 
 dbc_proc_callback_t::dbc_proc_callback_t( const item_t& i, const special_effect_t& e )
   : action_callback_t( e.player ),
-    item( i.parent_slot != SLOT_INVALID ? i : e.player->items[SLOT_MAIN_HAND] ),
+    // item( i.parent_slot != SLOT_INVALID ? i : e.player->items[SLOT_MAIN_HAND] ),
+    item( i ),
     effect( e ),
     cooldown( nullptr ),
     target_specific_cooldown( nullptr ),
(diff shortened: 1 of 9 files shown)


renanthera:
Three effects got blacklisted in the latest commit:
- Ripple in Space (Azerite Trait)
- Harmonic Dematerializer (BfA Punchcard?)
- Subroutine: Optimize (BfA Punchcard?)

Everything else successfully runs.

New problems:
- Effects that exclusively use the fallback effect db fail to operate correctly, but may be registered. (Potion of Chilled Clarity)
- Effects that are on-use are never used. Using them is non-trivial, as they are not equipped, and the shared cooldown between trinkets will be a problem.
- Role/spec-impacted effects won't run all of their effect variants. Not too big of an issue as it's still expected for this to be used with many specs.

renanthera:
Punchcard effects have a dependency on another item (the Pocket Sized Computational Device they are socketed in), as they are documented to scale based on the item level of the PSCD they are socketed in. As `enable_all_item_effects` does not alter the equipped gear of the actor(s) in the sim, this dependency is impossible to fulfill, and some bugs are introduced. 

Previously, the function `init_punchcard` returns a null `item_t` object if one of several checks fail:
- if `effect.enchant_data` is null.
- if no equipped item that can fit the gem is found.

The Punchcards Yellow Punchcard, Subroutine: Overclock already have short-circuits implemented in case a null `item_t` is returned. Others such as Subroutine: Recalibration, Subroutine: Optimization, Harmonic Dematerializer, did not.

Subroutine: Recalibration technically does not require such a condition to function with `enable_all_item_effects`, but it was added for consistency. A player would be hard-pressed to equip the Punchcard without modifying their character in "improper" ways to not also have the item Punchcards were socketed in anyway.

Null-checking a function that may return a null object is something I would consider good practice, thus a reasonable requirement for new and existing special effect implementations.

renanthera:
Closing this for now in favor of a different potential approach to solving the same problem.

PR #8551, opened 2023-12-26
[Player] Add `apply_player_auras`

nyterage:
Very similar to the `parse_buff_effects` mix-in, though player scoped rather than action scoped. Primarily meant to allow easier implementation of passive effects that modify total player stats, pet damage mod effects, and player wide damage mod effects. 

Current iteration is still VERY work in progress, and is more meant as a proof of concept to gauge interest in adding this.


nyterage:
Updated the scope a decent bit, as a lot of potential passive aura effects could also have their logic applied to buffs. 

Now allows buff aura parsing with a simple flag `set_parse_player_auras( true )`, and if its effect is accounted for, will apply the affecting auras while the buff is active, and automatically invalidate the proper caches. The only exception to this is if a buff has another aura that affects it. If that is the case, it will need to be applied via `apply_buff_aura_effects( buff, modifying_spells );`

Passive effects can be parsed utilizing `apply_passive_aura_effects( spell );`, similar to buffs, can also have modifier spells added to them with `apply_passive_aura_effects( spell, modifying_spells );`

PR #8574, opened 2024-01-06
[Death Knight] Add option to push back first AMS cast

Khazakdk:


engine/class_modules/sc_death_knight.cpp
```diff
@@ -1261,6 +1261,7 @@ struct death_knight_t : public player_t {
     std::vector<timespan_t> amz_use_time;
     bool amz_specified = false;
     double average_cs_travel_time = 0.4;
+    timespan_t first_ams_cast = timespan_t::from_seconds(20);
```
nyterage: could probably just be `timespan_t first_ams_cast = 20_s;`

engine/class_modules/sc_death_knight.cpp
```diff
@@ -8594,6 +8596,10 @@ std::string death_knight_t::create_profile( save_e type )
     {
       profile_str += "deathknight.amz_absorb_percent=" + util::to_string( options.amz_absorb_percent ) + term;
     }
+    if ( options.first_ams_cast != timespan_t::from_seconds( 20 ) )
```
nyterage: same as previous comment

engine/class_modules/sc_death_knight.cpp
Khazakdk: Is there a practical difference or just style?

PR #8603, opened 2024-01-17
[Warlock] Stat on use trinket logic fix

abousumer:
Updated the precombat variable for stat on use trinkets 
Aff: https://www.raidbots.com/simbot/report/5ys9jz6sJ2BcNgcVtJzLWs
Demo: https://www.raidbots.com/simbot/report/ugf9pnjycZnxUkYAHDcxGP
Destro: https://www.raidbots.com/simbot/report/kFbBAfDrLkXUnvCPYfrf1p

### DIFF
--- a/engine/class_modules/apl/warlock.cpp
+++ b/engine/class_modules/apl/warlock.cpp
@@ -50,8 +50,8 @@ void affliction( player_t* p )
   precombat->add_action( "augmentation" );
   precombat->add_action( "summon_pet" );
   precombat->add_action( "variable,name=cleave_apl,default=0,op=reset" );
-  precombat->add_action( "variable,name=trinket_1_buffs,value=trinket.1.has_use_buff.intellect|trinket.1.has_use_buff.mastery|trinket.1.has_use_buff.versatility|trinket.1.has_use_buff.haste|trinket.1.has_use_buff.crit" );
-  precombat->add_action( "variable,name=trinket_2_buffs,value=trinket.2.has_use_buff.intellect|trinket.2.has_use_buff.mastery|trinket.2.has_use_buff.versatility|trinket.2.has_use_buff.haste|trinket.2.has_use_buff.crit" );
+  precombat->add_action( "variable,name=trinket_1_buffs,value=trinket.1.has_use_buff" );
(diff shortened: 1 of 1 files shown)


engine/class_modules/apl/warlock.cpp
```diff
@@ -50,8 +50,8 @@ void affliction( player_t* p )
   precombat->add_action( "augmentation" );
   precombat->add_action( "summon_pet" );
   precombat->add_action( "variable,name=cleave_apl,default=0,op=reset" );
-  precombat->add_action( "variable,name=trinket_1_buffs,value=trinket.1.has_buff.intellect|trinket.1.has_buff.mastery|trinket.1.has_buff.versatility|trinket.1.has_buff.haste|trinket.1.has_buff.crit|trinket.1.is.mirror_of_fractured_tomorrows|trinket.1.is.spoils_of_neltharus|trinket.1.is.nymues_unraveling_spindle" );
-  precombat->add_action( "variable,name=trinket_2_buffs,value=trinket.2.has_buff.intellect|trinket.2.has_buff.mastery|trinket.2.has_buff.versatility|trinket.2.has_buff.haste|trinket.2.has_buff.crit|trinket.2.is.mirror_of_fractured_tomorrows|trinket.2.is.spoils_of_neltharus|trinket.2.is.nymues_unraveling_spindle" );
+  precombat->add_action( "variable,name=trinket_1_buffs,value=trinket.1.has_use_buff.intellect|trinket.1.has_use_buff.mastery|trinket.1.has_use_buff.versatility|trinket.1.has_use_buff.haste|trinket.1.has_use_buff.crit" );
```
Jayezi: has_use_buff is a general check on the trinket for any dps stat buff, the stat specification following it is actually ignored so using it like this is basically saying true|true|true|true|true. You could replace the entire line with just value=trinket.1.has_use_buff, or could even replace the entire variable with just trinket.1.has_use_buff in the other lines that use the variable.

abousumer: Got it! Vocox and I discussed it and decided to keep the variable in case we need to add trinkets to it in the future, like we previously did with Spoils and Mirror. Thank you!

PR #8612, opened 2024-01-21
[sim] Do not prevent the default actor from performing its actions.

renanthera:
Commit 4be94135517 caused Fluffy Pillow to no longer be the default actor used in all sims and the new default actor for DungeonSlice and DungeonRoute Dungeon is forced to sleep immediately upon initialization.

This causes tanks to not take damage in DungeonSlice sims, which makes the fight style perform poorly.

Given that DungeonRoute sims can have damage dealt to the actor configured, this may not be an appropriate change for DungeonRoute.

An alternative approach would be to provide all sims with an actor that deals damage to all tank-roled players that are active by default, with a global sim option to disable this behaviour.

### DIFF
--- a/engine/sim/sim.cpp
+++ b/engine/sim/sim.cpp
@@ -2804,7 +2804,6 @@ void sim_t::init()
     if ( main_target_str.empty() && target_list.empty() )
     {
       target = module_t::enemy()->create_player( this, "Dungeon" );
-      // target->initial.sleeping = target->base.sleeping = target->current.sleeping = true;
     }
     else
     {
--- a/engine/sim/sim.cpp
+++ b/engine/sim/sim.cpp
@@ -2330,11 +2330,11 @@ void sim_t::init_fight_style()
     case FIGHT_STYLE_PATCHWERK:
       raid_events_str.clear();
       break;
-    
+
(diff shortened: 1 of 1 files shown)


renanthera:
Apologies for the specific request for review. I clicked it, but GitHub did not allow me to un-click it.

renanthera:
Following a brief discussion with Koji, it was noted that this approach would likely introduce other issues.
> Q: Do you possibly have any recollection as to why you made the default actor sleep in DSlice and DRoute sims?
Does that do anything in sims for dps specs beyond "the actor is sleeping so the sim ostensibly runs faster"?
> A: So that we don't need to rely on janky invulnerability event retargeting. Which messes up precombat and timestamp 0. And some other persistent type debuff effects.

Currently rethinking the PR from the direction of deprecating "Target Dummy" NPCs dealing damage whatsoever and just using raid events, but further research is required.

I had already intended to do so, given some of the improved capabilities that raid events provide, along with no longer maintaining two independent systems to model damage intake.

renanthera:
https://github.com/simulationcraft/simc/pull/8617 and another work in progress PR will address this behaviour in a different way, opting instead to augment sim options and `raid_events` to achieve a comparable result.

PR #8621, opened 2024-01-24
[sim] Prevent additional enemies spawned by `desired_targets` from following tank damage action list by default.

renanthera:
`tank_dummy_enemy_t::init()` was missing an invocation of `enemy_t::init()` to complete initialization.

This resulted in uninitialized data to be consumed by the JSON writer, causing it to terminate mid-way through writing the JSON objects to disk.

> Additional targets created by any number of methods, but chiefly `desired_targets` sim opt would cause damage intake to scale linearly. This caused tank sims to change behaviour drastically as target count varied.
> 
> Two things are done to address this behaviour, but still allow some configurability in this space, pending further work on tank damage intake.
> 
> First, most common methods of creating enemy actors no longer follow the default tank enemy action list. The `tank_dummy_enemy_t` module has this behaviour preserved.
> 
> Fluffy Pillow is created using the `tank_dummy_enemy_t` module, whereas other enemies use the `enemy_t` module.
> 
> Secondly, a new sim opt `desired_tank_targets=k` is added which spawns `k-1` additional tank target dummies, as one is provided by default (Fluffy Pillow), for a total of `k` tank target dummies in the sim.
> 
> This new option is compatible with the existing `desired_targets=n`, which spawns up to `n` additional targets prior to `desired_tank_targets` in order for behaviour of the `desired_targets` opt to remain unchanged.
> 
> Results in a total number of targets `n + k - 1` targets, where both `n` and `k` default to 1.

### DIFF
--- a/engine/class_modules/sc_enemy.cpp
+++ b/engine/class_modules/sc_enemy.cpp
@@ -1445,9 +1445,10 @@ void enemy_t::init_target()
 
 // enemy_t::init_actions ====================================================
 
+// default provided to be overloaded for tank_dummy_enemy_t::generate_action_list()
 std::string enemy_t::generate_action_list()
 {
-  return generate_tank_action_list( tank_dummy_e::MYTHIC );
+  return "/";
 }
 
 void enemy_t::generate_heal_raid_event()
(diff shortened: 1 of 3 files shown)


seriallos:
Looks good locally!

PR #8624, opened 2024-01-26
[Reports] Move print_version_info

### LINKED ISSUE #8623: PTR Flag not working for spell_query
### Describe the bug
Running `simc ptr=1 spell_query="x"` will output the same as ptr=0

### Expected behavior
spell_query should run in ptr mode when ptr is set to 1 

### To Reproduce
Run the spell_query command manually or run the spelldatadump_ptr.bat file in /SpellDataDump, the output will be Live data and the build info will also output live.

### Additional information
* SimulationCraft Version (required)
Latest

* If SimulationCraft is crashing, please include your **dxdiag** file - Instructions here - 
N/A

N/A

JeremyMonk:
* Move print_version_info to after command arguments are parsed and setup of control
* Fixes #8623

@vituscze

### DIFF
--- a/engine/sc_main.cpp
+++ b/engine/sc_main.cpp
@@ -263,23 +263,13 @@ int sim_t::main( const std::vector<std::string>& args )
 
     sim_control_t control;
 
-    auto version_displayed = false;
-    auto _print_version_info = [ this, &version_displayed ] ( )
-    {
-      if ( !version_displayed )
-      {
-        print_version_info( *dbc );
-        version_displayed = true;
-      }
-    };
-
     try
     {
       control.options.parse_args( args );
     }
     catch ( const std::exception& )
     {
-      _print_version_info();
+      print_version_info( *dbc );
       fmt::print( "\n" );
       std::throw_with_nested( std::invalid_argument( "Incorrect option format" ) );
     }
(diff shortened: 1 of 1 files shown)


gastank:
This would result in any options parse failure not giving the version info at all. Version info should always be displayed for all outputs, valid or invalid. If you want to display the PTR string post-parsing the string can be moved to a post-parse function like print_build_info().

engine/sc_main.cpp
```diff
@@ -282,12 +293,14 @@ int sim_t::main( const std::vector<std::string>& args )
...
     {
+      _print_version_info();
       fmt::print( "\n" );
       std::throw_with_nested( std::runtime_error( "Setup failure" ) );
     }
 
-    // print version info
-    print_version_info( *dbc );
+    // print version info if it hasn't been displayed already
+    if ( !version_displayed )
```
vituscze: Any reason this check is here twice?

JeremyMonk: Forgot to remove it but I think the lambda is unnecessary anyway as code execution stops after throw_with_nested is called

vituscze:
Looks reasonable to me now.

PR #8638, opened 2024-02-02
Add health% timeline to JSON and HTML output

vituscze:
The enemy health graph tends to confuse people when simc runs with `fixed_time=1`. Instead of removing/hiding it, make the information about what's happening to health% explicit.


vituscze:
Example of the HTML report:
image

Issue #8639, opened 2024-02-03
[rng] Using Clamp on Random Values


renanthera:
In several locations, I have found `clamp` being used on randomly generated values.
https://github.com/simulationcraft/simc/blob/40ee6c8d181de01b0bd80c208ab37fee730d7882/engine/sim/raid_event.cpp#L1690-L1706
https://github.com/simulationcraft/simc/blob/40ee6c8d181de01b0bd80c208ab37fee730d7882/engine/sim/raid_event.cpp#L1708-L1715

This produces some likely undesired distributions of random values, such as the following:
<img src="https://github.com/simulationcraft/simc/assets/51387586/0528d74c-9d48-4d9e-bd65-b8ef1ab66ed1" width=50% height=50%>

I also have the following gist that provides steps for reproduction of this chart, and the code I used to generate the image:
<https://gist.github.com/renanthera/c0a97491cdef390483fd7334b2360651>

My biggest concern, thus the cause for this issue, is that this may be done elsewhere in the codebase. Additionally, other functions besides clamp being used on randomly generated values can result in very similar behaviour.

Incorrect input into these functions could result in very non-random results, and I am unsure how to go about finding any others that may exist.

renanthera:
CORRECTED BY #8652 

I have found two additional cases of similar behaviour:
https://github.com/simulationcraft/simc/blob/fe986ea18b4e108ba38cf4c2004f96aafbba94f0/engine/util/rng.hpp#L235-L237
https://github.com/simulationcraft/simc/blob/fe986ea18b4e108ba38cf4c2004f96aafbba94f0/engine/util/rng.hpp#L251-L257

I have experimentally verified the former leads to a similar behaviour as was documented in the original issue, and the latter is functionally the same logic, albeit without control via method parameters.
Both of these aforementioned methods have also been used in the current codebase.

Additionally, there are some concerns with the `timespan_t` specializations of `range`, `gauss` and `exgauss`, but I haven't finished looking through those or breadth of their usage at this time.

PR #8652, opened 2024-02-16
[rng] Truncated Gaussian Distribution

renanthera:
* Implemented additional methods `gauss_ab`, `gauss_a`, `gauss_b` to provide random values from the Truncated Gaussian Distribution, with different bounding behaviour. ([a,b], [a,\infty], [-\infty,b]).
 * Updated some code style in `rng` to make it more legible.
 * Updated usages of the old `gauss` parameter `truncate_low_end` to use the new equivalent method `gauss_a`.
 
 This PR addresses all concerns from #8639 that were isolated to the `rng` module, but does not extend further than that.

Here are some sample plots of these new generators with the parameters of min = -2, max = 2, mean = 0, stddev = 1, sampled 50000 times:
Figure_1

I have also requested review from Azevara and Gas, as I have touched modules you typically maintain.


Azevara (approved):
The use of the functions with truncate at 0 in the warlock module was to prevent time events from being scheduled in the past, so assuming the gauss_a distribution works the same with a=0, mean=0 as it does in the example with a=-2 I have no issues with this PR.

engine/util/rng.cpp
```diff
@@ -149,186 +149,6 @@ const char* xorshift1024_t::name() const noexcept
...
- * @brief The standard normal CDF, for one random variable.
- *
- *   Author:  W. J. Cody
- *   URL:     http://www.netlib.org/specfun/erf
- *   Source:  http://home.online.no/~pjacklam/notes/invnorm/
- *
- * This is the erfc() routine only, adapted by the
- * transform stdnormal_cdf(u)=(erfc(-u/sqrt(2))/2;
- */
-double stdnormal_cdf( double u )
```
Dorovon, line 162: Why were this and `stdnormal_inv` moved to the header and made part of `basic_rng_t`?

engine/util/rng.hpp
```diff
@@ -194,66 +420,82 @@ double basic_rng_t<Engine>::range( double min, double max )
...
+{
+  assert( stddev >= 0.0 && "Stddev must be non-negative." );
+  assert( min < max && "Minimum must be less than maximum." );
+  assert( mean >= min && mean <= max && "Mean must be contained within the interval [min, max]" );
+
+  if ( stddev == 0.0 )
+    return mean;
+
+  double min_cdf      = stdnormal_cdf( ( min - mean ) / stddev );
+  double max_cdf      = stdnormal_cdf( ( max - mean ) / stddev );
```
Dorovon, line 277: I wonder how much time could be saved in some cases by not calculating these CDF values every time the function is called for the cases where `min` and `max` are going to be constant for the entire simulation.

Given that (a) I think it would require the caller to do things very differently and (b) I don't think the rate that this is called is going to be that high compared to things like `real`, it's probably not worth doing at the moment, but I thought it was worth pointing out.

engine/class_modules/sc_druid.cpp
```diff
@@ -8579,15 +8570,14 @@ struct convoke_the_spirits_t : public druid_spell_t
                    };
 
     cast_list.insert( cast_list.end(),
-                      static_cast<size_t>( rng().range( guidance ? 2.5 : 4, guidance ? 7.5 : 9 ) ),
+                      rng().range<size_t>( guidance ? 2.5 : 4, guidance ? 7.5 : 9 ),
```
Dorovon: This results in the 2.5 and 7.5 being truncated to 2 and 7, which changes the behavior of this code.

engine/class_modules/sc_druid.cpp
```diff
@@ -8579,15 +8570,14 @@ struct convoke_the_spirits_t : public druid_spell_t
...
                       CAST_OFFSPEC );
 
     insert_exceptional( CAST_FERAL_FRENZY );
 
     cast_list.insert(
         cast_list.end(),
-        _clamp_and_cast( rng().gauss( guidance ? 3 : 4.2, guidance ? 0.8 : 0.9360890055, true ), 0,
-                         max_ticks - cast_list.size() ),
+        rng().gauss_ab<size_t>( guidance ? 3 : 4.2, guidance ? 0.8 : 0.9360890055, 0, max_ticks - cast_list.size() ),
```
Dorovon: See above, this should use the double version and cast to size_t after returning.

engine/util/rng.hpp
```diff
@@ -194,66 +420,82 @@ double basic_rng_t<Engine>::range( double min, double max )
...
   return - std::log( 1 - x ) * nu;
 }
 
 /// Exponentially Modified Gaussian Distribution
 template <typename Engine>
 double basic_rng_t<Engine>::exgauss( double gauss_mean, double gauss_stddev, double exp_nu )
 {
-  double v = gauss( gauss_mean, gauss_stddev ) + exponential( exp_nu );
-  return v > 0.0 ? v : 0.0;
+  return gauss( gauss_mean, gauss_stddev ) + exponential( exp_nu );
```
Dorovon, line 308: player_t::total_reaction_time should be changed to ensure no negative values can happen, although with the default parameters the probability of that is extremely low.

renanthera:
Gauss templates are currently fundamentally flawed, as they expect the parameters to be of the same type as is selected using the angle bracket notation.

`gauss_ab<int>( 0.0, 1.0, -0.5, 0.5 )` would first demote the parameters to integers, then cast them to doubles, calling the function `gauss( 0.0, 1.0, 0.0, 0.0)` (if I'm recalling demotion/promotion rules properly), which does not match up with intent.

I can also leave the current methods as-is so auto-specialization works fine, and this will pick up the manually type-specified cases.

Decoupling type-parameter of return and arguments would fix this, which would have a signature looking more like this:

```c++
template <typename R, typename P, typename std::enable_if<std::numeric_limits<T>::is_integer && !std::numeric_limits<T>::is_signed>
R gauss( P mean, P stddev )
```

This is helpful in the following case:
https://github.com/simulationcraft/simc/blob/663441e820848fa698ab3092139f2769bdc35bdf/engine/class_modules/sc_druid.cpp#L8578-L8581

engine/util/rng.hpp
```diff
@@ -66,26 +244,67 @@ struct basic_rng_t
...
+        gauss_a( static_cast<double>( mean ), static_cast<double>( stddev ), static_cast<double>( min ) ) );
+  }
+
+  template <typename T,
+            typename std::enable_if<std::numeric_limits<T>::is_integer && std::numeric_limits<T>::is_signed, T>::type>
+  T gauss( T mean, T stddev )
+  {
+    return static_cast<T>( gauss_a( static_cast<double>( mean ), static_cast<double>( stddev ),
+                                    static_cast<double>( std::numeric_limits<T>::min() ) ) );
+  }
```
Dorovon: I think the `is_signed` checks here and above look backwards. Truncating here should only happen when the distribution is unsigned.

Dorovon (changes requested):

engine/util/rng.cpp
renanthera: They're necessary for the computation of `gauss_ab` and such. Including `rng.cpp` in `rng.hpp` seems quite wrong, but I can't say I explored that specific change very deeply.

I can take a look at it tomorrow if you don't have any immediate recommendations.

Dorovon:
> Gauss templates are currently fundamentally flawed, as they expect the parameters to be of the same type as is selected using the angle bracket notation.

Ah, I didn't see this comment before I submitted that review. For those two cases, I think sticking to single type templates makes more sense, since having to specify the template to pick the return type isn't really much different from casting the type afterwards and having two template types in this case is not very standard.

engine/util/rng.hpp
renanthera: I agree that caching `stdnormal_cdf` values could potentially be valuable, and further profiling could be done on that front w.r.t the expense of `stdnormal_cdf`.

engine/class_modules/sc_druid.cpp
renanthera: Yes, I mentioned that and a potential solution here: https://github.com/simulationcraft/simc/pull/8652#issuecomment-1948313138

engine/util/rng.hpp
renanthera: I actually need to verify some behaviour on this front, but I'll address it tomorrow.

engine/util/rng.hpp
renanthera: I would agree that they are backwards. I have made the appropriate local fix, which will be included in the next commit.

renanthera:
In that case, I would be in favor of removing all templates for `gauss` and all related functions except for the case that covers unsigned integers, to avoid the demotion issue altogether and mandate all casting is explicitly handled, hopefully forcing the author to think about the issue.

I would agree that they're very similar (adding the extra type parameter and just static casting it afterwards yourself), but the former is definitely convenient.

Dorovon:
Yeah, I agree with removing the other gauss templates.

renanthera:
Addresses concerns on type demotion and usage of `is_signed`, but does not build.

I'll investigate what removal of the `gauss` templates breaks shortly.

engine/util/rng.cpp
Dorovon: I believe if you just moved the declarations that were in the header to the top of the rng namespace then the functions could stay in the .cpp file.
```cpp
double stdnormal_cdf( double );
double stdnormal_inv( double );
```

gastank (approved):

engine/util/rng.hpp
renanthera: After further review, I see two potential solutions:
- Provide an overload for unsigned parameter inputs in which `gauss_a( gauss_mean, gauss_stddev, 0.0)` is called instead of `gauss(...)`.
- Opt for `gauss_a`'s usage in `exgauss` instead of `gauss` altogether (no conditioning on signedness of type).

For consistency's sake, the former seems most reasonable.

EDIT:

As `player_t::total_reaction_time` is a `timespan_t`, I opted instead to implement the `timespan_t` specialization of `gauss` and `exgauss` with `gauss_a`. Providing non-negative values by default for all time intervals seems reasonable.

Additionally, it is not possible for either `gauss_a( ..., 0.0 )` nor `exponential( nu )` for `nu > 0` to return negative values, so their sum will be non-negative as well. I feel like that should be safe enough for usage in `timespan_t`, short of ill-formed, albeit still legal, parameters.

engine/util/rng.cpp
renanthera: I realized as much this afternoon when I thought about it a bit more. Will do.

renanthera:
It also proved feasible to remove templates for Range as well. The only case that comes to mind where numeric promotion/demotion would be insufficient is for indexing.

engine/util/rng.hpp
```diff
@@ -65,33 +68,45 @@ struct basic_rng_t
...
+  {
+    return static_cast<T>( gauss_a( static_cast<double>( mean ), static_cast<double>( stddev ), 0.0 ) );
+  }
+
+  template <typename T, typename std::enable_if_t<!std::numeric_limits<T>::is_signed>>
+  T exgauss( T gauss_mean, T gauss_stddev, T exp_nu )
+  {
+    return static_cast<T>( gauss_a( static_cast<double>( gauss_mean ), static_cast<double>( gauss_stddev ), 0.0 ) +
+                           exponential( static_cast<double>( exp_nu ) ) );
+  }
```
vituscze: This won't work. The template line needs to be:
```cpp
template <typename T, typename = std::enable_if_t<!std::numeric_limits<T>::is_signed>>
```
Note the additional equals sign.

engine/util/rng.hpp
```diff
@@ -65,33 +68,45 @@ struct basic_rng_t
...
-  template <typename T, typename = std::enable_if_t<std::is_integral<T>::value>>
-  T range( T min, T max ) {
-    return static_cast<T>(range(static_cast<double>(min), static_cast<double>(max)));
-  }
-
-  /// Uniform distribution in the range [0..max)
-  template <typename T, typename = std::enable_if_t<std::is_integral<T>::value>>
-  T range( T max ) {
-    return range<T>( T{}, max );
-  }
```
Dorovon, line 78: I think we do still need these range templates, or rather we still need the second one as well as this version with a floor to correct the bug with `min < 0` to replace the first one. In addition to fixing that bug, not having these templates to do the integer to double static casts results in a significant number of new warnings with some compilers (e.g., MSVC).
```cpp
  /// Uniform distribution in the range [min..max)
  template <typename T, typename = std::enable_if_t<std::is_integral<T>::value>>
  T range( T min, T max ) {
    return static_cast<T>( std::floor( range( static_cast<double>( min ), static_cast<double>( max ) ) ) );
  }
```

engine/util/rng.hpp
renanthera: Thanks, I was using partial specializations previously (which ended out identical in template to what I had, albeit with a forward declaration above) and had not realized that they're syntactically different.

Dorovon (approved):

renanthera:
Quickly threw together some histograms of generated random values. Found a bug with `timespan_t` random generators as well, which got fixed in the previous commit.

Definitely not a thorough test, but checks some cases for non-negative values and everything looks good to me and matches what I would expect for each of these distributions, given the selected parameters.

Additionally, all of the desired behaviour w.r.t unsigned overloads enabling for the appropriate parameters appears to be working in these charts.

Figure_1

Relevant sources for generating this image can be found at:
https://github.com/renanthera/simc/tree/rng_fixes-8639-view-rng-generators

PR #8657, opened 2024-02-19
[Frost DK] Updated BoS Food, Regen Profile

Solindvian:
Re-ran food sims below post recent apl and BiS changes.

100 - https://www.raidbots.com/simbot/report/vbhadRNtnSfRGKbu4RVDEx
70 - https://www.raidbots.com/simbot/report/g9tmUvRFHicZmXAHv3hUW8
50 - https://www.raidbots.com/simbot/report/hwZxWTucrraX7UfijW44hK
0 - https://www.raidbots.com/simbot/report/88pGHyMDLeVxyfAYz1qF3P


engine/class_modules/apl/apl_death_knight.cpp
```diff
@@ -41,7 +41,7 @@ std::string food( const player_t* p )
 
   if ( p->true_level >= 61 )
   {
-    frost_food  = ( p->dual_wield() ) ? "thousandbone_tongueslicer" : "sizzling_seafood_medley";
+    frost_food  = ( p->dual_wield() ) ? "sizzling_seafood_medley" : "sizzling_seafood_medley";
```
nyterage: If both 2h and dw are using the same food, probably best to remove the inline if statement and just have it as 
```suggestion
    frost_food  = "sizzling_seafood_medley";
```

nyterage (changes requested):

engine/class_modules/apl/apl_death_knight.cpp
Solindvian: That's true, to be honest I only left it in to thoughtlessly swap it back in the future but might as well just remove it as it's not complicated anyway.

PR #8660, opened 2024-02-21
[tanks] Mark non-validated fight styles.

renanthera:
Marks Dungeon Slice and Dungeon route as non-validated fight styles for all tanks, until at minimum tank damage intake defaults are fixed.

### DIFF
--- a/engine/class_modules/monk/sc_monk.cpp
+++ b/engine/class_modules/monk/sc_monk.cpp
@@ -7750,6 +7750,22 @@ namespace monk
     sample_datas.stagger_pct_timeline.add( sim->current_time(), stagger_pct_val * 100.0 );
   }
 
+  bool monk_t::validate_fight_style( fight_style_e style ) const
+  {
+    if ( specialization() == MONK_BREWMASTER )
+    {
+      switch ( style )
+      {
+      case FIGHT_STYLE_DUNGEON_ROUTE:
+      case FIGHT_STYLE_DUNGEON_SLICE:
+        return false;
+      default:
+        return true;
+      }
+    }
+    return true;
+  }
+
   // monk_t::init_spells ======================================================
   void monk_t::init_spells()
   {
--- a/engine/class_modules/monk/sc_monk.cpp
+++ b/engine/class_modules/monk/sc_monk.cpp
@@ -7763,6 +7763,19 @@ namespace monk
         return true;
       }
     }
+    if ( specialization() == MONK_WINDWALKER )
+    {
+      switch ( style )
+      {
+        case FIGHT_STYLE_PATCHWERK:
+        case FIGHT_STYLE_CASTING_PATCHWERK:
+        case FIGHT_STYLE_DUNGEON_ROUTE:
+        case FIGHT_STYLE_DUNGEON_SLICE:
+          return true;
+        default:
+          return false;
+      }
+    }
     return true;
   }
 
--- a/engine/class_modules/paladin/sc_paladin.cpp
+++ b/engine/class_modules/paladin/sc_paladin.cpp
@@ -2533,9 +2533,9 @@ bool paladin_t::validate_fight_style( fight_style_e style ) const
     {
       case FIGHT_STYLE_DUNGEON_ROUTE:
       case FIGHT_STYLE_DUNGEON_SLICE:
-        return true;
-      default:
         return false;
+      default:
+        return true;
     }
   }
   return true;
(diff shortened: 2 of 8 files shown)


engine/class_modules/monk/sc_monk.cpp
```diff
@@ -7750,6 +7750,22 @@ namespace monk
...
+      switch ( style )
+      {
+      case FIGHT_STYLE_DUNGEON_ROUTE:
+      case FIGHT_STYLE_DUNGEON_SLICE:
+        return false;
+      default:
+        return true;
+      }
+    }
+    return true;
```
Hinalover, line 7779: I know this is for Brewmaster, but for Windwalker, Patchwork, Casting Patchwork, Dungeon Route and Dungeon Slice are the only ones that are valid for us. Add Cleave and Hectic Add Cleave are not supported due to the APL not holding CDs for add spawning. Also any movement is not supported as well.

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2524,6 +2524,23 @@ void paladin_t::init_action_list()
...
+  if ( specialization() == PALADIN_PROTECTION )
+  {
+    switch ( style )
+    {
+      case FIGHT_STYLE_DUNGEON_ROUTE:
+      case FIGHT_STYLE_DUNGEON_SLICE:
+        return true;
+      default:
+        return false;
+    }
```
msserrano, line 2539: This seems backwards - I think you have true / false flipped here

engine/class_modules/paladin/sc_paladin.cpp
```diff
@@ -2524,6 +2524,23 @@ void paladin_t::init_action_list()
...
+  }
+  return true;
+}
+
```
mserrano:
```suggestion
bool paladin_t::validate_fight_style( fight_style_e style ) const
{
  if ( specialization() == PALADIN_PROTECTION )
  {
    switch ( style )
    {
      case FIGHT_STYLE_DUNGEON_ROUTE:
      case FIGHT_STYLE_DUNGEON_SLICE:
        return false;
      default:
        return true;
    }
  }
  return true;
}

```

mserrano (changes requested):

engine/class_modules/sc_death_knight.cpp
```diff
@@ -10405,6 +10406,24 @@ void death_knight_t::init_finished()
...
+    {
+    case FIGHT_STYLE_DUNGEON_ROUTE:
+    case FIGHT_STYLE_DUNGEON_SLICE:
+      return false;
+    default:
+      return true;
+    }
+  }
+  return true;
+}
```
Khazakdk:
```suggestion
bool death_knight_t::validate_fight_style( fight_style_e fight ) const
{
  if ( specialization() == DEATH_KNIGHT_BLOOD )
  {
    switch ( fight )
    {
    case FIGHT_STYLE_PATCHWERK:
    case FIGHT_STYLE_CASTING_PATCHWERK:
      return true;
    default:
      return false;
    }    
  }
  if ( specialization() == DEATH_KNIGHT_FROST )
  {
    switch ( fight )
    {
    case FIGHT_STYLE_PATCHWERK:
    case FIGHT_STYLE_CASTING_PATCHWERK:
    case FIGHT_STYLE_DUNGEON_SLICE:
    case FIGHT_STYLE_DUNGEON_ROUTE:
      return true;
    default:
      return false;
    }
  }

  if ( specialization() == DEATH_KNIGHT_UNHOLY )
  {
    switch ( fight )
    {
    case FIGHT_STYLE_PATCHWERK:
    case FIGHT_STYLE_CASTING_PATCHWERK:
    case FIGHT_STYLE_DUNGEON_SLICE:
    case FIGHT_STYLE_DUNGEON_ROUTE:
      return true;
    default:
      return false;
    }
  }
  return false;
}
```

I'd like to leverage `validate_fight_style` for DPS specs as well and I think the easiest way would be to drop the new function into your PR.

engine/class_modules/paladin/sc_paladin.cpp
renanthera: Manually corrected for in conjunction with combining changes from #8665.

engine/class_modules/sc_death_knight.cpp
renanthera: Manually corrected for paladin in conjunction with combining changes from https://github.com/simulationcraft/simc/pull/8665.

PR #8663, opened 2024-02-21
[Warlock] Fix occasional 6th Drain Soul tick

Fatherdotter:
Re-ordered multiplication, hopefully preserves some precision before multiplying into timespans.

Appears to resolve the issue of occasional 6-tick DS and is reflected in logs and haste scaling plots matching expectations.


Azevara (commented):
Builds off of fix originally done in #8030. No issues with this change.

PR #8665, opened 2024-02-22
[Death Knight] Add validate_fight_style

Khazakdk:
Adds a warning to text output/raidbots if an unsupported fight style is set.


Khazakdk:
Closing to avoid conflicts with #8660. We will merge these changes into that PR instead

PR #8673, opened 2024-02-28
[rng] Add additional timespan_t gauss_* methods and remove uses of clamp on randomly generated values.

renanthera:
[rng]
* Explicitly provide timespan_t gauss_a, gauss_b and gauss_ab.

[other]
* Remove uses of clamp on randomly generated values.

The only class module this touches is Mage. I'm confident this accurately reproduces the intent behind the previous implementation, but would appreciate an extra pair of eyes there.

### DIFF
--- a/engine/class_modules/sc_enemy.cpp
+++ b/engine/class_modules/sc_enemy.cpp
@@ -986,18 +986,12 @@ struct pause_action_t : public action_t
 
   timespan_t execute_time() const override
   {
-    timespan_t duration = sim->rng().gauss( base_execute_time, duration_stddev );
-
-    duration = clamp( duration, duration_min, duration_max );
-
-    return duration;
+    return sim->rng().gauss_ab( base_execute_time, duration_stddev, duration_min, duration_max );
   }
 
   void update_ready( timespan_t /* cd_duration */ ) override
   {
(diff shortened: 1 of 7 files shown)


Dorovon:
I think the clamp cases that were changed to `gauss_b` should be `gauss_ab`, unless I'm misunderstanding something.

renanthera:
Agreed, got the `timespan_t gauss(...)` implementation stuck in my head.

renanthera:
I've updated `timespan_t gauss_b` to guarantee non-negative generated values, along with making intent a bit more explicit in the external uses by instead opting to use gauss_ab.

renanthera:
[b5f6bec](https://github.com/simulationcraft/simc/pull/8673/commits/b5f6becceac884dd6fc49fa9bae16cfe0b439106) addresses CI failures for DungeonSlice fight style.

[4dbce04](https://github.com/simulationcraft/simc/pull/8673/commits/4dbce043aaf542424ae175f0fc07c231eb9bac58) addresses the easier concern Dorovon mentioned in Discord. (expressions sampling duration and cooldown interval instead of using the current value)

Does not address uses of `dynamic_cast` nor strcomp that occurs for current DRoute support.
Likely addressed in a future PR that will overhaul `raid_event` in more depth.

PR #8677, opened 2024-03-02
[Warlock] Pass raw tyrant CD to builds which do not take GFG

Fatherdotter:
Raidbots example: https://www.raidbots.com/simbot/report/e8wW9Giw8J7B9cgxHUbZuS

Minor adjustment, pass raw Tyrant cooldown instead of synced Tyrant timings to builds that are not running GFG. On average about 2-3% gain for non-GFG builds, should have no effect on GFG build behavior.


Fatherdotter:
Tyrant_cd variable is tied to recent Dungeon Slice changes. PR closed until the behavior can be separated.

PR #8683, opened 2024-03-04
[Warlock] Improve dreadstalkers behavior

millanzarreta:
The simc behavior of the [Dreadstalkers](https://www.wowhead.com/spell=104316/call-dreadstalkers) is not completely faithful to the in-game behavior. The main differences could be categorized into the nexts:
- Distance should affect dreadstalkers, since when they are summoned they charge towards the target, which will take longer depending on the distance. Currently in simc dreadstalkers start attacking immediately after being summoned.
- Regardless of this travel time, the first hit of the dreadstalkers has a certain random delay (although this does not apply if they are summoned at melee) that must be added to the travel time. This is not currently being implemented in simc.
- Although the duration of the dreadstalkers is supposedly 12 seconds, the actual duration in-game is extended a bit randomly, and this is not being taken into account in simc. This random extension time is directly related to the delay of the first melee attack delay mentioned above in a not-very-intuitive way. The reason for this relation is unknown, although it is supposed to be related to how the WoW server processes and updates events.

In addition to these differences, until recently in simc the [Demonic Inspiration](https://www.wowhead.com/spell=386858/demonic-inspiration) talent was also incorrectly applied to dreadstalkers (which should only affect the main pet), increasing its attack speed by 5%. This was already corrected several weeks ago in [this other PR](https://github.com/simulationcraft/simc/pull/8656).

All of these factors affect the number of melee attacks dreadstalkers do in each summon, which is quite relevant for simulations in ST and even in AoE due to [Dreadlash](https://www.wowhead.com/spell=264078/dreadlash), [Carnivorous Stalkers](https://www.wowhead.com/spell=386194/carnivorous-stalkers), and [The Houndmaster's Stratagem](https://www.wowhead.com/spell=267170/the-houndmasters-stratagem) talents.

With this PR, multiple changes are made to the dreadstalkers, adding travel time, the random delay of the first melee attack, and the random extra duration, so they behave similarly to how they do in-game. Although in-game at a distance of 5yd the dreadstalkers still charge, since this is the minimum distance value that the user can set (at least for the warlock) in simc, then for distance values less than or equal to 5yd it will be considered that dreadstalkers are summoned at melee range.

### DIFF
--- a/engine/class_modules/warlock/sc_warlock.hpp
+++ b/engine/class_modules/warlock/sc_warlock.hpp
@@ -123,6 +123,7 @@ public:
   int incinerate_last_target_count; // For use with T30 Destruction tier set
   double shadow_invocation_proc_chance; // 2023-09-10: Annoyingly, at this time there is no listed proc chance in data for Shadow Invocation
   double doom_brand_accumulator;
+  timespan_t dreadstalker_first_attack_delay_rng; // Used to set the same rng delay for the first melee auto-attack of the last pair of summoned dreadstalkers
   std::vector<event_t*> wild_imp_spawns; // Used for tracking incoming imps from HoG
 
   unsigned active_pets;
--- a/engine/class_modules/warlock/sc_warlock_demonology.cpp
+++ b/engine/class_modules/warlock/sc_warlock_demonology.cpp
@@ -502,8 +502,42 @@ struct call_dreadstalkers_t : public demonology_spell_t
     demonology_spell_t::execute();
 
     unsigned count = as<unsigned>( p()->talents.call_dreadstalkers->effectN( 1 ).base_value() );
+    
+    // 2024-03-03: The dreadstalker summon has an extra random duration that will be calculated on each summon
+    timespan_t extra_duration = 0_ms;
+    
+    // At range this value is also used to calculate the dreadstalkers' extra duration, although in these cases it is not a hard maximum
+    const timespan_t max_extraduration_sync_ts = 820_ms;
 
-    auto dogs = p()->warlock_pet_list.dreadstalkers.spawn( p()->talents.call_dreadstalkers_2->duration(), count );
+    const double dist = p()->get_player_distance( *target );
+    // The minimum distance a user can set for the player is 5yd, which in simc is usually considered melee range
+    if (dist >= 5.5) {
+      // 2024-03-03: When the summon is cast at range, there will be a certain delay in the dreadstalkers' first melee attack
+      // This delay is independent of travel time (will be added to this), which will be taken into account within the dreadstalker pet class itself
+      // The delay of the dreadstalkers' first melee varies in a random range of approximately 30ms-990ms (last checked 2024-03-03)
+      p()->dreadstalker_first_attack_delay_rng = timespan_t::from_millis( rng().range( 30.0, 990.0 ) );
+
+      // 2024-03-03: The extra duration of dreadstalkers, although random within a range, is closely related to the delay of the previous first melee delay
+      // It is unknown why these values are related, although it is assumed that it has to do with how the WoW server internally processes the timings of these events
+
+      // There is a threshold (depending on the distance, varying 24_ms per yard approx) from which the extra duration can be related with first melee delay in a way that is quite close to the real in-game behavior
+      double part_threshold_dblval = 560.0 - std::min(dist, 40.0) * 24.0;
+      while(part_threshold_dblval < 0.0) part_threshold_dblval += as<double>(max_extraduration_sync_ts.total_millis());
+      const timespan_t part_threshold_ts = timespan_t::from_millis( rng().gauss( part_threshold_dblval, 25.0) );
+      if (p()->dreadstalker_first_attack_delay_rng < part_threshold_ts) {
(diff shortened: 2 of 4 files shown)


millanzarreta:
Let's expand the info about this topic with more detail in this comment:

This is a figure (made from someone in warlock discord some time ago before all the changes, including [the Demonic Inspiration PR](https://github.com/simulationcraft/simc/pull/8656)) where the DPS per secondary stat point is displayed:
imagen

While the ``Crit``, ``Versa`` and ``Mastery`` behave as expected, the ``Haste`` value shows a large variation, with peaks and valleys. It is assumed that this variation is related in some strange ways to various thresholds related to pets. The highest peaks of Haste (marked with a purple circle) are directly related to the number of melee attacks the dreadstalkers do in each summon. Using an old version of simc and a custom APL (to prevent the use of Tyrant and the summon of Dreadstalker at the end of fight) we obtain the old number of melee hits of the dreadstalkers for each invocation, getting the following figure:
imagen

It can be seen that according to the old simulation, an additional hit was obtained at approximately 12% haste, 27% haste and 43% haste. The 27% haste threshold is especially well known, which commonly produced simulations with more DPS due to this supposed extra attack of the dreadstalker, showing how important it is to more accurately simulate the real behavior of dreadstalkers.

After [the PR fixing Demonic Inspiration](https://github.com/simulationcraft/simc/pull/8656) was carried out a few weeks ago, an error was fixed in which 5% more attack speed was incorrectly provided to dogs. After this, the figure of the number of attacks per invocation is the following, where these threshold values are simply shifted (12% to 17%, 25% to 33%, ...):
imagen

***

Anyway, we need to extract the actual behavior of the dreadstalkers from the game to be able to replicate it in sims. In order to extract the in-game behavior of the dreadstalkers, an extensive battery of tests had to be carried out. Values between 0% and 44% of haste were selected (varying by 1% in each interval), for which tests were carried out at 0yd, 10yd, 15yd, 20yd, 25yd, 30yd, 35yd, 40yd. At least 100 dreadstalkers summon calls were made for each case. This is enough to have relatively good precision for building a figure and extracting data:
imagen

Although some precision errors can be seen in the figure, a good overall behaviour of the dreadstalkers' in-game behavior is obtained. The tests were carried out with training dummies in Orgrimmar, and as usually happens in WoW, behavior can vary for NPCs with different hitbox sizes even if you are at the same range. The following [WA](https://wago.io/7U5xcYCLm) was used to obtain both general and detailed data during testing. It may be interesting in the future to expand the number of samples (and do so for intervals smaller than 1% haste) to add greater precision, especially for the points near to the haste breakpoint values.

***

The travel time has been extracted from this obtained data and can also be verified by analyzing the time and frames of the [following video](http://youtu.be/70uXXYXO3Mg). This travel speed is not constant, but rather has a certain acceleration, so the average speed is different for each distance. A simple solution was chosen where the average speeds were obtained for some distances [0yd, 10yd, 15yd, 20yd, 25yd, 30yd, 35yd, 40yd] and the others were interpolated, storing them in a lookup table.

From the extracted data was deduced that after the dogs are summoned and they travel to melee range, there is then a random delay (in the approximately [0-1] second range) to do their first attack. This delay does not happen if they are summoned at melee range, and as expected, there is also no travel time if they are summoned at melee range. In simc the minimum distance value the user can set is 5yd (at least for warlocks), so we consider that 5yd is melee range concerning to dreadstalkers' behavior.

Although the duration of the dogs is theoretically 12 seconds, in the game this time is actually a little longer randomly (this also happens with several other non-controllable summoned pets). Although this extra time appears to be in a random range between approximately [0.0-0.82] seconds, there is a strange relationship between this extension and the aforementioned delay time. It is not known for sure why this is like this, but it is supposed to be related to how the WoW server processes and updates events. In any case, through the large amount of data extracted, it has been possible to recreate this extra time and its relationship with the delay of the first attack quite accurately.

This delay and extra duration time are (almost) always the same for the two dogs, that arise, travel, attack and demise at the same timestamps in a synchronized manner.

Finally, it remains to carry out tests with the new changes to check if the new behavior is similar to in-game observations. The following figures show for different distances the results of the simulations (solid line) concerning those observed in-game (dash-dotted line), and it can be seen how the results are quite similar:
__- 0yd:__
imagen
__- 10yd:__
imagen
__- 15yd:__
imagen
__- 20yd:__
imagen
__- 25yd:__
imagen
__- 30yd:__
imagen
__- 35yd:__
imagen
__- 40yd:__
imagen

millanzarreta:
At last, I leave in this comment the files with the data obtained with the in-game test battery.

[wl_demonology_dreadstalkers_overall_data_03032024.xlsx](https://github.com/simulationcraft/simc/files/14485094/wl_demonology_dreadstalkers_overall_data_03032024.xlsx)
[wl_demonology_dreadstalkers_detailed_data_03032024.xlsx](https://github.com/simulationcraft/simc/files/14485095/wl_demonology_dreadstalkers_detailed_data_03032024.xlsx)

Azevara:
Closing this PR as a more flexible skeleton for handling the same issues was implemented in #8698 . Please consider reviewing those changes and suggesting modifications to the added functionality as a new PR for areas where improvement in matching to the data can be found.

PR #8712, opened 2024-03-25
[Warlock] 30 yards default

Facenomore23:
Changed default distance to target from 40 to 30 yards.

Need to change the wiki too : https://github.com/simulationcraft/simc/wiki/Characters#optional


Azevara:
Updated distance description to note that warlock is now 30 yards.

PR #8721, opened 2024-03-28
[Warlock] Dreadstalkers leap and melee attacks fixes and improvements

millanzarreta:
Some modifications as continuation of [#8683](https://github.com/simulationcraft/simc/pull/8683) and [#8698](https://github.com/simulationcraft/simc/pull/8698).

It is intended to correct some errors of [#8698](https://github.com/simulationcraft/simc/pull/8698) PR:
- ``position()`` function is used incorrectly to try to get the distance, but that function returns the position relative to the enemy (front, back, ranged_front, ranged_back), not the distance range.
- The distance check to establish the ``melee_on_summon`` flag must be done at the beginning of ``dreadstalker_t::arise()`` (before the ``warlock_pet_t::arise()`` call), not at the end.
- The melee auto in ``dreadstalker_leap_t`` should only be cancelled/scheduled if the ``melee_on_summon`` flag is ``false``, since otherwise melee auto-attacks are performed normally just after summon, not after the leap.

Additionally, the following change has been added to this PR:
- Now at distance <= 5yd the dreadstalker summon is considered at melee range (the melee auto-attacks starts immediately), instead of 1yd. The 5yd was selected because is the minimum that the user can establish in simc, and is usually considered melee range in simc.

### DIFF
--- a/engine/class_modules/warlock/sc_warlock_pets.cpp
+++ b/engine/class_modules/warlock/sc_warlock_pets.cpp
@@ -1570,7 +1570,9 @@ struct dreadstalker_leap_t : warlock_pet_t::travel_t
 
   void schedule_execute( action_state_t* s ) override
   {
-    debug_cast<warlock_pet_t*>( player )->melee_attack->cancel();
+    if (!debug_cast<warlock_pet_t*>( player )->melee_on_summon) {
+      debug_cast<warlock_pet_t*>( player )->melee_attack->cancel();
+    }
 
     warlock_pet_t::travel_t::schedule_execute( s );
   }
@@ -1580,10 +1582,12 @@ struct dreadstalker_leap_t : warlock_pet_t::travel_t
     warlock_pet_t::travel_t::execute();
 
     // There is an observed delay of up to 1 second before a melee attack begins again for pets after a movement action like the leap (possibly server tick?)
-    make_event( sim, debug_cast<dreadstalker_t*>( player )->server_action_delay, [ this ]{
-      debug_cast<warlock_pet_t*>( player )->melee_attack->reset();
-      debug_cast<warlock_pet_t*>( player )->melee_attack->schedule_execute();
-    } );
+    if (!debug_cast<warlock_pet_t*>( player )->melee_on_summon) {
+      make_event( sim, debug_cast<dreadstalker_t*>( player )->server_action_delay, [ this ]{
+        debug_cast<warlock_pet_t*>( player )->melee_attack->reset();
+        debug_cast<warlock_pet_t*>( player )->melee_attack->schedule_execute();
+      } );
+    }
   }
 };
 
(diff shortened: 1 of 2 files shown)


engine/class_modules/warlock/sc_warlock_pets.cpp
```diff
@@ -1570,7 +1570,9 @@ struct dreadstalker_leap_t : warlock_pet_t::travel_t
 
   void schedule_execute( action_state_t* s ) override
   {
-    debug_cast<warlock_pet_t*>( player )->melee_attack->cancel();
+    if (!debug_cast<warlock_pet_t*>( player )->melee_on_summon) {
```
Azevara: This change is incorrect - any leap should cancel melee attacks as they should not resume until after landing.

millanzarreta: In the case that the ``melee_on_summon`` flag is ``true`` the auto-attack begins immediately after being summoned, and in this case the leap should neither cancel nor start new auto-attacks. Actually in this case the leap would behave more like the parent ``travel_t``, it would simply serve to move the pet to melee range (in ``travel_t`` parent class there is no ``melee_attack->cancel()``).

Actually, the most correct thing would be not to do the leap if the ``melee_on_summon`` flag is ``true``, because in reality no leap occurs, but in my understanding is not possible to not execute the leap only in these cases. So, by checking that ``melee_on_summon`` flag and preventing the ``melee_attack->cancel()`` (and also the schedule of a new one) we ensure that the behavior of the auto-attack is not modified when it is summoned at melee, and the "leap" (acting like a "travel") would only serve to move the pet into melee.

Azevara: Apologies for not returning to this sooner due to illness. The appropriate way to control the leap behavior is to either adjust the action list string for the dreadstalker to put a condition on the leap, or to adjust the ready() method of the leap action so that it does not execute under certain circumstances (such as being too close).

In general, the leap action should always cancel auto attacks as it is a generic travel action that we do not observe auto attacks during. If there are scenarios in which the leap action should not take place, those should be controlled by the alternate means mentioned above.

millanzarreta: Hope everything goes well. Taking into account what you said I have made some changes. Now the ``action_list`` of dreadstalkers includes both the leap and the normal travel of the pets (``"leap/travel/dreadbite"`` instead of ``"leap/dreadbite"``). In the ``leap`` action class in the ``ready()`` method is now where it is checked if the leap is made (the ``melee_on_summon`` flag check). In addition, I have also added a variable counter (similar to how they are done for other actions) to be sure that the leap is only performed once. If the leap's ``ready()`` method is ``false``, the usual ``travel`` action would be executed instead. This ``travel`` action would not be executed if the ``leap`` is being performed before, since at the end of the leap the pet is in melee range and this makes the travel's ``ready()`` method ``false``. I've been debugging and testing it and it seems to work correctly, but I'd appreciate it if you could fast-check the changes to see if these agree with your suggestions. Thanks and take care!

PR #8754, opened 2024-04-17
[Monk] Update Generator with New SpellIDs

JeremyMonk:


renanthera:
Verified which spells were picked up automatically and pruned them. The remaining changes are still missing from dbc export.

PR #8811, opened 2024-05-02
[TMI] Remove TMI

renanthera:
Removes TMI.
TMI has been largely ignored for quite a while at this point, and I am not aware of any spec that supports it, nor anyone who advocates its use.
It regularly causes other headaches, so this may be a good time to just remove it altogether.

I removed all mentions of `tmi` and `theck_meloree_index`, while being a bit sensitive to the surrounding code, but there may be other dead codepaths that I have not found as of the initial commit.

As of the initial commit, it compiles and runs for my BrM test cases.


renanthera:
As of the day this PR was authored the previous iteration was tested by Saeldur against Aug and Topple against VDH. At that point, everything worked time.

Since then, I fixed a very minor detail that broke two charts in the HTML report, rebased it against current TWW to verify nothing weird had occurred, and removed whitespace-only changes.

PR #8813, opened 2024-05-04
Update DF4_Death_Knight_Blood food & talents

Thaiax:
Updates talents and food for ~1500 dps increase in standard 5min 1 boss patchwerk


nyterage:
changing these values in the final profile will end up with them just being replaced as soon as the profiles regenerate. they would need to be changed in `\engine\class_modules\apl\apl_death_knight.cpp` to have a permanent effect. 

would also appreciate example result links from raidbots as a sanity check.

PR #8815, opened 2024-05-06
[rogue] Initial implementation for core fatebound talents

weswigham:
Aside from the talents still `[NYI]` in-game, I also have yet to implement a bug in how Edge Case or Double Jeopardy combine - mostly because I wanted to write down the inexplicable layering behavior I've been seeing in-game first.

Double Jeopardy, like the text says, makes the first finisher you use *after* you break stealth add two stacks to the coin it rolls instead of one. This is going to trip people up, because it doesn't apply to the finisher you break stealth with, like outlaw often does with crackshot now, but the finisher *after* that. Usually.* We'll get to that asterisk.

Edge case, like the text says, makes your next finisher after you use ADR/deathmark roll both faces, thus getting 1 stack of both coin buffs. And if you never layer it with double jeopardy, that's exactly what it does.

With both talents, when you press ADR/Deathmark and break stealth with a finisher, that finisher gives one stack to both buffs. OK, nothing weird here. Your next finisher, the double jeopardy one, then *also* always rolls 1 stack of both buffs, _instead_ of two stacks of one. The text does not capture this, nor does this behavior fall out from the text, yet... that's how it seems to work.
If you instead break stealth with a builder and then use a finisher, you'll get the expected edge case/double jeopardy layering, where you get two heads and two tails stacks from a single finisher.

Not sure if that behavior is worth simming yet. It definitely seems buggy, since it's not what you'd expect to happen from any of the talent text.

### DIFF
--- a/engine/class_modules/sc_rogue.cpp
+++ b/engine/class_modules/sc_rogue.cpp
@@ -332,6 +332,10 @@ public:
       actions::rogue_attack_t* deathmark_garrote = nullptr;
       actions::rogue_attack_t* deathmark_rupture = nullptr;
     } sanguine_blades;
+    struct {
+      actions::rogue_attack_t* fatebound_coin_tails = nullptr;
+      actions::rogue_attack_t* fatebound_coin_tails_delivered = nullptr;
+    } fatebound;
   } active;
 
   // Autoattacks
@@ -386,6 +390,14 @@ public:
     buff_t* sepsis;
     buff_t* subterfuge;
     buff_t* thistle_tea;
+
+    // Hero
+    // Fatebound
+    damage_buff_t* fatebound_coin_heads;
+    buff_t* fatebound_coin_tails;
+    stat_buff_t* fatebound_lucky_coin;
+    buff_t* edge_case; // not a buff in-game, but useful to track as a buff
+    buff_t* double_jeopardy; // not a buff in-game, but useful to track as a buff
     
     // Assassination
     buff_t* blindside;
@@ -518,6 +530,7 @@ public:
     gain_t* shrouded_suffocation;
     gain_t* the_first_dance;
     gain_t* t31_subtlety_4pc;
+    gain_t* deal_fate;
 
   } gains;
 
@@ -560,6 +573,12 @@ public:
     const spell_data_t* shadowstep_buff;
     const spell_data_t* subterfuge_buff;
     const spell_data_t* vanish_buff;
+
+    // Hero Spells
+    const spell_data_t* fatebound_coin_heads_buff;
+    const spell_data_t* fatebound_coin_tails_buff;
+    const spell_data_t* fatebound_coin_tails;
+    const spell_data_t* fatebound_lucky_coin_buff;
 
   } spell;
 
@@ -952,6 +971,29 @@ public:
 
     } shared;
 
+    struct fatebound_talents_t
+    {
+      player_talent_t hand_of_fate;
+
+      player_talent_t chosens_revelry;
+      player_talent_t tempted_fate;     // TODO: Defensive
+      player_talent_t mean_streak;
+      player_talent_t inexorable_march; // NYI in-game
+      player_talent_t deaths_arrival;   // NYI in-game
+
+      player_talent_t deal_fate;
+      player_talent_t edge_case;        // TODO: Double jeopardy + edge case stealth break overlap bug
+      player_talent_t threads_of_fate;  // NYI in-game
+
+      player_talent_t delivered_doom;
+      player_talent_t inevitability;    // NYI in-game
+      player_talent_t destiny_defined;  // TODO: Outlaw also gets the poison proc rate the text says is for assa? Verify in-game.
+      player_talent_t double_jeopardy;  // TODO: Double jeopardy + edge case stealth break overlap bug
+
+      player_talent_t fateful_ending;   // TODO: Add tertiary stats
+
+    } fatebound;
+
   } talent;
 
   // Masteries
@@ -1203,6 +1245,8 @@ public:
     if ( set_bonuses.t31_outlaw_2pc->ok() )
       proc_chance += set_bonuses.t31_outlaw_2pc->effectN( 1 ).percent();
     proc_chance += buffs.skull_and_crossbones->stack_value();
+    if ( talent.fatebound.destiny_defined->ok() )
+      proc_chance += talent.fatebound.destiny_defined->effectN( 2 ).percent();
     return proc_chance;
   }
 
@@ -1640,6 +1684,8 @@ public:
     ab::apply_affecting_aura( p->talent.subtlety.secret_stratagem );
     ab::apply_affecting_aura( p->talent.subtlety.dark_brew );
 
+    ab::apply_affecting_aura( p->talent.fatebound.destiny_defined );
+
     // Dynamically affected flags
     // Special things like CP, Energy, Crit, etc.
     affected_by.shadow_blades_cp = ab::data().affected_by( p->talent.subtlety.shadow_blades->effectN( 2 ) ) ||
@@ -1818,6 +1864,7 @@ public:
     register_damage_buff( p()->buffs.summarily_dispatched );
     register_damage_buff( p()->buffs.symbols_of_death );
     register_damage_buff( p()->buffs.the_rotten );
+    register_damage_buff( p()->buffs.fatebound_coin_heads );
 
     register_damage_buff( p()->buffs.t29_assassination_4pc );
     register_damage_buff( p()->buffs.t29_outlaw_2pc );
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_rogue.cpp
```diff
@@ -7878,8 +7979,11 @@ void actions::rogue_action_t<Base>::trigger_seal_fate( const action_state_t* sta
   if ( !p()->rng().roll( p()->talent.assassination.seal_fate->effectN( 1 ).percent() ) )
     return;
 
-  trigger_combo_point_gain( as<int>( p()->talent.assassination.seal_fate->effectN( 2 ).trigger()->effectN( 1 ).base_value() ),
-                            p()->gains.seal_fate );
+  auto gain = as<int>( p()->talent.assassination.seal_fate->effectN( 2 ).trigger()->effectN( 1 ).base_value() );
+  if ( procs_deal_fate() && p()->talent.fatebound.deal_fate->ok() ) {
+    gain += as<int>( p()->talent.fatebound.deal_fate->effectN( 1 ).base_value() );
```
weswigham: Reminder for myself: this should probably be a secondary `trigger_combo_point_gain` call, so the gain is tracked separately from `Seal Fate` in the stats, like it is for outlaw's version.

engine/class_modules/sc_rogue.cpp
```diff
@@ -10072,6 +10268,12 @@ void rogue_t::init_spells()
   spell.subterfuge_buff = ( talent.rogue.subterfuge->ok() || talent.outlaw.underhanded_upper_hand->ok() ) ? find_spell( 115192 ) : spell_data_t::not_found();
   spell.vanish_buff = spell.vanish->ok() ? find_spell( 11327 ) : spell_data_t::not_found();
 
+  // Hero Talent Background Spells
+  spell.fatebound_coin_heads_buff = talent.fatebound.hand_of_fate->ok() ? find_spell( 452923 ) : spell_data_t::not_found();
+  spell.fatebound_coin_tails_buff = talent.fatebound.hand_of_fate->ok() ? find_spell( 452917 ) : spell_data_t::not_found();
```
weswigham, line 10342: Oh, an aside: The spell data for these says they have a 3600 second expiry - an hour. This is not shown in-game, and I have yet to idle with one for an hour to see if it actually disappears. Something to check later, I suppose. It won't matter for any sims - an hour is effectively forever for a sim.

engine/class_modules/sc_rogue.cpp
```diff
@@ -10072,6 +10268,12 @@ void rogue_t::init_spells()
...
+  spell.fatebound_coin_tails = talent.fatebound.hand_of_fate->ok() ? find_spell( 452538 ) : spell_data_t::not_found();
```
weswigham, line 10343: `simc`'s automatic buff tracking let me know this is being affected by `Lethality`'s crit damage bonus right now. No idea why, since it's not a builder. Just something to be aware of, I guess.

engine/class_modules/sc_rogue.cpp
```diff
@@ -10675,6 +10886,22 @@ void rogue_t::create_buffs()
...
+        buffs.fatebound_lucky_coin->trigger();
+    } );
+  buffs.fatebound_coin_tails = make_buff( this, "fatebound_coin_tails_buff", spell.fatebound_coin_tails_buff )
+    ->set_stack_change_callback( [this]( buff_t*, int, int new_stacks ) {
+      if ( new_stacks == 7 )
+        buffs.fatebound_lucky_coin->trigger();
+    } );
+  buffs.fatebound_lucky_coin = make_buff( this, "fatebound_lucky_coin_buff", spell.fatebound_lucky_coin_buff )
+    ->set_pct_buff_type( STAT_PCT_BUFF_AGILITY )
+    ->set_default_value_from_effect_type( A_MOD_TOTAL_STAT_PERCENTAGE ); // TODO: Add omni-stat buff type
```
weswigham: Definitely need some direction here - this buff has effects for buffing like 9 different stats by 1%. (agi, stam, crit, vers, mastery, attack speed (? haste?), leech, avoidance, movement speed). I feel like the automatic buff parser *should* be able to handle it, but... I don't know if it does? The summary doesn't display any affected stats on the buff details, anyway, even with the above two lines trying to at least force an agi buff. 🫤

engine/class_modules/sc_rogue.cpp
```diff
@@ -8225,6 +8332,75 @@ void actions::rogue_action_t<Base>::trigger_restless_blades( const action_state_
...
+
+      p()->buffs.fatebound_coin_heads->expire();
+      p()->buffs.fatebound_coin_tails->increment();
+    }
+    return;
+  }
+
+  double odds = 0.5;
+  if (biased) {
+    // TODO: Validate how these stack with the presumed base 50/50 chance and one another
```
weswigham, line 8416: I have no idea if these are additive (with either one another or the base 50/50 odds) or multiplicative. And the only way to know for sure would seem to be a *lot* of log analysis. I think the pessimistic option would be swapping these to be additive with one another but multiplicative with the base 0.5.  So
```suggestion
    // TODO: Validate how these stack with the presumed base 50/50 chance and one another
    double bias = 1.0;
    if (p()->talent.fatebound.mean_streak->ok()) {
      bias += p()->talent.fatebound.mean_streak->effectN( 1 ).percent();
    }
    if (p()->talent.fatebound.destiny_defined->ok()) {
      bias += p()->talent.fatebound.destiny_defined->effectN( 3 ).percent();
    }
    odds *= bias;
```
This works out to be 69% in favor of a match, while the current stacking is 88% - much higher.

engine/class_modules/sc_rogue.cpp
```diff
@@ -10675,6 +10886,22 @@ void rogue_t::create_buffs()
...
+  // Hero
+  // Fatebound
+  buffs.fatebound_coin_heads = make_buff<damage_buff_t>( this, "fatebound_coin_heads_buff", spell.fatebound_coin_heads_buff );
+  buffs.fatebound_coin_heads->set_stack_change_callback( [this]( buff_t*, int, int new_stacks ) {
+      if ( new_stacks == 7 )
+        buffs.fatebound_lucky_coin->trigger();
+    } );
+  buffs.fatebound_coin_tails = make_buff( this, "fatebound_coin_tails_buff", spell.fatebound_coin_tails_buff )
+    ->set_stack_change_callback( [this]( buff_t*, int, int new_stacks ) {
+      if ( new_stacks == 7 )
```
weswigham: 🤦Reminder to self: This needs to also check for the capstone talent's presence.

engine/class_modules/sc_rogue.cpp
```diff
@@ -1203,6 +1243,8 @@ class rogue_t : public player_t
     if ( set_bonuses.t31_outlaw_2pc->ok() )
       proc_chance += set_bonuses.t31_outlaw_2pc->effectN( 1 ).percent();
     proc_chance += buffs.skull_and_crossbones->stack_value();
+    if ( talent.fatebound.destiny_defined->ok() )
+      proc_chance += talent.fatebound.destiny_defined->effectN( 1 ).percent();
```
weswigham: Effect 1 is clearly the poison proc rate, and effect 2 (and 3!) is also a value 5 effect with a dummy label and server side script tag. I should probably operate on the assumption effect 2 is supposed to be the opportunity proc rate addition, rather than reusing effect 1 here (and that effect 3 is the bonus coinflip matching rate).

engine/class_modules/sc_rogue.cpp
renanthera: I'd recently talked to Gas about this, actually.

You can invoke multiple `set_pct_buff_type`'s on a single buff, assuming they all provide the same value. If they don't provide the same value, you're out of luck.

It will end out registering the invalidation for each of the mapped `pct_buff_type` caches and add it to the matching buff vector for that `pct_buff_type`, as seen here:

https://github.com/simulationcraft/simc/blob/dbea3b61ccbdd95bf9e8ac82409872eccb19cd05/engine/buff/buff.cpp#L1143-L1154

Here's two examples of that being used in that way:

https://github.com/simulationcraft/simc/blob/dbea3b61ccbdd95bf9e8ac82409872eccb19cd05/engine/player/player.cpp#L4143-L4155

weswigham: Hm, I guess that works here since all the buffs are 1% - but there's no PCT_BUFF enum members for the 3 tertiary stats. Should I work on adding them? Or is there a different method for handling movement speed/avoidance/leech buffs?

renanthera: I personally wouldn't bother, as many of those effects aren't [entirely] implemented and don't have an impact on any results you are likely to care about. You could if you'd like, though.

PR #8823, opened 2024-05-08
[Warlock] Better adjust the dreadstalkers values and calculations to ingame behaviour

millanzarreta:
This PR is the last of several modifications that aim to improve the behavior of the demonology warlock dreadstalkers to adjust them more closely to real ingame behavior.

In the original PR #8683 all the intended modifications are discussed extensively, as well as all the documentation on the data extracted from ingame behaviour. That PR was closed without merge because a more flexible skeleton was developed in other PR (#8698), but partially implementing only some modifications and leaving others for future PRs. In the most recent PR #8721 some errors from the previous PR (#8698) were fixed and some improvements were added.

The last pending thing (that is intended to be resolved in this PR) is to better adjust the ingame behavior of the extra duration of the dreadstalkers and their leap speed:
- As for the extra duration, although the relation between the first auto delay and the extra duration was already implemented, it is also necessary to relate it to the distance and establish the maximum extra duration time as observed ingame. Based on data collected from the game, the maximum despawn extra duration for dreadstalkers is about ``820ms``. The initial offset point between the extra duration and the first auto delay is ``260ms``, increasing ``24ms`` with each yard. There is also some variance in the delay-extra_duration relation that can be modeled fairly closely using a normal distribution.
- As for the speed of the leap, it is not constant, and since there is a certain acceleration, its mean value is dependant on distance. As it is difficult to obtain accurate values regarding this acceleration, a simple solution was chosen instead, where the average speeds were obtained for some distances ``[0yd, 5yd, 10yd, 15yd, 20yd, 25yd, 30yd, 35yd, 40yd]`` and the others were interpolated, storing them in a lookup table. In the leap action, the appropriate speed will be obtained according to its distance on each occasion. How to extract these speeds from the game and their results was something already discussed in the original PR (#8683), although this was tested again just in case.

These changes have been implemented in this PR while maintaining the skeleton of the previous PRs.

***

Finally, tests need to be carried out to verify the changes. We will first check the results for the current, without the implementation of this PR (dashed line, labeled as *OLD*), concerning those observed in-game (dash-dotted line, labeled *INGAME*). The y-axis is the total number of dreadstalkers melee hits per summon, the x-axis is the haste amount, incrementing by 1% each step. Each figure shows the results for different distances, from 40yd to 0yd:

__- 40yd (*INGAME* vs *OLD*):__
imagen
__- 35yd (*INGAME* vs *OLD*):__
imagen
__- 30yd (*INGAME* vs *OLD*):__
imagen
__- 25yd (*INGAME* vs *OLD*):__
imagen
__- 20yd (*INGAME* vs *OLD*):__
imagen
__- 15yd (*INGAME* vs *OLD*):__
imagen
__- 10yd (*INGAME* vs *OLD*):__
imagen
__- 0yd (*INGAME* vs *OLD*):__
imagen

We can see that the results only coincide sometimes and are not entirely accurate (although they improved compared to how they were before all the previous PRs modifications), and it is important to highlight that the breakpoints do not coincide exactly on many occasions (this is something that historically had a lot of relevance in the final sim results).

Now we will show the results of the sims with this PR implemented (solid line, labeled as *PR*) concerning those observed in-game (dash-dotted line, labeled *INGAME*), and it can be seen how the results are quite similar:

__- 40yd (*INGAME* vs *PR*):__
imagen
__- 35yd (*INGAME* vs *PR*):__
imagen
__- 30yd (*INGAME* vs *PR*):__
imagen
__- 25yd (*INGAME* vs *PR*):__
imagen
__- 20yd (*INGAME* vs *PR*):__
imagen
__- 15yd (*INGAME* vs *PR*):__
imagen
__- 10yd (*INGAME* vs *PR*):__
imagen
__- 0yd (*INGAME* vs *PR*):__
imagen

The results of the sims with this PR modifications are now much more similar to those observed ingame.

***

Finally, to leave a more general image, this is the figure with the results for all the distances of the data obtained ingame:
imagen

And this would be the same figure but for the results obtained with simc after this PR:
imagen

### DIFF
--- a/engine/class_modules/warlock/sc_warlock_demonology.cpp
+++ b/engine/class_modules/warlock/sc_warlock_demonology.cpp
@@ -503,13 +503,27 @@ struct call_dreadstalkers_t : public demonology_spell_t
 
     unsigned count = as<unsigned>( p()->talents.call_dreadstalkers->effectN( 1 ).base_value() );
 
-    // Set a randomized offset on first melee attacks after travel time. Make sure it's the same value for each dog so they're synced
-    timespan_t delay = rng().range( 0_s, 1_s );
+    timespan_t delay;
+    timespan_t dur_adjust;
 
-    timespan_t dur_adjust = duration_adjustment( delay );
+    const double dist = p()->get_player_distance( *target );
+    // The summon is considered at melee range if the distance is less than or equal to 5 yards
+    if (dist > 5.0)
+    {
+      // Set a randomized offset on first melee attacks after travel time. Make sure it's the same value for each dog so they're synced
+      delay = rng().range( 0_s, 1_s );
+      // Adjust the extra duration, that appear to be offset from the melee attack check in a correlated manner
+      dur_adjust = duration_adjustment( delay, std::min(dist, 40.0) );
+    }
+    else
+    {
+      // There is no delay on the first melee attack when summoned from melee
+      delay = 0_ms;
+      // In this case the extra duration of the dreadstalkers can be assumed random between the minumum (0ms) and the maximum (820ms) (last tested 2024-05-08)
+      dur_adjust = timespan_t::from_millis( rng().range( 0.0, 820.0 ) );
+    }
 
     auto dogs = p()->warlock_pet_list.dreadstalkers.spawn( p()->talents.call_dreadstalkers_2->duration() + dur_adjust, count );
-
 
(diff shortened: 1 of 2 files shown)


Azevara:
Closing as this was for the Dragonflight branch and a new PR is required for The War Within due to underlying module changes.

PR #8841, opened 2024-05-13
[Monk] Shado-Pan Hero Tree

JeremyMonk:
- Against all Odds
- Efficient Training
- Flurry Strikes
- Martial Precision
- One vs Many
- Pride of Pandaria


engine/class_modules/monk/sc_monk.cpp
```diff
@@ -568,6 +598,24 @@ void monk_action_t<Base>::impact( action_state_t *s )
 
     if ( !ab::result_is_miss( s->result ) && s->result_amount > 0 )
     {
+      if ( p()->talent.shado_pan.flurry_strikes->ok() )
+      {
+        double damage_contribution = s->result_amount;
+
+        if ( p()->talent.shado_pan.one_versus_many->ok() && ( ab::data().id() == 117418 || ab::data().id() == 121253 ) )
+          damage_contribution *= 2.0f;
```
Hinalover: I would actually utilize the spell effect for this. The effect has a base value of 100, so I would doing
1 + p()->talent.shado_pan.one_versus_many->effect(1).percent();

JeremyMonk: Missed that one, fixed!

PR #8850, opened 2024-05-15
[stagger] Stagger mixin

renanthera:
Spinning off stagger so it can be more readily used in other class modules. Demon Hunter (Fel-Scarred Set Fire to the Pain) and Preservation Evoker (Time Dilation) are two known near-copies of the stagger subsystem that exists in current or near-future endgame content.

Early WIP, seeking comments on the following:
- Source file location (in `engine/class_modules/monk/` or in `engine/player/mixins/` are my current thoughts)
- Design it as a decorator class (ala `parse_<...>_effects`) or a mixin.
  * Decorator would be far less "DIY" (very easy to get working), but potential for confusing effect ordering exists.
  * Mixin requires more modifications to the class module to get it working, but very clear effect ordering.
  
Notes exist at the top of `sc_stagger.hpp` w.r.t other details and implementation required before complete.


renanthera:
TODO: Rebase when #8848 is merged.

renanthera:
At this point, I'm happy with this first revision.

There are still some outstanding changes I'd like to make:
- Redesign away from templates (maybe, I'm still thinking about it. there are some interface requirements that are best mandated via templates (?))
- Stagger Guard Absorb support.
- Improve debug messaging.
- Clean up and finish documentation.

PR #8855, opened 2024-05-20
[rogue] TWW Outlaw APL Improvements

weswigham:
1. Now that we don't get shadow dance, holding vanish for ADR is a substantial dps improvement.
3. When I removed shadow dance from the APL, I accidentally left part of the condition on KiR around it and prevented KiR from being used unless you had 6 buffs.


soloxcx:
for now, you can move the stealth prepull action above SnD, just so the coin flip from SnD doesn't prevent the actor from using stealth

weswigham:
@soloxcx I've done a bit better - `stealth` is now `precombat_seconds`-able (and cd tracked), _cancellable_ during precombat, and hand of fate now assumes you won't mess up and target something during precombat. If the precombat finisher is slice and dice, anyway. And the APL opener makes use of that.

PR #8865, opened 2024-05-22
[rogue] Fix precombat stealth cd reduction, improve fatebound fake+real durationless buff reporting

weswigham:
I missed skipping passing the options to parse into the base class ctor (oops - there really should be an assert if you add an option after parsing is already done!) and decided to swap the storage of the option to timespan_t while I was there - both double and timespan are in use in the rogue module, but I think timespan is probably better. While I was debugging this, I realized the example sequences weren't showing the fatebound buffs, even if they were in the buff aggregate table - I guess that's as good a reason as any to set the never constant flag.

### DIFF
--- a/engine/class_modules/sc_rogue.cpp
+++ b/engine/class_modules/sc_rogue.cpp
@@ -6353,13 +6353,13 @@ struct vanish_t : public rogue_spell_t
 
 struct stealth_t : public rogue_spell_t
 {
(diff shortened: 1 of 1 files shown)


weswigham:
I'm... at least somewhat sure a segfault in the demon hunter module isn't caused by this PR. Probably.

PR #8883, opened 2024-06-03
[Outlaw] War Within APL cleanup

soloxcx:
Merges recent changes from DF branch, removes single target Blade Flurry, removes very old tier set rules like T30, updates Keep it Rolling usage to initially promising condition after losing T31.

tagging @weswigham

### DIFF
--- a/engine/class_modules/apl/rogue/outlaw.simc
+++ b/engine/class_modules/apl/rogue/outlaw.simc
@@ -126,5 +126,5 @@ actions.stealth_cds=vanish,if=talent.crackshot&variable.finish_condition&(buff.a
 # Hidden Opportunity builds without Crackshot use Vanish if Audacity is not active and when under max Opportunity stacks
 actions.stealth_cds+=/vanish,if=!talent.crackshot&talent.hidden_opportunity&!buff.audacity.up&buff.opportunity.stack<buff.opportunity.max_stack&variable.ambush_condition
 # Builds without Crackshot and Hidden Opportunity use Vanish when not at finish condition to trigger Double Jeopardy, or to trigger Take 'em by Surprise
-actions.stealth_cds+=/vanish,if=!talent.crackshot&!talent.hidden_opportunity&(!variable.finish_condition|!talent.double_jeopardy)&(!buff.take_em_by_surprise.up|!talent.take_em_by_surprise)
+actions.stealth_cds+=/vanish,if=!talent.crackshot&!talent.hidden_opportunity&(!variable.finish_condition&talent.double_jeopardy|!buff.take_em_by_surprise.up&talent.take_em_by_surprise)
 actions.stealth_cds+=/shadowmeld,if=variable.finish_condition&!cooldown.vanish.ready
--- a/engine/class_modules/apl/rogue/outlaw.simc
+++ b/engine/class_modules/apl/rogue/outlaw.simc
@@ -7,7 +7,7 @@ actions.precombat+=/food
 actions.precombat+=/snapshot_stats
 actions.precombat+=/blade_flurry,precombat_seconds=3,if=talent.underhanded_upper_hand
 actions.precombat+=/stealth,precombat_seconds=2
-# Cancels the stealth we just set up to trigger double jeopardy
+# Cancel Stealth to activate Double Jeopardy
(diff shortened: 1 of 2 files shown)


engine/class_modules/apl/rogue/outlaw.simc
```diff
@@ -123,7 +124,7 @@ actions.stealth+=/ambush,if=talent.hidden_opportunity
 actions.stealth_cds=variable,name=vanish_opportunity_condition,value=talent.fan_the_hammer.rank+talent.quick_draw+talent.audacity<talent.count_the_odds+talent.keep_it_rolling
 # Hidden Opportunity builds without Crackshot use Vanish if Audacity is not active and when under max Opportunity stacks
 actions.stealth_cds+=/vanish,if=talent.hidden_opportunity&!talent.crackshot&!buff.audacity.up&(variable.vanish_opportunity_condition|buff.opportunity.stack<buff.opportunity.max_stack)&variable.ambush_condition
-# Crackshot builds or builds without Hidden Opportunity use Vanish at finish condition and hold for ADR if they have UHUH
-actions.stealth_cds+=/vanish,if=(!talent.hidden_opportunity|talent.crackshot)&variable.finish_condition&(!talent.underhanded_upper_hand|buff.adrenaline_rush.up)
+# Crackshot builds or builds without Hidden Opportunity use Vanish at finish condition and hold for ADR if they have Underhanded Upperhand + Without a Trace
+actions.stealth_cds+=/vanish,if=(!talent.hidden_opportunity|talent.crackshot)&variable.finish_condition&(!talent.underhanded_upper_hand&!talent.without_a_trace|buff.adrenaline_rush.up)
 # Crackshot builds use Dance at finish condition
```
weswigham: ...maybe we should delete this comment.

engine/class_modules/apl/rogue/outlaw.simc
```diff
@@ -123,7 +124,7 @@ actions.stealth+=/ambush,if=talent.hidden_opportunity
 actions.stealth_cds=variable,name=vanish_opportunity_condition,value=talent.fan_the_hammer.rank+talent.quick_draw+talent.audacity<talent.count_the_odds+talent.keep_it_rolling
 # Hidden Opportunity builds without Crackshot use Vanish if Audacity is not active and when under max Opportunity stacks
 actions.stealth_cds+=/vanish,if=talent.hidden_opportunity&!talent.crackshot&!buff.audacity.up&(variable.vanish_opportunity_condition|buff.opportunity.stack<buff.opportunity.max_stack)&variable.ambush_condition
-# Crackshot builds or builds without Hidden Opportunity use Vanish at finish condition and hold for ADR if they have UHUH
-actions.stealth_cds+=/vanish,if=(!talent.hidden_opportunity|talent.crackshot)&variable.finish_condition&(!talent.underhanded_upper_hand|buff.adrenaline_rush.up)
+# Crackshot builds or builds without Hidden Opportunity use Vanish at finish condition and hold for ADR if they have Underhanded Upperhand + Without a Trace
```
weswigham: Hm, thinking about it, vanish on finish condition just seems wrong for anything not crackshot now.

Builds without crackshot probably don't want to vanish at finish condition, at least for fatebound. The double jeopardy buff will only apply *after* the stealth break, not on the skill that breaks it. So if you want to vanish for double jeopardy value, you need to break with a builder for the following finisher to benefit.

Realistically, a non-crackshot build just doesn't need to vanish. I guess you can vanish for ambush value if you're still on HO, but that should also be vanish on build, not on finish.

soloxcx: Yeah I've not explored Vanish for Double Jeopardy but there is surely something to be found for non-Crackshot builds. I guess since the Tails coin is so strong it could be used to further Tails stacks that we already have instead of just forcing Tails to appear from nothing. Perhaps we could sit on Vanish if you only have a Heads coin unless we hit max charges.

soloxcx:
Attempted to be as exhaustive as possible with Vanish rules. In my testing, just not using Vanish at finish condition is enough for Double Jeopardy (+0.4% dps or so). Not had much success trying to hold Vanishes for the Tails coin. 

Builds without Crackshot, Hidden Opportunity, Double Jeopardy, and Take 'em by Surprise don't use Vanish at all.

PR #8906, opened 2024-06-13
[Rogue] Add TWW default consumables

weswigham:
These are basically all cribbed from the DK module for now, but they're also probably just the right picks in the end, maybe excepting the temporary weapon enchant. I did run a quick
```
copy="algari mana oil"
temporary_enchant=main_hand:algari_mana_oil_3/off_hand:algari_mana_oil_3

copy="bubbling wax"
temporary_enchant=main_hand:bubbling_wax/off_hand:bubbling_wax

copy="ironclaw whetstone"
temporary_enchant=main_hand:ironclaw_whetstone_3/off_hand:ironclaw_whetstone_3

copy="oil of deep toxins"
temporary_enchant=main_hand:oil_of_deep_toxins_3/off_hand:oil_of_deep_toxins_3
```
on my local testing profile, and the mana oil did win. [Bubbling wax](https://www.wowhead.com/beta/item=220156/bubbling-wax) in particular is interesting, and doesn't seem to have a functioning effect in simc yet (unlike the oil of deep toxins, which shows up in the damage log, the wax does not).

### DIFF
--- a/engine/class_modules/apl/apl_rogue.cpp
+++ b/engine/class_modules/apl/apl_rogue.cpp
@@ -5,7 +5,8 @@ namespace rogue_apl {
 
 std::string potion( const player_t* p )
 {
-  return ( ( p->true_level >= 61 ) ? "elemental_potion_of_ultimate_power_3" :
+  return ( ( p->true_level >= 71 ) ? "tempered_potion_3" :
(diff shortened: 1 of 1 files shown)


weswigham:
cc @EvanMichaels

PR #8914, opened 2024-06-14
[monk] Reorder ctor parameters.

renanthera:


renanthera:
`actor_target_data_t` needs to be dereferenced twice, for some reason, but beyond that ctor order is consistent.

I somehow broke most of BrM, I shall investigate tomorrow.

renanthera:
Turns out returning `true` if stagger is ticking regardless of Purifying Brew being off cooldown is bad.

PR #8942, opened 2024-06-22
Automate text dumps (APL/SpellData) via Github Actions

gastank:


renanthera:
Instead, could we publish these in the form of either packages, or a separate git repository?

The latter was discussed the last time this came up, a few months ago in Discord, and I recall there being some agreement to a separate repository for this sort of thing being more accessible to users looking for APLs as reference material.

PR #9034, opened 2024-07-16
[monk] Monk Review 14

renanthera:
Switch to `effect_mask_t`.
Begin switching to using `CONSUME_BUFF`.


renanthera:
I'll reopen this when there is less prepatch/TWW release time pressure.

PR #9097, opened 2024-07-25
Revert "Revert "[Warlock] Update beta Channel Demonfire to reflect pre-patch note""

Azevara:
Reverts simulationcraft/simc#9096


Azevara:
The hotfix to a split build between beta and pre-patch is so much fun. Further testing found that in the live game, this is in fact true. Porting to beta simc with the expectation that it will be in the TWW release, and will be more readily noticeable if hotfix was missed rather than the other way around.

PR #9103, opened 2024-07-25
[shaman] Infer deck-of-cards chances from spell data

wfarr:
This preserves the current options behavior by having sane defaults and
allowing the user to override these values without having to delve into
the complexity of simc overrides.

Instead of hardcoding proc-chances though, we now fallback to defaults
inferred from the spell data.

This is still somewhat fragile because Blizzard does not guarantee
consistent ordering of spell effects, but should at least _sometimes_
result in fewer manual changes to simc to keep up with in-game changes.


wfarr:
Bump for review @HawkCorrigan @Bloodmallet

Bloodmallet (approved):

PR #9136, opened 2024-07-31
[shaman] stub temporary hotfixes for future beta build

wfarr:
DO NOT MERGE

This change only exists so that it's easy for the Elemental
theorycrafters to work against the same copy of the class module while
we explore the impacts of the not-yet-released beta changes.


engine/class_modules/sc_shaman.cpp
```diff
@@ -7499,7 +7499,7 @@ struct earthquake_damage_t : public earthquake_damage_base_t
     earthquake_damage_base_t( player, "earthquake_damage", player->find_spell( 77478 ), parent )
   {
     // Earthquake modifier is hardcoded rather than using effects, so we set the modifier here
-    spell_power_mod.direct = 0.3915;
+    spell_power_mod.direct = 0.3195 * 0.85;
```
wfarr, line 7502: See #9138 for SP mod change

PR #9137, opened 2024-07-31
[shaman] Pull Earthquake and Earthquake Overload from SDB

wfarr:
We've been intending to do this for a long time but have kept kicking
the can down the road.

The discovery of a typo in the Earthquake SP coefficient that was
hardcoded since last year has pushed me to finally look deeper into
doing this.


wfarr:
I've tagged you on this one @navv1234 as I'm sure you'll have some thoughts on this pattern. This is mostly lifted from the unique gear role-based variable extraction, but it's also not really going to tolerate any failure to match on the regex.

Please feel free to make changes directly to this PR if you have some preferences or better ideas on how to handle this.

navv1234:
I'm ok with using regexp for that if you want. Those values in the past have been so static that once set, it is usually set for the whole expansion so have not seen it worthwhile to implement the regex pattern search for it.

engine/class_modules/sc_shaman.cpp
```diff
@@ -7461,11 +7466,19 @@ struct earthquake_overload_damage_t : public earthquake_damage_base_t
   earthquake_overload_damage_t( shaman_t* player, earthquake_base_t* parent ) :
     earthquake_damage_base_t( player, "earthquake_overload_damage", player->find_spell( 298765 ), parent )
   {
-    // Earthquake modifier is hardcoded rather than using effects, so we set the modifier here
-    spell_power_mod.direct = 0.3195 * player->talent.mountains_will_fall->effectN( 1 ).percent();
-    // TODO: implement spellpower coefficient extracttion from spell variable
-    // auto spell_desc = player->dbc->spell_desc_vars( this->data().id() );
-    // spell_desc.desc_vars()
+
+      if ( auto vars = player->dbc->spell_desc_vars( this->data().id() ).desc_vars() )
```
navv1234: Pretty sure this is not going to work. Spells 298765 for overload (and the real one 77478) do not contain the Spell Description Variable you are looking for. It's contained in the base earthquake spell, which would be 61882 or 462620.

engine/class_modules/sc_shaman.cpp
```diff
@@ -7461,11 +7466,19 @@ struct earthquake_overload_damage_t : public earthquake_damage_base_t
...
+      {
+          std::cmatch m;
+          std::regex get_var( R"(\$damage=\$\{\$SPN\*([\d\.]+)\*.*\})" );
+
+          if ( std::regex_search( vars, m, get_var ) )
+          {
+              auto coeff = util::to_double( m.str( 1 ) ) * player->talent.mountains_will_fall->effectN( 1 ).percent();
+              player->sim->print_debug( "parsed Earthquake Overload spell power coefficient to be '{}'", coeff );
```
navv1234: Player-relevant debug output generally contains the actor name in it, so you should add it to the debug message.

engine/class_modules/sc_shaman.cpp
```diff
@@ -7461,11 +7466,19 @@ struct earthquake_overload_damage_t : public earthquake_damage_base_t
   earthquake_overload_damage_t( shaman_t* player, earthquake_base_t* parent ) :
     earthquake_damage_base_t( player, "earthquake_overload_damage", player->find_spell( 298765 ), parent )
   {
-    // Earthquake modifier is hardcoded rather than using effects, so we set the modifier here
-    spell_power_mod.direct = 0.3195 * player->talent.mountains_will_fall->effectN( 1 ).percent();
-    // TODO: implement spellpower coefficient extracttion from spell variable
-    // auto spell_desc = player->dbc->spell_desc_vars( this->data().id() );
-    // spell_desc.desc_vars()
+
```
navv1234: Should this still have a default? As far as I know, this code will now make Earthquake / overload do 0 damage if the regex fails. One option would be to add (an expansion specific) default value that can be updated manually, so 0.3195 in tww. Another option would be to assert if the coefficient is 0, this should trigger CI if it gets changed. Third option would be to add some error message but allow the sim to run of course, which would probably get noticed by someone .. at some point.

wfarr:
OK I reckon I've addressed all the feedback here.

Primary changes:

* Add an option, which defaults to the current value.
* If the option is set to a non-default value, use that as the SP coefficient for the base Earthquake spell
* Else, attempt to parse the SP coefficient via regex from the spell database
* Finally, assert the coefficient is > 0
* Additionally, include the actor name in debug logs
* Refer to the correct Spell ID for the lookups

I opted for a singular option rather than one for each spell since in theory we shouldn't ever need to override Mountains Will Fall independent of an override on the effect that specifies the percentage itself.

engine/class_modules/sc_shaman.cpp
```diff
@@ -7467,11 +7475,28 @@ struct earthquake_overload_damage_t : public earthquake_damage_base_t
   earthquake_overload_damage_t( shaman_t* player, earthquake_base_t* parent ) :
     earthquake_damage_base_t( player, "earthquake_overload_damage", player->find_spell( 298765 ), parent )
   {
-    // Earthquake modifier is hardcoded rather than using effects, so we set the modifier here
-    spell_power_mod.direct = 0.3195 * player->talent.mountains_will_fall->effectN( 1 ).percent();
-    // TODO: implement spellpower coefficient extracttion from spell variable
-    // auto spell_desc = player->dbc->spell_desc_vars( this->data().id() );
-    // spell_desc.desc_vars()
+      auto coeff = 0.0;
+      auto default_option_value = (new shaman_t::options_t())->earthquake_spell_power_coefficient;
```
wfarr: Unsure if there's a nicer way to get at the default for an option than reinitializing the struct here, but this should be relatively fine I think.

Bloodmallet:
I welcome not having to double check this value. I forgot this regularly and it bit us.

engine/class_modules/sc_shaman.cpp
navv1234: Is there an issue with just using `player->option.earthquake_spell_power_coefficient` that i'm not understanding?

navv1234:
There a reason we duplicate the coefficient handling code? Could it not be implemented as a method in `earthquake_damage_base_t` and then just invoked from constructor?

engine/class_modules/sc_shaman.cpp
wfarr: > Is there an issue with just using player->option.earthquake_spell_power_coefficient that i'm not understanding?

Mostly just trying to ensure we only utilize the option if it has been overridden. The alternative I guess would be setting the default value to 0.0 and only threading the coefficient from the option through if it is >0.

wfarr:
> There a reason we duplicate the coefficient handling code? Could it not be implemented as a method in `earthquake_damage_base_t` and then just invoked from constructor?

Done.

HawkCorrigan (approved):
LGTM and we need the functionality

PR #9169, opened 2024-08-05
[Mage] Arcane APL Update and Consumeables for TWW

ArcanePorom:
Arcane APL complete re-write for TWW
Updated some level 80 consumeables

### DIFF
--- a/engine/class_modules/apl/mage.cpp
+++ b/engine/class_modules/apl/mage.cpp
@@ -36,7 +36,7 @@ std::string rune( const player_t* p )
 {
   return p->true_level >= 80 ? "crystallized"
        : p->true_level >= 70 ? "draconic"
-	   : p->true_level >= 60 ? "veiled"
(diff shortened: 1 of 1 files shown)


engine/class_modules/apl/mage.cpp
```diff
@@ -7,41 +7,44 @@ namespace mage_apl {
...
        :                       "disabled";
 }
 
 std::string rune( const player_t* p )
 {
-  return p->true_level >= 70 ? "draconic"
-       : p->true_level >= 60 ? "veiled"
+  return p->true_level >= 80 ? "crystallized"
+       : p->true_level >= 70 ? "draconic"
+	   : p->true_level >= 60 ? "veiled"
```
vituscze: Could you fix the tab here and replace it with spaces?

ArcanePorom: Should be good now, sorry!

PR #9190, opened 2024-08-08
[Monk] Explicitly set Weapons of Order buff cooldown to zero

justindriggers:
The Weapons of Order buff has the same spell ID as the ability, so it has a cooldown of 2 minutes. When WoO's cooldown is reduced by other effects, the buff's cooldown remains at 2 minutes. This cooldown override was removed as a part of a recent refactor [#8951](https://github.com/simulationcraft/simc/pull/8951/files#diff-2912bd33a4bb4d7aa74376e8eb6428fcf319f56eb84e2ccec85845982ec1ad72L7662).

I prefer setting the buff cooldown to zero rather than applying various CDR effects manually to both the buff and the triggering ability.

@renanthera for 👀


renanthera:
Tested locally and looks good to me.

PR #9213, opened 2024-08-10
[ele] fix lava surge proc rate

HawkCorrigan:
Logs are indicating a 10% proc rate

grafik

This seems to be matching 0.5/(1+e^(11-x)), normalized to target count (the above image is a 2 target test).

This implementation seems to be underperforming, only proccing 90% of the intended proc rate.


HawkCorrigan:
fixed my problems

datamining didnt ignore casts so it overestimated tick count

PR #9217, opened 2024-08-11
[rng] Multiple-Success Deck of Cards.

renanthera:
Initial implementation for multiple-success deck of cards.
Incompatible with the current version of `proc_rng_t::trigger()`, as it necessarily requires a non-covariant return type.

TODO:
- Improve random element selection.
- Add rng type to `rng_type_e`.


renanthera:
Discussion on Discord suggested that switching the return type of `trigger` was reasonably safe, and `int` was suggested for a more convenient error value (-1), if one ends out being useful. New implementation does not benefit from it, at this time.

This was also moved back toward my original vector implementation, and `shuffled_rng_t` was converted to be a special case of `shuffled_rng_multiple_t`.

I have also tested that the behaviour appears accurate at least for Guardian Druid Convoke the Spirits, after stepping through a sim's `trigger`/`reset` invocations, and checking the current `entries` and `position` values.

PR #9241, opened 2024-08-14
[Monk] Adjust Restore Balance / RJW

JeremyMonk:


Hinalover (approved):
This looks good and I swear I had the pandemic refresh before. So not sure when it was removed.

renanthera (approved):

PR #9263, opened 2024-08-18
[Buff] Do not extend pandemic buffs above their pandemic limit

Fluttershy-git:
`buff.extend_duration` has been extending pandemic buffs up to infinity, while they should only have a maximum duration of 1.3 times their buff length.

### DIFF
--- a/engine/buff/buff.cpp
+++ b/engine/buff/buff.cpp
@@ -2175,6 +2175,13 @@ void buff_t::extend_duration( player_t* p, timespan_t extra_seconds )
 
   assert( expiration.size() == 1 );
 
+  // Do not refresh above pandemic buff limit
+  if ( refresh_behavior == buff_refresh_behavior::PANDEMIC )
+  {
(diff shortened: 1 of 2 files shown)


engine/buff/buff.cpp
```diff
@@ -2175,6 +2175,13 @@ void buff_t::extend_duration( player_t* p, timespan_t extra_seconds )
 
   assert( expiration.size() == 1 );
 
+  // Do not refresh above pandemic buff limit
+  if ( refresh_behavior == buff_refresh_behavior::PANDEMIC )
+  {
+    extra_seconds = std::min( expiration.front()->remains() + extra_seconds, buff_duration() * 1.3 ) -
+                    expiration.front()->remains();
+  }
+
```
seanpeters86, line 2184: Does not seem to be the case for Voidform which can be extended much beyond Pandemic in-game.

Fluttershy-git:
Buff in question (Shake the Heavens) doesn't even pandemic, it extends up to 15+ seconds just fine. So everything is just fine :)

PR #9289, opened 2024-08-21
[WARLOCK] [AFFLICTION AND DEMONOLOGY] Update TWW1_Generate_Warlock.simc

Mlklock:
Gear Profiles for TWWW1 Demonology Warlock as Diabolist and for Affliction as both Hellcaller and Soul Harvester as they spec talents for the hero talents change.


Azevara:
Not creating gear profiles at the moment for Warlock - will revisit in the days after Early Access launch.

PR #9291, opened 2024-08-21
[Warlock] [Destruction] Update TWW1_Generate_Warlock.simc for Destruction Warlock

Mlklock:
This one is more of a temporary Profile for Destruction, but want to push a Gear Profile alongside the APL for generation.


Azevara:
Not creating gear profiles at the moment for Warlock - will revisit in the days after Early Access launch.

PR #9298, opened 2024-08-22
[Warlock] Cleaning Warlock Apls  and adds #comments

Mlklock:
Mostly removing Older Trinkets from the APL, but i left some for the sake of exemple in some of the automatic Trinket Logic. 

Added  Comments on th Demonology APL and in regards to the automatic use of Trinkets Variable.


Mlklock (commented):
I made a mistake while adding the comments, but upon reviewing other exemples of comments i updated the PR.

PR #9300, opened 2024-08-22
[Warrior] Initial Prot Warrior profiles

Armadk:


srliao:
Hey, 

I was just looking the prot warrior apl and for this line: `shield_block,if=buff.shield_block.duration<=10` (https://github.com/ebonhold/simc/blob/7dbf4114d5fd165b515f81d3015712b1c0ff016e/engine/class_modules/apl/apl_warrior.cpp#L278) shouldn't it say `.remains` instead of `.duration`? I'm assuming the intention here is stack shied block any time buff has less than 10s left?

I think `.duration` is just causing it to press sb on cd? albeit the difference in dps result seems to be trivial

I was going to put in a PR but I saw this PR and felt might be more relevant here

Armadk:
Great!  Thanks.  Updated the PR

PR #9327, opened 2024-08-25
[spell_info/parse_effects] Rename Spell Critical Multiplier Misc Type to Spell Critical Bonus Multiplier.

renanthera:
Requesting feedback on potentially renaming Spell Critical Multiplier (15)/`P_CRIT_DAMAGE` to a more clear name. Present naming has implications that it applies only to damaging effects, while it just modifies the bonus for any critically-striking effect.

Current proposed name is Spell Critical Bonus Multiplier (15)/`P_CRIT_BONUS`.

Consists of strictly cosmetic changes, modifying output of Spell Queries, `parse_effects` and some comments.
Could be extended to modifying `P_CRIT_DAMAGE`, but that would require more changes and devs keeping it in mind. Not sure how deep relevant changes would propagate.


renanthera:
Discussion occurred in Discord and it was agreed that renaming `P_CRIT_BONUS` was worth pursuing, but tackling other instances of similar language used in relationship to subtypes, as they are broken out more specifically than they are for misc values.

PR #9351, opened 2024-08-28
Update apl_warrior.cpp

hrothmar:
Adding Bladestorm to single target rotation, to match the WoWhead guide's priority.


hrothmar:
It was found that current Single Target sims are not casting Bladestorm at all, see here:
https://www.raidbots.com/simbot/report/47Hqa1SXXjBXNH2KJmaAyK

Looking over the `actions.single_target` table, Bladestorm doesn't appear to be listed at all for Arms:
image

Resulting in one (1) cast being made over a 5min period, as seen here:
Untitled

hrothmar:
Closing this PR, I discussed with @Archimtiros who confirmed updates are already being worked on. 

Thanks for the good work 🥇

PR #9359, opened 2024-08-29
[monk] Fix CotC Crashes and remove DF tier sets

renanthera:


renanthera:
Requires updates of the WW APL before it can be merged, or all WW sims using the default APL will fail.

PR #9368, opened 2024-08-30
[Monk] Mistweaver implementation

acornellier:


engine/class_modules/monk/sc_monk.cpp
```diff
@@ -1186,6 +1186,7 @@ struct tiger_palm_t : public overwhelming_force_t<monk_melee_attack_t>
 {
   bool face_palm;
   action_t *tigers_ferocity;
+  action_t *tiger_palm_jadefire_proc;
```
renanthera: This is no longer used.

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -1767,6 +1771,16 @@ struct blackout_kick_t : overwhelming_force_t<charred_passions_t<monk_melee_atta
       p()->gain.bok_proc->add( RESOURCE_CHI, base_costs[ RESOURCE_CHI ] );
   }
 
+  int n_targets() const override
+  {
+    // The Ancient Concordance buff's spell data modified number of targets hit, but the value is 1 instead of 3, so it
+    // is useless. We set the value manually to 3.
+    if ( p()->buff.ancient_concordance->check() )
+      return 3;
```
renanthera, line 1779: It's not necessarily useless, as it may provide 2 stacks in game. Note that the effect has the "Aura Points on Client" bit set, which has to do with behaviours related to effect stacking.

Would need to perform some in game science, as it is also not logged.

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -7313,8 +7326,12 @@ void monk_t::init_spells()
...
   shared.rushing_jade_wind = _priority( talent.windwalker.rushing_jade_wind, talent.brewmaster.rushing_jade_wind );
 
-  shared.teachings_of_the_monastery =
-      _priority( talent.windwalker.teachings_of_the_monastery, talent.mistweaver.teachings_of_the_monastery );
+  if ( talent.windwalker.teachings_of_the_monastery->ok() )
+    shared.teachings_of_the_monastery = talent.windwalker.teachings_of_the_monastery;
+  else if ( baseline.mistweaver.teachings_of_the_monastery->ok() )
+    shared.teachings_of_the_monastery = baseline.mistweaver.teachings_of_the_monastery;
+  else
+    shared.teachings_of_the_monastery = spell_data_t::not_found();
```
renanthera, line 7096: If you're not comfortable writing a new equivalent function to `_priority` that is more friendly to multiple parameter types, let me know and I can put something together.

This feels gross, even I'm generally trying to get away from shared spell pointers.

acornellier: sure, no idea how to rewrite `_priority` tbh

renanthera:
The only things left to double check are TP OF working (I'm unsure if that's resolved) and whether or not all of the appropriate actions are using SP conversion properly when MW.

At that point, I would consider this good to merge (even prior to SEF completion).

A more flexible `_priority` is certainly a TODO, but not mandatory. It's in the same category of priority as my intended spell data storage changes.

PR #9374, opened 2024-08-30
[Warlock] Avoid rounding error in Channel Demonfire

Azevara:


Azevara:
Should hopefully be resolved by 14c1796 instead.

PR #9454, opened 2024-09-08
[shaman] `debug()` for temp enchants over imbues

wfarr:
This is a bit noisy, and there might be some better ways we could handle
this in the future given different shaman specializations have different
constraints with regards to imbues (eg. Elemental only taking them when
certain talents are present, but Enhancement almost always taking them,
etc).

For now, make this a `debug()` log statement to get rid of the warning
bloat at the top of the page.


wfarr:
cc @simulationcraft/shaman for interest

PR #9455, opened 2024-09-08
[elemental] extract APLs and consumables

wfarr:
Keeping in line with other specializations, and seeking to make it
easier to contribute APL changes to simc for folks not used to
navigating a big pile of C++, extract the Elemental APL out into the APL
module and use code generators to turn the raw APL into C++.

I've additionally set up a separate APL for PTR, as with the substantial
changes coming to Elemental on the PTR, we will surely want to begin
developing an entirely separate APL.

I've left Restoration and Enhancement mostly alone for now (the one
exception being the Rune), but the requisite skeletons are mostly there
so it would be trivial to lift-and-shift these over should the folks who
work on them want to do so.

Finally, I've removed support for sub-71 consumables for Elemental. I
personally see very little value in continuing to support them now that
the expansion has been out for several weeks.


Bloodmallet:
Awesome!

PR #9461, opened 2024-09-08
[shaman] if PTR, icefury can proc from LB/CL

wfarr:
The 11.0.5 PTR has enabled Icefury to be proc'd off of Lightning Bolt
and Chain Lightning in addition to Lava Burst.

These only work with normal cast variants (not background casts).

The chance to proc has not changed so the rng object configuration has
been left as-is.

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -5865,6 +5865,17 @@ struct chain_lightning_t : public chained_base_t
   {
     chained_base_t::execute();
 
+    if ( exec_type == spell_variant::NORMAL && p()->specialization() == SHAMAN_ELEMENTAL )
+    {
+      if ( p()->is_ptr() )
+      {
+        if ( p()->rng_obj.icefury->trigger() )
+        {
+          p()->buff.icefury_cast->trigger();
+        }
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_shaman.cpp
```diff
@@ -5865,6 +5865,17 @@ struct chain_lightning_t : public chained_base_t
   {
     chained_base_t::execute();
 
+    if ( exec_type == spell_variant::NORMAL && p()->specialization() == SHAMAN_ELEMENTAL )
```
wfarr, line 5868: I'm not sure if we have any sources of background Chain Lightnings today actually, but feels better safe than sorry to start scoping foreground only effects in a check for that.

engine/class_modules/sc_shaman.cpp
```diff
@@ -6803,6 +6813,14 @@ struct lightning_bolt_t : public shaman_spell_t
         p()->buff.stormkeeper->decrement();
       }
       p()->sk_during_cast = false;
+
+      if ( p()->is_ptr() )
+      {
```
wfarr, line 6818: I know this can be condensed into a singular conditional, but I think when it comes time to prune all these after the PTR lands that this will be cleaner to diff.

PR #9462, opened 2024-09-08
[shaman] Ascendance changes for PTR

wfarr:
This should cover nearly all the Ascendance changes for PTR, modulo the Flame Shock spreading.

Note: I fully suspect the implementation of how I did the +1 Overloads to maybe be Weird(TM) with Chained Lightning, but that can be dealt with later I think.


wfarr:
Opting to merge this as-is. The tracking issue #9463 details the remaining work to be done, but this covers _most_ of the Ascendance implementation for single-target.

PR #9467, opened 2024-09-09
[shaman] Add support for new PTR talent Herald of the Storms

wfarr:
This talent takes over the CDR that previously belonged to Flash of
Lightning, but applies only to Stormkeeper rather than all nature
spells.


wfarr:
Best I can tell looking at rogue stuff this may Just Work(TM) without the need for any special setup around the talent only existing in the PTR DBC. Needs testing still to verify it works w/ an updated talent string for the Stormbringer profile with ptr=1, but it should run cleanly on live at least with no negative impact.

PR #9470, opened 2024-09-09
[shaman] Mark Skybreaker's Fiery Demise as removed on PTR

wfarr:
Additionally, disable its effects always if `ptr=1` even though the
talent guards should handle that anyway.

The intent here is to make it easy to identify all code blocks which can
safely be deleted after the PTR is released as a live version.

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -953,7 +953,7 @@ public:
     player_talent_t first_ascendant;
     player_talent_t preeminence;
     player_talent_t fury_of_the_storms;
-    player_talent_t skybreakers_fiery_demise;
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_shaman.cpp
```diff
@@ -8017,7 +8017,13 @@ struct flame_shock_t : public shaman_spell_t
   {
     double m = shaman_spell_t::composite_crit_chance();
 
-    m += p()->talent.skybreakers_fiery_demise->effectN( 3 ).percent();
+    if ( !p()->is_ptr() )
```
wfarr, line 8020: This guard isn't necessary, but is left as a reminder to delete this code after the PTR for 11.0.5 is mainlined.

PR #9484, opened 2024-09-10
[gear] Update role mults for Sikran's Endless Arsenal, Shadowed Essence, Mad Queen's Mandate, Skyterror's Corrosive Organ

renanthera:


renanthera:
TODO: Implement a fallback for `role_mult` in case it is not found on the effect.

PR #9515, opened 2024-09-15
[Mage] Frost GS and Flurry over FOF

Manather:
https://www.raidbots.com/simbot/report/3ChwLgQ5YhCuyEvmzqnh3F
not my changes, but good update


Manather:
I did not separate my pull requests, whoops. If you dont want to change distance for frost just decline and I'll open a new one for the ST apl updates.

PR #9520, opened 2024-09-15
[shaman] Add 11.0.5 PTR Primordial Wave support

wfarr:
On 11.0.5 PTR, Primordial Wave has been updated to generate maelstrom
and proc Lava Surge in addition to all of its current effects.

I am unsure if the Lava Surge is actually on projectile hit, but going
off the tooltip it seems somewhat likely that it may be. Worst case,
this is fairly trivial to change.


HawkCorrigan (dismissed):
please move surge trigger to execute

PR #9527, opened 2024-09-15
[Warlock] Demonology Warlock APL Update

Mlklock:
Currently, and even after the hotfixes next week casting Demonic Strength or Bilescourge Bombers in single target is a dps loss or dps neutral. 

This happens because Demonic Strength isn't strong enough to out-value  the benefits of a GCD being spent on something else + the 4 seconds where the Felguard can't Legion Strike or Auto-attack. 

I believe part of the reason is that Shadowbolt now is a much stronger cast than in Dragonflight, being a 1,5s cast  and the Soul Shard having that much more value in both Hero Talents. 

https://www.raidbots.com/simbot/report/1MgUhcWxZJ7C7aM3urKhDC 

Even afternext hotfix 15% buff to Felguard you would still have Demonic Strength only breaking even or at a 0.1% worse than just not casting it in single target.

As Such, i am limiting Bilescourge Bombers, Guillotine and Demonic Strength to only be cast at 2 targets or more.

Additionally,  i opted to delete two lines who don't seems to have much function anymore in the APL and who i previously failed to notice. 

Thanks for Sumer in the contributions towards this changes, i appreciate it a lot.


Mlklock:
Update:  Redone the sims with latest and no overriding spelldata https://www.raidbots.com/simbot/report/5nn3VknGcj59KVrVcgBQX6  

similar situation,  casting BB in single target is for sure a dps loss, DSTR is within a 0.1% in favor of not casting Demonic Strength in single target.

PR #9560, opened 2024-09-20
[Gear] add concoction fallback buff

Shurtugal:
With the new Concoction: Kiss of Death behaviour, you have to jump to cancel it.
While it would be better if this was directly handled by simc, I wanted the buff to be trackable by an APL to at least be able to have an APL condition to do movement when the player has to jump.

### DIFF
--- a/engine/player/unique_gear_thewarwithin.cpp
+++ b/engine/player/unique_gear_thewarwithin.cpp
@@ -4557,6 +4557,9 @@ void candle_confidant( special_effect_t& effect )
 // 440235 stun, NYI
 void concoction_kiss_of_death( special_effect_t& effect )
 {
+  if ( unique_gear::create_fallback_buffs( effect, { "concoction_kiss_of_death" } ) )
+    return;
+
   struct concoction_kiss_of_death_buff_t : public stat_buff_t
   {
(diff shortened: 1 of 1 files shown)


gastank:
The current implementation assumes you will automatically succeed in cancelling the buff before the stun. `thewarwithin.concoction_kiss_of_death_buff_remaining_max` and `thewarwithin.concoction_kiss_of_death_buff_remaining_min` are used to set how much to shorten the buff duration by to account for you cancelling before the end.

Shurtugal:
Indeed, but it doesn't account the fact that, since yesterday, you need to jump to actually cancel the buff. So I wanted to model that movement in the APL. And to have a buff.concoction_kiss_of_death.remains condition in the APL, without having errors if the trinket is not equipped, we need to add a fallback buff, right ?

gastank:
Ya you need the fallback to check the buff via the APL. There's just no APL way to cancel the buff atm and it'll simply time out at the full determined time. We don't have a 'jump' action per se nor really any way to force movement via the APL other than relying on assumptions like the player will jump/move during an instant gcd. If you want that level of control via the APL you'll want to further adjust the implementation with possibly something like a specific action for cancelling the buff.

PR #9589, opened 2024-10-01
[shaman][elemental] add renamed talent as "new" talent

Bloodmallet:
This allows live simulations to keep existing and provide no change to ptr, while ptr uses the renamed talent to still add its benefit to the simulation.


wfarr (approved):
Looks good to me!

PR #9596, opened 2024-10-02
[shaman][elemental] fix Earthquake Overload damage calculation

Bloodmallet:
Live and ptr simulations are overvalueing Earthquake Overload damage. This fix applies our Overload multiplier and Echo Chamber properly.


HawkCorrigan (approved):
LGTM

PR #9603, opened 2024-10-03
Add Enhancement to CI

NGAndersson:
Seeing if Enhancement can be added to CI without issues.


Pewtro:
Enhancement has since been added to the CI

PR #9607, opened 2024-10-04
[Death Knight] 1105 Deathbringer rework

Khazakdk:


engine/class_modules/sc_death_knight.cpp
```diff
@@ -9573,10 +9650,10 @@ struct howling_blast_t final : public death_knight_spell_t
                           p()->gains.rage_of_the_frozen_champion );
     }
 
-    if ( p()->talent.deathbringer.dark_talons.ok() && p()->buffs.rime->check() && p()->buffs.icy_talons->check() &&
+    if ( p()->talent.deathbringer.dark_talons.ok() && p()->talent.icy_talons->ok() &&
```
Armadk: I think we still need the rime check here

engine/class_modules/sc_death_knight.cpp
```diff
@@ -10639,6 +10735,17 @@ struct soul_reaper_action_t final : public soul_reaper_t
       }
     }
   }
+
+  double cost() const override
+  {
+    double c = death_knight_melee_attack_t::cost();
+    if ( p()->is_ptr() && p()->talent.deathbringer.reaper_of_souls.ok() && p()->buffs.reaper_of_souls->up() )
```
Armadk, line 10747: FYI, in game there is currently no cost reduction.  It's in the tooltips, but you still get charged the rune for soul reaper on cast.

engine/class_modules/sc_death_knight.cpp
```diff
@@ -15050,6 +15221,11 @@ void death_knight_t::apply_affecting_auras( buff_t& buff )
   buff.apply_affecting_aura( talent.sanlayn.frenzied_bloodthirst );
 
   // Deathbringer
+  if ( is_ptr() )
+  {
+    buff.apply_affecting_aura( talent.deathbringer.reapers_onslaught );
+    buff.apply_affecting_aura( talent.deathbringer.bind_in_darkness );
```
Armadk: This doesn't really hurt anything, but at present the talent doesn't modify anything in the buff.

engine/class_modules/sc_death_knight.cpp
Khazakdk: Good catch

PR #9627, opened 2024-10-09
[shaman][elemental] implement Flame Shock variants more thoroughly

Bloodmallet:
NORMAL and PRIMORDIAL_WAVE can trigger Surge of Power.
NORMAL and SURGE_OF_POWER generate Maelstrom.


HawkCorrigan (approved):

HawkCorrigan:
looks fine, could throw a default value at the type but I dont really feel strongly about that anyway

PR #9648, opened 2024-10-18
[ele] Update PTR gear profile

MadBubble:


wfarr:
This was included in changes in #9653

PR #9656, opened 2024-10-20
Added Aftershock logic to the single target spenders

The-PaleRider:
Checked to see if the player was running aftershock over surge of power, and if so just send the spenders


HawkCorrigan:
should be covered with #9663

PR #9697, opened 2024-10-25
[SS Frost] AoE Rework for Spellslinger

wingsco:
Split AoE rotation between SS and FF and modified the action list calls appropriately.

SS Frost has now a completely new and reworked AoE apl.

Also removed the now unnecessary time_to_coc variable and replaced the variable in aoe_ff with its numeric value.

Cleaned up the cleave_ss apl from 3t specific conditions (since ss is doing aoe at 3 anyway) and a few freeze-specific conditions without change in functionality against unfreezables and a very minor loss against freezable targets.

Refined the variable that determines for FF when to switch into the FrostfireBolt spam build.

Edit1: also slightly refined Ray of Frost usage for SS when specced into an M+ build (st_aoebuild)

Edit2: ill get mana to do it, im too 2head for this ;_;

-Gosto


wingsco:
*forgot: slightly refined Ray of Frost usage for SS when specced into an M+ build (st_aoebuild)

PR #9716, opened 2024-10-29
[Destro] Update TWW1 Destro lock profile

SzynaWoW:
It is updating the current Destro Lock profile that uses Diabolist hero tree.

Below is the sim report 
https://www.raidbots.com/simbot/report/5UZqCAq4VHC5bmKSjKV5AE

Most of the gear and talent changes are a result of buffs to spenders. The race change fixes an oversight from previous iterations of this profile, as Dwarf has always siming better than Orc for Destro due to the crit interaction with Chaos Bolt.

As for the APL change, there has been a rare edge-case in the Shadowburn build, when Demonic Arts could've been consumed with the shadowburn instead of the chaos bolt.

image

This is a sanity check (and minor ~~0.1% gain) to ensure such things never happen.


SzynaWoW:
I'm sorry for the multiple commits. I wanted to make sure everything checks out and found a gem mismatch. It is the final version now, please squash when merging if possible :)

PR #9731, opened 2024-11-01
[Hunter] Updates hunter generated profiles

CrankyDragon:
Updates the loadouts for hunter specs adds back generated profiles.


Pewtro:
I personally don't really like the way profiles are used by external sources, and I think it's a misuse of them in general, so I won't be merging this. If others want to merge it in, I'm not opposed to it and won't be removing them (unless they're invalid for an extended period of time like I did with 11.0.5). I, personally, just want to have as little to do with profiles as possible.

PR #9759, opened 2024-11-23
[ele] adding some bugs

HawkCorrigan:


HawkCorrigan:
Logic for LB buff stuff looks like this:

SK takes precedence over AD
SK takes precedence over NS

If you do not have AD or NS active
then
   If you have no SK charges remaining when LB hits the boss, consume Storm Frenzy

PR #9827, opened 2024-12-31
[Warlock][Diabolist] Adds Flame of Xoroth to Guardians and a small change in effect order.

Mlklock:
Attempt at adding Flame of Xoroth damage modifier to Guardian Pets,  additionally, i noticed the effect 4 is the one responsible to buffing primary pet, and effect 3 is for guardians. 

I  hope i have done this right...

### DIFF
--- a/engine/class_modules/warlock/sc_warlock.cpp
+++ b/engine/class_modules/warlock/sc_warlock.cpp
@@ -363,11 +363,8 @@ double warlock_t::composite_player_pet_damage_multiplier( const action_state_t* 
       m *= 1.0 + talents.summoners_embrace->effectN( 2 ).percent();
   }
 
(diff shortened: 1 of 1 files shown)


Azevara:
This was apparently hotfixed on Oct 1, 2024. See #9590 . Similar updates for Xalan's Cruelty and Xalan's Ferocity (both Hellcaller spells) should be included as well for completeness.

Mlklock:
> This was apparently hotfixed on Oct 1, 2024. See #9590 . Similar updates for Xalan's Cruelty and Xalan's Ferocity (both Hellcaller spells) should be included as well for completeness.

Yeah, before that hotfix those effects had some very weird implementation ingame, effectively it didn't work for guardian pets while having some double dipping interactions with player damage.

Azevara:
Replaced by #9844

PR #9863, opened 2025-01-14
[monk] Conditionally disable effect 8 of HotJS if specialization is Windwalker.

renanthera:
TODO: double check that this is actually what is going on :)


renanthera:
Haha just kidding, this got reverted.

PR #9880, opened 2025-01-24
[Warlock] [Demonology] Change Soul Strike Priority.

Mlklock:
After performing some tests i have noticed that Soulstrike priority is lower than Legion Strike, in a way that if Felguard have enough energy to cast Legion Strike it will wait to cast Soul Strike

https://github.com/user-attachments/assets/7df6aa4c-ee5c-4731-aed5-b024e63ed135

In this video i allow the Felguard to be at 200 energy before turning Legion Strike with the Soul Strike avaliable, this results in 3 Legion Strikes being cast before Felguard casts 1 Soul Strike.


Mlklock:
https://imgur.com/a/xS52Cym Alternative for the video above.

PR #9900, opened 2025-02-05
TWW - Season 2 - Survival Tier Set

syrifgit:
First time PRing here, pls check carefully lol. I am one of the Hekili Devs but C++ is new to me.

### DIFF
--- a/engine/class_modules/sc_hunter.cpp
+++ b/engine/class_modules/sc_hunter.cpp
@@ -428,6 +428,8 @@ public:
 
     spell_data_ptr_t tww_s2_mm_2pc;
     spell_data_ptr_t tww_s2_mm_4pc;
+    spell_data_ptr_t tww_s2_sv_2pc;
+    spell_data_ptr_t tww_s2_sv_4pc;
   } tier_set;
 
   struct buffs_t
@@ -493,6 +495,8 @@ public:
     buff_t* harmonize; // BM 4pc
     // TWW - S2
     buff_t* jackpot; // MM 2pc
+    buff_t* winning_streak;  // SV 2pc - Wildfire Bomb damage stacking buff
+    buff_t* strike_it_rich;  // SV 4pc - Mongoose Bite damage buff, consuming it reduces Wildfire Bomb cooldown
 
     // Hero Talents 
 
@@ -1249,6 +1253,9 @@ public:
     affected_by.lead_from_the_front = parse_damage_affecting_aura( this, p->talents.lead_from_the_front_buff );
 
     affected_by.tww_s2_mm_2pc = parse_damage_affecting_aura( this, p -> tier_set.tww_s2_mm_2pc->effectN( 2 ).trigger() );
+    affected_by.tww_s2_sv_2pc = parse_damage_affecting_aura( this, p -> tier_set.tww_s2_sv_2pc ); // Winning Streak affects Wildfire Bomb
+    affected_by.tww_s2_sv_4pc = parse_damage_affecting_aura( this, p -> tier_set.tww_s2_sv_4pc ); // Strike It Rich affects Mongoose Bite
+
 
     // Hunter Tree passives
     ab::apply_affecting_aura( p -> talents.specialized_arsenal );
@@ -1290,6 +1297,8 @@ public:
     ab::apply_affecting_aura( p -> tier_set.tww_s1_mm_2pc );
     ab::apply_affecting_aura( p -> tier_set.tww_s1_mm_4pc );
     ab::apply_affecting_aura( p -> tier_set.tww_s1_sv_2pc );
+    ab::apply_affecting_aura( p -> tier_set.tww_s2_sv_2pc );
+    ab::apply_affecting_aura( p -> tier_set.tww_s2_sv_4pc );
 
     // Hero Tree passives
     ab::apply_affecting_aura( p->talents.sentinel_precision );
@@ -6242,6 +6251,9 @@ struct melee_focus_spender_t: hunter_melee_attack_t
     double am = hunter_melee_attack_t::action_multiplier();
 
     am *= 1 + p() -> buffs.mongoose_fury -> stack_value();
+        
+    if ( p()->buffs.strike_it_rich->check() )
+      am *= 5.0;
 
     return am;
   }
@@ -6298,6 +6310,12 @@ struct melee_focus_spender_t: hunter_melee_attack_t
     p()->buffs.howl_of_the_pack_leader_cooldown->extend_duration( p(), -p()->talents.dire_summons->effectN( 4 ).time_value() );
 
     p()->buffs.hogstrider->expire();
+
+        if ( p()->tier_set.tww_s2_sv_4pc.ok() && p()->buffs.strike_it_rich->check() )
+    {
+        p()->buffs.strike_it_rich->expire();
+        p()->cooldowns.wildfire_bomb->adjust( -10.0 );
+    }
   }
 
   void impact( action_state_t* s ) override
@@ -7858,6 +7876,15 @@ struct wildfire_bomb_base_t: public hunter_spell_t
     {
       p()->cooldowns.butchery->adjust( -timespan_t::from_seconds( p()->talents.covering_fire->effectN( 2 ).base_value() ) );
     }
+
+    
+    if ( p()->tier_set.tww_s2_sv_2pc.ok() )
+    {
+      if ( p()->buffs.winning_streak->check() && rng().roll( 0.15 ) )
+      {
+          p()->buffs.winning_streak->expire();  // Consume 2pc buff
+          p()->buffs.strike_it_rich->trigger(); // Apply 4pc buff
+    }
   }
 };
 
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_hunter.cpp
```diff
@@ -9424,6 +9468,39 @@ void hunter_t::init_special_effects()
...
+      };
+
+      auto const effect = new special_effect_t( this );
+      effect->name_str = "winning_streak";
+      effect->spell_id = tier_set.tww_s2_sv_2pc->id();
+      effect->custom_buff = buffs.winning_streak;
+      special_effects.push_back( effect );
+
+      auto cb = new winning_streak_cb_t( *effect, this );
+      cb->initialize();
```
Jayezi, line 9472: should be able to just set this up exactly like jackpot above without defining a new callback, this callback is pretty simple so initialization should take care of defaults like proc chance and flags and handle triggering what's set to custom_buff when it procs

engine/class_modules/sc_hunter.cpp
```diff
@@ -9013,6 +9042,16 @@ void hunter_t::create_buffs()
     = make_buff( this, "jackpot", tier_set.tww_s2_mm_2pc->effectN( 2 ).trigger() )
       ->set_default_value_from_effect( 1 );
 
+  buffs.winning_streak =
```
Jayezi, line 9042: normally won't want to hard code them but max stacks and duration will get picked up automatically anyway so you can leave them off here

engine/class_modules/sc_hunter.cpp
syrifgit: Ok, thanks!

engine/class_modules/sc_hunter.cpp
```diff
@@ -1290,6 +1297,8 @@ struct hunter_action_t: public Base
     ab::apply_affecting_aura( p -> tier_set.tww_s1_mm_2pc );
     ab::apply_affecting_aura( p -> tier_set.tww_s1_mm_4pc );
     ab::apply_affecting_aura( p -> tier_set.tww_s1_sv_2pc );
```
Jayezi: these calls would set up passive auras to modify base values of spell damage etc, but these will be dynamic effects based on buffs, so there's nothing in the tier set driver spelldata to pick up and apply here
you'll need to add appropriate mods in the multiplier calculations below appropriate to the buff, such as winning streak has a direct damage modifier

engine/class_modules/sc_hunter.cpp
```diff
@@ -1249,6 +1253,9 @@ struct hunter_action_t: public Base
     affected_by.lead_from_the_front = parse_damage_affecting_aura( this, p->talents.lead_from_the_front_buff );
 
     affected_by.tww_s2_mm_2pc = parse_damage_affecting_aura( this, p -> tier_set.tww_s2_mm_2pc->effectN( 2 ).trigger() );
+    affected_by.tww_s2_sv_2pc = parse_damage_affecting_aura( this, p -> tier_set.tww_s2_sv_2pc ); // Winning Streak affects Wildfire Bomb
```
Jayezi: these will need to look at the buff spelldata instead, as those have the effects that apply the damage bonus mods to the list of spells

engine/class_modules/sc_hunter.cpp
```diff
@@ -6242,6 +6251,9 @@ struct melee_focus_spender_t: hunter_melee_attack_t
     double am = hunter_melee_attack_t::action_multiplier();
 
     am *= 1 + p() -> buffs.mongoose_fury -> stack_value();
+        
+    if ( p()->buffs.strike_it_rich->check() )
+      am *= 5.0;
```
Jayezi: when applying damage mods you'll want to look up the bonus from spelldata, this case it would be from the buff spelldata (you could also refer to the buff's value() if you set up the buff with a default value pulled from its data)
doing lookups rather than hard numbers keeps everything up to date with no effort added when there are spelldata changes

syrifgit: So like this? 

```C++
if ( p()->buffs.strike_it_rich->check() )
      am *= 1 + p() -> buffs.strike_it_rich -> value();
```

It looks like the value I want is stored in effect # 1 (which I will add to the buff definition). But it's stored as 400%, is this going to read it as `400` or `4`?

Jayezi: would depend on what's set as the default value for the buff, in this case it's dealing with a % bonus so using percent() on it (effect 1 from the buff's spelldata) returns the value / 100, which you will want to use with set_default_value on the buff, the other methods to set the default normally handle it but this effect uses a dummy value rather than a proper damage mod effect

engine/class_modules/sc_hunter.cpp
```diff
@@ -7858,6 +7876,15 @@ struct wildfire_bomb_base_t: public hunter_spell_t
     {
       p()->cooldowns.butchery->adjust( -timespan_t::from_seconds( p()->talents.covering_fire->effectN( 2 ).base_value() ) );
     }
+
+    
+    if ( p()->tier_set.tww_s2_sv_2pc.ok() )
+    {
+      if ( p()->buffs.winning_streak->check() && rng().roll( 0.15 ) )
```
Jayezi: can look up the proc_chance() from the 2pc spelldata for this
p()->tier_set.tww_s2_sv_2pc->proc_chance()

engine/class_modules/sc_hunter.cpp
```diff
@@ -6298,6 +6310,12 @@ struct melee_focus_spender_t: hunter_melee_attack_t
     p()->buffs.howl_of_the_pack_leader_cooldown->extend_duration( p(), -p()->talents.dire_summons->effectN( 4 ).time_value() );
 
     p()->buffs.hogstrider->expire();
+
+        if ( p()->tier_set.tww_s2_sv_4pc.ok() && p()->buffs.strike_it_rich->check() )
+    {
+        p()->buffs.strike_it_rich->expire();
+        p()->cooldowns.wildfire_bomb->adjust( -10.0 );
```
Jayezi: time values can also be pulled from a spelldata effect, here the spelldata uses a dummy value 10000 which is the ms value of the adjustment, so time_value() can be used when looking up effect 2 of the buff's spelldata since it assumes a millisecond value
here for example -p()->buffs.strike_it_rich->data().effectN( 2 ).time_value()

syrifgit:
I think I fixed everything @Jayezi, ty for looking at it.

Where can I find docs on the spell data stuff to learn more about what's available to me, such as `set_default_value_from_effect( 1 );`? Something to help correlate between the functions, the web interface on WoWhead and the spell data dump in the files.

renanthera:
There's very little documentation of simc from the developer perspective. All of the major available members of `buff_t` (the object you're referencing) will be available in https://github.com/simulationcraft/simc/blob/thewarwithin/engine/buff/buff.hpp.

Effects will be enumerated by index in the same way they are in spell data dumps/spell queries.

All of the relevant getters for spell data objects can be found here: https://github.com/simulationcraft/simc/blob/thewarwithin/engine/dbc/spell_data.hpp#L143

If you have questions, the Discord is usually your best bet.
My biggest tip is if you think the answer to solving a problem is hard-coding some number, pause and think very carefully, as it's nearly always incorrect and a better solution exists :)

engine/class_modules/sc_hunter.cpp
```diff
@@ -6242,6 +6246,9 @@ struct melee_focus_spender_t: hunter_melee_attack_t
     double am = hunter_melee_attack_t::action_multiplier();
 
     am *= 1 + p() -> buffs.mongoose_fury -> stack_value();
+        
+    if ( p()->buffs.strike_it_rich->check() )
+      am *= 1 + p() -> buffs.strike_it_rich -> value().percent();
```
Jayezi: value() is a double so this won't compile, percent() is a function of a spelldata effect to pull the effect value interpreted as a percent (usually value / 100)
since you're pulling the damage bonus from the buff which has the value set from the spelldata effect already, you won't need to do anything else
the alternative where percent() would be needed is instead of looking at the buff you look at the spelldata effect, examples are elsewhere in the code but basically would look like p()->tier_set.tww_s2_sv_4pc_buff.effectN( 1 ).percent()

engine/class_modules/sc_hunter.cpp
```diff
@@ -9199,6 +9233,11 @@ void hunter_t::init_rng()
   
   rppm.shadow_hounds = get_rppm( "Shadow Hounds", talents.shadow_hounds );
   rppm.shadow_surge = get_rppm( "Shadow Surge", talents.shadow_surge );
+  if ( tier_set.tww_s2_sv_2pc.ok() )
```
Jayezi: this doesn't exist anywhere so would need to be removed

engine/class_modules/sc_hunter.cpp
syrifgit: ah ok, was wondering about that thanks. I tried checking other examples but a lot of them seemed to come from the effect fields.

I saw this one came from the aura proc chance field but wasn't sure how to actually retrieve it, which was the prompt for my question about the docs haha.

syrifgit:
> All of the relevant getters for spell data objects can be found here: https://github.com/simulationcraft/simc/blob/thewarwithin/engine/dbc/spell_data.hpp#L143

This is exactly what I was looking for thanks!

syrifgit:
@Jayezi Think I figured it all out, managed to compile it locally as well. Thanks for the help, Putro too.

PR #9920, opened 2025-02-14
Update Augmentation APL

KYZ124:
Minor fix as discussed in the Wyrmrest Temple Discord that avoids Tomorrow, Today showing as a DPS decrease over spending the point on a non-DPS increasing node.


Saeldur:
This is just the auto-generated APL file and not the actual apl file. The real apl file is [https://github.com/simulationcraft/simc/blob/thewarwithin/engine/class_modules/apl/apl_evoker.cpp](here).

PR #9946, opened 2025-02-20
[Gear] Tome of Light's Devotion

Armadk:


engine/player/unique_gear_thewarwithin.cpp
```diff
@@ -6919,6 +6919,239 @@ void zees_thug_hotline( special_effect_t& effect )
...
+      {
+        cumulative = false;
+        set_default_value( data->effectN( 5 ).average( e ) );
+      }
+    };
+    buff_t* ward_of_devotion_buff;
+
+    const spell_data_t* data;
+
+    inner_resilience_cb_t( const special_effect_t& e )
```
nyterage: Probably dont need a custom callback for this if all it does is trigger a buff, can use `effect.custom_buff = buff_t*` to trigger it whenever the callback executes.

engine/player/unique_gear_thewarwithin.cpp
```diff
@@ -6919,6 +6919,239 @@ void zees_thug_hotline( special_effect_t& effect )
...
+        radiance_verses_tracker->set_expire_callback( [ this ]( buff_t* b, int, timespan_t ) {
+                                      resilience_phase = true;
+                                      inner_radiance_buff->expire();
+                                      inner_resilience_buff->trigger();
+                                      resilience_verses_tracker->trigger();
+                                      } );
+
+        auto inner_resilience = new special_effect_t( effect.player );
+        inner_resilience->name_str = "inner_resilience";
+        inner_resilience->spell_id = 450706;
```
nyterage: Should probably have some asserts when searching these special effect spell ids just in case they change them for whatever reason.

engine/player/unique_gear_thewarwithin.cpp
```diff
@@ -6919,6 +6919,239 @@ void zees_thug_hotline( special_effect_t& effect )
...
+        inner_resilience->proc_flags2_ = PF2_ALL_HIT;
+        inner_resilience->disable_buff();
+        effect.player->special_effects.push_back( inner_resilience );
+
+        inner_resilience_cb = new inner_resilience_cb_t( *inner_resilience );
+        inner_resilience_cb->activate_with_buff( resilience_verses_tracker, true );
+
+        auto inner_radiance = new special_effect_t( effect.player );
+        inner_radiance->name_str = "inner_radiance";
+        inner_radiance->spell_id = 450720;
```
nyterage: Same as above

engine/player/unique_gear_thewarwithin.cpp
```diff
@@ -6919,6 +6919,239 @@ void zees_thug_hotline( special_effect_t& effect )
...
+        radiance_buff = rb;
+        base_dd_min = base_dd_max = data->effectN( 3 ).average( e.item );
+      }
+
+      double composite_da_multiplier( const action_state_t* s ) const override
+      {
+        double m = generic_proc_t::composite_da_multiplier( s );
+
+        if ( radiance_buff->check() )
+            m *= 2;
```
nyterage, line 7018: Is this value not contained in spell data?

Armadk: sadly no.  Just says increased effectiveness.  Value was found via manual testing.

engine/player/unique_gear_thewarwithin.cpp
Armadk: Original intention was for the callback to handle the doubling of the shield via radiance.  Implemented this and pushed.

PR #9947, opened 2025-02-20
[Destro] Fire and Brimstone can proc Demonfire Infusion

Saeldur:
Add the Demonfire Infusion roll from Incinerate to Fire and Brimstone Impact


Azevara:
Is there a log or screenshot example of a single cast proccing multiple cdf projectiles?

Saeldur:
Scratch this - Seems like it works completely differently now. Would need large testing. Chance is potentially increased but no multi-proc scenarios occured

PR #10030, opened 2025-03-02
[Hunter] Adding corrected TWW1 profiles

Samos2b:
Adding corrected TWW1 Beast Mastery profile to match the 11.1 changes.


Pewtro (changes requested):
The changes required would be in this file: 
https://github.com/simulationcraft/simc/blob/thewarwithin/profiles/generators/TWW1/TWW1_Generate_Hunter.simc

The APL lines are not necessary

PR #10032, opened 2025-03-03
Fix casc_extract on Python >= 3.12 (`ConfigParser.readfp()` -> `read_file()`)

micolous:
`ConfigParser.readfp()` was removed in Python 3.12 (released 2023).

[`ConfigParser.read_file()`]( https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read_file) is available on Python 3.2 and later. Given that was released in 2011, I don't think there's any need for backward compatibility that far back. :)


seriallos:
Thanks!

PR #10036, opened 2025-03-03
[Warrior] Update TWW2 Fury profile

pauldrinn:


pauldrinn:
Archimtiros added both Arms and Fury so closing this

PR #10051, opened 2025-03-04
[Druid] Balance APL: use built-in expressions instead of variables

syrifgit:
Instead of defining a new variable, use the already-defined `eclipse.X` expressions. This change should have no effect on the rotation. One could also be built to replace `variable.eclipse_remains`, but I will leave this PR at the existing expressions.

This type of change helps greatly with a downstream project.


gastank:
Druid APLs are handled by making changes to https://github.com/dreamgrove/dreamgrove/blob/master/sims/owl/balance.txt

PR #10058, opened 2025-03-06
[Warlock] Wither tick time needs manual adjustment

Azevara:


Azevara:
This will be fixed in core auto-parsing instead.

PR #10090, opened 2025-03-15
[Gear] Add new option to Soulletting Ruby trinket

millanzarreta:
The new option allows setting a fixed health percentage for the target when using the SR trinket

### DIFF
--- a/engine/player/unique_gear_shadowlands.cpp
+++ b/engine/player/unique_gear_shadowlands.cpp
@@ -958,12 +958,14 @@ void soulletting_ruby( special_effect_t& effect )
     stat_buff_t* buff;
     double base_crit_value;
     double max_crit_bonus;
+    double fixed_percentage_targethealth;
 
     soulletting_ruby_t( const special_effect_t& e, stat_buff_t* b ) :
       proc_spell_t( "soulletting_ruby", e.player, e.player->find_spell( 345802 ) ),
       buff( b ),
       base_crit_value( e.player->find_spell( 345807 )->effectN( 1 ).average( e.item ) ),
-      max_crit_bonus( e.player->find_spell( 345807 )->effectN( 2 ).average( e.item ) )
+      max_crit_bonus( e.player->find_spell( 345807 )->effectN( 2 ).average( e.item ) ),
+      fixed_percentage_targethealth( e.player->sim->shadowlands_opts.soulletting_ruby_fixed_percent_targethealth )
     {}
 
     void execute() override
@@ -971,7 +973,7 @@ void soulletting_ruby( special_effect_t& effect )
       proc_spell_t::execute();
 
       make_event( *sim, travel_time(), [ this, t = execute_state->target ] {
-        double bonus_mul = 1.0 - ( t->is_active() ? t->health_percentage() * 0.01 : 0.0 );
+        double bonus_mul = 1.0 - ( ( fixed_percentage_targethealth < 0.0 ) ? ( t->is_active() ? t->health_percentage() * 0.01 : 0.0 ) : ( fixed_percentage_targethealth ));
 
         buff->stats.back().amount = base_crit_value + max_crit_bonus * bonus_mul;
         buff->trigger();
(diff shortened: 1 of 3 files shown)


millanzarreta:
The [Soulletting Ruby](https://www.wowhead.com/item=178809/soulletting-ruby) trinket from Theater of Pain dungeon (Shadowlands) is now available to players in TWW due to the dungeon's addition to the TWW Season 2 M+ pool.

This trinket provides a variable critical hit value based on the target's health percentage when used (less target HP% = more crit). This behavior is already implemented in ``simc`` and does not need adjustments.

It may sometimes be interesting for players to simulate the use of this trinket based on a fixed health percentage (instead of the real target health percentage). This allows them to simulate additional situations during a game session in a more personalized way and/or obtain other useful information, such as: setting an appropriate mean value to better sim m+ situations when player is sniping low-health mobs, to know the maximum value the trinket can provide with optimal use, to know the minimum value the trinket can provide in the worst-case scenario, etc.

In any case, this parameter is optional and is disabled by default (thus is also disabled if the user does not use this parameter). If disabled, the trinket in sim behaves the same way it did until now.

Therefore, the new optional option ``shadowlands.soulletting_ruby_fixed_percent_targethealth`` has been added, allowing a fixed health percentage value to be set for all uses of the trinket. Its default value is ``-1.0``, and with this value the option is disabled and the health percentage value of the current target on which the trinket is used will be taken in every trinket use. This option is enabled by setting a value between ``0.0`` and ``1.0`` for a fixed health percentage value between 0% and 100% respectively for all trinket uses.

At last, if this PR is accepted and merged, the [Wiki page](https://github.com/simulationcraft/simc/wiki/ExpansionOptions) should be updated with the info of this new option.

nyterage:
this is generally already an option for players, utilizing `enemy_fixed_health_percentage=x` or `enemy_initial_health_percentage=x` options implemented in simc
more details about those options here. 
https://github.com/simulationcraft/simc/wiki/Enemies#health-related-options

millanzarreta:
It's true that these options could be used for the same purpose. However, these options don't just affect the trinket, but the entire simulation, and they are very likely to affect the results due to other reasons (different priorities in the APL based on the percentage of health, buffs and/or spells that are only used/affected at certain percentages of health, executes, etc.).

This is why I still think this new option is useful. I think the new option would allow for more isolated testing of the differences in the trinket's impact based on the percentage of health it is used on.

renanthera:
The behaviours are fundamentally linked, making this produce clearly invalid output.

If you want to decouple the two effects, you would be able to spawn an invulnerable target at a separate fixed hp iirc, but there may be strange behaviours down that rabbit hole.

millanzarreta:
> The behaviours are fundamentally linked, making this produce clearly invalid output.

The same could be said for most of the trinkets/sets/effects options on [this page](https://github.com/simulationcraft/simc/wiki/ExpansionOptions). In many of these effects, changing the default value also produces an "invalid output". However, I don't see it that way: when using these extra options, you are enforcing behavior that exists in-game (usually related to player behavior) but is not easy (or even impossible) to replicate in the simulator.

To give one example of many possible options, the already available `dragonflight.nymue_forced_immobilized` option allows you to set the trinket to always deal the extra damage it does against rooted/stunned targets. By default, this option is set to `false`, and the simulator has implemented a check for rooted/stunned targets in the code. Theoretically, this new option isn't necessary, as roots/stuns can be set on the target in other ways. This option, however, is simple and effective. It's disabled by default (so it doesn't produce unexpected results), and deliberately enabling it easily achieves the desired goal in the simulation, which is to check the results if we always make sure to use the trinket on rooted/stunned targets. I wouldn't consider this an "invalid output". There are more examples, but this one seemed quite similar to this one.

Likewise, the option proposed in this PR allows for a simple and effective option to transfer to the simulator the desired in-game behavior (e.g. when sniping other targets with the trinket, targets that are often not the primary target), I would not consider this an invalid output, or no less than other options already available. The option is disabled by default, so no unexpected results occur for those who ignore it.

> If you want to decouple the two effects, you would be able to spawn an invulnerable target at a separate fixed hp iirc, but there may be strange behaviours down that rabbit hole.

I suppose something along those lines could be attempted, but it's very likely to produce various unintended effects (changes in the APL due to the number of targets, effects that shouldn't affect that other unit, etc.). Both these options and others related to `enemy_fixed_health_percentage=x` or `enemy_initial_health_percentage=x` would be feasible, but they present certain problems, in addition to being quite complex solutions.

I still believe that this new option is simple and clear in what it seeks to achieve, it doesn't affect those who don't explicitly want to use it, it allows you to adjust certain in-game behaviors in the simulator as desired, and the alternatives present certain complications in addition to not being simple.

PR #10093, opened 2025-03-16
[Outlaw] Edit KIR rules and RTB usage for HO builds

soloxcx:
KIR fishing rules without a 5 roll are superfluous in the end.


soloxcx:
Drafting, some changes to Hidden Opportunity rolls need to be ironed out and might as well add it to this PR.

PR #10101, opened 2025-03-22
[Monk] Fix BoC tracking for BoF and PtA+RSK

justindriggers:
Also fixes unconsumed BoC when BoF is used before KS


engine/class_modules/monk/sc_monk.cpp
```diff
@@ -1658,6 +1663,7 @@ struct charred_passions_t : base_action_t
       background = dual = proc = true;
       may_crit                 = false;
       base_multiplier          = data().effectN( 1 ).percent();
+      school                   = SCHOOL_FIRE;
```
renanthera: Technically it "worked" previously on technicality given the override to mults. It being the wrong school indicates that it wasn't based on the correct spell.

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -3570,23 +3619,37 @@ struct breath_of_fire_dot_t : public monk_spell_t
...
+
+  state_t* cast_state( action_state_t* s )
+  {
+    return static_cast<state_t*>( s );
+  }
+
+  const state_t* cast_state( const action_state_t* s ) const
+  {
+    return static_cast<const state_t*>( s );
   }
```
renanthera, line 3625: I prefer directly using `debug_cast` as `cast_state` may be conflated to "a state object/ptr for the cast of the spell" which doesn't exist.
Also given the minimal use of these functions.

Also `using custom_state_t` is preferable as it makes the fact that the state is non-default more clear, although I'm not entirely stuck on that one.

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -3631,34 +3696,71 @@ struct breath_of_fire_t : public monk_spell_t
   double action_multiplier() const override
   {
     double am = monk_spell_t::action_multiplier();
+
+    if ( blackout_combo )
+      am *= 1.0 + p()->buff.blackout_combo->data().effectN( 5 ).percent();
+
```
renanthera: Blackout Combo is already handled via `parse_effect` here:
https://github.com/simulationcraft/simc/blob/560c527933beab95f893c047c63f489a04377635/engine/class_modules/monk/sc_monk.cpp#L169

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -3631,34 +3696,71 @@ struct breath_of_fire_t : public monk_spell_t
...
+  state_t* cast_state( action_state_t* s )
+  {
+    return static_cast<state_t*>( s );
+  }
+
+  const state_t* cast_state( const action_state_t* s ) const
+  {
+    return static_cast<const state_t*>( s );
+  }
+
```
renanthera, line 3688: See https://github.com/simulationcraft/simc/pull/10101/files#r2029854282

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -3631,34 +3696,71 @@ struct breath_of_fire_t : public monk_spell_t
...
       dragonfire_brew->execute();
 
-    if ( p()->buff.blackout_combo->up() )
-      p()->proc.blackout_combo_breath_of_fire->occur();
-    // defer boc consumption to be handled by bof dot
+    monk_spell_t::execute();
   }
 
-  void impact( action_state_t *state ) override
+  void impact( action_state_t *s ) override
```
renanthera: Full names when at all possible :)

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -1653,11 +1658,11 @@ struct charred_passions_t : base_action_t
   struct damage_t : monk_spell_t
   {
     damage_t( monk_t *player, std::string_view name )
-      : monk_spell_t( player, fmt::format( "charred_passions_{}", name ), player->talent.brewmaster.charred_passions )
+      : monk_spell_t( player, fmt::format( "charred_passions_{}", name ),
+                      player->talent.brewmaster.charred_passions_damage )
     {
       background = dual = proc = true;
-      may_crit                 = false;
```
renanthera, line 1659: Swapping to the correct damage spell made overriding this flag redundant.

engine/class_modules/monk/sc_monk.cpp
```diff
@@ -1692,15 +1708,11 @@ struct charred_passions_t : base_action_t
       return;
 
     base_action_t::p()->proc.charred_passions->occur();
-    chp_damage->base_dd_min = chp_damage->base_dd_max = state->result_amount;
-    chp_damage->execute();
+    chp_damage->execute_on_target( state->target, state->result_amount );
```
renanthera, line 1711: For reasons unbeknownst to all, the `execute_on_target` method can also handle setting `base_dd_min = base_dd_max = amt`. Previously, ChP would only ever hit the main target as well, oops!

engine/class_modules/monk/sc_monk.cpp
justindriggers: I definitely noticed the conflated words, but truth be told I was just copying what other modules were doing. I like the assertions provided by `debug_cast` - I'll start using that in the future.

engine/class_modules/monk/sc_monk.cpp
justindriggers: Totally missed that, but I don't believe this is working. Given a simple APL:

```
actions=auto_attack
actions+=/blackout_kick
actions+=/breath_of_fire
```

Breath of Fire's damage is the same even if I switch Blackout Combo (`CwQAAAAAAAAAAAAAAAAAAAAAAAAAAAjxYGGWmx2MzMDAAAAAAALLgYmBmhBGjtZmZYWGzM2mlZabb2sNzMMLAAwGCAAgZbWamZmFmhFA`) for Pretense of Instability (`CwQAAAAAAAAAAAAAAAAAAAAAAAAAAAjxYGGzyM2mZmZAAAAAAAYZBEzMwMMwYsNzMDzyYmx2sMTbbzmtxMMLAAwGCAAgZbWamZmFmhFA`)

renanthera:
Oh duh, that should have been obvious...

engine/class_modules/monk/sc_monk.cpp
renanthera: You'll also notice the same issue with Dragonfire Brew prior to 24aed20eb0122de1157c1f1bd14164bc3af61fc5.

Moving Blackout Combo `expire` prior to `execute` causes the `parse_effect` buff check to always fail.

Shifting it to after should not be a problem given that `impact` does not reference the buff state, only the locally held bool.

Default APL wouldn't also make this behaviour super clear, so I just tested with the following parameter sets:

`profiles/TWW2/TWW2_Monk_Brewmaster.simc html=/tmp/boc.html debug=0 output=/dev/null enable_all_talents=1 actions=blackout_kick actions+=/breath_of_fire`

`profiles/TWW2/TWW2_Monk_Brewmaster.simc html=/tmp/noboc.html debug=0 output=/dev/null enable_all_talents=1 actions=blackout_kick actions+=/breath_of_fire spec_talents=blackout_combo:0`

As expected, the former average bof hit is nearly exactly 50% larger than the latter.

justindriggers:
Ahh that'll do it. Good catch.

PR #10103, opened 2025-03-22
[Warlock] [APL] Fixes some usage of Trinkets

Mlklock:
Funhouse Lens is a secondary stat steroid trinket but without a trinket.1.is.funhouse_lens the automagic does not recognize it. 

Elixir of imperfect ascendancy wasn't being used due to the gcd conditional


engine/class_modules/apl/warlock.cpp
```diff
@@ -51,8 +51,8 @@ void affliction( player_t* p )
 
   precombat->add_action( "summon_pet" );
   precombat->add_action( "variable,name=cleave_apl,default=0,op=reset" );
-  precombat->add_action( "variable,name=trinket_1_buffs,value=trinket.1.has_use_buff", "Used to set Trinket in slot 1 as Buff Trinkets for the automatic logic" );
-  precombat->add_action( "variable,name=trinket_2_buffs,value=trinket.2.has_use_buff", "Used to set Trinkets in slot 2 as Buff Trinkets for the automatic logic" );
+  precombat->add_action( "variable,name=trinket_1_buffs,value=trinket.1.has_use_buff|trinket.1.is.funhouse_lens", "Used to set Trinket in slot 1 as Buff Trinkets for the automatic logic" );
```
nyterage, line 54: Signet of the Priory also has this issue. 

Should also add them to `trinket_x_buff_duration` variables.

Mlklock: This should do it right?

PR #10135, opened 2025-03-28
[Warlock] Fix talent interaction with Greater Dreadstalkers

millanzarreta:
Currently in-game, the **[[The Expendables](https://www.wowhead.com/spell=387600/the-expendables)]** talent never applies to **[Greater Dreadstalkers]** (possible bug?), so sims should reflect this same behavior.

As for the **[[The Houndmaster's Gambit](https://www.wowhead.com/spell=455572/the-houndmasters-gambit)]** talent, it works differently with **[Greater Dreadstalkers]** than with regular **[Dreadstalkers]**.
For regular **[Dreadstalkers]**, what is stated and expected for the talent occurs: it doesn't matter whether the **[Dreadstalkers]** are summoned before or after the **[Vilefiend]** (or its variations), while the **[Vilefiend]** is active (and only while it is active), the **[Dreadstalkers]** receive the +50% damage bonus.
However, **[Greater Dreadstalkers]** only receive this bonus if the **[Vilefiend]** was already active when they were summoned. That is, if the **[Vilefiend]** is summoned after the **[Greater Dreadstalkers]** are active, they will not receive this damage bonus. This could also be a bug currently ingame, but it's how it works, and that's how it should be reflected in the sims.


millanzarreta:
In this commit, I've been continuist using the `PET_FELHUNTER` identifier for the `pet_type` (`enum pet_e`) for Greater Dreadstalkers (as it was already being done currently) to distinguish them from the other regular Dreadstalkers (`PET_DREADSTALKER`).

This shouldn't cause major problems, since the felhunter isn't used in demonology spec; instead, the felguard is always used. Although I suppose that the most strictly correct thing would be to use a new pet type identifier, but perhaps that is not what is desired. So, if preferred, I can make a new commit creating a new pet type identifier in that enum (`PET_GT_DREADSTALKER`, for example) and use it instead.

engine/class_modules/warlock/sc_warlock_pets.cpp
```diff
@@ -166,7 +166,8 @@ double warlock_pet_t::composite_player_multiplier( school_e school ) const
   if ( ( pet_type == PET_DREADSTALKER || pet_type == PET_FELHUNTER ) && o()->talents.dread_calling.ok() )
     m *= 1.0 + buffs.dread_calling->check_value();
 
-  if ( o()->talents.the_expendables.ok() )
+  // 2025-03-28: The Expendables talent does not apply to Greater Dreadstalkers (maybe a bug?)
+  if ( ( pet_type != PET_FELHUNTER || !bugs ) && o()->talents.the_expendables.ok() )
```
Azevara: This change is not necessary, check_stack_value() will return 0 when there is no buff so the multiplier will remain unchanged.

millanzarreta: Indeed, that's right. I'd already seen that it was unnecessary, but since there was already a check to determine whether the player has the talent or not, I think it can be a good idea to add the other check anyway.

In any case, I've updated the code in a new commit to remove that change.

PR #10153, opened 2025-04-04
[Outlaw] Add Trickster optimizations & misc changes to APL

soloxcx:
as discussed in RH:

Trickster optimizations-
1. Prio Sinister Strike during Disorienting Strikes (mainly for KIR builds)
2. Avoid refreshing AR and using Deft during Disorienting Strikes unless in stealth
3. Hold Vanish during Disorienting Strikes and 3 stacks of Escalating Blade. KIR 1 FTH builds can hold Vanish at 2 stacks though it's just neutral compared to 3 atm. There is prob a way to make 2 show a gain

Miscellaneous changes-
1. Ghostly strike can be used at max CP for 2 FTH builds again
2. Deft Maneuvers logic at 3-4 targets is not worth anymore
3. Moved normal KS and Coup into finisher sublist, directly called the finisher sublist again right before Vanish to catch them
4. 1 FTH builds don't have to specifically consume FTH at max stacks
5. Holding vanish for KS cooldown 15s -> 20s, just because i've been telling people 20s for a while now
6. Adds low-CP vanish for AR uptime as a separate priority above normal vanish usage 
7. Let Junkmaestro's Mega Magnet trinket stack up to avoid disrupting another on-use trinket
8. Add pooling to KS so it doesn't get skipped at low energy


soloxcx:
Drafting. The double coup was hotfixed so I need to review how necessary these rules still are.

PR #10157, opened 2025-04-06
[Demonology][Warlock] Excludes Greater Dreadstalkers from Wild Imps Expendable Buff

Mlklock:
Greater Dreadstalkers does not benefit of stacks of The Expendables when Wild Imps vanish.

This is a small but noticeable consideration in AOE, as Imploding 10 imps will cause Greater Dreadstalkers to gain 10 stacks, increasing its damage by 10% in the simulation but not in the game.


Mlklock:
Closing Pull Request, a PR who does this already exists.

PR #10160, opened 2025-04-08
[Warlock] Dreadstalker and Greater Dreadstalker adjustments and improvements

millanzarreta:


millanzarreta:
This PR is intended to be the culmination of a series of past PRs aimed at more accurately sim the Dreadstalker in-game behavior, which ended with only a partial implementation.

The previous PRs were, in chronological order, :one:#8683 &#8594; :two:#8698 &#8594; :three:#8721 &#8594; :four:#8823. We'll briefly describe them below, noting their role in this PR:

- :one:#8683: The original PR that describes in great detail the Dreadstalker in-game behavior, its importance in the results, how the tests were conducted, and their results. It also proposes code that implements this behavior. This PR was closed because a more flexible skeleton that handles the same issues was implemented in :two:#8698.

- :two:#8698: This PR implements a more flexible skeleton for these changes, although this implementation is only partial and has some bugs. These changes were successfully merged into the main DF branch of simc and later successfully transferred to the TWW branch.

- :three:#8721: This fixes some bugs from the previous PR, as well as adding the feature that if a distance of `<=5yd` is defined, Dreadstalkers will be considered summoned at melee range (they behave somewhat differently when summoned at melee range: they don't leap and have no delay on the first attack). Although these changes were merged into the main DF branch, the TWW branch was already under development, and these changes were never ported from the DF branch to the TWW branch.
  - This new PR includes the changes made here so they can be properly transferred to the current main branch of TWW.

- :four:#8823: This PR was intended to make the final adjustments to the behavior of the Dreadstalkers that were still pending, establishing a more accurate adjustment of the leap speed and extra duration to better match the in-game behavior. The final results are shown in several comparison figures. However, this PR was not merged because, as it was a late PR in DF, the development branch was already focused on TWW, and this PR was for DF.
  - This new PR implements the features from here so they can be included in the main branch of TWW.

It was also pending to review the WoW changes due to the TWW expansion (including talents and spells) and check if they affect the Dreadstalker behavior in some way. The following conclusions have been reached:

- Dreadstalker behavior and timing remain unchanged in TWW. In DF, an intensive battery of ingame tests was performed (varying distance and haste) to extract data, understand how they work, and then implement them in simc (results in :one:#8683). This battery of tests involved over 200 hours of playtime. Therefore, repeating them all is complicated and somewhat unnecessary. To verify that there are no changes in Dreadstalker's behavior, speed, or timings compared to DF, several options were chosen at random and tested, thus verifying that there are no changes.

- The [Flametouched](https://www.wowhead.com/spell=453699/flametouched) talent increases the attack speed of Dreadstalkers by 10%. Similar to the previous one, several random in-game tests were performed to verify that the in-game values match those of the simc, showing that the talent is correctly implemented in simc and no further adjustments are necessary in this PR.

- The Season 2 tier of TWW gives to player the chance to occasionally summon an improved version of Dreadstalkers, known as Greater Dreadstalkers. Another objective of this PR is to analyze the in-game behavior and timings of these Greater Dreadstalkers and properly implement them in simc. After performing and reviewing several tests, it has been determined that the behavior is the same as the regular Dreadstalkers: the jump speed is the same, the delay of the first attack is the same, and the extended duration is the same. In fact, if they are summoned simultaneously, the timestamps are exactly the same. Currently in simc, these Greater Dreadstalkers attack instantly and have a fixed duration, which doesn't match the observed in-game behavior.
  - This PR implements the necessary changes to ensure that Greater Dreadstalkers' behavior also matches the ingame behavior.

Finally, tests need to be carried out to verify the changes. We will check the results for the current simc behavior, without the implementation of this PR (dashed line, labeled as `OLD`), concerning those observed in-game (dash-dotted line, labeled `INGAME`), and concerning the ones of this implemented PR (solid line, labeled as `NEW`). These tests were performed without the [Flametouched](https://www.wowhead.com/spell=453699/flametouched) talent selected. The y-axis is the total number of dreadstalkers melee hits per summon, the x-axis is the haste amount, incrementing by 1% each step. Each figure shows the results for different distances, from 40yd to melee:

**- 40yd**
imagen

**- 35yd**
imagen

**- 30yd**
imagen

**- 25yd**
imagen

**- 20yd**
imagen

**- 15yd**
imagen

**- 10yd**
imagen

**- 0yd**
imagen

We can see that the new results after this PR are much more accurate to the ingame results than the previous ones. 

Finally, it's also interesting to verify that this implementation applied to Greater Dreadstalkers instead of Dreadstalkers is correct. Below are the results for 30 yd, where it can be seen that the behavior of the big dogs is also correct (note that the measurement accuracy is somewhat lower, but the result is consistent):

**- 30yd (Greater Dreadstalkers)**
imagen

PR #10168, opened 2025-04-10
Add attack_haste and spell_haste to the APL expressions

IIeTpoc:
Add `attack_haste `and `spell_haste` so they can be used in APL


nyterage:
These are already exposed to the APL in player.cpp, lines 11156 and 11162 respectively.

IIeTpoc:
Thanks, sorry for disturbing)

PR #10177, opened 2025-04-12
[Havoc] Winning streak and Chaos Brand Fix

Aezeor:
winning streaks convert to residuals after the action deals all damage events
chaos brand is applied soulscar and wounded quarry through a server side script to override spell flags that prevent in sims


ToppleTheNun:
handled in a83aaf34a62b771cd866c080194d46801bd5bf81 and c4e12dd191649b7c24fbb4dc8b4f097d6739c42b.

PR #10182, opened 2025-04-14
[Affliction] Count Jackpot UA for dot count

Saeldur:
Noticed this was not being counted when i was trying to test something briefly for Wark
`https://www.raidbots.com/simbot/report/9RKEeJbx7v6LNGepfh1KN8` was the APL
(Skip UA on pull and get the jackpot from DGL instead. Very basic test case.)

Not the bugged behaviour just a "normal" opener


Saeldur:
Finally found where this was being actually added. Oops.
My bad. That is some extra spaghetti bugs. Will ask Wark to test if they all still apply.

PR #10186, opened 2025-04-15
[Warlock] Dimension Ripper talent hotfix 2.25%

millanzarreta:
This change would be a temporary semi-solution until we fully understand how this talent's probabilities work ingame and adjust them accordingly in the simc.

Currently, selecting the [Dimension Ripper](https://www.wowhead.com/spell=457025/dimension-ripper) talent oversimulates more than expected. It has been tested that the average proc chance is 2.25% (maybe 2.26%?), not 5%.

The [Dimension Ripper](https://www.wowhead.com/spell=457025/dimension-ripper)  talent appears to use some sort of threshold-based accumulator, but further testing is still needed to determine its behavior and appropriate values. Currently, the simc implementation simply uses a base roll probability directly with every immolate/wither tick (which is not correct). Therefore, deeper adjustments to this talent in the warlock module are really needed. In the meantime, this hotfix should at least normalize the procs to the expected average.

### DIFF
--- a/engine/class_modules/warlock/sc_warlock.cpp
+++ b/engine/class_modules/warlock/sc_warlock.cpp
@@ -1021,7 +1021,18 @@ struct warlock_module_t : public module_t
   { return new warlock_t( sim, name, r ); }
 
   void register_hotfixes() const override
-  { }
+  {
+    // TODO: On average, the Dimension Ripper chance is 2.25% (maybe 2.26%?), not 5%.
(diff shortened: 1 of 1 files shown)


Khazakdk:
If this is truly still a draft, it would be good to clarify what needs to happen for this PR to be ready for review and merging so it doesn't sit waiting too long. Hotfixes should generally be able to be quickly reviewed and merged.

Azevara:
Handled by #10194

PR #10191, opened 2025-04-16
[Warlock] Adjustments to spells affected by Wicked Maw and Shadowtouched

millanzarreta:
As of 2025-04-16, the interaction of the **Wicked Maw** and **Shadowtouched** talents with the warlock and his pet spells ingame is as indicated in the following table (excel file also attached):
imagen
[wl_demon_wm_and_st_modifiers.xlsx](https://github.com/user-attachments/files/19772302/wl_demon_wm_and_st_modifiers.xlsx)

Some of these interactions appear to be bugged ingame, as they don't behave as expected. However, in this PR the behavior is recreated in simc as it works ingame, indicating in each case if necessary that it may be a bug. It would be a good idea to review it again in the future to check if Blizzard makes any related changes or fixes.

Regarding the spells, we highlight the following:
- It seems that **Implosion** previously had strange interactions with these two talents. Currently, it works with both as expected, receiving a +20% modifier with Wicked Maw and another 20% with Shadowtouched.
- **Bilescouge Bombers** works strangely (perhaps it's a bug). If you only have the Wicked Maw talent without Shadowtouched, it has no effect at all. However, if Wicked Maw is used with Shadowtouched, this spell will gain the bonus from both effects (20% for each), making the damage bonus modifier 1.20*1.20=1.44.
- The **Demonic Soul** hero talent is affected by Shadowtouched, which seems to be a bug. Additionally, the bonus is 10% for the primary target, while for other targets it is 20%.
- The **Wicked Reaping** hero talent is affected by Shadowtouched, which seems to be a bug. Additionally, the bonus is 50%.
- All pet spells that do not have a Shadow component are not affected by Shadowtouched, as expected. Pet spells with a Shadow component are affected by Shadowtouched, as expected, with the exception of the **Wicked Cleave** spell from *Overlord*, which is not affected even though it is from the Shadowflame school (probably a bug).


millanzarreta:
As of 2025-07-27, all of the Warlock spells (listed in the table above) have been retested on the WoW PTR version ``11.2.0 Build 62253``.

All spells continue to behave the same as originally stated in this PR for the previous WoW version (11.1.X), except for the Pit Lord pet spell **Felseeker**, which for some strange reason seems to have a ``1.25`` modifier for the **[Wicked Maw debuff (+Shadowtouched talent)]** instead of the expected ``1.20``.

Logs showing that behavior:
https://www.warcraftlogs.com/reports/6fDbZ7qB3FCPRwGr?fight=last&type=damage-done&source=10&ability=434404&view=events
https://www.warcraftlogs.com/reports/Njnf9vtkTpMCabdX?fight=2&type=damage&source=1&ability=434404&view=events

For reference, the new table for the 11.2.0 patch with the warlock spells and how **Wicked Maw** and **Shadowtouched** affect them would be the following:

imagen

A new commit has been made to include this change to the **Felseeker** spell, affecting only version >= 11.2.0 (it would be convenient at some point in the future when 11.2.0 goes live to remove this check in the code).

PR #10202, opened 2025-04-19
[Warlock] Adjust Felguard Fiendish Wrath behavior

millanzarreta:
The ingame damage of [Fiendish Wrath](https://www.wowhead.com/spell=386702/fiendish-wrath) is a simple `Deal Weapon Damage` aoe effect (excluding the primary target). So, in other words, the damage of each [Fiendish Wrath](https://www.wowhead.com/spell=386702/fiendish-wrath) hit the same damage amount as the Felguard's **melee** damage.

In simc the damage of [Fiendish Wrath](https://www.wowhead.com/spell=386702/fiendish-wrath) attack is currently implemented by setting its damage to the `result_raw` damage of the parent **melee** attack. This `result_raw` damage already includes increases from modifiers, so we need to prevent these from being reapplied again to [Fiendish Wrath](https://www.wowhead.com/spell=386702/fiendish-wrath). This is already partially done in the current code, but more modifiers still need to be added (and that is what is done in this PR).

This `result_raw` does not include armor damage reduction or any crit damage increases, but this is not a problem because they will also be calculated independently for [Fiendish Wrath](https://www.wowhead.com/spell=386702/fiendish-wrath). Just remember that for this reason, the `STATE_TGT_ARMOR`, `STATE_CRIT`, and `STATE_TGT_CRIT` flags should not be disabled.

Finally, as additional information, note that the crit roll chance of each [Fiendish Wrath](https://www.wowhead.com/spell=386702/fiendish-wrath) attack is independent of the crit roll chance of the parent **melee** attack, which is not a problem because that is already how it behaves in simc in this case.

### DIFF
--- a/engine/class_modules/warlock/sc_warlock_pets.cpp
+++ b/engine/class_modules/warlock/sc_warlock_pets.cpp
@@ -494,27 +494,9 @@ struct felguard_melee_t : public warlock_pet_melee_t
   {
     fiendish_wrath_t( warlock_pet_t* p ) : warlock_pet_melee_attack_t( "Fiendish Wrath", p, p->o()->talents.fiendish_wrath_dmg )
     {
+      weapon_multiplier = 1.0;
       background = dual = true;
       aoe = -1;
-    }
-
-    void init_finished() override
-    {
-      warlock_pet_melee_attack_t::init_finished();
-
-      // Disable multipliers: its base damage comes from its parent felguard's melee base damage after modifiers and should not be applied again
-      // Its crit roll is independent of its parent felguard's melee crit roll. Keep STATE_CRIT and STATE_TGT_CRIT flags: required to be able to crit
-      // Keep STATE_TGT_ARMOR flag to be affected by armor
-      snapshot_flags &= ~STATE_AP;
-      snapshot_flags &= ~STATE_MUL_SPELL_DA;
-      snapshot_flags &= ~STATE_MUL_SPELL_TA;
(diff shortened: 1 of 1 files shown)


millanzarreta:
I've made another commit to better tune the spell.

The previous implementation used the `result_raw` of the parent melee hit damage as the base damage, which already includes most multipliers, including the target multipliers. The child spell must then invalidate its own modifiers so that they are not applied again, and this way we avoid double dipping. The problem with this approach was that the target of the melee attack is different from the targets of the child spell, so the child spell shouldn't be subject to the multipliers of the parent's target, but rather its own.

The new implementation takes an approach more similar to what the spell does ingame, which is simply to apply weapon damage from the base (the `Deal Weapon Damage` effect), and not negate the effects of any multipliers. To do this, the damage of this spell is simply adjusted to deal the same base damage as the melee weapon hit (`weapon_multiplier = 1.0` and `base_dd_multiplier = 1.42`, the same parameters as the parent). This way, [Fiendish Wrath](https://www.wowhead.com/spell=386702/fiendish-wrath)'s damage is the same as the parent melee damage, but the target's multipliers will be applied correctly based on the target of each hit, just as they work ingame.

For practical purposes, the difference between the previous and the new implementation is small and will only be noticeable when the parent melee target and the child spell's targets have different multipliers (for example, some of them have [Mystic Touch](https://www.wowhead.com/spell=113746/mystic-touch) and others don't).

PR #10211, opened 2025-04-21
[DungeonRoute] Mark unused keystone and hp percent options as obsolete

Khazakdk:
These options have no bearing on the sim, but their presence in an input implies that users could sim a higher or lower keystone by changing a single number. Marking these as obsolete for now as popular methods of generating DungeonRoute strings still include them and need time to update. The goal is to improve clarity on how the strings are generated by making it obvious that the mob hp values for each pull are what need to change to simulate higher/lower keys.

### DIFF
--- a/engine/sim/sim.cpp
+++ b/engine/sim/sim.cpp
@@ -1473,8 +1473,6 @@ sim_t::sim_t()
     dbc_override( std::make_unique<dbc_override_t>() ),
     timewalk( -1 ),
     scale_to_itemlevel( -1 ),
-    keystone_level( 10 ),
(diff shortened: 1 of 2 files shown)


Jayezi:
while currently not used explicitly, there are generally useful for debugging based on just getting the input from rb error reports, and otherwise when the occasion calls for it, affix implementations where we need to spawn enemies that are not defined in mdt, such as what used to be done for the encrypted affix https://github.com/simulationcraft/simc/blob/c68c9d83425e437c27d314423347b1d904efe5ef/engine/sim/raid_event.cpp#L408

PR #10194, opened 2025-04-17
[Warlock] Halve Dimensional Ripper proc chance to match testing

Azevara:

PR #10213, opened 2025-04-21
[Warlock] Dimension Ripper talent adjust to 2.25%

millanzarreta:
1️⃣#10186 &#8594; 2️⃣#10194 &#8594; 3️⃣**this**

In the latest merged PR 2️⃣#10194, the [Dimension Ripper](https://www.wowhead.com/spell=457025/dimension-ripper) proc value is adjusted to half of its original value (`5 / 2 = 2.5`), but this is not the good value. The observed value in the ingame tests for this proc is `2.25`. In this PR, this is adjusted to give the expected result (`5 * 0.45 = 2.25`).

### DIFF
--- a/engine/class_modules/warlock/sc_warlock.hpp
+++ b/engine/class_modules/warlock/sc_warlock.hpp
@@ -459,7 +459,7 @@ public:
     const spell_data_t* chaos_barrage_tick;
     const spell_data_t* chaos_tear_summon; // This only creates the "pet"
     const spell_data_t* rift_chaos_bolt; // Separate ID from Warlock's Chaos Bolt
-    player_talent_t dimension_ripper; // TODO: After 11.1 goes live, removed outdated RNG option and outdated trigger flags
+    player_talent_t dimension_ripper; // TODO: implement the correct proc behavior based on the spell data
 
     player_talent_t decimation; // Crits can proc Soul Fire cooldown reset. Proc chance is not in spell data
     const spell_data_t* decimation_buff;
@@ -807,7 +807,7 @@ public:
 
     // Destruction
     rng_setting_t decimation = { 0.10, 0.10, "decimation" };
-    rng_setting_t dimension_ripper = { 0.05, 0.05, "dimension_ripper" };
+    rng_setting_t dimension_ripper = { 0.0225, 0.0225, "dimension_ripper" };
 
     // Diabolist
 
--- a/engine/class_modules/warlock/sc_warlock_actions.cpp
+++ b/engine/class_modules/warlock/sc_warlock_actions.cpp
@@ -341,35 +341,6 @@ using namespace helpers;
         p()->procs.decimation->occur();
       }
 
-      if ( destruction() && triggers.dimension_ripper && !p()->min_version_check( VERSION_11_1_0 ) && rng().roll( p()->rng_settings.dimension_ripper.setting_value )  )
-      {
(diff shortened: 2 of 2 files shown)


Azevara:
The issue with updating the code to function this way is that it's hardcoding an unexplained value into the operational code. This has the knock-on effects of making spell data overrides have unexpected behavior and also making future maintenance more obscured.

The choice of 1/2ing the rate tracks with the averaged behavior for an accumulator with rng(0,0.05). The fact this value still does not line up means that there must be another issue at play (ICD?) reducing the total proc-rate in game.

If there were no spell data indicating a supposed "known" value, the rng struct in warlock.hpp is one of the few well-defined places for hardcoded values. Both it and linking to traditional spell data in a simple manner are easy enough to override for users wishing to test the effects of a slightly altered proc rate.

millanzarreta:
> The issue with updating the code to function this way is that it's hardcoding an unexplained value into the operational code. This has the knock-on effects of making spell data overrides have unexpected behavior and also making future maintenance more obscured.

I also think the same, which is why in the first PR, instead of hardcoding it, I implemented it as a hotfix. I've seen some cases where the DBC data doesn't match the game's and it's implemented there, but, indeed, this may not be the most appropriate approach for this case.

> The choice of 1/2ing the rate tracks with the averaged behavior for an accumulator with rng(0,0.05). The fact this value still does not line up means that there must be another issue at play (ICD?) reducing the total proc-rate in game.

There's been some past discussion on the Warlock Discord about how the talent proc could work, but no definitive conclusion has been reached. In fact, one of the assumptions is that it uses an accumulator with rng(0,0.05) as you mention, but for this, the effect value in the spelldata should be 4.5, not 5. More testing and data are probably necessary to draw a conclusion.

> If there were no spell data indicating a supposed "known" value, the rng struct in warlock.hpp is one of the few well-defined places for hardcoded values. Both it and linking to traditional spell data in a simple manner are easy enough to override for users wishing to test the effects of a slightly altered proc rate. The only thing that has been made clear in the tests is that the average proc chance is 2.25%.

Although the best thing would undoubtedly be to figure out how it works and implement it that way based on the spell data effect value, in the meantime, it would be best to make that proc chance a fixed 2.25%. I did a new commit modifying `warlock.hpp` to add the expected value there as suggested. There was already an `rng_setting_t dimension_ripper` option in the `warlock.hpp` file that was being used for the previous pre-11.1 version, so I reused it for the new version of Dimension Ripper post-11.1. Additionally, since it's closely related, I followed the suggestion of this comment `TODO: After 11.1 goes live, removed outdated RNG option and outdated trigger flags`, removing the outdated trigger flags and removing the code related to the old behavior of this talent.

I hope I did all properly, but if not, feel free to do modifications / comment what to do.

Azevara:
There's some additional code in the module to let users override rng settings, check what other rng_setting structs are doing and add that where appropriate, then this should be good to go.

millanzarreta:
I don't think that is necessary since I'm reusing an already existing `rng_setting_t`, which was the `rng_setting_t dimension_ripper` and was used for the previous talent chance proc of the old talent version (pre 11.1.0). Therefore, all the code that allows the user to set a custom value for this option is already present.

Just in case, I've tested setting the `rng_dimension_ripper=` option from the input to different values ​​in several tests and have verified that it is being taken into account.

Azevara:
Ah, whoops, that's my bad - mixed up that you were removing old code with adding new stuff. All good.

PR #10219, opened 2025-04-24
[Warlock] Diabolic Ritual and Touch of Rancora talents adjustments

millanzarreta:
This PR aims to implement/adjust several ingame behaviors for the [Diabolic Ritual](https://www.wowhead.com/spell=428514/diabolic-ritual) and [Touch of Rancora](https://www.wowhead.com/spell=429893/touch-of-rancora) talents in the Diabolist hero tree, both for the Destruction and Demonology specs.

- [Diabolic Ritual](https://www.wowhead.com/spell=428514/diabolic-ritual):
  - The [Diabolic Ritual](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;431944:432815:432816) and [Demonic Art](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;428524:432794:432795) buffs are non-cancellable.
  - When the [Diabolic Ritual](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;431944:432815:432816) buff ends, the [Demonic Art](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;428524:432794:432795) buff activates. In simc this was done instantly, while ingame it has a certain delay.
  - The effect of [Demonic Art](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;428524:432794:432795) triggers on buff expiration (either due to time or because it is consumed by casting one of the spells that activate it), but only if the player is in combat.
  - The [Demonic Art](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;428524:432794:432795) effect does not trigger if the buff is replaced by another [Demonic Art](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;428524:432794:432795).
  - There are several spells that consume [Demonic Art](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;428524:432794:432795) and trigger it, with [Rain of Fire](https://www.wowhead.com/spell=5740/rain-of-fire) and [Shadowburn](https://www.wowhead.com/spell=17877/shadowburn) being instant cast, and [Chaos Bolt](https://www.wowhead.com/spell=116858/chaos-bolt) and [Hand of Gul'dan](https://www.wowhead.com/spell=105174/hand-of-guldan) having a certain cast time. Casted spells only trigger [Demonic Art](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;428524:432794:432795) if any [Demonic Art](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;428524:432794:432795) buff is active when the cast begins, and as long as any [Demonic Art](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;428524:432794:432795) buff is still active at the end of the cast.

- [Touch of Rancora](https://www.wowhead.com/spell=429893/touch-of-rancora):
  - Similar to what was discussed above, casted spells affected by [Touch of Rancora](https://www.wowhead.com/spell=429893/touch-of-rancora) ([Chaos Bolt](https://www.wowhead.com/spell=116858/chaos-bolt) and [Hand of Gul'dan](https://www.wowhead.com/spell=105174/hand-of-guldan)) must have one of the [Demonic Art](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;428524:432794:432795) buffs active at the start of the cast to be affected by [Touch of Rancora](https://www.wowhead.com/spell=429893/touch-of-rancora). This means that if a [Demonic Art](https://www.wowhead.com/spells?filter-any=14:14:14;3:3:3;428524:432794:432795) buff is active at the end of the cast but was not active at the start of the cast, the spell will not benefit from [Touch of Rancora](https://www.wowhead.com/spell=429893/touch-of-rancora) (nor will it consume the buff, as discussed in the previous section).
  - Additionally, if the buff was present at the start of the cast but expires before completing the cast, the spell will still be empowered by [Touch of Rancora](https://www.wowhead.com/spell=429893/touch-of-rancora).
  - In what appears to be an ingame bug, when [Hand of Gul'dan](https://www.wowhead.com/spell=105174/hand-of-guldan) hits multiple enemies, [Touch of Rancora](https://www.wowhead.com/spell=429893/touch-of-rancora)'s increased damage effect only applies to one of the targets hit. The target that receives the increased damage may or may not be the primary target, the choice of who receives the increased damage is random.
  - In what may also be an ingame bug, [Touch of Rancora](https://www.wowhead.com/spell=429893/touch-of-rancora)'s damage amp effect is multiplicative for all affected spells except for [Hand of Gul'dan](https://www.wowhead.com/spell=105174/hand-of-guldan), where the damage amp effect appears to be additive. This was already implemented in simc, added to the multiplier after `action_multiplier()`. Although we haven't rechecked the exact correct point at which this effect should be added to the multiplier, after the changes we made in this PR we want to maintain the same behavior as before, which is that this effect is additive after `action_multiplier()`.

Regarding these last two possible bugs, related report here: https://github.com/SimCMinMax/WoW-BugTracker/issues/1331


Azevara:
I believe, rather than setting a bool in schedule_execute, behavior should be captured with a custom state and snapshot_state etc.

Demonbolt and Demonic Core tracking may be a good example for this.

millanzarreta:
I wasn't familiar with the use of these custom states, snapshot_state, etc., but their use certainly seems more appropriate in this case, since I was actually using the booleans as if they were a state (needing to reset them after use and stuff like that).

I've made a new commit using these custom states. After running several tests I always get the same result as with the previous implementation, which had already been tested extensively before, so I think they're working as they should. Additionally, I did several more tests and debugging step-by-step to be sure and everything seems to be OK.

Note that the use of these custom states in this case is somewhat different from the other examples present in the warlock module, since the condition of whether the buffs are present must be checked when the cast begins (`schedule_execute`), and in this function there is usually no valid action_state yet (`s` is a `nullptr`), so we make use of the `pre_execute_state` if necessary in this case.

Let me know if you have any thoughts, comments or suggestions.

PR #10223, opened 2025-04-26
[Mage] Implement Intuition BLP

pinepinepinepinepine:


engine/class_modules/sc_mage.cpp
```diff
@@ -339,6 +339,7 @@ struct mage_t final : public player_t
     buff_t* high_voltage;
     buff_t* impetus;
     buff_t* intuition;
+    buff_t* intuition_tracker;
```
vituscze: There are some use cases for "fake" buffs like this, but for a simple counter, it would be better to just add a new field to the `state_t` object instead.

engine/class_modules/sc_mage.cpp
```diff
@@ -2275,9 +2276,15 @@ struct mage_spell_t : public spell_t
       trigger_frostfire_mastery();
 
     // Needs to be triggered with a delay so that ABar doesn't eat its own proc
-    if ( !background && harmful && rng().roll( p()->talents.intuition->effectN( 1 ).percent() ) )
-      make_event( *sim, [ this ] { p()->buffs.intuition->trigger(); } );
-  }
+    if ( !background && harmful ) {
+      if ( p()->buffs.intuition_tracker->at_max_stacks() ||  rng().roll( p()->talents.intuition->effectN( 1 ).percent() )  ) {
```
vituscze: Bunch of code style violations. `{` goes on a new line, there should be just a single space around operators (you have two after `||`), etc.

engine/class_modules/sc_mage.cpp
```diff
@@ -3513,9 +3520,19 @@ struct arcane_barrage_t final : public dematerialize_spell_t
     // Arcane Charge from the Orb cast increases Barrage damage, but does not change
     // how many targets it hits. Snapshot the buff stacks before executing the Orb.
     snapshot_charges = p()->buffs.arcane_charge->check();
-    if ( rng().roll( snapshot_charges * p()->talents.orb_barrage->effectN( 1 ).percent() ) )
+    if ( rng().roll( snapshot_charges * p()->talents.orb_barrage->effectN( 1 ).percent() ) ) {
       orb_barrage->execute_on_target( target );
 
+      //so i had to repeat this block of code three times (ae echo, orb barrage, sft orb), which is pretty ugly, and i think its preferable to leave it within mage_spell's execute,
```
vituscze: Comments like this should be in the PR, not in source code.

engine/class_modules/sc_mage.cpp
```diff
@@ -3513,9 +3520,19 @@ struct arcane_barrage_t final : public dematerialize_spell_t
     // Arcane Charge from the Orb cast increases Barrage damage, but does not change
     // how many targets it hits. Snapshot the buff stacks before executing the Orb.
     snapshot_charges = p()->buffs.arcane_charge->check();
-    if ( rng().roll( snapshot_charges * p()->talents.orb_barrage->effectN( 1 ).percent() ) )
+    if ( rng().roll( snapshot_charges * p()->talents.orb_barrage->effectN( 1 ).percent() ) ) {
       orb_barrage->execute_on_target( target );
 
+      //so i had to repeat this block of code three times (ae echo, orb barrage, sft orb), which is pretty ugly, and i think its preferable to leave it within mage_spell's execute, 
+      //but i don't know how to cleanly refer to those three as theyre considered within the background. please lemme know if there's a way to refer to a specific action being executed within mage_spell.
```
vituscze: There is (see how we handle Scorch/IB in `handle_hot_streak`), but it would be much easier of you just refactored this into a new method (i.e. something like `trigger_intuition`) somewhere in `arcane_mage_spell_t`/`mage_spell_t`.

engine/class_modules/sc_mage.cpp
```diff
@@ -3513,9 +3520,19 @@ struct arcane_barrage_t final : public dematerialize_spell_t
...
+      if ( p()->buffs.intuition_tracker->up() )  // it's possible, on 10count (for intuition blp), if you barrage + proc orb, without this it'll barr > intuition gained + 0count > orb barrage registers + 1count, but in game, it'll barr at 10count > orb barrage > intuition gained from blp + 0count. same story with ae echo.
```
vituscze: Comments like this are fine, but please format it better (and don't write it all on a single gigantic line). Also, adding note about "same story with ae echo" doesn't help anyone reading the AE echo code understand what's going on.

engine/class_modules/sc_mage.cpp
```diff
@@ -3751,6 +3768,16 @@ struct arcane_explosion_t final : public arcane_mage_spell_t
       p()->buffs.static_cloud->expire();
     p()->buffs.static_cloud->trigger();
 
+    if ( background )  { // echo and recon increments blp, might be clearer to write "type == ae_type::ECHO || type == ae_type::ENERGY_RECON", but "background" is neater, i dunno
```
vituscze: `background` is fine.

engine/class_modules/sc_mage.cpp
```diff
@@ -3513,9 +3520,19 @@ struct arcane_barrage_t final : public dematerialize_spell_t
...
+      //so i had to repeat this block of code three times (ae echo, orb barrage, sft orb), which is pretty ugly, and i think its preferable to leave it within mage_spell's execute, 
+      //but i don't know how to cleanly refer to those three as theyre considered within the background. please lemme know if there's a way to refer to a specific action being executed within mage_spell.
+      if ( p()->buffs.intuition_tracker->up() )  // it's possible, on 10count (for intuition blp), if you barrage + proc orb, without this it'll barr > intuition gained + 0count > orb barrage registers + 1count, but in game, it'll barr at 10count > orb barrage > intuition gained from blp + 0count. same story with ae echo.
+        p()->buffs.intuition_tracker->increment();
+      
+      if ( p()->buffs.intuition_tracker->at_max_stacks() ) {
+        make_event( *sim, [ this ] { p()->buffs.intuition->trigger(); } );
+        p()->buffs.intuition_tracker->expire();
+      }
+    }
```
vituscze: Also, why is this code in ABar when it's the orb that triggers Intuition? It should be there instead.

engine/class_modules/sc_mage.cpp
```diff
@@ -7191,8 +7218,15 @@ struct splinter_t final : public mage_spell_t
...
-      if ( p()->action.spellfrost_arcane_orb && p()->target )
+      if ( p()->action.spellfrost_arcane_orb && p()->target ) {
         p()->action.spellfrost_arcane_orb->execute_on_target( p()->target );
+
+        p()->buffs.intuition_tracker->increment();
+        if ( p()->buffs.intuition_tracker->at_max_stacks() ) { 
+          make_event( *sim, [ this ] { p()->buffs.intuition->trigger(); } );
+          p()->buffs.intuition_tracker->expire();
+        }
+      }
```
vituscze: Again, would be better if this was in the Orb class rather than here.

engine/class_modules/sc_mage.cpp
```diff
@@ -8453,6 +8487,8 @@ void mage_t::create_buffs()
                                       ->set_default_value_from_effect( 1 )
                                       ->modify_default_value( talents.aether_fragment->effectN( 1 ).percent() )
                                       ->set_chance( talents.intuition.ok() );
+  buffs.intuition_tracker         = make_buff( this, "intuition_tracker", find_spell( 384455 ) ) //uses arcane harmony's id, scuffed but works well enough (i think, but it's a HUGE hack-job), im not sure how to set up an imaginary buff without literally adding it into spelldata. previously did use intuition's id, had to change its duration as to not have it expire, but it's also way neater within the html report -- i dunno
```
vituscze: There are ways of creating "fake" buffs like this, you've already discovered that you can manually set stacks. You can do the same thing with duration etc. But as I mentioned earlier, this would work much better as a field in the `state` object.

vituscze (changes requested):
Added some comments. Also, use a meaningful PR/commit name, e.g. `[Mage] Implement Intuition BLP`

pinepinepinepinepine:
Thank you very much, sorry for the informal comments. bye.

engine/class_modules/sc_mage.cpp
```diff
@@ -2471,6 +2464,24 @@ struct mage_spell_t : public spell_t
 
     p()->state.trigger_glorious_incandescence = false;
   }
+
+  void trigger_intuition( bool BLP_exclude_initial )
+  {
+    if ( !p()->talents.intuition.ok() )
```
pinepinepinepinepine, line 2473: I'm not sure if this is necessary since a player will always be talented with intuition. I suppose maybe if you're too low of a level or something? Dunno. I left it in just to be safe.

engine/class_modules/sc_mage.cpp
```diff
@@ -2471,6 +2464,24 @@ struct mage_spell_t : public spell_t
...
+      return;
+
+    const int BLP_threshold { 11 };
```
pinepinepinepinepine: I had used a constant here as opposed to out-right including 'if condition >= 11' for no real reason, it's the only time this variable is used -- was a bit worried that it's somehow possible to get ahold of BLP's cap within intuition's properties. I dunno if this was the right call.

engine/class_modules/sc_mage.cpp
```diff
@@ -574,6 +574,7 @@ struct mage_t final : public player_t
     bool heat_shimmer;
     int embedded_splinters;
     int magis_spark_spells;
+    int intuition_BLP_count;
```
vituscze: IIRC BLP is not capitalized in simc source (i.e. it would be `intuition_blp_count`), but that's really just a nitpick.

engine/class_modules/sc_mage.cpp
vituscze: Yeah, this is fine. No reason to break weird/low level sims if we can help it.

engine/class_modules/sc_mage.cpp
vituscze: I don't think it's in spell data, so this is about as good as it gets, maybe make it constexpr while you're at it (i.e. `constexpr int blp_threshold = 11;`)

engine/class_modules/sc_mage.cpp
```diff
@@ -2464,6 +2464,24 @@ struct mage_spell_t : public spell_t
...
   }
+
+  void trigger_intuition( bool BLP_exclude_initial )
+  {
+    if ( !p()->talents.intuition.ok() )
+      return;
+
+    const int BLP_threshold { 11 };
+    // If an action and it's direct after-effect can increment intuition's BLP,
+    // exclude the second incoming incrementation if it already gained intuition from its previous associated action
```
vituscze: You're trying to explain what `BLP_exclude_initial` is for, yes? If so, it would be better to have it in front of the method rater than somewhere in the middle.

engine/class_modules/sc_mage.cpp
```diff
@@ -3398,8 +3416,11 @@ struct arcane_orb_bolt_t final : public arcane_mage_spell_t
 
 struct arcane_orb_t final : public arcane_mage_spell_t
 {
-  arcane_orb_t( std::string_view n, mage_t* p, std::string_view options_str, ao_type type = ao_type::NORMAL ) :
-    arcane_mage_spell_t( n, p, p->find_specialization_spell( "Arcane Orb" ) )
+  ao_type type;
```
vituscze: This can be `const`.

engine/class_modules/sc_mage.cpp
```diff
@@ -3515,7 +3539,7 @@ struct arcane_barrage_t final : public dematerialize_spell_t
     snapshot_charges = p()->buffs.arcane_charge->check();
     if ( rng().roll( snapshot_charges * p()->talents.orb_barrage->effectN( 1 ).percent() ) )
       orb_barrage->execute_on_target( target );
-
+
```
vituscze: Trailing whitespace here.

vituscze (changes requested):

PR #10230, opened 2025-05-02
Immediately enter Carved Blazikon Wax light

silverwind:
I tested around ~30~ 50 procs and the candle always spawned within player range, so change these constants to make the "In Light" buff 100% uptime, assuming the player does not move.

Sample spawns across my testing:

furthest case:
image

closest case:
image


silverwind:
I see this project is not welcoming contribution, I'll make sure to never try again.

PR #10238, opened 2025-05-06
[Mage] Fix Intuition's Base Proc Chance

pinepinepinepinepine:
Intuition's base proc only applies to arcane spells (as well as Arcane Explosion), and doesn't trigger from shifting power.

Just moved the previous 'trigger_intuition()' into arcane_mage_spell_t::execute().

### DIFF
--- a/engine/class_modules/sc_mage.cpp
+++ b/engine/class_modules/sc_mage.cpp
@@ -2603,7 +2603,7 @@ struct arcane_mage_spell_t : public mage_spell_t
     if ( p()->talents.intuition.ok() && !p()->buffs.intuition->check() && triggers.intuition )
     {
       constexpr int blp_threshold = 11;
-      constexpr int base_proc_chance = 0.01;
+      constexpr double base_proc_chance = 0.01;
 
       p()->state.intuition_blp_count += 1;
       if ( p()->state.intuition_blp_count >= blp_threshold
--- a/engine/class_modules/sc_mage.cpp
+++ b/engine/class_modules/sc_mage.cpp
@@ -1924,7 +1924,7 @@ struct mage_spell_t : public spell_t
   {
     bool chill = false;
     bool clearcasting = false;
-    bool intuition_blp = false;
+    bool intuition = false;
     bool from_the_ashes = false;
     bool frostfire_infusion = true;
     bool frostfire_mastery = true;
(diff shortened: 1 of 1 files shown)


pinepinepinepinepine:
Hey. 
Terribly, terribly sorry for the initial PR I put in regarding intuition, we were over-simming by a decent amount because, apparently, intuition's base proc chance is 1%, not 5%. 
I had just fixed some minor issues about what spells can trigger the BLP/proc chance (specifically Shifting Power, but also Fire Blast + Frostbolt + the utility group spells -- they all cannot trigger it). Regardless, I'll be more diligent next time, sorry sorry.

I did leave the 5% chance to proc it if the input specifies bugs=0 for no real reason, but I figured it's kind of a neat thing to sim -- I dunno. Anyways, thanks.

engine/class_modules/sc_mage.cpp
```diff
@@ -1924,6 +1924,7 @@ struct mage_spell_t : public spell_t
   {
     bool chill = false;
     bool clearcasting = false;
+    bool intuition_blp = false;
```
vituscze: Why is this called `intuition_blp` if it also controls the non-BLP triggers?

engine/class_modules/sc_mage.cpp
```diff
@@ -2620,6 +2596,25 @@ struct arcane_mage_spell_t : public mage_spell_t
...
+  {
+    mage_spell_t::execute();
+
+    if ( p()->talents.intuition.ok() && !p()->buffs.intuition->check() && triggers.intuition_blp )
+    {
+      constexpr int blp_threshold = 11;
+
+      p()->state.intuition_blp_count += 1;
+      if ( p()->state.intuition_blp_count >= blp_threshold
+        || ( !background && rng().roll( p()->bugs ? 0.01 : p()->talents.intuition->effectN( 1 ).percent() ) ) )
```
vituscze: I take issue with how `bugs` is used here. The tooltip states it's 5%, with BLP alone it's 9% (and more if you include the base proc chance). Reducing base proc chance to account for extra procs from BLP is fairly normal. The bug here is, arguably, that even with the reduced proc chance the overall proc rate is FAR above the stated 5%.

So either we treat the entire BLP system as a bug or none of it. Picking out just the base chance doesn't make sense here.

vituscze (changes requested):

pinepinepinepinepine:
Added a new commit, had copy and pasted the 'constexpr int' above when I wanted to initialize it w/ a double. sorry!

engine/class_modules/sc_mage.cpp
```diff
@@ -2620,6 +2596,26 @@ struct arcane_mage_spell_t : public mage_spell_t
...
   }
 
+  void execute() override
+  {
+    mage_spell_t::execute();
+
+    if ( p()->talents.intuition.ok() && !p()->buffs.intuition->check() && triggers.intuition )
+    {
+      constexpr int blp_threshold = 11;
+      constexpr double base_proc_chance = 0.01;
```
vituscze, line 2607: Maybe add a comment here explaining why we're not using the spell data value, otherwise it looks ready to merge.

vituscze (changes requested):

PR #10252, opened 2025-05-08
[Gear] Set ICD for Moxie trinket activation buff proc

millanzarreta:
This PR is a small change that sets an internal cooldown for the [effect (471548)](https://www.wowhead.com/spell=471548/mugs-moxie-jug) that triggers the activation (the first stack) of [Moxie Frenzy buff (474285)](https://www.wowhead.com/spell=474285/moxie-frenzy) from the [Mug's Moxie Jug trinket](https://www.wowhead.com/item=230192/mugs-moxie-jug).

Once the buff is active, the proc effect that increases the stacks will be performed by another [effect (474376)](https://www.wowhead.com/spell=474376/mugs-moxie-jug) (the hidden buff driver), and the main activation proc [effect (471548)](https://www.wowhead.com/spell=471548/mugs-moxie-jug) will not occur in the meantime. In the current implementation of simc, the activation proc effect continues to occur even if the buff is active, adding a stack if the proc's rppm rng results in activation.

It has been verified ingame that the main activation [effect (471548)](https://www.wowhead.com/spell=471548/mugs-moxie-jug) doesn't occur while the buff is active, or if it does, it does so silently without anything happening. We know this because no increase in the number of buff stacks has been detected outside the hidden buff driver's internal cooldown time window (750ms). We also know that the proc that triggers the buff's initial effect doesn't share an ICD with the hidden buff driver: there may be less than a 750ms difference between stacks 1 and 2 (1 from main activation effect, 2 from hidden debuff driver), but not for any other stack increment (e.g.: 2 to 3, 4 to 5, 5 to 6, 8 to 9, ...).

Therefore, this PR sets an ICD for the main item activation effect equal to the buff duration (currently 15 seconds) in simc, thus preventing the main item activation proc effect from being triggered while the buff is active.

### DIFF
--- a/engine/player/unique_gear_thewarwithin.cpp
+++ b/engine/player/unique_gear_thewarwithin.cpp
@@ -6895,6 +6895,7 @@ void mugs_moxie_jug( special_effect_t& effect )
   auto second_proc = new dbc_proc_callback_t( effect.player, *buff_driver );
   second_proc->activate_with_buff( crit_buff, true );
 
+  effect.cooldown_    = effect.duration();
   effect.proc_flags_  = PF_ALL_DAMAGE | PF_ALL_HEAL;
   effect.proc_flags2_ = PF2_ALL_HIT | PF2_ALL_CAST | PF2_PERIODIC_DAMAGE | PF2_PERIODIC_HEAL | PF2_LANDED;
   effect.custom_buff  = crit_buff;


Saeldur:
I *think* how this works ingame is it can still proc but just attempts to add a stack if its already out. Testing it is a pain since they reworked it several times on PTR and now it stacks up on a lot of background actions, you'd need to use a periodic effect but not have any buffs/other events that can trigger from that to test it over along period of time i think.

millanzarreta:
After performing several tests, I have verified that the trinket's [main driver](https://www.wowhead.com/spell=471548/mugs-moxie-jug) indeed has a 15-second ICD after activation, preventing the driver from being triggered again while the [buff](https://www.wowhead.com/spell=474285/moxie-frenzy) is active.

The trinket has two drivers: the [main driver](https://www.wowhead.com/spell=471548/mugs-moxie-jug) is the one that initially activates the buff, hereafter referred to as **MD**; while the [child driver or secondary driver](https://www.wowhead.com/spell=474376/mugs-moxie-jug) is the one that increases the stacks while the buff is active, hereafter referred to as **SD**.

Taking advantage of the fact that the MD can be triggered by DoT damage, while the SD cannot, the following tests were performed:

- With a priest, the [Shadow Word: Pain DoT](https://www.wowhead.com/spell=202896/shadow-word-pain) is renewed on a dummy every 14 seconds. This means that if the MD indeed has a 15-second ICD, the [Moxie Frenzy buff](https://www.wowhead.com/spell=474285/moxie-frenzy) can only reach a maximum of 2 stacks: one from the activation of the MD, and one from the renewal of the DoT (done by the SD). On the other hand, if the MD doesn't have any type of ICD, the buff could occasionally reach 3 or even more stacks, since if it activates sporadically during these 15 seconds, the buff would gain an extra stack. [Here's the test log](https://www.warcraftlogs.com/reports/H4CvNph8JGWjAVt9?fight=last&type=auras&source=1&view=events&ability=474285) for 1 hour on a dummy, where you can see that the buff never gains a third stack, showing that the MD does have an ICD that prevents it from activating during the buff.

- We also ran another additional test. In this test, we constantly refresh the [Shadow Word: Pain DoT](https://www.wowhead.com/spell=202896/shadow-word-pain), and when the MD is activated, either by the direct damage or by the DoT damage, we refresh the DoT one last time to establish it to the max duration and then we perform a `/cancelaura`, then we stop all actions for the remaining duration of the DoT (more than 15 seconds). This way, we can see if the DoT, which can only trigger the MD, ever triggers the trinket again within 15 seconds of its last activation. In short, by canceling the buff and stopping performing actions, leaving only the DoT, we see if the MD can activate the trinket again within 15 seconds to check if it has an ICD. [Here's the log](https://www.warcraftlogs.com/reports/YbBtan4HC2Q8vpdZ?fight=last&type=auras&source=1&ability=474285&view=events&pins=2%24Separate%24%23244F4B%24auras-gained%24-1%240.0.0.Any%2415198769.0.0.Priest%24true%240.0.0.Any%24true%24474285%2433) of the 2-hour test on a dummy, where you can see that the MD never activates within a period of less than 15 seconds, thus showing that it does have an ICD of the same duration as the buff. In fact, if we order the times between MD activations from shortest to longest, the times are as follows `[15.092, 15.120, 15.154, 15.182, 15.189, 15.235, 15.301, 15.780, 15.806, 15.882, 16.152 ...]`, clearly showing that it can be activated again at any time just after the 15 seconds ICD, but never before.

- Finally, although it's somewhat less conclusive, I'm also leaving [here the log](https://www.warcraftlogs.com/reports/nBpg1TPZ4yRxXqGz?boss=-3&difficulty=0&type=auras&source=5&ability=474285&view=events&pins=2%24Off%24%23244F4B%24auras-gained%24-1%240.0.0.Any%24163876261.0.0.Warlock%24true%240.0.0.Any%24true%24474285%2459) of what I mentioned in my original comment. This is a normal log of several M+ doing normal actions. You can see that the time between stack 1 (activated by the MD) and stack 2 (activated by the SD) of the buff can be any time, since the MD doesn't share the SD's 750ms ICD. However, if you check all the other stack gains (e.g.: 2 to 3, 4 to 5, 5 to 6, 8 to 9, ...), they always occur at least 750ms apart, which indicates that they were all activated by the SD and never by the MD. If the MD could be activated while the buff was active and didn't have a 15-second ICD, there would occasionally be some stack gains with a shorter time than 750ms, since, as we've already stated before observing the times between stack 1 and 2, the MD doesn't share the 750ms ICD of the SD, but this doesn't happen in the entire log.

Finally, I'd like to point out that if you want to run any of these tests yourself, it's difficult to get a proper setup. Although DoT damage doesn't trigger the SD, there are thousands of background effects (some even from hidden auras) that seem to trigger it, and the same for MD. With the Priest, I had to use the Holy spec and [these specific talents](https://www.wowhead.com/talent-calc/priest/holy/archon/DAMFFVVVFVVVVUgCABOVUVQVBEVVFQKAVBUEFVVUBA), but, for example, when using the Disci spec, some unknown periodic background effect was constantly triggering the MD and SD, and on Shadow, there was also some kind of unknown periodic background effect that triggered the MD and SD approximately every 5 seconds.

After all these tests, I believe we can conclude that the MD has a 15-second ICD after activation, so I consider the changes in this PR to be adequate to reflect this behavior in simc.

Issue #10116, opened 2025-03-25
[Trinket] Mug's Moxie Jug


Mlklock:
### Describe the bug
Mug's Moxie Jug procs and trigger stacks beyond just spellcasting, this means a spec like Demonology in sims will generate less stacks than its Ingame Counterpart. 

### Expected behavior
> The behavior in game seems to grant stacks from things like Apply Buff (Demonic Cores, Imp Gang Boss transformation), Applying Debuffs (Wicked Maw debuff), Spell Casting (Shadow Bolt, Call Dreadstalkers), Damage Events (Shadow bolt, Hand of Guldan).
>
> There are some background and other effects who also grant it, but those are beyond the scope of simc (defensives, summoning pet, some skills). 
>
> Of course, the granting of stacks does seems to obey the 0.75s ICD, but it is possible to Proc and Gain a stack in quick sucession and at a interval smaller than those 0.75s (screenshot/log provided in this report). 

### To Reproduce
>
> Proc the Trinket/Gain Stacks of the trinket across different actions and then check a combatlog in warcraftlogs to compare.

### Additional information
> Log where nothing is done but it procs and gains stacks from simply Wild Imps spawning from the Inner Demon Talent /becoming Imp Gang Bosses : https://www.warcraftlogs.com/reports/dHhLaDGnyfQR9gbp?fight=14&type=summary&source=16&pins=0%24Separate%24%23244F4B%24auras-gained%240%240.0.0.Any%240.0.0.Any%24true%24152690636.0.0.Warlock%24false%24474285&start=7632321&end=7671378&view=events 
>
>Log where it procs from both the Shadowbolt Cast at 1:16  bu also from its hit at 1:17.6 : https://www.warcraftlogs.com/reports/PNcwm9BMgDTZzjr8?fight=10&type=summary&source=16&pins=0%24Separate%24%23244F4B%24auras-gained%240%240.0.0.Any%240.0.0.Any%24true%24152690636.0.0.Warlock%24false%24474285%5E2%24Off%24%23909049%24expression%24not+ability.name+%3D+%22firebolt%22&view=events 
>
>Screenshot of Log 1 where Wild Imp 9 spawns and gains Imp Gang Boss buff from the warlock activating it and stacking it to 2,  https://i.imgur.com/7UUZYgE.png
>
> Screenshot of Log 2 where Shadowbolt cast and damage grant stacks https://i.imgur.com/6Gl3zeJ.png
>
[...]

PR #10267, opened 2025-05-16
[Gear] Alternative option for Moxie trinket buff stacking

millanzarreta:
The [Mug's Moxie Jug](https://www.wowhead.com/item=230192/mugs-moxie-jug) trinket presents some [issues](https://github.com/simulationcraft/simc/issues/10116) in Simc. This trinket has two drivers: the [main driver](https://www.wowhead.com/spell=471548/mugs-moxie-jug), which is responsible for processing the buff activation (15sec ICD), and the [secondary driver](https://www.wowhead.com/spell=474376/mugs-moxie-jug), which increases the stacks while the buff is active (750ms ICD). Both effects can be triggered by a multitude of player actions: healing effects, damage effects, energize effects, some gain/drop/refresh auras, and even some triggers from periodic hidden auras/spells that don't appear in the combatlog or the logs.

Simc doesn't include the implementation of many of these effects, which causes some issues with the trinket. For the main driver, this isn't a big problem, as it's RPPM-based. In Simc it has been established for main driver that it can be triggered by many casts, damage effects, and healing effects. This should be sufficient for it to trigger periodically based on RPPM, so no major changes are needed. The problem is the secondary driver that increases stacks while the buff is present, which is not RPPM-based and always increases the buff by one stack with any of its activations (is 100% chance, so it is activating all the time with all these effects/spells, as long as they occur after its 750ms ICD). In Simc, many of these effects are not implemented, so the buff doesn't increase stacks as quickly as it does ingame for many specs, and that's the problem.

Commit 46eea1600ecc42f00279902b9e6f17f44a1fb907 has attempted to make the secondary driver increase stacks more quickly, including most damage and healing casts and effects. This isn't entirely accurate, but it makes its effectiveness more similar to ingame.

While this 46eea1600ecc42f00279902b9e6f17f44a1fb907  commit is a great improvement, it may still be insufficient for some specs or classes. For example, for the Demonology Warlock in ST and 5T, the following average uptimes for each buff stack is currently being simulated (normalizing in this context means transforming average buff-stack uptimes relative to the entire fight, which is what Simc reports, to average buff-stack uptimes relative to only when the buff is active; we only really care about the normalized values):
- <ins>**ST (1 Target) (simc current implementation):**</ins>
imagen
- <ins>**5 Targets (simc current implementation):**</ins>
imagen

It can be seen that it rarely reaches 14 or 15 stacks in sims. However, looking at the logs, we can see that the ingame effect is greater. [Here](https://www.warcraftlogs.com/reports/ca2G3fRLMPDWj9Nd?fight=last&source=4&type=auras&target=4&ability=474285&view=events) is an example of a 1-hour log on a ST dummy (ST Raid Talents and rotation), and [here](https://www.warcraftlogs.com/reports/TmPkbpHxM9h2jZNF?fight=last&type=auras&source=1&target=1&ability=474285&view=events) is another 1-hour log on 5T dummies (M+ AoE Talents and rotation). The normalized average uptimes of this trinket for these logs are the following:
- <ins>**1T (extracted from [logs](https://www.warcraftlogs.com/reports/ca2G3fRLMPDWj9Nd?fight=last&source=4&type=auras&target=4&ability=474285&view=events)):**</ins>
imagen
- <ins>**5T (extracted from [logs](https://www.warcraftlogs.com/reports/TmPkbpHxM9h2jZNF?fight=last&type=auras&source=1&target=1&ability=474285&view=events)):**</ins>
imagen

We can see that the average buff-stack uptime difference between the current simulations and the ingame results (logs) is significant, at least for this spec.

This PR adds an additional and completely optional setting that lets you configure the trinket buff to gain stacks at time intervals instead of tying stack gains to player actions. This new mode is more difficult to use and requires more setup work, but allows for more accurate results when configured properly. To activate this new mode, the user must set the new option `thewarwithin.moxie_frenzy_buff_stack_mode=timed`. If set to `normal`, the functionality will remain exactly the same as the current implementation without any changes. `normal` mode is the default mode, so the new mode would only be activated if the user manually sets this option.

In order for the user to set the time interval between stack gains (only for the new `timed` mode) they must use the new option `thewarwithin.moxie_frenzy_buff_stack_gain_interval`. Because of how the ingame trinket works, the first activation (stack 1) is done by the main driver, and the second activation (stack 2) and subsequent ones are done by the secondary driver, and these two drivers do not share the same ICD. This is why ingame the time of this buff from stack 1 to stack 2 is much faster than the rest of the stack time intervals. This is why we also included an additional option `thewarwithin.moxie_frenzy_buff_stack_first_tick` that allows you to set the time between the interval of this first-stack-to-the-second. Finally, to give them a less rigid timings more similar to the stack gains intervals of the game, you can also set a standard deviation for these two intervals (`thewarwithin.moxie_frenzy_buff_stack_gain_interval_stddev` and `thewarwithin.moxie_frenzy_buff_stack_first_tick_stddev` options), so that the intervals vary slightly in each tick depending on rng. You can always set these values to `0` for fixed time intervals.

With this new mode, a user who knows approximately how often the trinket gains stacks for their character (this data can be obtained by analyzing logs, for example) can set these new option values to the desired ones, thus achieving a simulation of the trinket that more accurately reflects what happens to their character ingame. This trinket is one of the best for many classes, and many people now with dinars want to simulate it to compare with others. Therefore, with this new option, an advanced user can better test its actual effectiveness on their character. Also, remember that these options are completely optional and disabled by default, so for normal users, the current simulation implementation will continue to be used, but for those who specifically want to use them, it gives them the opportunity to achieve more consistent results with those observed ingame.

At last, since we previously discussed the case of the demonology warlock for 1T and 5T, I'm posting here the simulation results after using this new option with an appropriate configuration for this character (according to data extracted from the logs):

<ins>**Configuration used in Simc (for both, 1T and 5T):**</ins>
```
thewarwithin.moxie_frenzy_buff_stack_mode=timed
thewarwithin.moxie_frenzy_buff_stack_gain_interval=0.900
thewarwithin.moxie_frenzy_buff_stack_gain_interval_stddev=0.250
thewarwithin.moxie_frenzy_buff_stack_first_tick=0.025
thewarwithin.moxie_frenzy_buff_stack_first_tick_stddev=0.350
```
<ins>***Results:**</ins>
- <ins>**1T (simc new trinket mode):**</ins>
imagen
- <ins>**5T (simc new trinket mode):**</ins>
imagen

We can see that the normalized average uptimes are now more consistent with the ones seen in the logs. These examples were for the Demonology Warlock, for which the "normal" implementation doesn't give particularly accurate results. It's possible that for other specs the "normal" implementation will be sufficient and the new mode may not be necessary, but there are surely some other specs that could also benefit from this new mode (with its correspondingly appropriate configuration) to achieve more accurate results.

### DIFF
--- a/engine/player/player.cpp
+++ b/engine/player/player.cpp
@@ -12922,6 +12922,11 @@ void player_t::create_options()
   add_option( opt_string( "thewarwithin.jastor_diamond_ally_stat", thewarwithin_opts.jastor_diamond_ally_stat ) );
   add_option( opt_float( "thewarwithin.suspicious_energy_drink_bonus_chance",
                          thewarwithin_opts.suspicious_energy_drink_bonus_chance, 0, 1 ) );
+  add_option( opt_string( "thewarwithin.moxie_frenzy_buff_stack_mode", thewarwithin_opts.moxie_frenzy_buff_stack_mode ) );
+  add_option( opt_timespan( "thewarwithin.moxie_frenzy_buff_stack_gain_interval", thewarwithin_opts.moxie_frenzy_buff_stack_gain_interval, 0_ms, 14_s ) );
+  add_option( opt_timespan( "thewarwithin.moxie_frenzy_buff_stack_gain_interval_stddev", thewarwithin_opts.moxie_frenzy_buff_stack_gain_interval_stddev, 0_ms, 14_s ) );
+  add_option( opt_timespan( "thewarwithin.moxie_frenzy_buff_stack_first_tick", thewarwithin_opts.moxie_frenzy_buff_stack_first_tick, 0_ms, 14_s ) );
+  add_option( opt_timespan( "thewarwithin.moxie_frenzy_buff_stack_first_tick_stddev", thewarwithin_opts.moxie_frenzy_buff_stack_first_tick_stddev, 0_ms, 14_s ) );
   add_option( opt_timespan( "thewarwithin.additional_gcd_time", thewarwithin_opts.additional_gcd_time, 0_s, 10_s ) );
 }
 
--- a/engine/player/player.hpp
+++ b/engine/player/player.hpp
@@ -886,6 +886,16 @@ struct player_t : public actor_t
     player_option_t<std::string> mister_locknstalk_mode = "dynamic";
     player_option_t<std::string> jastor_diamond_ally_stat = "none";
     double suspicious_energy_drink_bonus_chance           = 0;
+    // Which method to use to determine Moxie Frenzy buff stack gain. Accepts "normal" (default) or "timed"
+    // "Normal": Most casts, damage effects and healing effects increase the buff stack
+    // "Timed": The buff gain a stack periodically based on custom 'interval' and 'first_tick' times
+    // Whichever method is selected, the ICD of buff stack gain will always be respected
+    player_option_t<std::string> moxie_frenzy_buff_stack_mode = "normal";
+    timespan_t moxie_frenzy_buff_stack_gain_interval          = 900_ms;
+    timespan_t moxie_frenzy_buff_stack_gain_interval_stddev   = 250_ms;
+    timespan_t moxie_frenzy_buff_stack_first_tick             = 25_ms;
+    timespan_t moxie_frenzy_buff_stack_first_tick_stddev      = 350_ms;
+    // Additional GCD time
     timespan_t additional_gcd_time                        = 0_s;
   } thewarwithin_opts;
 
--- a/engine/player/unique_gear_thewarwithin.cpp
+++ b/engine/player/unique_gear_thewarwithin.cpp
@@ -6885,15 +6885,46 @@ void mugs_moxie_jug( special_effect_t& effect )
                        ->add_stat_from_effect( 1, effect.driver()->effectN( 1 ).average( effect ) )
                        ->set_refresh_behavior( buff_refresh_behavior::DISABLED );
 
-  auto buff_driver          = new special_effect_t( effect.player );
-  buff_driver->name_str     = util::tokenize_fn( effect.driver()->effectN( 1 ).trigger()->name_cstr() );
-  buff_driver->spell_id     = effect.driver()->effectN( 1 ).trigger_spell_id();
-  buff_driver->proc_flags2_ = PF2_ALL_HIT | PF2_ALL_CAST | PF2_PERIODIC_DAMAGE | PF2_PERIODIC_HEAL;
-  buff_driver->custom_buff  = crit_buff;
-  effect.player->special_effects.push_back( buff_driver );
+  if ( !( effect.player->thewarwithin_opts.moxie_frenzy_buff_stack_mode == "timed" ) )
+  {
+    auto buff_driver          = new special_effect_t( effect.player );
+    buff_driver->name_str     = util::tokenize_fn( effect.driver()->effectN( 1 ).trigger()->name_cstr() );
+    buff_driver->spell_id     = effect.driver()->effectN( 1 ).trigger_spell_id();
+    buff_driver->proc_flags2_ = PF2_ALL_HIT | PF2_ALL_CAST | PF2_PERIODIC_DAMAGE | PF2_PERIODIC_HEAL;
+    buff_driver->custom_buff  = crit_buff;
+    effect.player->special_effects.push_back( buff_driver );
 
-  auto second_proc = new dbc_proc_callback_t( effect.player, *buff_driver );
-  second_proc->activate_with_buff( crit_buff, true );
+    auto second_proc = new dbc_proc_callback_t( effect.player, *buff_driver );
+    second_proc->activate_with_buff( crit_buff, true );
+  }
+  else
+  {
+    const timespan_t opt_s_interval = effect.player->thewarwithin_opts.moxie_frenzy_buff_stack_gain_interval > effect.trigger()->internal_cooldown()
+                                          ? effect.player->thewarwithin_opts.moxie_frenzy_buff_stack_gain_interval
+                                          : effect.trigger()->internal_cooldown();
+    const timespan_t opt_s_interval_stddev = effect.player->thewarwithin_opts.moxie_frenzy_buff_stack_gain_interval_stddev;
+    const timespan_t opt_s_first_tick = effect.player->thewarwithin_opts.moxie_frenzy_buff_stack_first_tick;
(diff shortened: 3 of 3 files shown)


nyterage:
What about demonology is causing the stacks to not increment properly? is it pet summon spells? pet casts themselves? if those things are triggering this, that means they can also trigger other things, and its likely so many more things are inaccurate due to not being triggered by the module. I feel like that should be looked into more thoroughly.

millanzarreta:
> What about demonology is causing the stacks to not increment properly? is it pet summon spells? pet casts themselves? if those things are triggering this, that means they can also trigger other things, and its likely so many more things are inaccurate due to not being triggered by the module. I feel like that should be looked into more thoroughly.

This isn't just something unique to the Demonology Warlock. I've used this class and spec just as an example, and because it's the one I'm currently using and one of the ones I know is affected by this issue, but the fact that the trinket undersims is something that surely also affects other specs/classes. Ingame there are MANY things (a lot of them not implemented in simc) that trigger the effects of this trinket, and that affect all classes. These things include, for example, many leech effects, energize effects, auras/effects/spells that trigger when taking damage, the gain/change/loss of stacks of some auras, and even some hidden auras/effects that run periodically in the background without leaving a trace in either the combat log or the logs. For obvious reasons, many of these effects aren't implemented in Simc.

For example, the Discipline/Holy Priest with the Oracle hero talents has some kind of periodic hidden aura (which doesn't appear in either the combat log or the logs) that triggers this trinket repeatedly, as you can see here (where only by being in combat without do any actions triggers the trinket constantly):

https://github.com/user-attachments/assets/f222bebf-4059-4d33-8825-9724b2b1fa9f

A similar thing happens with other spells/talents/effect from other classes/specs. For example, the Shadow Priest also has some kind of periodic hidden aura that triggers the trinket, but in this case only every 5 seconds.

In the specific case of the Demonology Warlock, in the discussion [here](https://github.com/simulationcraft/simc/issues/10116) are more details on some of the effects that trigger it. Among them, I can point out [Soul Leech](https://www.wowhead.com/spell=108366/soul-leech), or [this](https://www.wowhead.com/spell=387549/infernal-command) aura applied to demons that still exists in the game even though it no longer has an effect (its effect was removed, but instead of removing the spell, they left it with 0% value). The gain/loss of [Demoncores](https://www.wowhead.com/spell=264173/demonic-core) can also be affected. Other examples are Wild Imps spawning from the [Inner Demons](https://www.wowhead.com/spell=267216/inner-demons) Talent or they becoming [Imp Gang Bosses](https://www.wowhead.com/spell=387445/imp-gang-boss). There are even some pet effects that, ingame, even though the pet applies them, actually count as being from the player, so they also trigger the trinket. Some of these things aren't implemented in Simc or aren't implemented in the exact same way as in the game. But this doesn't really matter that much, implementing they more accurately in the module would not improve the result much (besides being very hard work), because with effects that don't usually affect dps and aren't worth implementing (like [Soul Leech](https://www.wowhead.com/spell=108366/soul-leech)) and also some unknown periodic hidden auras/effects, it's actually very hard to the stacks to be incremented in simc (with the normal implementation) at the same rate as ingame. That's why I find the new optional mode implemented in this PR convenient, as it, while not exact and requiring more customized configuration, allows for more accurate results.

One question remains. _For which specs does the current normal implementation of this trinket in simc work very similarly to the ingame?_ Well, is a hard question and requires testing, but I guess that the simc results are similar to the ingame ones for those specs that don't have too many hidden auras/effects ingame triggering the trinket and the buff stacking rely more on casts and damage/healing effects. I also assume that the greater the number of damage/healing events and/or casts (whether dots or direct damage) and the higher the APM the spec has, the more likely the activations will be closer to the shortest possible theoretical time, which is the ICD of 750ms, and the closer they will be to their actual ingame effect.

nyterage:
While i understand its a lot of work to properly model things in simc for each classes module, stuff like this is why its important to do so, even if its a lot of work. This wont be the only trinket to ever work like this, and its likely the inaccuracies are affecting other things as well. All this PR does is provide an overly complex solution to a symptom, rather than trying to solve the root of the issue.

millanzarreta:
> While i understand its a lot of work to properly model things in simc for each classes module, stuff like this is why its important to do so, even if its a lot of work. This wont be the only trinket to ever work like this, and its likely the inaccuracies are affecting other things as well. All this PR does is provide an overly complex solution to a symptom, rather than trying to solve the root of the issue.

While I agree that it's always best to have the most faithful implementation possible in the modules, this doesn't really matter too much for this issue. Ingame, Blizzard has too many hidden things running periodically in the background that affect this trinket's trigger, and we can't model some of them in simc because we don't even know what they are. I don't think it's entirely possible to solve the root of this issue.

To me, this PR's solution is a good alternative for those who want a more precise way to test the power of this trinket. Since it's an alternative solution not entirely based on how the trinket actually works ingame, it's not the default option, and only the user who specifically chooses to use these options will end up using it. Furthermore, this new mode is completely optional, and if the new option isn't established, it will function just as it did in the current simc normal implementation. It only adds, not subtracts. If a sufficiently accurate normal implementation of trinket were to be achieved in the future (which I doubt is possible), this would still be the default so no problem; and in that hypothetical case, this implementation could even be removed. In the meantime this alternative is useful.

I do not agree that this PR is a complex solution: it's really just about setting a timing for the "tick rate" at which stacks are gained, simply as that. There's also an option for the first tick to have a different timing, but this could be ignored if desired and left at the default value. The `stddev` options could also be ignored or set to 0, I simply added them to include a bit of random variance in the stack increment (as many other timing options in simc also include), but it's not really that necessary.

nyterage:
at the very least, id recommend adding the pet summon spells for pets that trigger on proc, rather than having their own action. and see how that pans out for demo. probably still wont be 100% of the way there, but those i know from experience do trigger a lot of other item effects, and it being missing could be impacting more than just this trinket.

millanzarreta:
I'll try to analyze the wl demonology module to see how it can be improved (within reason). Let's see if we can at least improve it a bit by adding some pet spells, or, as you mentioned, pet summoning spells. There are other things like hidden auras or healing effects like Soul Leech that would be very difficult to test/add, so they'll probably be left out.

nyterage:
Pet summon spells should be relatively easy at least, heres a few examples of proc pet spawn spells i wrote in DK. 
https://github.com/simulationcraft/simc/blob/8b087976a56925fa2054642426843560ce62ff17/engine/class_modules/sc_death_knight.cpp#L5103

Saeldur:
No longer necessary due to https://github.com/simulationcraft/simc/commit/53ce36e879c3be599b82c27b355d8b09d6c035f7 however this is still a large workaround hack.
Having the summon spells added to Warlock in the future would help prevent workarounds being necessary.

PR #10271, opened 2025-05-18
[Marksmanship] - Change lunar storm tracking aura

syrifgit:
This change swaps references of `buff.lunar_storm_ready.up` to `buff.lunar_storm_cooldown.down`. While these conditions are logically equivalent, using `buff.lunar_storm_cooldown.down` is significantly less complex for the downstream Hekili addon to manage in-game. I understand it may not be accepted, but the change would be much appreciated.

Of note: This condition should be combined with hero tree checks for sentinel to avoid false positives on dark ranger when the aura is down, but all 4 entries _already_ have a hero tree check.


Pewtro:
No promise this change will be permanent as people work on the APLs

PR #10281, opened 2025-05-22
[config] Remove CISO646

renanthera:
Clang has begun to warn about the deprecation of CISO646 in C++17, and its removal for C++20.

This PR exists to test whether or not CI passes for other compiler configurations, and to remove it if MSVC supports alternate operator representations with the configuration we currently provide.

### DIFF
--- a/engine/config.hpp
+++ b/engine/config.hpp
@@ -82,19 +82,6 @@
 
 
 // ==========================================================================
-// Compiler Workarounds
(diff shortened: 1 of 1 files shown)


renanthera:
MSVC passed, suggesting either no alternative operator representations are used, it's resolved in the compiler or msvc is configured with permissive for us.

PR #10301, opened 2025-06-01
[Shaman] Suppress a Conversion Warning

Saeldur:
Fixes a type conversion warning caused by it incorrectly using `rng().range` on `double,double` assumed types due to the inputs not matching by forcing it to use it for size_t, which then converts implicitly without any precision loss fine.
Shouldn't cause any issues but I haven't confirmed if behaviour is altered (negatively or positively)

### DIFF
--- a/engine/class_modules/sc_shaman.cpp
+++ b/engine/class_modules/sc_shaman.cpp
@@ -6623,7 +6623,7 @@ struct chained_base_t : public shaman_spell_t
         auto new_idx = i;
         do
         {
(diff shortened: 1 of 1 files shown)


renanthera:
Alternatively, you could replace nearly the whole method with `rng().shuffle( ++tl.begin(), tl.end() );`, but it certainly has subtly different behaviour to the current implementation, even with the shuffle-thoroughness variable set to `1.0`.

PR #10313, opened 2025-06-08
[player] Deduplicate expressions that resolve identically.

renanthera:
Would it be reasonable to discard unknown spell tokenizations if a known tokenization was found?


renanthera:
Discussion in Discord concluded that that discarding unknown spell tokenizations would not be a good idea.

PR #10339, opened 2025-06-20
[player] Allow dumping complete Assisted Combat APLs

renanthera:
Requires that action expr parsing gets skipped, as complete Assisted Combat APLs are presently not formable into valid APL expr syntax.


renanthera:
Ended out not being as useful as I was surmising for future updates. Closing for now.

PR #10359, opened 2025-06-25
[Death Knight] First attempt at empower BoS

Khazakdk:
Lifted the empower bits right from evokers. Still some work to be done on making sure Breath is reporting properly, and the empower CDR effects needs implementing but this is a good checkpoint for seeing how the empower infrastructure is set.


engine/class_modules/sc_death_knight.cpp
```diff
@@ -4527,7 +4529,7 @@ struct whitemane_pet_t final : public horseman_pet_t
       background              = true;
       aoe                     = 20;
       attack_power_mod.direct = 0;
-      impact_action           = get_action<epidemic_whitemane_main_t>( "epidemic", p );
+      //impact_action           = get_action<epidemic_whitemane_main_t>( "epidemic", p );
```
Khazakdk: Needed to stop an assert during debugging, will remove in final PR

engine/class_modules/sc_death_knight.cpp
```diff
@@ -4607,7 +4611,7 @@ struct whitemane_pet_t final : public horseman_pet_t
   void create_actions() override
   {
     death_knight_pet_t::create_actions();
-    epidemic = new epidemic_whitemane_t( "epidemic", this );
+    epidemic = new epidemic_whitemane_t( "epidemic_cast", this );
```
Khazakdk: Stops a debug assert from whitemane having multiple actions named epidemic

PR #10371, opened 2025-06-29
[Mage] TWW3 Arcane Tier Effect

pinepinepinepinepine:
Hey. I figured now's a good time to put a PR in with all the non-hero changes simmering in. Otherwise, I don't mind closing this.

Only thing I have of note is that technically some of the talent checks are redundant. Specifically for Missile tick's harmony + hero_spellslinger's tier effects (checks for splintering sorcery -- SF's are already exclusive as is), but I figured it just looks neater, and you wouldn't have to manually disable Sunfury's tier effect via override whenever running Spellslinger -- can just keep Sunfury's enabled (it's nice, imo, but it is technically useless).

Anyways, best of luck. Lemme know if I should close this PR. Thank you(s)!

### DIFF
--- a/engine/class_modules/sc_mage.cpp
+++ b/engine/class_modules/sc_mage.cpp
@@ -409,6 +409,8 @@ public:
     buff_t* mana_cascade;
     buff_t* spellfire_sphere;
     buff_t* spellfire_spheres;
+    buff_t* lesser_time_warp;
+    buff_t* flame_quills;
 
 
     // Shared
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_mage.cpp
```diff
@@ -9736,6 +9774,42 @@ bool mage_t::trigger_clearcasting( double chance, timespan_t delay, bool never_p
   return success;
 }
 
+void mage_t::trigger_harmony( double chance, bool spell_requires_extra_increment )
```
pinepinepinepinepine, line 9777: Only reason this is here instead of using stack_change_callback is because of the (supposed) bug for Missiles, where it requires a 21st increment; whenever I was using callback instead of this, it results in it being kinda disorienting (could also be just me being messy), so I opted to just abandon it altogether.

pinepinepinepinepine:
Closing this. This is all garbage because Phoenix's damage was changed, this commit here doesn't buff him up with stuff like flame quills + clarity. sorry sorry.

PR #10382, opened 2025-07-03
[Mage] Remove Intuition Damage Bug

pinepinepinepinepine:


pinepinepinepinepine:
shouldve probably checked right before i put this PR in but it looks like you're already working on fixing stuff, and so this PR is probably pointless because it's just removing one small thing. honestly, might be easier to just ignore this entirely. sorry.

i'll close this asap if/when i see the same thing pop up within commits.

PR #10418, opened 2025-07-19
[Havoc] Art of the Glaive timings

Aezeor:
Using glaive flurry as havoc causes all subsequent death sweeps and blade dances to  gain the first art of the glaive buff, even after all related reavers glaives buffs expire until either chaos strike of annihilation are used.
Consuming rending strike as havoc during a death sweep before the final hit causes thefinal death sweep hit to also gain the second art of the glaive damage buff. This applies that buff temporarily during the rending strike event to ensure the final death sweep hit also gains the buff.

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -2259,7 +2259,7 @@ struct art_of_the_glaive_trigger_t : public BASE
         // 2025-07-19 -- Using glaive flurry as havoc causes all subsequent death sweeps and blade dances to
         // gain the first art of the glaive buff, even after all related reavers glaives buffs expire until
         // either chaos strike of annihilation are used.
-        if ( !BASE::p()->specialization() == DEMON_HUNTER_HAVOC )
+        if ( BASE::p()->specialization() != DEMON_HUNTER_HAVOC )
         {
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -2256,7 +2256,13 @@ struct art_of_the_glaive_trigger_t : public BASE
 
       BASE::p()->buff.glaive_flurry->expire();
       make_event( *BASE::p()->sim, thrill_delay, [ this, second_ability ] {
-        BASE::p()->buff.art_of_the_glaive_first->expire();
+        // 2025-07-19 -- Using glaive flurry as havoc causes all subsequent death sweeps and blade dances to
+        // gain the first art of the glaive buff, even after all related reavers glaives buffs expire until
+        // either chaos strike of annihilation are used.
+        if ( !BASE::p()->specialization() == DEMON_HUNTER_HAVOC )
```
ToppleTheNun:
```suggestion
        if ( BASE::p()->specialization() != DEMON_HUNTER_HAVOC )
```

ToppleTheNun (changes requested):

ToppleTheNun (approved):

PR #10432, opened 2025-07-25
[Mage] Implement BoP and GI bug + Fix Arcane's Spellfire Spheres tracking

pinepinepinepinepine:


engine/class_modules/sc_mage.cpp
```diff
@@ -9882,19 +9892,26 @@ void mage_t::trigger_spellfire_spheres()
...
     return;
 
-  int max_stacks = buffs.spellfire_spheres->max_stack();
-
   buffs.spellfire_spheres->trigger();
 
-  auto check_stacks = [ this, s = max_stacks ]
+  // Expiration of Spellfire Spheres needs to occur instantly as to have Blast (while at 5 spheres) with a queued Barrage end with 1 spheres.
+  // Otherwise, the 15ms delay will nullify the queued increment because spheres are at max stacks for an extra 10ms.
+  bool max_stacks = buffs.spellfire_spheres->at_max_stacks();
```
pinepinepinepinepine, line 9899: So, the way I wrote this is to write around the current Implementation of Spheres instead of directly changing it, I suppose. 

What I'm trying to ask is whether or not I should rewrite 9900-9914 to handle the delay + expiration all in one block instead of separating them -- anything works with me. Let me know, thanks.

PR #10452, opened 2025-07-29
[Warlock][Affliction] Add Effect 26 and 27 of Aff Aura

Mlklock:
Aff Aura adds 75% damage to wicked reaping and 55% damage to Demonic Soul.


Mlklock:
Converting this to Draft because i am not sure if those effects are properly affecting affliction in ptr.

PR #10456, opened 2025-07-31
[Warlock] Optimize S3 profiles

SzynaWoW:
Gems and gear are optimised. Need to double-check trinekts after buffs before merging, but apart from that, they are good to go. 

Destro: https://mimiron.raidbots.com/simbot/report/9t4yAR2i2fiDinvjFKbpqj
Aff: https://mimiron.raidbots.com/simbot/report/rmsinPtpsH358oCRMsyasd
Demo: https://mimiron.raidbots.com/simbot/report/sMQSG1bkKDWTjicgh9fZvd

PR #10457, opened 2025-07-31
[Warlock] Fix typos in TWW3_Generate_Warlock.simc

SzynaWoW:
I made a typo when filling in item names to gear search results in that PR https://github.com/simulationcraft/simc/pull/10456. I'm sorry for that!


nyterage:
should add `13504` as a bonus id to the boots while youre at it. Blizz reworked the reshii grace boots to work off bonus ids, so if you have the equipped without the bonus id, they do nothing.

SzynaWoW:
> should add `13504` as a bonus id to the boots while youre at it. Blizz reworked the reshii grace boots to work off bonus ids, so if you have the equipped without the bonus id, they do nothing.

Gotcha! On it

SzynaWoW:
@nyterage So it should be something like `feet=interlopers_silken_striders,id=243305,bonus_id=1533/12361/12239/13504,enchant_id=7424`, right?

SzynaWoW:
Turns out there were also a few typos in item names. Also found better demo profile

PR #10464, opened 2025-08-02
[DBC] Add threat reduction (103) to consumables

soloxcx:
- So it picks up the new Umbral Essentia combat potion


soloxcx:
I'm being very lazy and not regenerating spell data myself, instead letting it regenerate on the next hotfix or whenever, so I suppose there is some chance that this doesn't work

PR #10465, opened 2025-08-02
[Warlock] Add destruction aura modifiers to xalan's ferocity/cruelty

Twinkeli:


engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -402,6 +409,12 @@ double warlock_t::composite_player_pet_damage_multiplier( const action_state_t*
     // 2024-07-06 Summoner's Embrace only affects main pet
     if ( !guardian && talents.summoners_embrace.ok() )
       m *= 1.0 + talents.summoners_embrace->effectN( 2 ).percent();
+
+    if ( hero.xalans_ferocity.ok() )
+      m *= 1.0 + hero.xalans_ferocity->effectN( guardian ? 5 : 3 ).percent();
+
+    if ( hero.xalans_cruelty.ok() )
+      m *= 1.0 + hero.xalans_cruelty->effectN( guardian ? 3 : 4 ).percent();
```
nyterage: these effect indexes are incorrect here, they are referencing a spell direct damage modifier, and a spell periodic modifier. 

these should be 2 and 5 respectively.

nyterage (changes requested):

engine/class_modules/warlock/sc_warlock.cpp
Twinkeli: True, fixed

PR #10496, opened 2025-08-07
[Warlock] Don't trigger expendables on diabolist demons/sh tier

Twinkeli:
Succulent soul doesn't buff mt if at 1 stack, nightfall +dmg doesn't work on hellcaller ds, allow vilefiend to stack to 2 for reporting

### DIFF
--- a/engine/class_modules/warlock/sc_warlock_init.cpp
+++ b/engine/class_modules/warlock/sc_warlock_init.cpp
@@ -749,7 +749,7 @@ namespace warlock
     buffs.dreadstalkers = make_buff( this, "dreadstalkers" )->set_max_stack( 8 )
                           ->set_duration( talents.call_dreadstalkers_2->duration() );
 
-    buffs.vilefiend = make_buff( this, "vilefiend" )->set_max_stack( 1 )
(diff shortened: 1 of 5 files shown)


engine/class_modules/warlock/sc_warlock_init.cpp
```diff
@@ -749,7 +749,7 @@ namespace warlock
     buffs.dreadstalkers = make_buff( this, "dreadstalkers" )->set_max_stack( 8 )
                           ->set_duration( talents.call_dreadstalkers_2->duration() );
 
-    buffs.vilefiend = make_buff( this, "vilefiend" )->set_max_stack( 1 )
+    buffs.vilefiend = make_buff( this, "vilefiend" )->set_max_stack( 2 )
```
seanpeters86, line 752: is this not in spelldata anywhere?

seanpeters86 (approved):

engine/class_modules/warlock/sc_warlock_init.cpp
Twinkeli: No, it’s just for the sake of reporting afaik (Its a custom buff that doesn't exist ingame)

PR #10596, opened 2025-09-02
Resolve several -Wunused-paramter warnings

NeonArray:


gastank:
looks like the dk warnings were fixed in cca0b3e5fed0967df5d21a82b25865ea9a4a309f

the evoker warnings are from code that is auto-generated via ConvertAPL.py, and the generator is fixed in 6c1d72aa80d7c9330ef4e74de0ac994629a57c23

PR #10597, opened 2025-09-03
Add ~find_best~ profileset_cull feature for profileset early elimination

thebigjc:
Raidbots was struggling today, so I tried running simc locally and noticed it was trying to hit the same error for each run, even when it's clear that a profileset is not going to be the 'winner'. If the user cares about finding the 'best' configuration over a number of profilesets, this will speed it up greatly. 

Implements statistical confidence-based early elimination for profilesets to avoid simulating clearly inferior gear combinations. Uses confidence intervals to compare candidates against the current best, with configurable safety margins.

Options:
- profileset_cull=1 to enable
- profileset_cull_verbose=N for verbosity control
- profileset_cull_method= {t_test (default), ci}
- profileset_cull_min_iterations=N (default 100, min 30 before t-test breaks)
- profileset_cull_alpha=N (default 0.05, used as the alpha for the t-test)
-

### DIFF
--- a/engine/report/json/report_json.cpp
+++ b/engine/report/json/report_json.cpp
@@ -1022,6 +1022,17 @@ void profileset_json2( const profileset::profilesets_t& profileset, const sim_t&
 
                      obj[ "iterations" ] = as<uint64_t>( result.iterations() );
 
+                     // find_best elimination metadata (MVP)
+                     if ( sim.find_best.enabled )
+                     {
+                       obj[ "find_best_best" ] = ( sim.find_best.best_name == profileset->name() );
+                       if ( sim.find_best.best_name == profileset->name() )
+                       {
+                         obj[ "find_best_best_error" ] = sim.find_best.best_error;
+                       }
+                       // We cannot directly know elimination status from parent result; use mean_error heuristic not available here.
+                     }
+
                      if ( profileset->results() > 1 )
                      {
                        auto results2 = obj[ "additional_metrics" ].make_array();
--- a/engine/report/json/report_json.cpp
+++ b/engine/report/json/report_json.cpp
@@ -1024,11 +1024,6 @@ void profileset_json2( const profileset::profilesets_t& profileset, const sim_t&
 
                      if ( sim.profileset_cull.enabled )
                      {
-                       obj[ "profileset_cull_best" ] = ( sim.profileset_cull.best_name == profileset->name() );
-                       if ( sim.profileset_cull.best_name == profileset->name() )
-                       {
-                         obj[ "profileset_cull_best_error" ] = sim.profileset_cull.best_error;
-                       }
                        if ( profileset->culled() )
                        {
                          obj[ "culled" ] = true;
@@ -1089,11 +1084,6 @@ void profileset_json3( const profileset::profilesets_t& profilesets, const sim_t
                      // Profileset culling metadata at profileset level
                      if ( sim.profileset_cull.enabled )
                      {
-                       obj[ "profileset_cull_best" ] = ( sim.profileset_cull.best_name == profileset->name() );
-                       if ( sim.profileset_cull.best_name == profileset->name() )
-                       {
-                         obj[ "profileset_cull_best_error" ] = sim.profileset_cull.best_error;
-                       }
                        if ( profileset->culled() )
                        {
                          obj[ "culled" ] = true;
@@ -1274,7 +1264,7 @@ void to_json( const ::report::json::report_configuration_t& report_configuration
     cull[ "enabled" ] = true;
     cull[ "method" ] = sim.profileset_cull.method_name();
     cull[ "min_iterations" ] = sim.profileset_cull.min_iterations;
-    if ( sim.profileset_cull.method == sim_t::profileset_cull_state_t::T_TEST )
+    if ( sim.profileset_cull.uses_alpha() )
       cull[ "alpha" ] = sim.profileset_cull.alpha;
     else
       cull[ "margin" ] = sim.profileset_cull.margin;
@@ -1357,6 +1347,26 @@ void to_json( const ::report::json::report_configuration_t& report_configuration
   add_non_zero( stats_root, "total_dmg", sim.total_dmg );
   add_non_zero( stats_root, "total_heal", sim.total_heal );
   add_non_zero( stats_root, "total_absorb", sim.total_absorb );
+
+  if ( sim.profileset_cull.enabled )
+  {
+    const std::string best_name = sim.profileset_cull.best_name.empty()
+                                    ? sim.profileset_multiactor_base_name
+                                    : sim.profileset_cull.best_name;
+
+    auto cull_stats = stats_root[ "profileset_cull" ];
+    cull_stats[ "method" ] = sim.profileset_cull.method_name();
+    cull_stats[ "metric" ] = util::scale_metric_type_abbrev( sim.profileset_cull.metric );
+    
+    add_non_zero( cull_stats, "best_name", best_name );
+
+    if ( sim.profileset_cull.baseline_seeded )
+    {
+      cull_stats[ "best_error" ] = sim.profileset_cull.best_error;
+      cull_stats[ "best_iterations" ] = sim.profileset_cull.best_iterations;
+      cull_stats[ "best_mean" ] = sim.profileset_cull.best_mean;
+    }
+  }
 
   if ( sim.report_details != 0 )
   {
--- a/engine/report/json/report_json.cpp
+++ b/engine/report/json/report_json.cpp
@@ -1022,7 +1022,6 @@ void profileset_json2( const profileset::profilesets_t& profileset, const sim_t&
 
                      obj[ "iterations" ] = as<uint64_t>( result.iterations() );
 
-                     // find_best elimination metadata (MVP)
                      if ( sim.find_best.enabled )
                      {
                        obj[ "find_best_best" ] = ( sim.find_best.best_name == profileset->name() );
@@ -1030,7 +1029,6 @@ void profileset_json2( const profileset::profilesets_t& profileset, const sim_t&
                        {
                          obj[ "find_best_best_error" ] = sim.find_best.best_error;
                        }
-                       // We cannot directly know elimination status from parent result; use mean_error heuristic not available here.
                      }
 
                      if ( profileset->results() > 1 )
--- a/engine/report/json/report_json.cpp
+++ b/engine/report/json/report_json.cpp
@@ -1022,12 +1022,21 @@ void profileset_json2( const profileset::profilesets_t& profileset, const sim_t&
 
                      obj[ "iterations" ] = as<uint64_t>( result.iterations() );
 
-                     if ( sim.find_best.enabled )
+                     if ( sim.profileset_cull.enabled )
                      {
-                       obj[ "find_best_best" ] = ( sim.find_best.best_name == profileset->name() );
-                       if ( sim.find_best.best_name == profileset->name() )
+                       obj[ "profileset_cull_best" ] = ( sim.profileset_cull.best_name == profileset->name() );
+                       if ( sim.profileset_cull.best_name == profileset->name() )
                        {
-                         obj[ "find_best_best_error" ] = sim.find_best.best_error;
+                         obj[ "profileset_cull_best_error" ] = sim.profileset_cull.best_error;
+                       }
+                       if ( profileset->culled() )
+                       {
+                         obj[ "culled" ] = true;
+                         obj[ "culled_reason" ] = profileset->culled_reason();
+                         obj[ "culled_iterations" ] = profileset->culled_iterations();
+                         obj[ "culled_mean" ] = profileset->culled_mean();
+                         obj[ "culled_error" ] = profileset->culled_error();
+                         obj[ "culled_error_type" ] = profileset->culled_error_type_cstr();
                        }
                      }
 
@@ -1076,6 +1085,25 @@ void profileset_json3( const profileset::profilesets_t& profilesets, const sim_t
                      auto&& obj = results.add();
                      obj[ "name" ] = profileset->name();
                      auto results_obj = obj[ "metrics" ].make_array();
+
+                     // Profileset culling metadata at profileset level
+                     if ( sim.profileset_cull.enabled )
+                     {
+                       obj[ "profileset_cull_best" ] = ( sim.profileset_cull.best_name == profileset->name() );
+                       if ( sim.profileset_cull.best_name == profileset->name() )
+                       {
+                         obj[ "profileset_cull_best_error" ] = sim.profileset_cull.best_error;
+                       }
+                       if ( profileset->culled() )
+                       {
+                         obj[ "culled" ] = true;
+                         obj[ "culled_reason" ] = profileset->culled_reason();
+                         obj[ "culled_iterations" ] = profileset->culled_iterations();
+                         obj[ "culled_mean" ] = profileset->culled_mean();
+                         obj[ "culled_error" ] = profileset->culled_error();
+                         obj[ "culled_error_type" ] = profileset->culled_error_type_cstr();
+                       }
+                     }
 
                      for ( size_t midx = 0; midx < sim.profileset_metric.size(); ++midx )
                      {
@@ -1239,6 +1267,19 @@ void to_json( const ::report::json::report_configuration_t& report_configuration
   options_root[ "default_aura_delay_stddev" ] = sim.default_aura_delay.stddev;
   options_root[ "profileset_metric" ] = util::scale_metric_type_abbrev( sim.profileset_metric.front() );
   options_root[ "profileset_multiactor_base_name" ] = sim.profileset_multiactor_base_name;
+
+  if ( sim.profileset_cull.enabled )
+  {
+    auto cull = options_root[ "profileset_cull" ];
+    cull[ "enabled" ] = true;
+    cull[ "method" ] = sim.profileset_cull.method_name();
+    cull[ "min_iterations" ] = sim.profileset_cull.min_iterations;
+    if ( sim.profileset_cull.method == sim_t::profileset_cull_state_t::T_TEST )
+      cull[ "alpha" ] = sim.profileset_cull.alpha;
+    else
+      cull[ "margin" ] = sim.profileset_cull.margin;
+    cull[ "metric" ] = util::scale_metric_type_abbrev( sim.profileset_cull.metric );
+  }
 
   to_json( options_root[ "dbc" ], *sim.dbc );
 
--- a/engine/report/report_html_sim.cpp
+++ b/engine/report/report_html_sim.cpp
@@ -1158,6 +1158,43 @@ void print_profilesets( std::ostream& out, const profileset::profilesets_t& prof
   out << "<h2 class=\"toggle open\">Profile sets</h2>\n";
   out << "<div class=\"toggle-content\">\n";
 
+  // Profileset culling indicator and culled list
+  if ( sim.profileset_cull.enabled )
+  {
+    out << "<div class=\"note\" style=\"margin:6px 0;\">";
+    out << "Profileset culling enabled: method="
+        << ( sim.profileset_cull.method == sim_t::profileset_cull_state_t::T_TEST ? "t_test" : "ci" )
+  << ", min_iters=" << sim.profileset_cull.min_iterations;
+    if ( sim.profileset_cull.method == sim_t::profileset_cull_state_t::T_TEST )
+      out << ", alpha=" << sim.profileset_cull.alpha;
+    else
+      out << ", margin=" << sim.profileset_cull.margin;
+    out << "</div>\n";
+
+    // List culled profiles if any
+    bool any_culled = false;
+    for ( const auto& pset : profilesets.profilesets() )
+    {
+      if ( pset->culled() ) { any_culled = true; break; }
+    }
+    if ( any_culled )
+    {
+      out << "<div class=\"note\" style=\"margin:6px 0;\"><strong>Culled profiles:</strong><ul>";
+      for ( const auto& pset : profilesets.profilesets() )
+      {
+        if ( !pset->culled() ) continue;
+        out << "<li>" << util::encode_html( pset->name() )
+            << ": " << util::encode_html( pset->culled_reason() )
+            << " (iters=" << pset->culled_iterations()
+            << ", mean=" << util::round( pset->culled_mean(), 2 )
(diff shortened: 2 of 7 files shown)


renanthera:
I've always wanted this, but never sat down to actually get this going. Thanks for your work on this PR so far!

I find the naming of variables in this commit to be a little dense/hard to parse. I've oftentimes seen this sort of technique referred to as "profileset culling", or things like that.

Instead of checking whether or not the target error CIs are disjoint (which is more or less what is occurring in Raidbots), considering instead a two sample t-test to determine if the means are not equal may be a better strategy. I would also like the option to use a variety of culling methods, if only for the ability to experiment with alternate schemes.

Much of the code could become a member function of `find_best_state_t`, which could better allow for modularity for culling strategies.

engine/report/json/report_json.cpp
```diff
@@ -1022,6 +1022,15 @@ void profileset_json2( const profileset::profilesets_t& profileset, const sim_t&
...
 
+                     if ( sim.find_best.enabled )
+                     {
+                       obj[ "find_best_best" ] = ( sim.find_best.best_name == profileset->name() );
+                       if ( sim.find_best.best_name == profileset->name() )
+                       {
+                         obj[ "find_best_best_error" ] = sim.find_best.best_error;
+                       }
+                     }
+
```
vituscze: Is there any particular reason why this is repeated for each profileset? Seems like attaching it to the baseline sim would make more sense. Also, there should probably be some indication that the `find_best` mode is on in the HTML report.

engine/sim/profileset.cpp
```diff
@@ -137,6 +138,10 @@ void simulate_profileset( sim_t* parent, profileset::profile_set_t& set, sim_t*&
 
   if ( !ret || profile_sim -> is_canceled() )
   {
+    if ( profile_sim->find_best_eliminated )
+    {
+      fmt::print( stderr, "\nProfileset '{}' early-stopped: {}\n", set.name(), profile_sim->find_best_reason );
+    }
```
vituscze: This doesn't look like an early stop. You eliminate a profileset by telling simc that it ran enough iterations, not that the sim failed.

engine/sim/sim.cpp
```diff
@@ -2247,6 +2251,96 @@ void sim_t::analyze_error()
...
   }
 
+  if ( need_find_best && current_mean > 0 )
+  {
+    auto &s = parent->find_best;
+    // Convert relative percent error to absolute half-width
+    double abs_error = ( current_error / 100.0 ) * current_mean; 
+
+    // ensure enough iterations
+    if ( work_queue->progress().current_iterations >= s.min_iterations )
```
vituscze: Is it necessary to keep asking `work_queue->progress()` multiple times? Also, does it work correctly when `strict_work_queue` is on, in which case the children don't share the same work queue.

engine/sim/sim.hpp
```diff
@@ -96,6 +96,30 @@ struct sim_t : private sc_thread_t
...
+  struct find_best_state_t {
+    mutex_t mtx;
+    bool    enabled = false;
+    // Metric currently only supports primary profileset metric (DPS-family)
+    scale_metric_e metric = SCALE_METRIC_DPS;
+    std::string best_name;          // profileset name
+    double best_mean = 0.0;         // mean of current best
+    double best_error = 0.0;        // absolute half-width (same units as mean)
+    unsigned best_iterations = 0;   // iterations when last updated
+    bool best_precision_satisfied = false; // winner precision threshold reached
```
vituscze: This doesn't seem to do anything. You only ever write into it.

engine/sim/sim.hpp
```diff
@@ -96,6 +96,30 @@ struct sim_t : private sc_thread_t
...
+    // Configuration copied from options (stored here for child sims to read)
+    int    min_iterations = 500;          // minimum iterations before evaluating elimination
+    double winner_precision = -1.0;       // percent relative error threshold (same unit as target_error/current_error)
```
vituscze: This is only used with `best_precision_satisfied` (see above) and as such doesn't do anything.

engine/sim/sim.cpp
```diff
@@ -2247,6 +2251,96 @@ void sim_t::analyze_error()
...
+      AUTO_LOCK( s.mtx );
+      // Establish winner_precision threshold default if not set: 0.5*target_error if target_error>0 else 0.05%
+      if ( s.winner_precision <= 0 )
+      {
+        // Default: need at most half the target_error to start eliminating (faster unlock); fallback 0.05%
+        if ( parent->target_error > 0 ) s.winner_precision = parent->target_error / 2.0; else s.winner_precision = 0.05; // percent
+      }
+
+      // If no best yet, promote self
+      if ( s.best_name.empty() )
```
vituscze: What about the baseline sim? Seems like a waste to ignore it when it typically finishes (with full precision) before any of the profilesets even start.

engine/sim/sim.cpp
```diff
@@ -2247,6 +2251,96 @@ void sim_t::analyze_error()
...
+            }
+            if ( current_error > 0 && current_error <= s.winner_precision ) s.best_precision_satisfied = true;
+        }
+        else
+        {
+          // Candidate: test elimination vs current best 
+          // Eliminate immediately once candidate interval (with safety) cannot reach best lower bound
+          double safety = s.elim_safety_margin_frac > 0 ? s.elim_safety_margin_frac * s.best_mean : 0.0;
+          double candidate_upper = current_mean + abs_error + safety;
+          double best_lower = s.best_mean - s.best_error;
```
vituscze: As kate mentioned, this could really use a proper statistical test. Especially the safety margin seems pretty ad-hoc.

engine/sim/sim.cpp
```diff
@@ -2247,6 +2251,96 @@ void sim_t::analyze_error()
...
+          if ( candidate_upper < best_lower )
+          {
+            find_best_eliminated = true;
+            find_best_reason = fmt::format( "find_best: eliminated vs '{}' (cand_upper={:.2f} < best_lower={:.2f})", s.best_name, candidate_upper, best_lower );
+            if ( s.verbose >= 1 )
+            {
+              fmt::print( stderr, "\n{}\n", find_best_reason );
+            }
+            work_queue->unlock(); 
+            interrupt();
```
vituscze: `work_queue` uses a reentrant lock partly so that you can interrupt the sim without having to let the lock go.

engine/sim/sim.cpp
```diff
@@ -3773,6 +3867,13 @@ void sim_t::create_options()
   add_option( opt_int( "min_report_iteration_data", min_report_iteration_data ) );
   add_option( opt_bool( "average_range", average_range ) );
   add_option( opt_bool( "average_gauss", average_gauss ) );
+  // Find-Best (profileset early elimination) options (MVP)
+  add_option( opt_bool( "find_best", find_best.enabled ) );
+  add_option( opt_string( "find_best_metric", find_best_metric_str ) );
+  add_option( opt_int( "find_best_min_iterations", find_best.min_iterations ) );
+  add_option( opt_float( "find_best_winner_precision", find_best.winner_precision ) );
+  add_option( opt_float( "find_best_elim_safety_margin", find_best.elim_safety_margin_frac ) );
+  add_option( opt_int( "find_best_verbose", find_best.verbose ) );
```
vituscze: A minor point, but maybe a more descriptive name would be better for the options (and the whole mode). "Finding best" is kinda what the sim does normally, so this doesn't really provide any more info about what's going on.

thebigjc:
Thanks for the great feedback! I'll share some thoughts and another patch in the next day or so!

engine/sim/sim.cpp
thebigjc: Agreed. I've changed the approach to start from the baseline sim.

engine/sim/sim.cpp
thebigjc: There are some valid reasons to compare using a CI, but it's very (probably too) conservative. It was the easiest to implement given what you already had, but it's not the bet it could be. The biggest barrier on the t-test is that we don't have a good statistics library - I'll add in an quick implementation, but it would be easier if we brought in Boost.Math. If you'd rather do that, let me know after you see the next patch

engine/sim/sim.cpp
thebigjc: Thank you!

engine/sim/sim.cpp
thebigjc: I'm going to adopt profileset_cull as the prefix for everything going forward.

thebigjc:
Okay - I cleaned up the approach and added the t-test and made it the default. It's a little messy to pick between the CI approach and the t-test approach, but I've tried to keep it clean enough that it could be refactored into something more 'strategy pattern' like if we want to go down that path.

Also added the HTML and JSON report outputs. 

I like the t-test approach - it is more 'optimistic' than the CI approach. One 'theoretical' concern is that we are doing a lot of repeated comparisons so we might want to either make alpha smaller by default, or take an approach that is more theoretically sound for this kind of sequential arm search. (Interestingly enough, it looks more like the Confidence Interface approach, but with a different method for calculating the bounds)

Let me know if you have any other feedback here.

thebigjc:
Oh, I added a quick and dirty t-test into the code. If we'd rather use Boost.Math I can change that, but as I mentioned above I didn't see that anywhere in the project so I didn't want to add a dependency.

engine/sim/sim.cpp
```diff
@@ -2268,6 +2275,101 @@ void sim_t::analyze_error()
...
+            culled = true;
+            culled_reason = fmt::format( "profileset_cull: eliminated vs '{}' using {}", s.best_name, s.method_name() );
+            if ( s.verbose >= 1 )
+            {
+              fmt::print( stderr, "\n{}\n", culled_reason );
+            }
+            // Without this unlock the program hangs
+            work_queue -> unlock();
+            interrupt();
+            return;
```
vituscze: I think there was a misunderstanding regarding my previous comment on this bit of code. My point was that these two lines are in the wrong order (interrupt first, unlock after), not that you can skip the unlock. If you do it in this order, another thread could grab the lock before you interrupt and do some pointless work.

vituscze (commented):
I'll go over the code in more detail later, so I'm just commenting on some stuff that stood out when I took a quick look.

Regarding the statistical test: I played with it some more after writing the initial comment and while the CI overlap testing is pretty conservative (and perhaps not entirely statistically sound) it might not necessarily be a bad thing. After trying it with some dummy data, when the 95% CIs just barely didn't overlap, a proper statistical test (a one-tailed Welch's t-test) gave me p-values of around 0.002-0.003. So yeah, I agree with your point: if we're gonna be culling potentially thousands of profilesets, a lower alpha is probably warranted (I'm thinking 0.005?).

The general approach of simc is to keep the dependencies pretty light, so we'd rather avoid bringing in Boost.Math for a single statistical test.

engine/sim/sim.cpp
thebigjc: Thanks for clarifying!

thebigjc:
> Regarding the statistical test: I played with it some more after writing the initial comment and while the CI overlap testing is pretty conservative (and perhaps not entirely statistically sound) it might not necessarily be a bad thing. After trying it with some dummy data, when the 95% CIs just barely didn't overlap, a proper statistical test (a one-tailed Welch's t-test) gave me p-values of around 0.002-0.003. So yeah, I agree with your point: if we're gonna be culling potentially thousands of profilesets, a lower alpha is probably warranted (I'm thinking 0.005?).

The 'traditional' method would be to take a single alpha (0.05) and divide by the number of comparisons (profilesets, or arms, k), called a 'bonferroni correction' which is a much smaller value than you're suggesting (0.05 / 1000 -> 0.00005). As you saw, this is a VERY conservative number and wouldn't prune many, if any, arms. Other classical approaches refine this, but have requirements that don't fit into our 'sequential' model of testing each profileset (arm) (Holm-Bonferroni, which requires all the p-values to exist and be sorted)

There is a whole sub-field of stats here (SAVI) we could get into for the exact type of test that is right here if we want to very 'tight', or we could start with the CI approach, which is a little hand wavy, and add in the 'more correct' approach later. 

Or we could do the research and add in the best option now!

thebigjc:
We don't need to pull in the library, but https://github.com/gostevehoward/confseq is an example of the 'theoretically correct' way to do this.

engine/report/json/report_json.cpp
```diff
@@ -1022,6 +1022,24 @@ void profileset_json2( const profileset::profilesets_t& profileset, const sim_t&
 
                      obj[ "iterations" ] = as<uint64_t>( result.iterations() );
 
+                     if ( sim.profileset_cull.enabled )
+                     {
+                       obj[ "profileset_cull_best" ] = ( sim.profileset_cull.best_name == profileset->name() );
+                       if ( sim.profileset_cull.best_name == profileset->name() )
+                       {
+                         obj[ "profileset_cull_best_error" ] = sim.profileset_cull.best_error;
+                       }
```
vituscze: Perhaps I'm missing something, but is there a good reason to keep this information redundant? I can't think of a use case that wouldn't be covered by just including the best name/error in the JSON of the parent (non-profileset) sim.

engine/report/json/report_json.cpp
```diff
@@ -1231,6 +1268,19 @@ void to_json( const ::report::json::report_configuration_t& report_configuration
   options_root[ "profileset_metric" ] = util::scale_metric_type_abbrev( sim.profileset_metric.front() );
   options_root[ "profileset_multiactor_base_name" ] = sim.profileset_multiactor_base_name;
 
+  if ( sim.profileset_cull.enabled )
+  {
+    auto cull = options_root[ "profileset_cull" ];
+    cull[ "enabled" ] = true;
+    cull[ "method" ] = sim.profileset_cull.method_name();
+    cull[ "min_iterations" ] = sim.profileset_cull.min_iterations;
+    if ( sim.profileset_cull.method == sim_t::profileset_cull_state_t::T_TEST )
```
vituscze: Probably better to add something like `sim.profileset_cull.uses_alpha()` to keep the method checks in the same place. This place would be pretty easy to forget if we ever add another method.

engine/report/report_html_sim.cpp
```diff
@@ -1351,6 +1351,43 @@ void print_profilesets( std::ostream& out, const profileset::profilesets_t& prof
...
+  if ( sim.profileset_cull.enabled )
+  {
+    out << "<div class=\"note\" style=\"margin:6px 0;\">";
+    out << "Profileset culling enabled: method="
+        << ( sim.profileset_cull.method == sim_t::profileset_cull_state_t::T_TEST ? "t_test" : "ci" )
+  << ", min_iters=" << sim.profileset_cull.min_iterations;
+    if ( sim.profileset_cull.method == sim_t::profileset_cull_state_t::T_TEST )
+      out << ", alpha=" << sim.profileset_cull.alpha;
+    else
+      out << ", margin=" << sim.profileset_cull.margin;
```
vituscze, line 1176: You're using `method_name()` above, not sure why it's missing here. Also same as above regarding the alpha/margin.

engine/report/report_html_sim.cpp
```diff
@@ -1351,6 +1351,43 @@ void print_profilesets( std::ostream& out, const profileset::profilesets_t& prof
...
+            << ": " << util::encode_html( pset->culled_reason() )
+            << " (iters=" << pset->culled_iterations()
+            << ", mean=" << util::round( pset->culled_mean(), 2 )
+            << ", error=" << util::round( pset->culled_error(), 4 )
+            << ", type=" << pset->culled_error_type_cstr() << ")";
+        out << "</li>";
+      }
+      out << "</ul></div>\n";
+    }
+  }
```
vituscze: I think this should probably be below the profileset chart, since most people will be primarily interested in the profileset results. Perhaps it could even be collapsed by default, but that's a minor thing.

engine/sim/profileset.cpp
```diff
@@ -137,6 +138,10 @@ void simulate_profileset( sim_t* parent, profileset::profile_set_t& set, sim_t*&
 
   if ( !ret || profile_sim -> is_canceled() )
   {
+    if ( profile_sim->culled )
+    {
+      fmt::print( stderr, "\nProfileset '{}' culled: {}\n", set.name(), profile_sim->culled_reason );
+    }
```
vituscze: Is there a reason why you kept this one around? If the profileset sim ends in an error or is forcefully cancelled, I don't think we can say it was culled. You can refer to my earlier comment on this.

engine/sim/sim.cpp
```diff
@@ -2268,6 +2275,100 @@ void sim_t::analyze_error()
...
+
+    // Snapshot progress once to avoid repeated work_queue calls
+    sim_progress_t current_progress;
+    if ( strict_work_queue ) {
+      // In strict_work_queue mode, read iterations from this child sim's own counters
+      current_progress.current_iterations = current_iteration;
+      current_progress.total_iterations = iterations;
+    } else {
+      current_progress = work_queue->progress();
+    }
```
vituscze: The statistics (mean, stddev, error) are calculated from *all* sims, including the children. The issue with `strict_work_queue` is that the `work_queue` then only has iteration data for the current sim. Check how `n_iterations` is computed earlier in this function (in fact, you might just be able to reuse it here).

engine/sim/sim.cpp
```diff
@@ -3073,6 +3174,172 @@ void sim_t::analyze_iteration_data()
...
+  double absolute_error = relative_error * baseline_data.mean;
+  
+  // Get iterations from current simulation progress
+  auto current_progress = progress();
+  
+  // Seed the culling state with baseline values
+  {
+    std::lock_guard<mutex_t> lock( profileset_cull.mtx );
+        
+    profileset_cull.best_name = "Baseline";
```
vituscze: Unfortunately, `Baseline` is a valid profileset name. Empty string isn't, though, but that might not be the best way to handle this.

engine/sim/sim.cpp
```diff
@@ -3789,6 +4056,30 @@ void sim_t::create_options()
...
+  // Profileset culling (early elimination) options
+  add_option( opt_bool( "profileset_cull", profileset_cull.enabled ) );
+  add_option( opt_func( "profileset_cull_method", []( sim_t* sim, util::string_view, util::string_view value ) {
+    if ( util::str_compare_ci( value, "ci" ) )
+      sim->profileset_cull.method = sim_t::profileset_cull_state_t::CI_OVERLAP;
+    else if ( util::str_compare_ci( value, "t_test" ) )
+      sim->profileset_cull.method = sim_t::profileset_cull_state_t::T_TEST;
+    else {
+      sim->error( "Invalid profileset_cull_method '{}', valid options: ci, t_test", value );
+      return false;
```
vituscze: There are some enum helper functions that can automatically derive a parser (i.e. string -> enum function) when given an enum -> string function. Not really a big deal, but might be convenient in case more methods are added down the line.

engine/report/json/report_json.cpp
thebigjc: Nope - not missing something. Just cruft. Thanks - fixed.

engine/report/report_html_sim.cpp
thebigjc: Fixed

engine/report/json/report_json.cpp
thebigjc: Added uses_alpha()

engine/sim/profileset.cpp
thebigjc: Nope - this was cruft. Thank you.

engine/report/report_html_sim.cpp
thebigjc: Moved it down, made it collapsed.

engine/sim/sim.cpp
thebigjc: Changed to use the previously calculated values and remove the extra locals.

engine/sim/sim.cpp
thebigjc: Changed to use "", and added some special casing so we don't print ""

engine/sim/sim.cpp
thebigjc: To use those, I'd have to move the enum out of being declared inline, and touch a bunch of extra files. I cleaned this up by moving the parsing out of line here. If you want the full change (move the enum declaration to sc_enum.hpp, etc) I can do that too, but it seemed like more than you intended.

thebigjc:
PR comments addressed. Let me know if you have any other feedback!

thebigjc:
Rebased

vituscze (commented):
Finally had some time to go over it in more detail and I don't think I have any other comments. I'll try to play with it to get a feel for the culling with some real profiles and I'll poke another dev to take a look.

thebigjc:
Thanks for the update!

engine/sim/sim.cpp
```diff
@@ -2137,6 +2141,7 @@ void sim_t::analyze_error()
 
   double mean_total=0;
   int mean_count=0;
+  double current_standard_error = 0.0;
```
nyterage, line 2144: value of this variable outside of single_actor_batch sims is never set, leading to t_test always seeing a se of 0. Thus, breaking the t_test culling logic.

thebigjc: Hmm , that's a good catch, but I'm not exactly sure how to test a fix. 

I haven't used multi-actor simulation with profilesets. I see how to get a valid `se` in the other arm of !single_actor_batch, but I'm not really sure what the right behavior would be. Do you have a good test case I could validate against? 

Alternatively I can try to some multiactor tests and just figure it out.

thebigjc: It looks like there are already hacks in the case of multi actor metrics. Perhaps we should just fail if culling with multiactor is used? Happy to take a suggestion.

nyterage: only reason i noticed is because it came up in my testing. Doesn't even have to be multi-actor, just a base profile without `single_actor_batch=1` set, and the profilesets. 

Test case for me was as follows

Basic Unholy Death Knight profile
46k talent combination profilesets

best solution is *probably* throwing an error in the `is_multiactor_metric` check if its trying to use culling, and setting the variable value in the `else` function though.

nyterage: Thinking on the above comment a bit more, might be better to just set the variable no matter if its multi-actor or not, as throwing an error in multi-actor will just prevent it from working on augmentation sims.

thebigjc: Thanks - I'll have some time on the weekend to work on this. I'll make a change and push a patch once I have it.

PR #10685, opened 2025-10-25
Generalized Profileset Control

renanthera:
Derived from the concept found in #10597, but generalized to allow for custom implementation in class modules.

Significant outstanding issues:
- Many relevant player methods are useless at current `profileset_controller_t::evaluate_*` calls, as they only return sensible values during iteration runtime. Checking whether or not a player has a sufficient amount of a secondary stat, etc, is impossible as a result.

TODO:
- Implement error-based sim controller.
- Cherrypick from TeWaWi to Midnight.

Developer interface issues:
- Every derived `profileset_controller_t` requires a `using data_t = <derived from profileset_controller_data_t>;` type alias. Every controller probably should get and set a derived `profileset_controller_data_t`, just for the sake of reporting, as it is otherwise impossible to customize.
- `profileset_controller_t::get_data<T>()` and `profileset_controller_t::set_data<T>(...)` require a template parameter which is always equal to the type-aliased `data_t` for the derived `profileset_controller_t`. This kind of sucks, but I'm not sure of a way to avoid this without introducing manual mutex management and gratuitous casting.

Notes:
- Options are parsed from strings on construction/init of each profileset.
- Implementing `profileset_controller_t::create_options()` is kind of awful if the option value type isn't well supported.


renanthera:
Still researching best practices for statistical methods required to implement profileset culling in the most optimal/correct way.

The rest of the implementation is otherwise functional.

renanthera:
Finished preparing this phase of implementation core functionality along with a basic profileset controller example.

Allows for halting profilesets at both init and after any iteration, with a relatively small interface for anyone implementing a profileset controller.

See `set_bonus_enabled_t` as an example. Cancels profilesets after init that do not have a specified tier set and piece count enabled.

A followup will contain an implementation for dynamic profileset culling, a more granular form of the multi-stage simulation strategy.

PR #10686, opened 2025-10-30
Havoc tweaks

Aezeor:
corrected the spell id being used for glaive tempest
added condition to check fury to make sure glaive tempest can be cast
set the shattered souls that spawn when getting a killing to be havoc and vengeance only
essence break should only proc once per blade dance on the first hit
the eternal hunt buff should expire once blade dance is used
removed the felblade reset proc that was removed in alpha but not in notes
corrected the meta extension that occurs when using death sweep at the end of meta to only grant the time needed to execute the final hit

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -6826,10 +6826,12 @@ struct death_sweep_t : public demonsurge_trigger_t<demonsurge_ability::DEATH_SWE
   {
     assert( p()->buff.metamorphosis->check() );
 
-    // If Metamorphosis has less than 1s remaining, it gets extended so the whole Death Sweep happens during Meta.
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -8955,6 +8940,8 @@ demon_hunter_td_t::demon_hunter_td_t( player_t* target, demon_hunter_t& p )
 
 void demon_hunter_td_t::target_demise()
 {
+  if ( dh().specialization() != DEMON_HUNTER_HAVOC && dh().specialization() != DEMON_HUNTER_VENGEANCE )
```
ToppleTheNun:
```suggestion
  if ( dh().specialization() == DEMON_HUNTER_DEVOURER )
```

ToppleTheNun (changes requested):

ToppleTheNun (approved):

PR #10692, opened 2025-11-08
Demonsurge and havoc fixes

Aezeor:
move the demonsurge trigger for havoc from vola to correctly trigger always when pressing meta and only trigger off eyebeam when entering meta when not in meta
us a 700ms delay for death sweep and hardcast meta demonsurge triggers
first blood hits proc essence break


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -8642,6 +8651,11 @@ struct metamorphosis_buff_t : public demon_hunter_buff_t<buff_t>
       p()->buff.demonsurge_abilities[ demonsurge_ability::ANNIHILATION ]->trigger();
       p()->buff.demonsurge_abilities[ demonsurge_ability::DEATH_SWEEP ]->trigger();
       p()->buff.demonsurge_demonsurge->trigger();
+
+      if ( p()->talent.scarred.volatile_instinct->ok() )
```
ToppleTheNun, line 8963: this needs to go in `extend_duration_or_trigger` instead

ToppleTheNun (changes requested):

PR #10718, opened 2025-12-19
[Deathknight] Frigid Resolve, Blood MID1 tier set, Bloody Reflection

Armadk:


engine/player/player.cpp
```diff
@@ -8526,7 +8526,7 @@ void account_absorb_buffs( player_t& p, action_state_t* s, school_e school )
       if ( ab->up() )
       {
         // Consume the absorb and grab the effective amount consumed.
-        double absorbed = ab->consume( s->result_amount );
+        double absorbed = ab->consume( s->result_amount, s );
```
Armadk, line 8529: @gastank  Any particular reason you did not include the state with this one?  

I can't find any issues it should cause by including it.

Added in
https://github.com/simulationcraft/simc/commit/b8d701d43bb0e652f281f14159249011a3f1a708
and updated in
https://github.com/simulationcraft/simc/commit/d3030391ed9b4c5f76458a8be54dd7dde40fca57

PR #10740, opened 2026-01-05
[talents] Generic Apex Talent Detection

renanthera:


renanthera:
image

:)

renanthera:
`static children()` could be a member function if each `trait_data_t` were to have a `bool is_ptr` member. I opted for a slightly messier set of parameters instead of adding a member to each, as it would be inconsistent with the typical data handling strategy.

PR #10755, opened 2026-01-11
[Monk] Harmonic Surge, Potential Energy

justindriggers:


engine/class_modules/monk/sc_monk.cpp
```diff
@@ -710,15 +710,19 @@ struct harmonic_surge_t : public monk_spell_t
...
+      damage( new impact_t<monk_spell_t>( player, "harmonic_surge_damage", player->talent.master_of_harmony.harmonic_surge_damage ) ),
+      heal( new impact_t<monk_heal_t>( player, "harmonic_surge_heal", player->talent.master_of_harmony.harmonic_surge_heal ) )
   {
   }
 
   void execute() override
   {
     monk_spell_t::execute();
 
+    if ( !p()->buff.harmonic_surge->up() )
```
Hinalover: This could be moved to the ready override instead of the execute.
IE:
boolean ready() override
{
return p()->buff.harmonic_surge.up();
}

justindriggers: No longer a viable solution after implementing the templating suggested by @renanthera in DMs

PR #10761, opened 2026-01-12
[Monk] Swift as a Coursing River

justindriggers:
* Move Pretense of Instability and Special Delivery activation into brew_t
* Add parse_effects support for Melee Haste


renanthera:
Split out parse_effects changes into #10762

PR #10880, opened 2026-02-05
[DBC] Add Important Spell attribute (491)

prsauer:
Currently attribute 491 is marked as unknown -- "Unknown (491)"

I used https://wago.tools/files and a lua script in-game to deduce that this is the flag that determines which spells will return from the api C_Spell.IsSpellImportant

My lua script enumerated every spell id that this function returns true for and compared it to the spellmisc table of attributes. Using 491 as the flag results in a perfect mapping to the in game spell ids.


prsauer:
Closing - wrong target branch

PR #10882, opened 2026-02-05
[Monk] Niuzao's Resolve, Vital Flame

justindriggers:
* Also fixes Celestial Fortune amounts getting overwritten before scheduled execution


renanthera:
All of this drama to avoid one little `std::bind` ;)

PR #10883, opened 2026-02-06
Implement base Fired Up Talent.cpp

Preheat:
Adding basic functionality for the first point in Fire mage Apex talent (Fired Up).


Preheat:
Note: Dorovon is making his own, this is no longer needed.

PR #10887, opened 2026-02-07
[ci] What happens if branch that is being merged is named the same as default?

renanthera:


renanthera:
`head_ref` and `base_ref` are non-empty as expected, thus disambiguating from main.

PR #10888, opened 2026-02-07
[ci] Document flagging branches explicitly for CI and fix commits and merges into main not triggering tests.

renanthera:


renanthera:
Condition may be excessive, but should still correctly allow for CI to run against default branch.

PR #10904, opened 2026-02-10
[Core] Adjust max stack behavior for async buffs

Armadk:
In every case I have checked in war and dk, it seems like max stack behavior for async changed at some point over the last few years.

Current behavior in game
- If at max stacks, refresh the buff duration
- queue up a decrement event

In this scenario, we have more decrement events, than we have stacks, so we begin to lose stacks faster than expected.


Armadk:
Updated to make the new behavior opt in for buffs.

PR #10910, opened 2026-02-11
Fix missing parens around <? in arcane sunfury_hold_for_cds

elliottabirch:
The cooldown.touch_of_the_magi threshold was missing outer parentheses around the <? (min) expression, causing incorrect operator precedence.

Before: -(2*(react&cc))<?(orb&ae)  → min(4-A-B, C)
After:  -((2*(react&cc))<?(orb&ae)) → 4-A-min(B, C)

The cooldown.arcane_surge threshold already had the correct parens. This makes both instances consistent.

5*buff.overpowered_missiles.react=0  parses to (5*buff.overpowered_missiles.react)=0


ArcanePorom:
Hello, please do not undo previous APL updates in your PRs.
Your first commit does not match your note; the commit shows a deleted parens, none added.  Even adjusting the APL for your change changes no behavior. But I agree this is wrong currently, will be fixed in next update. Correction: I do see your correct parens, sorry!
Your second commit is just wrong, we use .react intentionally with buffs that have a random chance to happen, this avoids superhuman sim behaviors.
It would be highly preferable to communicate and discuss changes before commiting them, especially when using AI.

elliottabirch:
> Hello, please do not undo previous APL updates in your PRs. Your first commit does not match your note; the commit shows a deleted parens, none added. Even adjusting the APL for your change changes no behavior. But I agree this is wrong currently, will be fixed in next update. Correction: I do see your correct parens, sorry! Your second commit is just wrong, we use .react intentionally with buffs that have a random chance to happen, this avoids superhuman sim behaviors. It would be highly preferable to communicate and discuss changes before commiting them, especially when using AI.

I used a claude agent to throw up a quick pr, not work through the logic. 

i did not mean to undo other's changes. the merge conflicts were hard to see, so i think i just missed that change

Im aware of the purpose of react. There is a parsing problem that needs to be addressed, either with the parser, or the logic used here. The problem with `5*buff.overpowered_missiles.react=0` is that i parses incorrectly. the `=` is taking precedence over the parens, and it is mapping to the incorrect logic: `(5*buff.overpowered_missiles.react)=0`;

@ArcanePorom where is best place to discuss changes? I thought that creating a PR with the changes would give context and allow for discussion, but im happy to take a different approach if there is something else more standardized

PR #10956, opened 2026-02-15
[Warlock] Fix Demonology/Destruction infinite loops, remove version guard

Braxtar:
## Summary

All three Warlock specs are currently blocked from simming on the `midnight` branch. After removing the version guard, I found two runtime bugs that cause infinite event loops:

**Demonology — `summon_vilefiend`:**
- The talent spell `summon_vilefiend` (ID 1251778) has **0s duration** and **0s cooldown** in the Midnight DBC data. This is a trait_data mismatch — the actual gameplay values live on spell 1251781 (15s duration).
- Without a cooldown, the sim re-casts vilefiend every event tick at t=0, creating an infinite loop.
- Fix: use `find_spell(1251781)->duration()` for spawn duration/buff, hardcode 45s cooldown (matches in-game).
- Also added `set_default_duration` for vilefiends — it was the only pet spawner missing this call.

**Destruction — `dimensional_rift`:**
- The spell data already contains 3 charges / 45s recharge, but `dimensional_rift_t` never initializes the charge system in its constructor.
- Without charges, the sim treats it as a 0-cooldown spell and casts it infinitely.
- Fix: `cooldown->charges = as<int>( data().charges() );`

**Version guard:**
- Removed the `wowv_t( 13, 0, 0 )` check in `validate_actor()` since all three specs now sim successfully with these fixes.

## Test results

Ran with `load_default_gear=1 load_default_talents=1 default_actions=1 iterations=100`:

| Spec | DPS | Status |
|------|-----|--------|
| Affliction | ~21,000 | OK |
| Demonology | ~11,000 | OK (was infinite loop) |
| Destruction | ~30,000 | OK (was infinite loop) |

Note: some APL talent references (e.g. `vile_taint`, `eradication`, `the_houndmasters_gambit`) log warnings because these talents were renamed in Midnight. The APL still runs — expressions with missing talents just evaluate to false. A separate APL update pass would clean those up.

## Files changed

- `sc_warlock.cpp` — removed version guard
- `sc_warlock_actions.cpp` — vilefiend cooldown + duration fix, dimensional_rift charge init
- `sc_warlock_init.cpp` — vilefiend `set_default_duration` + buff duration fix

### DIFF
--- a/engine/class_modules/warlock/sc_warlock.cpp
+++ b/engine/class_modules/warlock/sc_warlock.cpp
@@ -639,16 +639,8 @@ stat_e warlock_t::convert_hybrid_stat( stat_e s ) const
 
 bool warlock_t::validate_actor()
 {
-  // TODO: Remove this when Midnight is properly supported
-  if ( sim->dbc->wowv() < wowv_t( 13, 0, 0 ) )
-  {
-    std::string patch = "Midnight prepatch";
-    if ( sim->dbc->wowv() > wowv_t( 12, 0, 0 ) )
-      patch = "Midnight";
-    throw sc_unsupported_specialization( fmt::format( "Warlock sims are non functional for {}", patch ) );
-    return false;
-  }
-
+  // Warlock Midnight support: guard removed — all three specs now sim
+  // with vilefiend and dimensional_rift fixes below
   return true;
 }
 
--- a/engine/class_modules/warlock/sc_warlock_actions.cpp
+++ b/engine/class_modules/warlock/sc_warlock_actions.cpp
@@ -3303,15 +3303,20 @@ using namespace helpers;
     {
       harmful = may_crit = false;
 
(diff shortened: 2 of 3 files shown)


millanzarreta:
The Warlock module hasn't been updated for Midnight yet, which is why there are so many errors.

There's a large pending PR that includes many changes for the initial implementation of the Warlock module for Midnight here: https://github.com/simulationcraft/simc/pull/10923

vituscze:
"Sim runs" and "sim gives accurate results" are two very distinct things. This does the former without doing the latter, which makes it basically useless.

Braxtar:
Thank you for the feedback. You're right — these are band-aid fixes that make specs run without ensuring accuracy. Closing in favor of #10923 which is the proper Midnight implementation. Appreciate the pointer\!

PR #10961, opened 2026-02-17
[Gear] Vaelgors final stare trinket

NN0323:


engine/player/unique_gear_midnight.cpp
```diff
@@ -2011,6 +2011,92 @@ void magisters_alchemist_stone( special_effect_t& e )
...
+
+    nullsight( const special_effect_t& e )
+      : stat_buff_t( e.player, "nullsight", e.player->find_spell( 1260459 ) ),
+        effect( e ),
+        value_spell( nullptr ),
+        current_tick( 0 ),
+        buff_val( 0 ),
+        decrease( 0 )
+      {
+      value_spell  = e.player->find_spell( 1260459 );
```
nyterage: Looks like you've got the incorrect spell id for the value here, should be `1259293` (the driver spell)

engine/player/unique_gear_midnight.cpp
```diff
@@ -2011,6 +2011,92 @@ void magisters_alchemist_stone( special_effect_t& e )
   new dbc_proc_callback_t( e.player, e );
 }
 
+// Vaelgor's Final Stare
+// 1259293 Driver
+// 1260459 Nullsight/value spell
+void vaelgors_final_stare( special_effect_t& effect ){
+  struct nullsight final : public stat_buff_t{
```
nyterage: Codestyle issue here, should be `nullsight_t`

nyterage (changes requested):
Few issues here which will mean its not properly modeled for sims

engine/player/unique_gear_midnight.cpp
NN0323: The driver spell does not have an effectN(3) and the duration and periodic damage is specified in nullsight it might be that i dont fully understand what the driver spell truly is the way its currently implemented works from what ive tested give an initial mastery buff for x amount then ticks down x/ticks amounts every tick = 1 second for 15 seconds

nyterage: so, when talking about the "value" for items like these, we do mean the spell that contains the coefficient and scaled value.

in the case of this trinket it would be 

```
Name             : Vaelgor's Final Stare (id=1259293) [Passive] 
School           : Physical
Spell Type       : None
Labels           : 292
                 : 3959: Item Effects
Attributes       : Passive (6), No Aura Icon (60), Can Proc From Procs (122), No Aura Log (202), Aura Points On Client (268), Scales with Casting Item's Level (354)
Effects          :
#1 (id=1268341)  : Apply Aura (6) | Dummy (4) | Scaling Class: Secondary Rating (-7)
                   Base Value: 0 | Scaled Value: 755.1957 (coefficient=5.761313) | PvP Coefficient: 0.5 | Target: Self (1)
```
                   
This spell contains the buffs value. While the other 3 data pieces youre looking for are in the buff spell themselves. That buff spell wont need the second lookup since its already the struct youre in, and can instead be called with `data().x`

so, this ctor should end up something like
```
auto n_ticks = data().duration() / data().effectN( 3 ).period();
buff_val     = e.driver()->effectN( 1 ).average( e );
decrease     = buff_val / n_ticks;
set_stat_from_effect( 1, buff_val );
set_tick_callback( [ this ]( buff_t*, int, timespan_t ) { recalculate(); } );
```

NN0323: ah okay i get it now i think ill make the proper corrections

engine/player/unique_gear_midnight.cpp
NN0323: Fixed

engine/player/unique_gear_midnight.cpp
NN0323: Should be fixed now

gastank:
We have a `.clang-format` that you should apply. Broadly speaking, block opening brackets are on a new line and function arguments are not on new lines.

PR #10965, opened 2026-02-17
druid: fix Guardian APL for Midnight talent tree

Braxtar:
## Summary

The Guardian Druid default APL references three talents that were removed in the Midnight talent tree, causing initialization errors (`exit=30`):

- `thorns_of_iron` — used in `variable.If_build` definition
- `ursocs_endurance` — used in `variable.If_build` definition  
- `rage_of_the_sleeper` — used as a talent check in an ironfur condition (the ability still exists but is no longer a selectable talent node)

## Changes

- **`variable.If_build`**: Set to `0` unconditionally since the two talents that defined it (`thorns_of_iron` + `ursocs_endurance`) no longer exist
- **ironfur condition**: Removed `talent.rage_of_the_sleeper.enabled` check (kept the `buff.rage_of_the_sleeper.up` check which still works)

Both `.simc` source and compiled `.inc` are updated.

## Test plan

- [x] `simc druid=test spec=guardian level=80 load_default_gear=1 load_default_talents=1 default_actions=1 iterations=500` — exits cleanly, ~24K DPS
- [x] No other talent references are broken (verified all 15 remaining talent refs resolve)
- [x] Compiled `.inc` matches `.simc` source

### DIFF
--- a/ActionPriorityLists/default/druid_guardian.simc
+++ b/ActionPriorityLists/default/druid_guardian.simc
@@ -11,7 +11,7 @@
 # Snapshot raid buffed stats before combat begins and pre-potting is done.
 actions.precombat=snapshot_stats
 # Executed before combat begins. Accepts non-harmful actions only.
-actions.precombat+=/variable,name=If_build,value=1,value_else=0,if=talent.thorns_of_iron.enabled&talent.ursocs_endurance.enabled
+actions.precombat+=/variable,name=If_build,value=0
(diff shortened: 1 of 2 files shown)


renanthera:
Mangling APLs to get them to run with no further validation of accuracy and optimality is not productive.

Braxtar:
Fair point — setting If_build=0 without validating that the resulting rotation is actually optimal for Midnight is just papering over the real issue. The APL needs a proper rewrite for the new talent tree, not a band-aid to make it parse.

Thanks for the feedback, closing this.

PR #10983, opened 2026-02-19
[Warlock] Destruction apex talent

stahlscott:
Initial implementation of Embers of Nihilam apex talent for warlock destruction spec

Needs verification:
- Rank 1 proc rate (Incinerate → Echo of Sargeras) is not in spell data: using 25% placeholder, configurable via rng_settings

Implementation notes:
- Visions of Nihilam (rank 2/3) scaling: spell data auto-scales base_value with talent rank via op=mul, values=(1,2), so .percent() alone gives the correct value (4% at rank 1, 8% at rank 2)
- Rank 4 uses a helper for all affected spells (CB, SB, RoF) to handle the case where there's no target present - even though CB/SB always has a target as of now, figured it was safer for future proofing
- Rank 1 (proc chance, no ICD) and rank 3 (guaranteed, 0.5s ICD) are intentionally independent as they are separate talent effects with different gating mechanics

### DIFF
--- a/engine/class_modules/warlock/sc_warlock_actions.cpp
+++ b/engine/class_modules/warlock/sc_warlock_actions.cpp
@@ -3234,6 +3234,13 @@ using namespace helpers;
         p()->procs.demonfire_infusion_inc->occur();
       }
 
+      if ( p()->talents.embers_of_nihilam_1.ok() && p()->rng().roll( p()->rng_settings.echo_of_sargeras.setting_value ) )
+      {
+        p()->proc_actions.echo_of_sargeras->execute_on_target( target );
+        p()->procs.echo_of_sargeras->occur();
+        p()->buffs.vision_of_nihilam->trigger();
+      }
+
       p()->buffs.backdraft->decrement();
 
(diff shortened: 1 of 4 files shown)


engine/class_modules/warlock/sc_warlock_actions.cpp
```diff
@@ -4562,6 +4527,38 @@ using namespace helpers;
...
+          continue;
+
+        if ( tdata->dots.immolate->is_ticking() || tdata->dots.wither->is_ticking() )
+          candidates.push_back( t );
+      }
+
+      if ( candidates.empty() )
+        return;
+
+      target = candidates[ p->rng().range( candidates.size() ) ];
```
nyterage: This can be simplified down to `target = p->rng().range( candidates );`

nyterage (approved):
Looks good generally, only one comment

engine/class_modules/warlock/sc_warlock_actions.cpp
stahlscott: fixed!

PR #10988, opened 2026-02-20
[Warlock] Affliction apex talent

stahlscott:
Shadow of Nathreza damage is triggered via Haunt's effect 5 (Periodic Trigger Spell every 1s), wired as a set_tick_callback on the Haunt debuff (similar to how Blackened Soul uses set_period + set_tick_callback)

Wrath of Nathreza uses the impact spell (1278047) for damage, not the trigger missile spell (1262028). The trigger spell is still loaded for reduced_aoe_targets data

### //  TOCHECK
- Does Wrath of Nathreza proc from the initial Haunt hit, periodic ticks, or both? (currently can proc from either based on my reading of the talent's wording, but could not get initial Haunt impact to trigger Wrath when testing on training dummies in beta. unsure if that's RNG or if it only triggers on dot ticks... at the same time it shouldn't be a huge change in output given its RPPM driven)
- Wrath of Nathreza RPPM rate: spell data suggests ~2 RPPM (hasted), should verify in-game

### DIFF
--- a/engine/class_modules/warlock/sc_warlock.cpp
+++ b/engine/class_modules/warlock/sc_warlock.cpp
@@ -791,9 +791,7 @@ void warlock_t::parse_player_effects()
     // Affliction Mastery
     parse_effects( warlock_base.potent_afflictions ); // 77215
     // Affliction Debuffs/DoTs
-    parse_target_effects( d_fn( &warlock_td_t::debuffs_t::haunt ), talents.haunt ); // 48181
-    parse_effects( talents.shadow_of_nathreza_2 ); // 1261990
-    parse_target_effects( d_fn( &warlock_td_t::debuffs_t::haunt ), talents.shadow_of_nathreza_2 ); // 1261990
+    parse_target_effects( d_fn( &warlock_td_t::debuffs_t::haunt ), talents.haunt, talents.shadow_of_nathreza_2 );
   }
 
   // Demonology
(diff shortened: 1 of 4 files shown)


engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -781,6 +792,8 @@ void warlock_t::parse_player_effects()
     parse_effects( warlock_base.potent_afflictions ); // 77215
     // Affliction Debuffs/DoTs
     parse_target_effects( d_fn( &warlock_td_t::debuffs_t::haunt ), talents.haunt ); // 48181
+    parse_effects( talents.shadow_of_nathreza_2 ); // 1261990
```
nyterage: `parse_effects` shouldnt be used for passive effects like talents (simc should actually be yelling at you about this) since they are parsed automatically.

engine/class_modules/warlock/sc_warlock.cpp
```diff
@@ -781,6 +792,8 @@ void warlock_t::parse_player_effects()
...
+    parse_target_effects( d_fn( &warlock_td_t::debuffs_t::haunt ), talents.shadow_of_nathreza_2 ); // 1261990
```
nyterage: this line is unnecessary, can just add shaodw of nathreza to the previous haunt `parse_target_effects` 

should end up looking like `parse_target_effects( d_fn( &warlock_td_t::debuffs_t::haunt ), talents.haunt, talents.shadow_of_nathreza_2 );`

though this is likely unnecessary as well, since talenting shadow of nathreza 2 should automatically apply the effect N modifiers to haunt.

nyterage (changes requested):

engine/class_modules/warlock/sc_warlock.cpp
stahlscott: removing both of these. i didn't see any warning around this when working on it 🤔

engine/class_modules/warlock/sc_warlock.cpp
stahlscott: I missed the last comment here - I'll check this later to see if I can remove it entirely

PR #10999, opened 2026-02-21
[Warlock] Affliction updates and module adjustments

millanzarreta:
Several changes have been made to the Warlock module (especially for Affliction). The functionality of various spells and their interactions with each other and with talents have been tested in-game and implemented in the sim accordingly. Since there are quite a few changes, here is a summary:

- Affliction:
  - Malefic Grasp and the extra ticks on DoTs have been reimplemented to behave more closely to how they do in-game, with their own spellIDs, multipliers, and interactions with different talents.
  - The hidden buff that increases the damage of some spells while Darkglare is active has been implemented (replacing the less accurate manual implementation).
  - Several adjustments have been made to Seed of Corruption and its related talents to work as they do in-game (including implementing existing in-game bugs).
  - Several talent interactions have been adjusted to behave as they do in-game, such as Death's Embrace, Cunning Cruelty, Fatal Echoes, and Shard Instability, among others.
  - Other small tweaks and fixes.
  
- Demonology:
  - The in-game bug that caused Hand of Gul'dan (Demonology) to consume two stacks of Succulent Soul (Soul Harvester) has been fixed, and the module has been updated accordingly.

- Destruction:
  - A bug that caused SimC to crash if Cataclysm was selected but Lake of Fire was not has been fixed.
  - The application of the Lake of Fire debuff has been adjusted to start immediately after Cataclysm is executed and to end with the last tick.
  - A new spell has been created for the Demonfire Infusion proc, simplifying the code and improving the final report.

- Hellcaller
  - The `devil_fruit` proc has been included in the final report.

- Soul Harvester
  - The `manifested_avarize` proc has been included in the final report.

- General:
  - ~~Nethermancy is now only active on armor match (as in-game).~~
  - Several `//TODO` tags have been reviewed, tested, and validated; as a result, some spells have been adjusted and/or the `TODO` tag removed where applicable.

### DIFF
--- a/engine/class_modules/warlock/sc_warlock_init.cpp
+++ b/engine/class_modules/warlock/sc_warlock_init.cpp
@@ -119,9 +119,6 @@ namespace warlock
     // NOTE: 2026-02-17 Mark of Perotharn is being applied twice in what appears to be a bug
     if ( bugs )
       parse_passive_effects( hero.mark_of_perotharn, true );
-
-    if ( matching_gear )
-      parse_passive_effects( warlock_base.nethermancy );
   }
 
   void warlock_t::init_spells_affliction()
--- a/engine/class_modules/warlock/sc_warlock_init.cpp
+++ b/engine/class_modules/warlock/sc_warlock_init.cpp
@@ -75,13 +75,42 @@ namespace warlock
       deregister_passive_effect( hero.gloom_of_nathreza->effectN( 2 ) );
 
     register_passive_effect_mask( hero.mark_of_xavius,
-      affliction() ? effect_mask_t( false ).enable( 1 ) : effect_mask_t( true ).disable( 1 ) );
+      affliction() ? effect_mask_t( false ).enable( 1, 3 ) : effect_mask_t( true ).disable( 1 ) );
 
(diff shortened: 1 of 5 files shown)


engine/class_modules/warlock/sc_warlock_init.cpp
```diff
@@ -75,21 +75,53 @@ namespace warlock
...
     parse_all_class_passives();
     parse_all_passive_talents();
     parse_all_passive_sets();
 
     // NOTE: 2026-02-17 Mark of Perotharn is being applied twice in what appears to be a bug
     if ( bugs )
       parse_passive_effects( hero.mark_of_perotharn, true );
+
+    if ( matching_gear )
+      parse_passive_effects( warlock_base.nethermancy );
```
nyterage: This is no longer needed, we fixed passive class effect parsing yesterday.

nyterage (changes requested):

engine/class_modules/warlock/sc_warlock_init.cpp
millanzarreta: fixed in the new commit

PR #11003, opened 2026-02-21
[Demon Hunter] Fix action state leak in Undying Embers

taherbert:
The Undying Embers re-trigger in `soul_immolation_base_t::last_tick` calls `get_state()` to allocate an action state, then passes it to `trigger_dot()` inside a `make_event` lambda. `trigger_dot()` copies the state into the dot via `copy_state()` but doesn't consume the source. The release call was missing, so one action state leaked from the pool every time Undying Embers procced.

Every other `get_state()` call in the DH module passes the state to `schedule_execute()`, which takes ownership. This was the only path using `trigger_dot()` directly.

The lambda needs `mutable` because `action_state_t::release()` takes a non-const `T*&` and value-captured pointers are const by default.


taherbert:
Duplicate of #11000 which was merged earlier today. Same fix.

PR #11009, opened 2026-02-22
[Gear] Initial/Partial Trinket - Wraps of Cosmic Madness

Lucasmingus:
Travel time not implemented due to projectile spell 1263614 not in the spell data yet.

Everything else seems to be working and scaling correctly.


engine/player/unique_gear_midnight.cpp
```diff
@@ -2128,6 +2127,52 @@ void locuswalkers_ribbon( special_effect_t& e )
...
+      base_dd_min = base_dd_max = missile_damage;
+      aoe = -1;
+    }
+
+    double action_multiplier() const override
+    {
+      double m = proc_spell_t::action_multiplier();
+
+      int extra_targets = sim->active_enemies - 1;
+      m *= 1.0 + std::min( 0.3 * extra_targets, 1.5 ); // 30% per additional target, cap 150%
```
nyterage: We actually have a built in function for this. Though is limited to structs of type `generic_aoe_proc_t`. 

Can skip the entire struct here, and instead change its creation to `create_proc_action<generic_aoe_proc_t>( "cosmic_barrage", e, "cosmic_barrage", e.driver()->effectN( 2 ).trigger(), true )`

The final bool set there is telling it to use the 30% damage increase per target, and split damage.

nyterage (changes requested):

renanthera:
I can also prepare a spell data rebuild to get you 1263614.

It'll possibly need to be dropped in a rebase prior to merge, depending on how long it takes to merge this PR. If help is required with that, I can assist.

gastank:
This trinket looks like a channel. If so you'll need to do some workarounds. Set `channel = true` in the main action and can look at examples like tidestorm codex from `unique_gear_bfa` or manic torch from `unique_gear_df` for the `last_tick()` override you'll need.

renanthera:
On that topic, options for the reset behaviour may be desirable.

PR #11023, opened 2026-02-23
[Gear] Implementation of trinket - Wraps of Cosmic Madness

Lucasmingus:
Spelldata is needed for the hidden spell "Cosmic Barrage - 1263614" to calculate travel time properly.

I don't know how to generate the spelldata.

Otherwise, this implementation is fully functional.

### DIFF
--- a/engine/player/unique_gear_midnight.cpp
+++ b/engine/player/unique_gear_midnight.cpp
@@ -2128,6 +2128,54 @@ void locuswalkers_ribbon( special_effect_t& e )
   new locuswalkers_ribbon_t( e );
 }
 
+// Wraps of Cosmic Madness
+// 1259153 Driver
+// 1263614 Missile
+// 1263393 Damage Spell (Cosmic Barrage)
+// 1259103 Equip Driver
+void wraps_of_cosmic_madness( special_effect_t& e )
+{
+  struct cosmic_madness_channel_t : public proc_spell_t
+  {
+    cosmic_madness_channel_t( const special_effect_t& e ) :
+      proc_spell_t( "wraps_of_cosmic_madness_channel", e.player, e.driver() )
+    {
+      unsigned equip_id = 1259103;
+      auto equip        = find_special_effect( e.player, equip_id );
+      assert( equip && "missing equip effect" );
+
+      channeled = true;
+
+      auto missile_spell = e.player->find_spell( 1263614 );
+      assert( missile_spell && "missing missile spell 1263614" );
+
+      auto missile_damage = equip->driver()->effectN( 1 ).average( e );
+      auto cosmic_barrage = create_proc_action<generic_aoe_proc_t>( "cosmic_barrage", e, "cosmic_barrage", missile_spell, true );
+      cosmic_barrage->school = SCHOOL_COSMIC;
+      cosmic_barrage->base_dd_min = cosmic_barrage->base_dd_max = missile_damage;
+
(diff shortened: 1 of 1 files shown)


engine/player/unique_gear_midnight.cpp
```diff
@@ -2128,6 +2128,54 @@ void locuswalkers_ribbon( special_effect_t& e )
...
+      assert( equip && "missing equip effect" );
+
+      channeled = true;
+
+      auto missile_spell = e.player->find_spell( 1263614 );
+      assert( missile_spell && "missing missile spell 1263614" );
+
+      auto missile_damage = equip->driver()->effectN( 1 ).average( e );
+      auto cosmic_barrage = create_proc_action<generic_aoe_proc_t>( "cosmic_barrage", e, "cosmic_barrage", missile_spell, true );
+      cosmic_barrage->school = SCHOOL_COSMIC;
```
nyterage: Rather than setting the school here, you should be setting the impact action of `cosmic_barrage` to the damage spell.

Should look something like `cosmic_barrage->impact_action = create_proc_action<generic_proc_t>( "cosmic_barrage", e, "cosmic_barrage", 1263393 );`

then set the base_dd_min/max for the impact_action to the damage amount from the driver.

engine/player/unique_gear_midnight.cpp
```diff
@@ -2128,6 +2128,54 @@ void locuswalkers_ribbon( special_effect_t& e )
...
+      auto equip        = find_special_effect( e.player, equip_id );
+      assert( equip && "missing equip effect" );
+
+      channeled = true;
+
+      auto missile_spell = e.player->find_spell( 1263614 );
+      assert( missile_spell && "missing missile spell 1263614" );
+
+      auto missile_damage = equip->driver()->effectN( 1 ).average( e );
+      auto cosmic_barrage = create_proc_action<generic_aoe_proc_t>( "cosmic_barrage", e, "cosmic_barrage", missile_spell, true );
```
nyterage: is this meant to be an AoE spell? sending a missile out to every target in the sim?

nyterage (changes requested):

nyterage:
Added the missile spell to the spell data generator, and regenerated for you. If you merge upstream youll be able to see its data.

engine/player/unique_gear_midnight.cpp
Lucasmingus: I've attempted to fix these issues in the latest commit

nyterage: Looks about right, though i just realized the impact action itself is what does the AoE, should be `generic_aoe_proc_t` with the split damage + increase per target bool set. 

so, `cosmic_barrage->impact_action = create_proc_action<generic_aoe_proc_t>( "cosmic_barrage", e, "cosmic_barrage", damage_spell, true );`

Lucasmingus: Ok, I will fix that. 

Would i need to change how cosmic_barrage is defined aswell or can that be kept as a `generic_proc_t` ?

nyterage: can keep that as generic_proc_t, since i assume the missiles only target your current target, while the impact does aoe damage. Least thats what the data suggests

Lucasmingus: alright, thanks, i've updated it

PR #11035, opened 2026-02-24
[Items] Implement Lightless Lament Heaven's Glaive proc

taherbert:
## Summary

Lightless Lament's equip proc (Heaven's Glaive, spell 1266257) is silently a no-op. The generic `special_effect_t` system cannot traverse the `E_TRIGGER_MISSILE` (type 32) on the trigger spell, so `usable_proc()` classifies it as `SPECIAL_EFFECT_ACTION_NONE` / `SPECIAL_EFFECT_BUFF_NONE` and disables the effect.

The actual damage is on a grandchild spell two levels deep:

```
1266257 driver (RPPM 3.0, A_PROC_TRIGGER_SPELL, coeff 9.69)
  → 1266591 missile (E_TRIGGER_MISSILE) ← generic system stops here
    → 1266592 damage (E_SCHOOL_DAMAGE, Cosmic, 8yd radius)
```

This adds a custom implementation that bypasses the missile spell, using `generic_aoe_proc_t` with the damage spell (1266592) directly. Damage is sourced from the driver's effect 1 coefficient, with `role_mult` applied.

## Testing

- Verified proc does not fire before the fix (HTML report shows zero Heaven's Glaive damage)
- Verified proc fires after the fix (~271 DPS ST, ~585 DPS at 5 targets)
- Verified AoE split damage scaling: ST hit ~2720, 5-target hit ~1169/target (expected: 2720 × 2.2 / 5 ≈ 1197)
- Verified RPPM: ~23 procs per 300s fight (3.0 base × haste)
- Verified school is Cosmic, radius 8yd

### DIFF
--- a/engine/player/unique_gear_midnight.cpp
+++ b/engine/player/unique_gear_midnight.cpp
@@ -2278,15 +2278,14 @@ void torments_duality( special_effect_t& effect )
 // 1266592 AoE damage (Cosmic, 8yd radius, triggered by missile)
 void lightless_lament( special_effect_t& effect )
 {
-  auto missile = create_proc_action<generic_proc_t>( "heavens_glaive_missile", effect,
-                                                     effect.trigger() );
-  auto damage  = create_proc_action<generic_aoe_proc_t>( "heavens_glaive", effect,
-                                                          effect.trigger()->effectN( 1 ).trigger() );
+  auto missile = create_proc_action<generic_proc_t>( "heavens_glaive_missile", effect, effect.trigger() );
+  auto damage = create_proc_action<generic_aoe_proc_t>( "heavens_glaive", effect,
+      missile->data().effectN( 1 ).trigger() );
   damage->base_dd_min = damage->base_dd_max = effect.driver()->effectN( 1 ).average( effect );
   damage->base_multiplier *= role_mult( effect );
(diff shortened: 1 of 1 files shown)


gastank:
You don't want to bypass any intermediate missile spells. The effect should execute the missile (1266591). The missile should have the damage (1266257) as the impact_action. You also don't need to hard-code the damage id, since it can be pulled from `effect.trigger()->effectN( 1 ).trigger()`

taherbert:
Thanks for the feedback @gastank. Updated in a16ab3fd:

- Execute the missile spell (1266591) as the `execute_action` instead of bypassing it
- Wire the AoE damage spell as `missile->impact_action`
- Derive the damage spell dynamically via `effect.trigger()->effectN(1).trigger()` — no hardcoded ID
- Set `missile->dual = true` and `missile->stats = damage->stats` so only the damage shows in the report

renanthera:
I'd recommend using add_child instead of modifying stats object pointers in nearly all cases.

Stat object pointer modification can obscure useful details while not appreciably reducing reporting complexity.

taherbert:
Good call, switched to `add_child` and dropped the `stats` pointer swap and `dual = true`.

PR #11077, opened 2026-03-01
Add Devastation

Preheat:
Updated evoker profile to level 90 and gear


renanthera:
Consumables should generally not be included in profiles. Defaults for these are set in the class module/apl implementation file itself, and it may be sensible to provide different defaults depending on hero talent choices, etc.

PR #11081, opened 2026-03-01
[frost dk] fix segfault

HawkCorrigan:


Khazakdk:
fixed in https://github.com/simulationcraft/simc/commit/42bdd06768eb48d29693cd81da2646b663330d89

PR #11091, opened 2026-03-03
[Monk] Adjust assisted combat

Hurstilol:
Current overwrite for the CJL conditions make the assisted combat list useless since it just spams cjl when it shouldnt even use it once, changing the return to 0 makes it skip the line as it is supposed to


renanthera:
Doesn't address underlying issue and instead addresses the symptom, keeping this open for now though.

PR #11099, opened 2026-03-05
[Mage] Augury BLP + Pulse Targetting

pinepinepinepinepine:


pinepinepinepinepine:
ignore the commit history. lemme know if you want whatever changes.

im not confident on two things: 
1. not sure whether the last 2 or 3 splinters are executed alongside one another from two augurys, but it's likely 2 -- if you want an explanation, lemme know;
2. not sure if augury is considered for a retrigger off a retrigger: the chance is so unbelievably small, so 3x in a row hasn't ever happened (the increments im not too sure of either, but it is likely... its complicated-ish). gimme more time and ill be a bit more confident, but one way or the other doesn't matter because of how UNFATHOMABLY rare it is. anyways.

PR #11103, opened 2026-03-05
[Gear] - Tangle of Vibrant Vines

Eein:
This PR changes:

- tangle of vibrant vines + override to apply debuff correctly

### DIFF
--- a/engine/player/unique_gear_midnight.cpp
+++ b/engine/player/unique_gear_midnight.cpp
@@ -2542,6 +2542,18 @@ void crucible_of_erratic_energies( special_effect_t& effect )
 
   new dbc_proc_callback_t( effect.player, effect );
 }
+
+void tangle_of_vibrant_vines( special_effect_t& effect )
+{
+  auto missile = create_proc_action<generic_proc_t>( "tangle_of_vibrant_vines_missile", effect, effect.trigger() );
+  auto debuff = create_proc_action<generic_proc_t>( "tangle_of_vibrant_vines", effect, 1255357 );
+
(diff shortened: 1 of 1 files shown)


engine/player/unique_gear_midnight.cpp
```diff
@@ -2425,7 +2425,7 @@ void mindpiercers_sigil( special_effect_t& effect )
   aoe->base_dd_min = aoe->base_dd_max = effect.driver()->effectN( 1 ).average( effect );
 
   auto buff = create_buff<stat_buff_t>( effect.player, effect.player->find_spell( 1264404 ) )
-    ->set_stat_from_effect_type( A_MOD_STAT, effect.driver()->effectN( 2 ).average( effect ) );
+    ->add_stat( STAT_INTELLECT, effect.driver()->effectN( 2 ).average( effect ) );
```
nyterage: `set_stat_from_effect_type` reads the attribute bits from the effect data. This would not give anything but Int because the attribute bits held in the effect's misc_value1 contain only intellect.

Eein: Let me double check it today - it was late and i swear i was noticing something strange.

In any case i'll simplify this PR without the change and test and submit evidence later if applicable.
Thanks!

gastank: This is unnecessary. The code reads the spell data and applies the appropriate stat as the spell data indicates. Do you have a sim showing it giving anything other than int?

Eein: See previous comment: 
https://github.com/simulationcraft/simc/pull/11103#discussion_r2890928581

PR #11160, opened 2026-03-12
fix: crucible of erratic energies never procs for melee specs

### LINKED ISSUE #11158: Crucible of Erratic Energies (264507) never procs for Subtlety Rogue due to incorrect proc_flags
### Describe the bug
Crucible of Erratic Energies (item 264507) never procs for melee specs. The driver spell (1272091) has
  proc_flags=0x00015500, which contains no melee proc types. When the proc callback is initialized, it registers only
  for those proc types, meaning melee actions never trigger the callback at all. The SX_ONLY_PROC_FROM_CLASS_ABILITIES
  check inside trigger() is never reached for melee actions because the dispatch filter eliminates them first.

### Expected behavior
The trinket should proc from class abilities for melee specs. In-game testing confirms it does proc live from specific
   class abilities but not from auto-attacks. This suggests the live engine may handle
  SX_ONLY_PROC_FROM_CLASS_ABILITIES differently than simc — possibly by broadening proc eligibility beyond what
  proc_flags alone would allow — but the exact live behavior is not fully confirmed.

### To Reproduce
Sim any melee spec with Crucible of Erratic Energies equipped. Enable debug logging and search for voidlust 

### Additional information
* SimulationCraft Version: 1201.01

[log.txt](https://github.com/user-attachments/files/25927195/log.txt)

The original bug report diagnosis (proc_flags are wrong) is incorrect. wago.tools confirms ProcTypeMask=0x00015500
  on spell 1272091. simc is reading the correct game data.
  - simc does implement SX_ONLY_PROC_FROM_CLASS_ABILITIES (spell_attribute index 415, maps to SPELL_ATTR12 bit 31 =
  0x80000000). The check exists in dbc_proc_callback_t::trigger() at line 210, but is never reached for melee actions.
  - Several class modules hardcode allow_class_ability_procs = true on auto-attacks (warrior, rogue, monk, dk, dh,
  druid, shaman, paladin, hunter). If the registration issue is fixed, these would cause auto-attacks to incorrectly
  proc Crucible — live testing shows auto-attacks do not proc it.

craned70:
Closes #11158

## Problem

  Crucible of Erratic Energies (264507, driver spell 1272091) never procs for
  melee specs. The driver spell's ProcTypeMask (0x00015500) contains no melee
  proc types, so the dbc_proc_callback registers only for spell/ranged/periodic
  action types. Melee actions never reach the callback at all, meaning the
  SX_ONLY_PROC_FROM_CLASS_ABILITIES check in trigger() is never evaluated for
  melee specs.

  The driver spell has SX_ONLY_PROC_FROM_CLASS_ABILITIES set. In-game testing
  confirms the trinket does proc from melee class abilities (e.g. Shadowstrike,
  Backstab) but not from auto-attacks, suggesting the live engine broadens proc
  eligibility beyond the raw ProcTypeMask when this flag is present.

  ## Fix

  In the existing crucible_of_erratic_energies handler, initialize proc_flags_
  from spell data and OR in PF_MELEE_ABILITY before the callback is created.
  This allows melee class abilities to reach the callback, where the existing
  SX_ONLY_PROC_FROM_CLASS_ABILITIES / allow_class_ability_procs gate correctly
  filters to class abilities only. Auto-attacks fire as PF_MELEE (not
  PF_MELEE_ABILITY) and are unaffected.

  ## Note

  This is an item-specific fix. The underlying issue seems to be that simc's general
  proc system does not handle SX_ONLY_PROC_FROM_CLASS_ABILITIES by expanding
  proc registration. It only uses it as a narrowing filter after proc_flags
  matching. A general fix in dbc_proc_callback_t::initialize() would be more
  correct long-term but carries broader risk. Flagging here for maintainer
  consideration.

  ## Testing

  Simulated Subtlety Rogue with Crucible equipped. Before fix:
  voidlust never appears in combat log. After fix: voidlust proc attempts on
  Backstab, Shadowstrike, Eviscerate with no auto-attack procs. Buff gained and
  tracked correctly.

### DIFF
--- a/engine/player/unique_gear_midnight.cpp
+++ b/engine/player/unique_gear_midnight.cpp
@@ -2912,6 +2912,10 @@ void crucible_of_erratic_energies( special_effect_t& effect )
   effect.custom_buff = buff;
 
   effect.proc_flags2_ = PF2_LANDED;
+  // Spell data proc_flags (0x00015500) has no melee ability types. The driver has
+  // SX_ONLY_PROC_FROM_CLASS_ABILITIES set, which gates procs on allow_class_ability_procs,
+  // but the callback is never registered for melee actions without this addition.
+  effect.proc_flags_ = effect.driver()->proc_flags() | PF_MELEE_ABILITY;
 
   new dbc_proc_callback_t( effect.player, effect );
 }


nyterage:
This is the incorrect fix for this issue. In game its missing these proc flags too, but triggers from things like buff applications, heals, etc. Things like those are often not implemented in sims, and thus it wont proc. Actual fix would be to implement these effects so that it triggers as expected.

EvanMichaels:
https://github.com/simulationcraft/simc/pull/11253/changes

This change should be the correct fix for this issue based on the description and discussions in the Rogue Discord, as I had proposed it was triggering from energize effects when looking at the logs.

PR #11181, opened 2026-03-14
[Mage] AC benefits Pulse's cast time

pinepinepinepinepine:


pinepinepinepinepine:
don't know what's preferable between the two commits. i was going to leave the 2nd commit local and push only if prompted, but yeah. i dunno if the norm is to look for the ptr flag, or check if the affected flag applies. anyways, lmk if i should revert.

edits because i gotta be dyslexic

might be a good idea to put it within the constructor for it to not repeatedly check, but i dunno.

pinepinepinepinepine:
another zinger of a commit. i don't know. just choose whichever, gonna have to delete it whenever 12.0.5 goes live -- it's all pointless! 
cant help but giggle, though.

Issue #2447, opened 2015-08-13
Replace hard-to-read Actions with CLC-like shorthand to ease priority experimentation for novices


navv1234:
Originally reported on Google Code with ID 2448

```
The APL is a scary wall of text to new users. Deciphering Actions such as:

actions.single+=/judgment,if=talent.empowered_seals.enabled&((seal.truth&buff.maraads_truth.remains<cooldown.judgment.duration*2)|(seal.righteousness&buff.liadrins_righteousness.remains<cooldown.judgment.duration*2))


is a deal breaker for non-coders. How many potential users abandon SC because they
can't understand the syntax? 

Solution: 
1) Create Text Templates to convert the Actions to CLC-like abbreviations
2) Include, either in the APL or another tab the abbreviations and their meaning in
English, not code.
3) Toggle using either the abbreviations or their English description in the APL, not
code.

Bonus:
4) Provide a UI interface to quickly change action priority positions.
5) Provide a UI interface to toggle available actions.
5.1) Example: Toggle T17 2pc
5.2) Example: Toggle (Action) 
6) Provide a UI interface to import/export CLC priority strings that directly modify
the APL. 


```

Reported by `Fogyreef` on 2015-02-20 16:31:45

PR #11182, opened 2026-03-14
feat: Adds the ability to use spaces in simc input

thevinter:
This PR adds whitespace stripping outside of double-quoted regions to both the C++ option parser (option.cpp) and the Python APL converter (ConvertAPL.py). This allows users to write more readable .simc files using spaces 

e.g.,
```
actions.ec_st += /sunfire, target_if= remains < 2 | refreshable & variable.eclipse_down
```
instead of 
```
actions.ec_st+=/sunfire,target_if=remains<2|refreshable&variable.eclipse_down
```

Important, while minor this is still a **breaking change**

This removes support for same-line option definitions that relied on spaces as token separators (e.g., `name=Vinter race=human`). 

Because spaces are now stripped, multiple definitions on one line will be merged into a single token. Each option must now be on its own line.

This tradeoff is intentional since same-line definitions aren't used anywhere in the current profiles, and allowing spaces in input significantly improves readability and lowers the barrier for others editing .simc files.

### DIFF
--- a/engine/class_modules/apl/ConvertAPL.py
+++ b/engine/class_modules/apl/ConvertAPL.py
@@ -28,6 +28,18 @@ ignored_comments = [
     '# Executed every time the actor is available.',
 ]
 
+def strip_whitespace_preserve_quotes(s):
+    """Strip whitespace outside of double-quoted regions."""
+    result = []
+    in_quotes = False
+    for c in s:
+        if c == '"':
+            in_quotes = not in_quotes
+            result.append(c)
+        elif in_quotes or not c.isspace():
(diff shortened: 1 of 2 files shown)


renanthera:
I am personally against any breaking change to the input parser, as I am not convinced that permitting whitespace in lines is sufficient enough to improve readability. I believe a vscode plugin exists for apl syntax highlighting, which provides much of the same benefit.

An apl-syntax pretty printer that yields expression trees would be very cool, but out of the scope of simc unless added as a reporting feature. A very old issue exists (#2447) to track this.

PR #11231, opened 2026-03-16
Create nix flake for compiling and developing

Denis101:
Adds a relatively simple `flake.nix` that enables the following with nix (and NixOS):
* Run `nix develop` in the root directory to get a development shell pre-installed with clang and llvm. Optionally run `nix develop .#simcqt` to get a devshell with Qt6 to develop the GUI with.
* Run `nix build` to compile `simc` into a nix derivation. The default package does not include the GUI, this can be built with `nix build .#simcqt`.
* Run `nix run . [options]` in the root directory to run the compiled binary in the derivation.

This can be used in flake-based NixOS configurations to include the `simc` package in theoretically any nix installation. I have only tested this on my own x86_64 NixOS (Linux 6.18) machine, so not sure if the compilation may fail in either Windows (via WSL) or macOS.


JoeGrn (approved):

renanthera:
I've never dove very deeply into flakes, and had @emallson provide some feedback, as they have a little more nix-specific experience than me. Here's a mixture of feedback from me and them.

> Run nix fmt to format github actions, docker, and nix files. I thought about including C/C++ formatters here but figured best left out for now.

File formatting in simc is very much up to individual maintainers, and there is no consistent formatting procedure for the entire repository.

Bad format commits will easily generate very massive whitespace changes across millions of lines, especially in data files.

> Run nix-shell in the root directory to get a development shell pre-installed with clang and llvm.

Allowing for a QT-less devshell would be desirable. Many developers skip building the GUI entirely, as many do not use it.

renanthera:
To sum up our discord conversation:
- Not hardcoding version string would be nice.
- Python environment handling for `casc_extract` and `dbc_extract3` would be nice.

PR #11232, opened 2026-03-16
[Shadow] Add Deaths Torment to SW:D execute condition

Urcra:
It's worth it to cast SW:D over mindflay if we are specced into DT after the buffs. Results and approach was discussed in https://github.com/WarcraftPriests/mid-shadow-priest/issues/7


seanpeters86:
@Urcra could you update/re-test after the buffs if you get a sec?

Urcra:
> @Urcra could you update/re-test after the buffs if you get a sec?

👍 I'll do that tomorrow

PR #11244, opened 2026-03-18
[DH] burning blades removed from felblade

Aezeor:
burning blades still is not applied from voidblade


engine/class_modules/sc_demon_hunter.cpp
ToppleTheNun: this needs to be `demon_hunter_attack_t::action_multiplier()`

PR #11249, opened 2026-03-19
[PvP] PvP scaling system with DBC coefficients, dampening, trinket set bonus

followingthefasciaplane:
Overrides are still not entirely implemented, but I would appreciate a review. Documentation is found in `/engine/sim/pvp_scaling.hpp`


renanthera:
Parsing of pvp data, both aura data and on effects can be handled by existing systems. All that is required is for said existing parse systems to support pvp effects.

The large set of coefficient options provided should not be configurable by users, as you cannot change these values in game directly.

Support for pvp would also require APLs, pvp talent implementations and likely additional features in the raid events system, none of which presently exist.

I'd be reticent to merge such a large set of changes from someone who isn't already somewhat active in the community around simc, due to concerns about continued support for the features.

PR #11250, opened 2026-03-19
Add buffed secondary stat ratings to HTML and JSON

seriallos:
Track secondary stat ratings in addition to percentages. Show the buffed ratings in the HTML report. Replace the raw gear amount with the snapshot buff amount in the JSON report.


seriallos:
Put the buffed rating in parens next to the percentages. Didn't feel worth a full new column just for those values.

image

PR #11253, opened 2026-03-19
[Action] trigger generic harmful procs from energize_power effects

gastank:
It seems 'Generic Helpful' can be triggered by various resource energize effects from actions such that items & effects described as proccing off "heals" or have otherwise incorrect proc flags can trigger off harmful attacks and spells.

action_t::suppress_callback_from_energize = true to disable for the action.

### DIFF
--- a/engine/action/action.cpp
+++ b/engine/action/action.cpp
@@ -498,6 +498,7 @@ action_t::action_t( action_e ty, util::string_view token, player_t* p, const spe
     pre_execute_state(),
     snapshot_flags(),
     update_flags( STATE_TGT_MUL_DA | STATE_TGT_MUL_TA | STATE_TGT_CRIT ),
+    energize_state( std::make_unique<action_state_t>( this, p ) ),
     target_cache(),
     options(),
     state_cache(),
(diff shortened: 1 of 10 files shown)


EvanMichaels:
I believe this should resolve the issue described here: https://github.com/simulationcraft/simc/pull/11160

EvanMichaels:
Think the only issue here that I see is that `action_t::gain_energize_resource` feels comparatively underutilized vs. direct calls to `p()->resource_gain()` ?

It will at least work intrinsically for ability `action_energize::` type ones which is probably enough to consistently proc things in stuff like the Rogue module, but I suspect many of the other resource gains would technically trigger it as well.

engine/action/action.cpp
```diff
@@ -5143,7 +5145,16 @@ player_t* action_t::get_expression_target()
...
 double action_t::gain_energize_resource( resource_e resource_type, double amount, gain_t* g )
 {
-  return player->resource_gain( resource_type, amount, g, this );
+  auto ret = player->resource_gain( resource_type, amount, g, this );
+
+  // energize_power can trigger generic helpful proc effects
+  if ( callbacks && caster_callbacks && ( !suppress_caster_procs || enable_proc_from_suppressed ) &&
+       !suppress_callback_from_energize )
+  {
+    player->trigger_callbacks( PROC1_NONE_HELPFUL, PROC2_HIT, this, energize_state.get() );
```
EvanMichaels: In follow-up to my comment on the main thread, I wonder if maybe this should just be moved into `player_t::resource_gain` instead, but only run if `action_t* action` is passed in? Many of the direct calls to `resource_gain` that are tied to an action pass in the action, similar to L5148 above.

e.g.
```
void actions::rogue_action_t<Base>::trigger_combo_point_gain( int cp, gain_t* gain )
{
  p()->resource_gain( RESOURCE_COMBO_POINT, cp, gain, this );
}
```

Similar in `priest_t::generate_insanity` and most Runic Power gains.

PR #11267, opened 2026-03-22
Update Devourer APL

misterwiki:
Made some updated to the Devourer APL to update reap usage for annihilator and trinket usage to enforce recklessness potion grants mastery


misterwiki:
https://www.raidbots.com/simbot/report/aZAvLbrgXiwJNodagCWXQW

PR #11271, opened 2026-03-23
[stats] Validate Sampled Distributions

renanthera:
- **[stats] Use `length-1` as all of our statistics are samples, not populations.**
- **[stats] Preliminary implementation to validate that an `extended_sample_data_t` object is normal.**


renanthera:
This would be nice to have but is really, really slow. Would need more thought on optimization, as well as possibly some tweaking to the stats system in order to apply these sorts of tests to more select sample data object instances.

PR #11291, opened 2026-03-25
[Gear] Gloom-Spattered Dreadscale

justindriggers:


engine/player/unique_gear_midnight.cpp
```diff
@@ -2956,6 +2956,84 @@ void tangle_of_vibrant_vines( special_effect_t& effect )
...
+    void execute_with_amount( double amount )
+    {
+      base_dd_min = base_dd_max = amount;
+      absorb_t::execute();
+    }
+  };
+
+  struct gloomspattered_dreadscale_t : public generic_aoe_proc_t
+  {
+    gloomspattered_dreadscale_absorb_t* absorb;
```
gastank: You don't need an absorb_t action at all - you can just make the absorb_buff_t and trigger it directly with the value via `->trigger( -1, <value> );` see brand of ceaseless ire in TWW for example.

engine/player/unique_gear_midnight.cpp
```diff
@@ -2956,6 +2956,84 @@ void tangle_of_vibrant_vines( special_effect_t& effect )
...
+    {
+      // Most spell data comes from the driver, but damage is determined by another spell
+      base_dd_min = base_dd_max = e.player->find_spell( 1260627 )->effectN( 1 ).average( e );
+    }
+
+    void execute() override
+    {
+      generic_aoe_proc_t::execute();
+
+      absorb->execute_with_amount( execute_state->result_amount * execute_state->n_targets );
```
gastank: The execute state is going to have the amount for the last hit, but not all targets may have taken the same amount of damage. If in-game you only get a single application of the shield, instead of one for each target hit, best way would be to store the value from each impact() on a counter then trigger the shield with that at end of execute.

engine/player/unique_gear_midnight.cpp
```diff
@@ -2956,6 +2956,84 @@ void tangle_of_vibrant_vines( special_effect_t& effect )
...
+
+  struct gloomspattered_dreadscale_t : public generic_aoe_proc_t
+  {
+    gloomspattered_dreadscale_absorb_t* absorb;
+
+    gloomspattered_dreadscale_t( const special_effect_t& e )
+      : generic_aoe_proc_t( e, "gloomspattered_dreadscale_damage", e.driver(), true ),
+        absorb( new gloomspattered_dreadscale_absorb_t( e.player ) )
+    {
+      // Most spell data comes from the driver, but damage is determined by another spell
```
gastank: This is a dual-effect item, so both the on-use 1260633 and the on-equip 1260627 are granted by the item. You want to register the on-equip to DISABLED_EFFECT() further down in register_special_effects(). You can get the on-equip effect via find_special_effect( effect.player, 1260627 ) + assert to ensure Blizzard hasn't changed the data without us knowing. There are several examples of this in the midnight file such as seed_of_the_devouring_wild. Once you have the on-equip effect you can grab the coefficient spell via effect.driver().

engine/player/unique_gear_midnight.cpp
justindriggers: Originally created a separate action so that it would be "pretty" in the healing stats HTML, but it's not important.

gastank: You don't need an action_t for that, just a stats_t will work. Can do set_absorb_source( effect.player->get_stats( <name> ) and that should make it show up on the HTML. Same with set_absorb_gain() if you want the hp 'gain' to show up on the resource gain section. Can see shadowflame_rage_t in dragonflight for example.

justindriggers: I did figure that much out, but is there a way to make it show the Wowhead link to the item rather than the plain string?

gastank: You can pass `this` as second argument to get_stats(), it should hold the item info that'll get properly decorated when the html processes. I'd also just name the generic_aoe_proc_t as "gloomspattered_dreadscale" so you don't get the dangling "_damage". Can do the same for the absorb buff as there's no other buff with the same name.

justindriggers: Perfect, thanks!

PR #11307, opened 2026-03-29
[Devourer] Vaelgor before Potion

Preheat:
Forcing Mastery on Reckless is a gain and also affects stat weights (specifically Haste vs Mastery) in Top Gear.

Report:
https://www.raidbots.com/simbot/report/mMFjS8RaN5X8D3w8eTUdiN


ToppleTheNun:
Needs to be in the .simc file too but I'll take care of that part.

PR #11340, opened 2026-04-06
[Gear] Umbral shift + Spell Data

Lucasmingus:


engine/player/unique_gear_midnight.cpp
```diff
@@ -3490,6 +3490,27 @@ void voidlight_bindings( special_effect_t& effect )
...
+// 253826 Mastery Buff
+void umbral_shift( special_effect_t& effect)
+{
+  auto equip = find_special_effects( effect.player, 1290152 );
+  assert( !equip.empty() && "Umbra-Weaver's Portent missing equip effect" );
+
+  auto buff = effect.driver()->effectN( 1 ).trigger();
+  auto stat_buff = create_buff<stat_buff_t>( effect.player, buff );
+
+  range::for_each( equip, [ stat_buff ]( auto effect ) {
```
gastank: Seems like 3pc is required for the buff to actually grant mastery? You'll want to check `effect.player->sets->has_set_bonus( effect.player->specialization(), MID_UWP, B3 )` before adding the stat buff.

Lucasmingus: Thanks for the headsup to add a check. It's added now

PR #11435, opened 2026-05-03
Add raid_events=absorb

mat-simon:
## What

- `raid_events=absorb` — new event that puts an `absorb_buff_t` on a target enemy. Options: `target=`, `amount=`, `school=` plus standard timing controls. No `amount=` means unbreakable for the window.
- `target.has_absorb` APL expression and `player_t::has_absorb()` / `current_absorb_amount()` helpers (scoped to enemy actors via `enemy_t` overrides).
- Attacker DPS attribution: when an action damages an enemy with an active absorb, the attacker's stats credit the rolled mitigated damage instead of just the post-absorb HP impact. Implemented as a virtual `credited_damage_amount` on `player_t`, overridden on `enemy_t` to add the absorbed delta. `result_amount` semantics are preserved.

## Use case

Encounter modeling — boss / add mechanics with absorb shields (Aleria-style fights). Analogous in shape to `raid_event=vulnerable`; class APLs and class modules can react to shielded targets via the helpers and the `target.has_absorb` expression.

### DIFF
--- a/engine/player/player.cpp
+++ b/engine/player/player.cpp
@@ -58,6 +58,7 @@
 #include "sim/plot.hpp"
 #include "sim/proc.hpp"
 #include "sim/proc_rng.hpp"
+#include "sim/raid_event.hpp"
 #include "sim/scale_factor_control.hpp"
 #include "sim/sim.hpp"
 #include "util/io.hpp"
@@ -4474,6 +4475,21 @@ void player_t::init_assessors()
     state->target->do_damage( state );
     return assessor::CONTINUE;
   } );
+
+  // Credit absorbed-on-enemy damage back to attacker stats when an absorb raid event is configured.
+  if ( !is_enemy() && range::any_of( sim->raid_events,
+                                     []( const auto& e ) { return e->type == "absorb"; } ) )
+  {
+    assessor_out_damage.add( assessor::TARGET_DAMAGE + 1, []( result_amount_type, action_state_t* s ) {
+      if ( s->target && s->target->is_enemy() )
+      {
+        double absorbed = s->result_mitigated - s->result_absorbed;
+        if ( absorbed > 0 )
+          s->result_amount += absorbed;
+      }
+      return assessor::CONTINUE;
+    } );
+  }
 
   // Logging and debug .. Technically, this should probably be in action_t::assess_damage, but we
   // don't need this piece of code for the vast majority of sims, so it makes sense to yank it out
@@ -12089,6 +12105,9 @@ std::unique_ptr<expr_t> player_t::create_expression( util::string_view expressio
 
   if ( expression_str == "is_enemy" )
     return expr_t::create_constant( "is_enemy", is_enemy() );
+
+  if ( expression_str == "has_absorb" )
+    return make_fn_expr( expression_str, [ this ] { return has_absorb() ? 1.0 : 0.0; } );
 
   if ( expression_str == "attack_haste" )
     return make_fn_expr( expression_str, [this] { return cache.attack_haste(); } );
--- a/engine/sim/raid_event.cpp
+++ b/engine/sim/raid_event.cpp
@@ -5,6 +5,7 @@
 
 #include "action/heal.hpp"
 #include "action/action.hpp"
+#include "action/action_state.hpp"
 #include "action/spell.hpp"
 #include "buff/buff.hpp"
 #include "dbc/dbc.hpp"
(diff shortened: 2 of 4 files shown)


mat-simon:
Pushed an update: added a `target.has_absorb` expression to the player expression dispatch (`player_t::create_expression` in `engine/player/player.cpp`). It returns `1.0` when the actor has any active `absorb_buff_t`, `0.0` otherwise.

This was originally scoped as a follow-up but I'm pulling it into this PR because a planned class-module follow-up (Voidweaver's Devour Matter) needs to gate an APL line on shield presence, and the expression is the cleanest way to expose that. Three lines added next to the `is_add` / `is_boss` / `is_enemy` predicates.

No diff impact on the rest of the PR.

engine/player/player.cpp
```diff
@@ -5888,6 +5888,23 @@ double player_t::composite_player_absorb_received_multiplier() const
...
+  return false;
+}
+
+double player_t::current_absorb_amount() const
+{
+  double total = 0;
+  for ( const absorb_buff_t* ab : absorb_buff_list )
+    if ( ab->check() )
+      total += ab->current_value;
+  return total;
```
renanthera: Simc's default implementation only supports absorbs that absorb all incoming damage up to a limit.

There are several other types that exist, specifically:
- absorbs a fraction of incoming damage, no limit
- absorbs a fraction of incoming damage, limit per event
- absorbs a fraction of incoming damage, total value limited

Without a fair bit of work to support these cases AND have each module that has player absorbs updated to use the new core absorb behaviour, this solution won't work.

Alternatively, this could be implemented strictly for enemy actors. 

EDIT: Moved out of player scope, resolved.

engine/player/player.cpp
```diff
@@ -8694,9 +8720,9 @@ void player_t::do_damage( action_state_t* incoming_state )
   collect_dmg_taken_data( *this, incoming_state,
                           incoming_state->result_mitigated - incoming_state->self_absorb_amount );
 
-  if ( incoming_state->result_amount > 0.0 )
+  if ( incoming_state->result_absorbed > 0.0 )
   {
-    actual_amount = resource_loss( RESOURCE_HEALTH, incoming_state->result_amount, nullptr, incoming_action );
+    actual_amount = resource_loss( RESOURCE_HEALTH, incoming_state->result_absorbed, nullptr, incoming_action );
```
renanthera: Not related specifically to this change, but the PR in general...

Historically, there were issues where some effects did not proc when the action attempting to trigger the proc was fully absorbed. The most notable case I immediately recall is Slimy Expulsion Boots during Dragonflight, most notably in Underrot (as it was available in M+ and the final boss has an absorb over its entire HP for the entire fight).

More investigation is definitely required before making this change.

engine/player/player.cpp
```diff
@@ -12090,6 +12116,9 @@ std::unique_ptr<expr_t> player_t::create_expression( util::string_view expressio
   if ( expression_str == "is_enemy" )
     return expr_t::create_constant( "is_enemy", is_enemy() );
 
+  if ( expression_str == "has_absorb" )
+    return make_fn_expr( expression_str, [ this ] { return has_absorb() ? 1.0 : 0.0; } );
```
renanthera, line 12110: An expression that yields the total amount of absorbs could also be useful, but it will be hard to avoid the trap that `UnitGetTotalAbsorbs` falls for given that there are multiple types of absorbs.

engine/player/player.cpp
```diff
@@ -5888,6 +5888,31 @@ double player_t::composite_player_absorb_received_multiplier() const
   return current.absorb_received_multiplier;
 }
 
+bool player_t::has_absorb() const
+{
+  // Scoped to enemy actors. Player-side absorb subtypes (fractional, per-event-
```
renanthera: Why is something "scoped to enemy actors" not in `sc_enemy.cpp`, the actual enemy implementation?

mat-simon: Good point, will fix

engine/action/action.cpp
```diff
@@ -2133,20 +2133,30 @@ void action_t::assess_damage( result_amount_type rt, action_state_t* state )
...
+  // Credit absorbed-on-enemy damage to attacker stats. (mitigated - absorbed)
+  // is per-event; self_absorb_amount would accumulate across DoT ticks.
+  double credited_amount = state->result_amount;
+  if ( state->target && state->target->is_enemy() )
+  {
+    double absorbed = state->result_mitigated - state->result_absorbed;
+    if ( absorbed > 0 )
+      credited_amount += absorbed;
+  }
+
```
renanthera: This (and the following changes at 2171-2179) could be an assessor that `enemy_t` registers potentially?

engine/sim/raid_event.cpp
```diff
@@ -2305,6 +2391,8 @@ std::unique_ptr<raid_event_t> raid_event_t::create( sim_t* sim, util::string_vie
     return std::unique_ptr<raid_event_t>( new stun_event_t( sim, options_str ) );
   if ( name == "vulnerable" )
     return std::unique_ptr<raid_event_t>( new vulnerable_event_t( sim, options_str ) );
+  if ( name == "shield" )
```
renanthera: Calling the raid event "absorb" would probably be more expected. Shield makes it sound like the enemy/enemies will block.

mat-simon: ok will change

engine/action/action.cpp
mat-simon: Is a method on enemy ok?

double enemy_t::credited_damage_amount( const action_state_t* s ) const override {
    double absorbed = s->result_mitigated - s->result_absorbed;
    return s->result_amount + std::max( 0.0, absorbed );
  }

mat-simon: I could make the assessor but I think the functionality would be the same?

gastank: There's no need to inject additional code & calculations into the main codepath that every damage goes through for something that only applies to a niche case. You can contain all the code entirely within `raid_events.cpp` by utilizing assessors and only have it apply to the player should the profile actually create a shield event.

Since you can't really search for existing assessors, something like `raid_event_t::init( sim )` would be a good place to search for the absorb event and add the assessor if necessary.

engine/player/assessor.hpp
```diff
@@ -24,6 +24,7 @@ namespace assessor
   {
     TARGET_MITIGATION = 100U,    // Target assessing (mitigation etc)
     TARGET_DAMAGE     = 200U,    // Do damage to target (and related functionality)
+    ABSORB_CREDIT     = 250U,    // Credit absorbed-on-enemy damage to attacker stats
```
renanthera: It'd be more common to set the assessor priority to `TARGET_DAMAGE+1` if you want the assessor to occur after `TARGET_DAMAGE` assessors, but beyond that don't mind the exact ordering. That is likely sufficient in this context.

mat-simon: Ok sounds good, dropped the enum entry and assigned it TARGET_DAMAGE+1

renanthera:
I've closed all previous reviews that are no longer relevant. I know less about the finer details of raid event implementation though, so I am not commenting on changes there.

gastank:
Instead of tracking just a single instance of the buff, you can use a `target_specific_t<buff_t>` to track across multiple targets in case there are multiples with the same name.

Other than that seems fine. Anything missed can be addressed if it comes up when ppl use the raid_event.

Edit: thinking more on it, it's prolly better to just use dynamic target searching (along with the above target_specific_t) for all fight style instead of just dungeon routes, and forgo the static parse_target() entirely. Users can attach add events to any fight style and you'd want to cover those cases where they want an add to have a shield.

mat-simon:
Yeah I like it, both done

mat-simon:
Caught a pet-damage credit bug while testing. raid_event_t::init runs before init_actors, so pets weren't getting the credit assessor. Moved the registration into player_t::init_assessors

gastank:
We really should have a way to attach callbacks to actor initialization from outside the actor, but that can be addressed later.

PR #11492, opened 2026-06-01
[Midnight] Darkmoon Blood: buff all secondaries tied for lowest

mat-simon:
On a tie for the lowest secondary, Blood should buff every tied stat, not just the first in priority order (confirmed in-game).


renanthera:
Addressed by #11493

PR #11516, opened 2026-06-19
[Mage] Fire mage 12.1 set bonus

Lucasmingus:


engine/class_modules/sc_mage.cpp
```diff
@@ -2384,7 +2384,13 @@ struct hot_streak_spell_t : public custom_state_spell_t<fire_mage_spell_t, hot_s
     if ( p()->buffs.hot_streak->check() || p()->buffs.hyperthermia->check() )
       return 0_ms;
 
-    return custom_state_spell_t::execute_time();
+    timespan_t t = custom_state_spell_t::execute_time();
+
+    // Fire Mage 12.1 Set Bonus 4pc
+    if ( p()->sets->has_set_bonus( MAGE_FIRE, MID2, B4 ) && p()->buffs.pyroclasm->check() )
```
vituscze: Since that effect is now always on (and the tier just modifies its value), the tier check shouldn't be here.

engine/class_modules/sc_mage.cpp
```diff
@@ -2384,7 +2384,13 @@ struct hot_streak_spell_t : public custom_state_spell_t<fire_mage_spell_t, hot_s
...
+      t *= 1.0 + p()->buffs.pyroclasm->data().effectN( 3 ).percent();
```
vituscze: This all should be in `execute_time_pct_multiplier`

engine/class_modules/sc_mage.cpp
```diff
@@ -2399,6 +2405,10 @@ struct hot_streak_spell_t : public custom_state_spell_t<fire_mage_spell_t, hot_s
 
     c += p()->buffs.hyperthermia->check_value();
 
+    // Fire Mage 12.1 Set Bonus 2pc
+    if ( p()->sets->has_set_bonus( MAGE_FIRE, MID2, B2 ) && p()->buffs.pyroclasm->check() )
```
vituscze: Same this as above wrt the tier check

engine/class_modules/sc_mage.cpp
```diff
@@ -2399,6 +2405,10 @@ struct hot_streak_spell_t : public custom_state_spell_t<fire_mage_spell_t, hot_s
...
+      c += p()->buffs.pyroclasm->data().effectN( 2 ).percent();
```
vituscze, line 2419: Is it how this works? Spell data says it's a percent modifier, not a flat one (so should be multiplied and not added)

vituscze (changes requested):

PR #11521, opened 2026-06-21
[Gear] Trinket - Vile Vial of Volatile Venom

Lucasmingus:


engine/player/unique_gear_midnight.cpp
```diff
@@ -4116,6 +4156,8 @@ void register_special_effects()
   set_min_version( wowv_t( 12, 0, 7 ) );
   register_special_effect( 1284696, trinkets::sporelords_mycelium );
   reset_version_check();
+  register_special_effect( 1293316, trinkets::vile_vial_of_volatile_venom );
```
nyterage, line 4159: the registers should be behind a `set_min_version` check similar to the other special effects to prevent trying to initialize it when its data doesnt exist.

engine/player/unique_gear_midnight.cpp
```diff
@@ -3079,6 +3079,46 @@ void sporelords_mycelium( special_effect_t& effect )
...
+    {
+      auto debuff_data = e.player->find_spell( 1295123 );
+
+      create_all_stat_buffs( e, debuff_data, 0, [ this ]( stat_e s, buff_t* b ) { debuffs[ s ] = b; } );
+
+      create_all_stat_buffs( e, e.driver(), 0, [ this ]( stat_e s, buff_t* b ) {
+        // The debuff rolls a stat independently of the stat granted by the buff
+        b->set_stack_change_callback( [ this ]( buff_t*, int old_, int new_ ) {
+          if ( old_ > 0 && new_ == 0 )
+            debuffs.at( player->rng().range( secondary_ratings ) )->trigger();
```
nyterage: can this roll the same stat as the buff? the current implementation would suggest it can? but wanted to check anyways.

nyterage (changes requested):

engine/player/unique_gear_midnight.cpp
nyterage: Also, this can be an expire_callback rather than a stack change callback, should simplify the logic.

Lucasmingus: yep, all is fully random, i checked on the ptr ingame

engine/player/unique_gear_midnight.cpp
Lucasmingus: i'll fix that!

PR #11524, opened 2026-06-23
[Mage] Frost mage 12.1 set bonus

Lucasmingus:


engine/class_modules/sc_mage.cpp
```diff
@@ -4206,6 +4207,7 @@ struct glacial_spike_t final : public frost_mage_spell_t
   action_t* duality_pyroblast = nullptr;
   int freezing_consume;
   shatter_source_t* shatter_source;
+  real_ppm_t* rppm_set_bonus = nullptr;
```
vituscze: This should go into mage_t and be initialized in `init_rng`, like all the other RNGs

engine/class_modules/sc_mage.cpp
```diff
@@ -7061,6 +7076,21 @@ int mage_t::trigger_shatter( player_t* target, action_t* action, int max_consump
       debuff->refresh();
   }
 
+  if ( consume_stacks > 0 && sets->has_set_bonus( MAGE_FROST, MID2, B2 ) )
```
vituscze: Was this tested? Using `consume_stacks` would mean it doesn't interact with FoF at all. Worth a comment if that is the case.

engine/class_modules/sc_mage.cpp
```diff
@@ -7061,6 +7076,21 @@ int mage_t::trigger_shatter( player_t* target, action_t* action, int max_consump
...
+  {
+    double chance = sets->set( MAGE_FROST, MID2, B2 )->effectN( 2 ).percent();
+    for ( int i = 0; i < consume_stacks; i++ )
+    {
+      if ( rng().roll( chance ) )
+      {
+        procs.icicle_from_set_bonus->occur();
+        // Icicles generated this way cannot grant the Glacial Spike buff on their own (bug?),
+        // the player still has to wait for the cyclic icicles proc to gain the buff
+        trigger_icicle( 1, false );
```
vituscze: Probably worth controlling this behavior with `player_t::bugs`

vituscze (changes requested):

engine/class_modules/sc_mage.cpp
Lucasmingus: It works with FoF. Was meant to use `shatter_stacks` instead of `consume_stacks`

PR #11527, opened 2026-06-23
[Mage] Arcane 12.1 changes + set bonus

Lucasmingus:
Arcane Pulse changes, Prismatic Bolt and 12.1 set bonus.
Added game version checks on everything related to Arcane Pulse changes.

### DIFF
--- a/engine/class_modules/sc_mage.cpp
+++ b/engine/class_modules/sc_mage.cpp
@@ -3051,11 +3051,9 @@ struct arcane_explosion_t final : public arcane_mage_spell_t
 struct arcane_pulse_t final : public arcane_mage_spell_t
 {
   action_t* arcane_pulse_echo = nullptr;
-  bool is_echo;
 
   arcane_pulse_t( std::string_view n, mage_t* p, std::string_view options_str, bool echo = false ) :
-    arcane_mage_spell_t( n, p, echo ? p->find_spell( 1243460 ) : p->talents.arcane_pulse ),
-    is_echo( echo )
+    arcane_mage_spell_t( n, p, echo ? p->find_spell( 1243460 ) : p->talents.arcane_pulse )
   {
     parse_options( options_str );
     aoe = -1;
@@ -3115,7 +3113,7 @@ struct arcane_pulse_t final : public arcane_mage_spell_t
       int charges_per_hit = 1;
 
       // Bug: with Impetus talented, the Reverberate echo generates double the arcane charges of the main cast
-      if ( is_echo && p()->bugs && p()->talents.impetus.ok() )
+      if ( background && p()->bugs && p()->talents.impetus.ok() )
         charges_per_hit = 2;
 
       p()->trigger_arcane_charge( num_targets_hit * charges_per_hit );
@@ -3148,7 +3146,7 @@ struct arcane_pulse_t final : public arcane_mage_spell_t
     }
     // On 12.1, the main cast no longer benefits from Impetus's arcane charge multiplier,
     // but the Reverberate echo still does. This is likely a bug.
-    else if ( is_echo && p()->bugs && p()->talents.impetus.ok() )
+    else if ( background && p()->bugs && p()->talents.impetus.ok() )
     {
       am *= arcane_charge_multiplier();
     }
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_mage.cpp
```diff
@@ -581,6 +583,7 @@ struct mage_t final : public player_t
 
     // Row 8
     player_talent_t touch_of_the_archmage_1;
+    player_talent_t prismatic_bolt_1;
```
vituscze, line 583: Add a TODO somewhere on TotA so that we don't forget to remove it in 12.1.

engine/class_modules/sc_mage.cpp
```diff
@@ -2945,6 +2961,73 @@ struct arcane_blast_t final : public arcane_mage_spell_t
...
+
+// The aoe part of Prismatic Bolt is its own spell (1295939)
+struct prismatic_bolt_aoe_t final : public arcane_mage_spell_t
+{
+  prismatic_bolt_aoe_t( std::string_view n, mage_t* p ) :
+    arcane_mage_spell_t( n, p, p->find_spell( 1295939 ) )
+  {
+    background = true;
+    aoe = -1;
+    radius = 8;
```
vituscze: This should be automatically picked up from spell data

engine/class_modules/sc_mage.cpp
```diff
@@ -2997,19 +3083,31 @@ struct arcane_pulse_t final : public arcane_mage_spell_t
...
+    }
 
     // TODO: radius increase?
     arcane_mage_spell_t::execute();
 
-    p()->trigger_arcane_charge( as<int>( data().effectN( 2 ).base_value() ) );
+    // TODO: Remove check when 12.1 goes live -- in 12.1, pulse generates charges per enemy hit in impact() instead.
+    if ( p()->dbc->wowv() < wowv_t{ 12, 1, 0 } )
+    {
+      p()->trigger_arcane_charge( as<int>( data().effectN( 2 ).base_value() ) );
```
vituscze: Better to generate the AC here with `num_targets_hit` (which is set after `execute()`). Saves some time by not having to do it one by one. Also, the Pulse echo behaves differently which isn't implemented here.

engine/class_modules/sc_mage.cpp
```diff
@@ -3020,18 +3118,39 @@ struct arcane_pulse_t final : public arcane_mage_spell_t
...
   }
 
+  void impact( action_state_t* s ) override
+  {
+    arcane_mage_spell_t::impact( s );
+
+    // 12.1: Arcane Pulse now generates 1 arcane charge per enemy hit
+    // TODO: Remove check when 12.1 goes live
+    if ( p()->dbc->wowv() >= wowv_t{ 12, 1, 0 } && result_is_hit( s->result ) )
+      p()->trigger_arcane_charge( 1 );
```
vituscze: See above

engine/class_modules/sc_mage.cpp
```diff
@@ -3020,18 +3118,39 @@ struct arcane_pulse_t final : public arcane_mage_spell_t
...
 
-    am *= arcane_charge_multiplier();
+    // TODO: Remove check when 12.1 goes live
+    if ( p()->dbc->wowv() < wowv_t{ 12, 1, 0 } )
+    {
+      am *= arcane_charge_multiplier();
+    }
 
+    // TODO: Remove check when 12.1 goes live
+    if ( p()->dbc->wowv() >= wowv_t{ 12, 1, 0 } )
```
vituscze: This check seems a bit meaningless? You're not getting cumulative power without already using 12.1 spell data

engine/class_modules/sc_mage.cpp
```diff
@@ -3146,6 +3265,9 @@ struct arcane_missiles_tick_t final : public custom_state_spell_t<arcane_mage_sp
   {
     custom_state_spell_t::execute();
 
+    if ( p()->sets->has_set_bonus( MAGE_ARCANE, MID2, B4 ) )
```
vituscze: No need to check for this again if the buff is disabled (trigger chance == 0) without the tier set

engine/class_modules/sc_mage.cpp
Lucasmingus: right, maybe i went a bit overboard with the version checks

PR #11529, opened 2026-06-24
[Gear] Trinket - Wavecaller's Seastone

criller7070:


engine/player/unique_gear_midnight.cpp
```diff
@@ -3079,6 +3079,49 @@ void sporelords_mycelium( special_effect_t& effect )
...
+
+// Wavecaller's Seastone
+// 1295058 Driver
+// 1295057 Tidal Insight Buff
+void wavecallers_seastone( special_effect_t& effect )
+{
+  auto buff = create_buff<stat_buff_t>( effect.player, effect.player->find_spell( 1295057 ) )
+                ->set_stat_from_effect_type( A_MOD_STAT, effect.driver()->effectN( 1 ).average( effect ) );
+
+  struct wavecallers_seastone_cb_t : public dbc_proc_callback_t
```
nyterage: entire callback isn't needed in this case. 

whole struct can be simplified down to

```
  auto buff = create_buff<stat_buff_t>( effect.player, effect.player->find_spell( 1295057 ) )
                  ->set_stat_from_effect_type( A_MOD_STAT, effect.driver()->effectN( 1 ).average( effect ) )
                  ->set_stack_change_callback( [ effect ]( buff_t* b, int, int new_ ) {
                      if ( new_  == b->max_stack() )
                         make_event( *effect.player->sim, 0_ms, [ b ] { b->set_reverse( true ); } );
                    } )
                  ->set_expire_callback( [ effect ]( buff_t* b, int, timespan_t ) {
                    make_event( *effect.player->sim, 0_ms, [ b ] { b->set_reverse( false ); } );
                  } );
```

nyterage (changes requested):

PR #11534, opened 2026-06-25
[Gear] Trinket - Fang of Umbral Malignance

criller7070:


engine/player/unique_gear_midnight.cpp
```diff
@@ -3134,7 +3134,57 @@ void vile_vial_of_volatile_venom( special_effect_t& effect )
...
+    action_t* dot;
+
+    fang_cb_t( const special_effect_t& e, action_t* burst_action ) : dbc_proc_callback_t( e.player, e )
+    {
+      dot                      = create_proc_action<generic_proc_t>( "umbral_malignance", e, 1305853 );
+      dot->dot_max_stack       = dot->data().max_stacks();
+      dot->base_td             = e.driver()->effectN( 1 ).average( e );
+      dot->base_td_multiplier *= role_mult( e );
+      dot->add_child( burst_action );
+      target_debuff            = e.player->find_spell( 1305853 );
```
nyterage: i don't see any reason why you'd need the debuff here? the DoT already handles everything? i dont see it being used for anything here

nyterage (changes requested):

engine/player/unique_gear_midnight.cpp
criller7070: the intention was to make the log clearer in showing the debuff stacks.

nyterage: the DoT already has the stack data. Shouldn't need to add an extra debuff to get stack data from that. having a secondary stacking effect being created on every target is also a performance concern when dot/debuff data is lazily initialized, only created when its first applied, basically costing 2x the initialization for every target its ever applied to during runtime.

PR #11535, opened 2026-06-25
[Gear] Trinket - Stormbound Emblem of Dazar

Lucasmingus:


engine/player/unique_gear_midnight.cpp
```diff
@@ -3276,6 +3276,64 @@ void font_of_venomous_rage( special_effect_t& effect )
...
+// 1295275 on-use driver (The King's Unyielding Wind), 2 sec channel ticking every 0.5 sec
+// 1294744 equip
+// 1294745 Haste buff
+void stormbound_emblem_of_dazar( special_effect_t& effect )
+{
+  // Each completed channel tick extends the Haste buff duration by 5sec,
+  // rather than the buff only being granted on full channel completion. 
+  // Interrupting the channel early keeps whatever duration has accumulated so far.
+  auto buff_spell = effect.player->find_spell( 1294745 );
+  auto value_spell = effect.player->find_spell( 1294744 );
```
nyterage, line 3291: just for safety should probably have an assert if either of these aren't found, if their IDs change or anything like that.

Lucasmingus: roger that

PR #11154, opened 2026-03-12
[DemonHunter] Voidfall proc rate is stack-dependent (per-spec)

taherbert:
## Summary

- Voidfall building proc chance decreases with current stack count, rather than being a flat rate from spell data (`effectN(3)` = 35%)
- WCL analysis of 72k+ eligible casts shows per-stack rates with high statistical significance for both specs
- Devourer procs ~2pp higher than Vengeance at all stack levels (z-test p&lt;0.02), so each spec gets its own defaults
- Defaults exposed as per-spec player options for tuning

### Vengeance (21,942 eligible Fracture casts, 325 players, S1 live)

| Stacks | Eligible | Procs | Rate | 95% CI | vs 35% |
|--------|----------|-------|------|--------|--------|
| 0 | 9,591 | 3,677 | **38.3%** | [37.4%, 39.3%] | p&lt;0.0001 |
| 1 | 7,471 | 2,432 | **32.6%** | [31.5%, 33.6%] | p&lt;0.0001 |
| 2 | 4,880 | 1,457 | **29.9%** | [28.6%, 31.2%] | p&lt;0.0001 |
| Overall | 21,942 | 7,566 | **34.5%** | [33.9%, 35.1%] | p=0.11 |

Chi-squared: X2=121.6, df=2 (p&lt;0.0001). Defaults: **0.385 / 0.325 / 0.30**

### Devourer (50,067 eligible Consume/Devour casts, 107 players)

| Stacks | Eligible | Procs | Rate | 95% CI | vs 35% |
|--------|----------|-------|------|--------|--------|
| 0 | 9,658 | 3,974 | **41.1%** | [40.2%, 42.1%] | p&lt;0.0001 |
| 1 | 22,638 | 7,712 | **34.1%** | [33.5%, 34.7%] | p=0.003 |
| 2 | 17,771 | 5,747 | **32.3%** | [31.7%, 33.0%] | p&lt;0.0001 |
| Overall | 50,067 | 17,451 | **34.9%** | [34.4%, 35.3%] | p=0.71 |

Chi-squared: X2=288.5, df=2 (p&lt;0.0001). Defaults: **0.41 / 0.34 / 0.325**

### Cross-spec comparison

| Stacks | Vengeance | Devourer | Delta | z-test |
|--------|-----------|----------|-------|--------|
| 0 | 38.3% | 41.1% | +2.8pp | p&lt;0.0001 |
| 1 | 32.6% | 34.1% | +1.5pp | p=0.017 |
[...]

PR #11538, opened 2026-06-26
[Devourer] Model Voidfall building procs as a pseudo-random distribution

taherbert:
Voidfall's building proc is currently a flat 35% roll. In game it's a pseudo-random distribution: the per-cast chance starts near 16% after a proc and climbs with each miss, averaging 35% and guaranteed by the 7th cast. This was checked against ~24k builder casts from logs (a controlled dummy plus ranked raids), where the per-attempt rate matches `C*N` with `C = prd::find_constant(0.35)`.

- Replace the flat roll in `voidfall_building_trigger_t` with `accumulated_rng`, the existing PRD primitive (same pattern as DK/mage/hunter), with `C` from `prd::find_constant` on the talent chance.
- Check `voidfall_spending` before the roll so the accumulator only advances on real builder attempts. The counter resets on a builder proc and does not advance during spending; guaranteed sources (Meteoric Rise, Mass Acceleration) trigger the buff directly and stay off the PRD.

DPS-neutral, since `find_constant` preserves the mean. Only the proc cadence is smoothed.

This closes: https://github.com/simulationcraft/simc/pull/11154

### DIFF
--- a/engine/class_modules/sc_demon_hunter.cpp
+++ b/engine/class_modules/sc_demon_hunter.cpp
@@ -1159,6 +1159,13 @@ public:
   {
   } shuffled_rng;
 
+  // Accumulated proc objects
+  struct accumulated_rngs_t
+  {
+    // Annihilator
+    accumulated_rng_t* voidfall = nullptr;
+  } accumulated_rng;
+
   // Special
   struct actives_t
   {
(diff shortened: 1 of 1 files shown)


engine/class_modules/sc_demon_hunter.cpp
```diff
@@ -1163,6 +1159,13 @@ class demon_hunter_t : public parse_player_effects_t
   {
   } shuffled_rng;
 
+  // Accumulated proc objects
+  struct accumulated_rngs_t
+  {
+    // Annihilator
+    accumulated_rng_t* voidfall = nullptr;
```
ToppleTheNun: don't need the `= nullptr` here

ToppleTheNun (changes requested):

PR #11539, opened 2026-06-26
[Gear] Venomcursed effects

Lucasmingus:
The corresponding bonus id for each buffed stat seems to be necessary to be on the item when simming for the effects to be detected. The venomcursed items themselves, don't seem to inherently have the effects.

Made `namespace items` since it felt awkward to put in any existing category.

### DIFF
--- a/engine/player/unique_gear_midnight.cpp
+++ b/engine/player/unique_gear_midnight.cpp
@@ -3675,6 +3675,32 @@ void rotmires_sporeheart( special_effect_t& effect )
 }
 }  // namespace armors
 
+namespace items
+{
+// Venomcursed <stat>, shared effect appearing on weapons, armor, and jewelry
+// Crit    : driver 1307906 / buff 1307910
+// Mastery : driver 1307923 / buff 1307922
+// Haste   : driver 1307928 / buff 1307927
+// <stat> increased, other 3 secondaries decreased
+custom_cb_t venomcursed( unsigned buff_id, stat_e primary )
+{
+  return [ = ]( special_effect_t& effect ) {
+    auto main_value    = effect.driver()->effectN( 1 ).average( effect );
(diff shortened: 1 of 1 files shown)


engine/player/unique_gear_midnight.cpp
```diff
@@ -3675,6 +3675,32 @@ void rotmires_sporeheart( special_effect_t& effect )
...
 }  // namespace armors
 
+namespace items
+{
+// Venomcursed <stat>, shared effect appearing on weapons, armor, and jewelry
+// Crit    : driver 1307906 / buff 1307910
+// Mastery : driver 1307923 / buff 1307922
+// Haste   : driver 1307928 / buff 1307927
+// <stat> increased, other 3 secondaries decreased
+custom_cb_t venomcursed( unsigned buff_id, stat_e primary )
```
nyterage, line 3685: Should be able to get the increased stat from effect 1 of the buff, and the reduced stats from effect 2. these are all present in the misc value for the effect type. 

Would much rather read from the data rather than using manual setting like done here.

nyterage (changes requested):

engine/player/unique_gear_midnight.cpp
```diff
@@ -3675,6 +3675,32 @@ void rotmires_sporeheart( special_effect_t& effect )
...
+{
+  return [ = ]( special_effect_t& effect ) {
+    auto main_value    = effect.driver()->effectN( 1 ).average( effect );
+    auto penalty_value = effect.driver()->effectN( 2 ).average( effect );
+
+    auto buff = create_buff<stat_buff_t>( effect.player, effect.player->find_spell( buff_id ) )
```
nyterage, line 3691: can also get the buff id from effectN( 1 ).trigger(), rather than manually passing the buff id into the callback.

nyterage (changes requested):

PR #11545, opened 2026-06-29
Squash warnings. Fix logical error with potion expression creation.

renanthera:


engine/player/player.cpp
```diff
@@ -12323,14 +12323,15 @@ std::unique_ptr<expr_t> player_t::create_expression( util::string_view expressio
...
   {
     std::string_view potion_view;
+    std::string default_potion = this->default_potion();
 
     if ( !potion_str.empty() )
     {
       potion_view = potion_str;
     }
-    else if ( default_potion().empty() )
+    else if ( !default_potion.empty() )
```
renanthera, line 12332: This is now consistent with the implementation this was likely based on, and makes more sense.

See `consumable.cpp:460`.

PR #11546, opened 2026-06-29
[Outlaw] Add 12.1 changes and set bonus

soloxcx:
- Update Zero In proc behavior with daggers
- Add MID2 set bonus for Outlaw


engine/class_modules/sc_rogue.cpp
```diff
@@ -7547,8 +7579,13 @@ void actions::rogue_action_t<Base>::spend_combo_points( const action_state_t* st
...
+  double cp_loss = max_spend;
+
+  if ( affected_by.mid2_outlaw_4pc && p()->buffs.mid2_outlaw_4pc->check() )
+    cp_loss *= 1.0 + p()->spec.mid2_outlaw_4pc_buff->effectN( 2 ).percent();
+
   ab::stats->consume_resource( RESOURCE_COMBO_POINT, max_spend );
-  p()->resource_loss( RESOURCE_COMBO_POINT, max_spend );
+  p()->resource_loss( RESOURCE_COMBO_POINT, cp_loss );
 
   p()->sim->print_log( "{} consumes {} {} for {} ({})", *p(), max_spend, util::resource_type_string( RESOURCE_COMBO_POINT ),
```
EvanMichaels:
```suggestion
  p()->sim->print_log( "{} consumes {} {} for {} ({})", *p(), cp_loss, util::resource_type_string( RESOURCE_COMBO_POINT ),
```

engine/class_modules/sc_rogue.cpp
```diff
@@ -10125,6 +10185,11 @@ void rogue_t::init_spells()
   set_bonuses.mid1_outlaw_4pc = sets->set( ROGUE_OUTLAW, MID1, B4 );
   set_bonuses.mid1_subtlety_2pc = sets->set( ROGUE_SUBTLETY, MID1, B2 );
   set_bonuses.mid1_subtlety_4pc = sets->set( ROGUE_SUBTLETY, MID1, B4 );
+  
+  set_bonuses.mid2_outlaw_2pc = sets->set( ROGUE_OUTLAW, MID2, B2 );
+  set_bonuses.mid2_outlaw_4pc = sets->set( ROGUE_OUTLAW, MID2, B4 );
+
+  spec.mid2_outlaw_4pc_buff = set_bonuses.mid2_outlaw_4pc->ok() ? set_bonuses.mid2_outlaw_4pc->effectN( 1 ).trigger() : spell_data_t::not_found();
```
EvanMichaels: Don't need the ok() check here I don't believe. `.trigger()` on a spell data will return `not_found` if the spell data doesn't exist.

For simplification purposes, it may be worth putting the chance here though with 
```
if( set_bonuses.mid2_outlaw_4pc->ok() )
{
  spec.mid2_outlaw_4pc_buff->set_chance( set_bonuses.mid2_outlaw_4pc->effectN( 1 ).percent() );
}
```

This sets the trigger chance on the buff and will automatically rng roll when calling buff.trigger(). Then you wouldn't have to check the chance at runtime above. You can just call `p()->buffs.mid2_outlaw_4pc->trigger();`

Could then just remove the entire `trigger_mid2_outlaw_4pc` function since you'd just need a buff trigger call in two places without any other logic.

engine/class_modules/sc_rogue.cpp
```diff
@@ -10922,7 +10987,10 @@ void rogue_t::create_buffs()
 
   // Set Bonus Items ========================================================
 
-  buffs.mid1_outlaw_4pc = make_buff<damage_buff_t>( this, "whirl_of_blades", set_bonuses.mid1_outlaw_4pc->effectN(2).trigger() );
+  buffs.mid1_outlaw_4pc = make_buff<damage_buff_t>( this, "whirl_of_blades", set_bonuses.mid1_outlaw_4pc->effectN( 2 ).trigger() );
+
+  buffs.mid2_outlaw_4pc = make_buff( this, "fang_strike", spec.mid2_outlaw_4pc_buff )
```
EvanMichaels: You may want to use make_buff_fallback here since it looks like you want to use this in the APL. Either the APL needs to have a prior `set_bonus&&` check to keep the buff.fang_strike conditional from evaluating or buff needs to be created as fallback so it exists even if the spell data doesn't.

PR #11547, opened 2026-06-29
[Gear] Trinket - Vashnik's Sanguine Rancor

criller7070:


engine/player/unique_gear_midnight.cpp
```diff
@@ -3276,6 +3276,40 @@ void font_of_venomous_rage( special_effect_t& effect )
...
+void vashniks_sanguine_rancor( special_effect_t& effect )
+{
+  auto damage = effect.driver()->effectN( 1 ).average( effect );
+
+  auto concentrated_bile = create_proc_action<generic_proc_t>( "concentrated_bile", effect, 1303484 );
+  concentrated_bile->base_td = damage * effect.driver()->effectN( 2 ).percent()
+                              * concentrated_bile->base_tick_time / concentrated_bile->dot_duration;
+  concentrated_bile->base_td_multiplier *= role_mult( effect );
+
+  auto crimson_bile = create_proc_action<generic_aoe_proc_t>( "crimson_bile", effect, 1303483 );
```
nyterage: since this effect splits damage and scales the 15% per target hit, should have the bool set for that after the spell id. 

` auto crimson_bile = create_proc_action<generic_aoe_proc_t>( "crimson_bile", effect, 1303483, true );`

engine/player/unique_gear_midnight.cpp
```diff
@@ -3276,6 +3276,40 @@ void font_of_venomous_rage( special_effect_t& effect )
...
+  crimson_bile->base_multiplier *= role_mult( effect );
+  crimson_bile->add_child( concentrated_bile );
+
+  effect.custom_buff = create_buff<buff_t>( effect.player, effect.player->find_spell( 1303479 ) )
+    ->set_expire_at_max_stack( true )
+    ->set_expire_callback( [ crimson_bile, concentrated_bile ]( buff_t* b, int stacks, timespan_t ) {
+      // only erupt when the buff was consumed at max stacks, not when expired during combat end
+      if ( stacks < b->max_stack() )
+        return;
+      auto n_targets = crimson_bile->target_list().size();
```
nyterage: this should use `available_targets()` instead, passing in the crimson bile target list as an arg so if distance targeting is enabled, it will know how many targets were actually in range.

nyterage (changes requested):

engine/player/unique_gear_midnight.cpp
criller7070: i believe whenever the bool is not set, it is taken from the base as true? its using the generic 30% per target up to 150%

engine/player/unique_gear_midnight.cpp
criller7070: using
`crimson_bile->set_target( b->player->target );`
this will force a refresh of the target cache, which will then make the target_list call check if distance_targeting is enabled

am i correct?

nyterage: while thats correct. not exactly clear code wise what its doing. would much prefer

```
auto n_targets = crimson_bile->available_targets( crimson_bile->target_list() );
```

as its simpler code wise, and incorporates the distance targeting check.

PR #11564, opened 2026-07-08
[Gear] Rotmire's Sporeheart is not split aoe damage

soloxcx:
https://www.warcraftlogs.com/reports/XpkLYjKr13m9TF4C?fight=1&type=damage-done&source=5&pins=2%24Separate%24%23244F4B%24damage%240%240.0.0.Any%24207714534.0.0.Paladin%24true%240.0.0.Any%24false%24-1285139&view=events

see single target hits (11:45.447, 11:47.449, 11:49.440) dealing the same damage as huge aoe hits (00:37.61, 22 targets)

### DIFF
--- a/engine/player/unique_gear_midnight.cpp
+++ b/engine/player/unique_gear_midnight.cpp
@@ -3715,10 +3715,10 @@ void rotmires_sporeheart( special_effect_t& effect )
     protective_toadstool_buff_t( const special_effect_t& e )
       : absorb_buff_t( e.player, "protective_toadstools", e.player->find_spell( 1285161 ) )
     {
-      damage = create_proc_action<generic_aoe_proc_t>( "bursting_toadstools", e, 1285163 );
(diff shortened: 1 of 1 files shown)


engine/player/unique_gear_midnight.cpp
```diff
@@ -3719,6 +3718,7 @@ void rotmires_sporeheart( special_effect_t& effect )
       damage = create_proc_action<generic_aoe_proc_t>( "bursting_toadstools", e, 1285163 );
       damage->base_dd_min = damage->base_dd_max = e.driver()->effectN( 3 ).average( e );
       damage->base_multiplier *= role_mult( e );
+      damage->split_aoe_damage = false;
```
nyterage: just this flag wouldnt adjust the aoe scaling as well. think it needs to be changed to the base type generic_proc_t and aoe set to -1 to properly capture that behavior.

nyterage (changes requested):

PR #11630, opened 2026-07-24
DungeonRoute: opt-in dynamic targeting (dungeon_route_dynamic_targeting)

Azortharion:
Adds `dungeon_route_dynamic_targeting` (default 0). With the option off nothing changes: same-seed DPS is bit-identical to master across DungeonRoute, DungeonSlice, HecticAddCleave and Patchwerk test runs.

With it on, during DungeonRoute pulls:

- `player_t::acquire_target` picks the living mob with the highest current HP instead of the first entry in `target_non_sleeping_list`. A 1s child event on the pull re-checks between deaths, and a 5% hysteresis band keeps the current target unless the candidate visibly leads it, so near-equal mobs dying together do not cause target flickering.
- Actions are allowed to swap off a still-alive target when the priority mob changes. Channels in progress and explicit `target_number` lines still pin.
- `prioritydps` counts damage into the current highest-hp mob instead of `BOSS_`-prefixed mobs.

Reasoning: the current behavior parks the player on the head of the HP-sorted spawn list until that mob dies, and the swap-with-back erase in `target_non_sleeping_list` then retargets to the smallest survivor rather than the next-biggest. A player pushing keys attacks whatever gates the end of the pull, which is the highest-hp mob at that moment, and equalized death times are the fastest clear since AoE is more efficient than single target. On a fitted ~30 min route with an MM hunter this measured route length -4.5%; a two-mob crossover pull went from deaths 91s apart to 1.6s apart and finished 17% sooner.

### DIFF
--- a/engine/player/player.cpp
+++ b/engine/player/player.cpp
@@ -14657,6 +14657,9 @@ void player_t::acquire_target( retarget_source event, player_t* context )
   player_t* candidate_target = nullptr;
   player_t* first_invuln_target = nullptr;
 
+  bool dynamic_targeting = sim->fight_style == FIGHT_STYLE_DUNGEON_ROUTE && sim->dungeon_route_dynamic_targeting;
+  double candidate_hp = -1.0;
+
   // TODO: Fancier system
   for ( auto enemy : sim->target_non_sleeping_list )
   {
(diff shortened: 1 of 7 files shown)


renanthera:
Why is this feature conditioned on the fight style being dungeon route? I feel like it'd be useful in all fight styles and the extra condition doesn't contribute anything.

This also can be handled in-apl with judicious use of targetif.

Azortharion:
> Why is this feature conditioned on the fight style being dungeon route? I feel like it'd be useful in all fight styles and the extra condition doesn't contribute anything.
> 
> This also can be handled in-apl with judicious use of targetif.

 I was a bit terrified of making a PR/contribution that would have any effect on anything outside of the narrow scope of "Hunter stuff", so it was important for me to make this opt-in.

If y'all think it's good enough to branch out to other fight styles, all the better, but that was my reason for giving it a narrow scope.

Target_if goes a long way but will not, for example, re-target pets for BM Hunters and so on. It also makes the APL harder to read, I think.

engine/player/player.cpp
Jayezi, line 7161: I would bet most of the gain from the new option would be covered by simply fixing this erase with one that preserves the order, assuming the existing smart targeting sort option is used it will correctly result in targeting down the highest to lowest mobs until dead (still only based on starting health order, but the post-fix gain of dynamically checking remaining hp would probably be minor in comparison).

```suggestion
    sim->target_non_sleeping_list.find_and_erase( this );
```

Otherwise I share the same sentiment about having the apl be the authority on retargeting outside of inevitable retargets like invulvn, demise etc. Ideally apls get every chance to make sure they are not being forced into very suboptimal targets swaps, such as significant debuffs/dots maintained on targets that will be lost.

This would imo better be accomplished with whatever adjustment to apl capability is needed for it to be able to do the retargeting on its own terms and direct pets as needed, like the `retarget_auto_attack` action some classes use https://github.com/simulationcraft/simc/blob/6da84b7bfd9e1fe8d3064f317cd4a15762cc6b01/engine/player/player.cpp#L10812

Azortharion: The erase is worth fixing regardless of this PR.

I measured your ordered version against the option (PR) on the six Hunter specs: on its own it's a minority of the gain (-0.3% to +3.2% by spec, against +4.4% to +10.6% for the option), and it doesn't touch pets or the prioritydps accounting. Default, opt-in, or not at all is the maintainers' call. I built it opt-in and off by default, so it costs nothing to leave in.

PR #11640, opened 2026-07-26
[Gear] Trinket - Knot of Writhing Serpents

criller7070:


engine/player/unique_gear_midnight.cpp
```diff
@@ -3636,6 +3636,46 @@ void vashniks_sanguine_rancor( special_effect_t& effect )
...
+    void impact( action_state_t* s ) override
+    {
+      generic_aoe_proc_t::impact( s );
+
+      if ( s->chain_target == 0 )
+        dot->execute_on_target( s->target );
+    }
+  };
+
+  effect.execute_action = create_proc_action<writhing_venom_t>( "writhing_venom", effect );
```
nyterage: due to some proc callbacks maybe triggering off the missiles, this should execute the missile action, and have the missle's impact action be the writhing venom damage itself. properly modeling all casts is generally ideal rather than setting the travel speed in the main spell due to these factors.

nyterage (changes requested):

PR #11677, opened 2026-08-07
[Priest] implement Tentacle Slam line targeting shape

Urcra:
Currently tentacle slam always hits everything even when we position bosses apart for spread cleave. This leads to slam oversimming on spread cleave as it does full dmg to everything.

This tries to simulate more how it works ingame. Splashing 8 yards around the target, and in a 8 yards wide rectangle from the caster to the target. The width is just assumed to be 8 yards, it feels a bit smaller ingame but I can't find any spelldata or confirmation of how wide it actually is, but the width is also not really that important for sims as long as it's roughly correct instead of hitting every target in a 40yard range.

Using this it seems to be worth dropping the tentacle slam talents on spread 2 target cleave like entombed sentinels:

Archon: drop Descending Darkness, Crushing Void. Take Spectral Horrors + Cthun. +1.42%
VW: drop Descending Darkness, Crushing Void. Take Maddening Touch + Mind Devourer. +1.27%

These were just some quick sims I did for testing there may be even more optimal talents for spread cleave. But just to give an idea that there is power to be gained. Without the change both of these changes would show as about -0.8% dps instead

### DIFF
--- a/engine/class_modules/priest/sc_priest_shadow.cpp
+++ b/engine/class_modules/priest/sc_priest_shadow.cpp
@@ -1707,6 +1707,33 @@ struct shadow_weaving_t final : public priest_spell_t
 // ==========================================================================
 // Tentacle Slam
 // ==========================================================================
+// Hits enemies on a line between the caster and the target plus a radius around the target.
+// Line width assumed to be around 8 yds
+static constexpr double tentacle_slam_line_half_width = 4.0;
+
+static bool in_tentacle_slam_area( const player_t* caster, const player_t* primary, const player_t* candidate,
+                                   double around_radius )
+{
+  if ( candidate == primary )
+    return true;
+  if ( candidate->get_position_distance( primary->x_position, primary->y_position ) <= around_radius )
+    return true;
+
+  double vx   = primary->x_position - caster->x_position;
+  double vy   = primary->y_position - caster->y_position;
+  double len2 = vx * vx + vy * vy;
+  double t    = 0.0;
+  if ( len2 > 0.0 )
+  {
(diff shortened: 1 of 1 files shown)


engine/class_modules/priest/sc_priest_shadow.cpp
```diff
@@ -1707,6 +1707,33 @@ struct shadow_weaving_t final : public priest_spell_t
 // ==========================================================================
 // Tentacle Slam
 // ==========================================================================
+// Hits enemies on a line between the caster and the target plus a radius around the target.
+// Line width assumed to be around 8 yds
+static constexpr double tentacle_slam_line_half_width = 4.0;
```
nyterage: should be pulling this from the spell data. there is a line width aspect there already

for spell id 1227621 line width in data is 8 yards, can just get the spell_data_t pointer, and point this at `data->line_width()` to get the data proper. This is ideal in case it changes rather than hard coded magic numbers. 

the length of the line is the effect N max radius.

so in the case of this spell, would be an 8 yard wide, 40 yard long tentacle.

Urcra: Nice didn't know it was available in the spelldata. I updated to use that. 

Though on the length part the tentacle will only extend to the target and not beyond(except for the splash radius). I think taking the 40 yards spell data would also give some weird results due to Phantom Reach increasing our range beyond 40 yards and in theory casting a slam that would not hit the target if we used the 40 yards from the spell data

But I updated to use the spelldata for the width

seanpeters86 (approved):

PR #11685, opened 2026-08-09
[druid] Guardian Druid 12.1

renanthera:
- **[druid] Clean up trailing whitespace so my editor doesn't try to do it on every file save.**
- **[druid] Remove manual Matted Fur coefficient, as the value is now properly driven straight from effect value.**
- **[druid] Remove Guardian Apex.**
- **[druid] Implement Gory Fur free Raze/Ravage/Maul procs. Add spell data to generator.**
- **[druid] Guardian Midnight Season 2 2pc set bonus**
- **[druid] Remove Guardian Apex even more.**
- **[druid] Guardian Midnight Season 2 4pc set bonus**
- **[druid] Offspec Starfire aoe damage bug is fixed.**
- **[druid] Trigger tier set 3 times with a 250ms delay between each trigger. Prevent aoe component from hitting primary target.**

## TODO:
- Killing Blow resource cost interaction.

Note, this PR does not include the spell data required for this PR to function.

### DIFF
--- a/dbc_extract3/dbc/generator.py
+++ b/dbc_extract3/dbc/generator.py
@@ -2439,8 +2439,8 @@ class SpellDataGenerator(DataGenerator):
           ( 1280133, 2 ), # Rake (unused?)
           # Guardian
           ( 1250913, 3 ), # Memory of Ysera heal
-          ( 1269633, 3 ), ( 1269645, 3 ), ( 1269648, 3 ), # Apex talent echoes
           ( 201671, 3 ), ( 1307881, 3 ), # Gory Fur
+          ( 1269616, 3 ), # Wild Guardian Usable Buff
           # Restoration
         ),
         # Demon Hunter:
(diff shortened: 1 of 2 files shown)


dbc_extract3/dbc/generator.py
```diff
@@ -2440,6 +2440,7 @@ class SpellDataGenerator(DataGenerator):
           # Guardian
           ( 1250913, 3 ), # Memory of Ysera heal
           ( 1269633, 3 ), ( 1269645, 3 ), ( 1269648, 3 ), # Apex talent echoes
```
indrih17: I think this is no longer necessary

engine/class_modules/sc_druid.cpp
```diff
@@ -5471,7 +5445,10 @@ struct mangle_t final : public use_fluid_form_t<BEAR_FORM,
   {
     base_t::execute();
 
-    p()->buff.gory_fur->trigger( this );
+    if ( p()->buff.answered_calling->up() )
+      p()->active.spirits_wrath->execute_on_target( target );
+
+    p()->buff.gory_fur_ironfur->trigger( this );
```
indrih17: Why does it proc from mangle?

indrih17 (commented):
Thank you for the great work done!

engine/class_modules/sc_druid.cpp
renanthera: Put it in the wrong spot, now fixed.

PR #11749, opened 2026-08-16
[Shaman] Fix Flametongue Weapon buff never applying

ahomescu:
Trying to sim my elemental shaman Traxxen with `use_blizzard_action_list=1` on raidbots after I picked _Flametongue Weapon_ produced some really low numbers (7k DPS), see the report at https://www.raidbots.com/simbot/report/2GQFRwTqKp2aQzCMbQog1s. All it did was cast that buff over and over the whole fight.

This patch fixes the issue. I tested a local build and it now correctly simulates the fight, and DPS went up to around 60k. I suspect this might also solve the really low numbers for enhancement at https://www.noxxic.com/wow/sims/enhancement-shaman/one-button-alt. Their build takes this talent (but not the elemental one, for some reason).


ahomescu:
It looks like the issue has been fixed another way in #11818, closing this one.

PR #11782, opened 2026-08-22
[Hunter] Simplify Rapid Fire logic for Precise Shots and No Scope with Unload

misterwiki:
Hotfixes applied on **08/20/2026** changed the interaction between **Rapid Fire**, **No Scope**, and **Unload**.

Previously, the **Precise Shots** buff was applied at the same time as the **Arcane Shot** triggered by Rapid Fire. As a result, the triggered Arcane Shot could consume an existing Precise Shots buff if one was active before Rapid Fire was cast.

The hotfix that addressed Arcane Shot clipping at the end of Rapid Fire also changed the order of these events. The triggered **Arcane Shot now fires before the new Precise Shots buff is applied**.

This means that with **Unload**, both Arcane Shots triggered during Rapid Fire can now benefit from **Precise Shots**, rather than the first Arcane Shot consuming the previously active buff before the second is fired.

This PR updates the relevant logic to account for the new post-hotfix behavior.

### DIFF
--- a/engine/class_modules/sc_hunter.cpp
+++ b/engine/class_modules/sc_hunter.cpp
@@ -5925,26 +5925,11 @@ struct rapid_fire_t: public hunter_ranged_attack_t
     hydra_target = p()->get_hydra_target( target );
     marked_targets.clear();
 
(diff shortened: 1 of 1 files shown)


ZingusWoW:
This is being worked on already. More work is required than this commit.

PR #11791, opened 2026-08-24
[Gear] Trinket - Sszorak's Ferocity

criller7070:


engine/player/unique_gear_midnight.cpp
```diff
@@ -4516,6 +4516,76 @@ void ophidian_bone_whistle( special_effect_t& effect )
...
+  auto ki_proc = new special_effect_t( effect.player );
+  ki_proc->name_str = "killer_instincts";
+  ki_proc->spell_id = 1307361;
+  effect.player->special_effects.push_back( ki_proc );
+
+  effect.player->callbacks.register_callback_execute_function(
+      ki_proc->spell_id,
+      [ follow_up, killer_instincts ]( dbc_proc_callback_t*, const spell_data_t*, player_t* t, action_state_t* ) {
+        killer_instincts->expire();
+        follow_up( t );
```
renanthera, line 4586: If Tempest is already active, Tempest cannot be triggered by Killer Instinct. KI can trigger Mutilate during Tempest, though.

I have been unable to find an instance of a natural proc occurring during Tempest. Does this munch ppm?

I'm unaware of a group of 6+ target dummies, which would be helpful for longer term testing to validate whether or not it munches ppm.

A direct comparison of time-normalized triggers of Mutilate (2-5t) and Tempest (6t+) against dummies without KI triggers would likely be sufficient. Tempest is the only ability based on my observations that would permit a potential proc munch, so there would hypothetically be a measurable difference in trigger count.

engine/player/unique_gear_midnight.cpp
```diff
@@ -4516,6 +4516,76 @@ void ophidian_bone_whistle( special_effect_t& effect )
...
+  mutilate->base_multiplier *= role_mult( effect );
+  mutilate->name_str_reporting = "Mutilate";
+
+  // no aoe scaling, unlike mutilate
+  auto tempest = create_proc_action<generic_aoe_proc_t>( "sszoraks_tempest", effect, 1307324, false );
+  auto tempest_ticks = as<unsigned>( tempest->dot_duration / tempest->base_tick_time ) + tempest->tick_on_application;
+  tempest->base_td = effect.driver()->effectN( 3 ).average( effect ) / tempest_ticks;
+  tempest->base_td_multiplier *= role_mult( effect );
+  tempest->split_aoe_damage = false;
+  tempest->name_str_reporting = "Tempest";
```
renanthera, line 4545: This should probably be a ground effect, otherwise the unique Tempest behaviour is a little more awkward (read: O(n)) to implement.

renanthera (changes requested):

engine/player/unique_gear_midnight.cpp
criller7070: I went and tested this. Pulled 10 mobs, waited until Tempest was ticking, then killed one so Killer Instincts came up with 9 still alive. The next attack gave me another Tempest and it refreshed on all 9.

I checked the munching too. Ran about 5 minutes on dummies and twice a normal proc refreshed Tempest while it was still ticking, so it isn't being eaten.

engine/player/unique_gear_midnight.cpp
criller7070: I'm pretty sure it isn't a ground effect. I had mobs run away while Tempest was on them and they kept taking full damage the whole way out. Nothing ever picked it up by walking into it either. So it's a debuff that sticks to the target rather than something sitting on the ground.

engine/player/unique_gear_midnight.cpp
renanthera: 5 minutes would be wildly insufficient to verify that.

Either way, I have also seen evidence that a natural proc can trigger Tempest while Tempest is ongoing. That removes any concern with proc munching.

engine/player/unique_gear_midnight.cpp
renanthera: In retrospect, agreed.

renanthera:
For additional clarity, trigger condition is satisfied so it can occasionally schedule multiple execute events. As the execute event hasn't occurred, expiration hasn't occurred.

There's a few ways to correct this behaviour, but dd0250b is one possible (and in this case fairly minimal) fix.

PR #11859, opened 2026-09-09
[Warlock] Update Destro APL and Demo and Destro profiles

SzynaWoW:
As usual, all the research has been documented on the locks' Discord


engine/class_modules/apl/warlock.cpp
```diff
@@ -339,10 +353,12 @@ void destruction( player_t* p )
   aoe_hc->add_action( "incinerate" );
 
   aoe_dia->add_action( "summon_infernal" );
-  aoe_dia->add_action( "chaos_bolt,if=talent.diabolic_ritual&(demonic_art|(variable.ritual_length<action.chaos_bolt.execute_time))&(active_enemies<=(10-2*talent.destructive_rapidity))" );
+  aoe_dia->add_action( "shadowburn,if=demonic_art" );
+  aoe_dia->add_action( "chaos_bolt,if=demonic_art" );
+  aoe_dia->add_action( "(demonic_art|(variable.ritual_length<action.chaos_bolt.execute_time))&(active_enemies<=(4-talent.destructive_rapidity))" );
```
renanthera: this line strictly contains an if condition and appears to be inaccurately converted into c++ apl syntax

renanthera (changes requested):

engine/class_modules/apl/warlock.cpp
SzynaWoW: Thanks! Mistake during copying. It is fixed now

PR #11863, opened 2026-09-10
[Gear] tidal insight sim bugfix

Lucasmingus:
example where sample iteration has normal behavior: https://www.raidbots.com/simbot/report/1ctmkDesAYXZ7k2DhkKpFA
example where sample iteration has abnormal behavior: https://www.raidbots.com/simbot/report/4iWX9RwcUSpL3WcmSSm9Pe

Every iteration that ended while the buff was diminishing would affect the next one, making the next iteration's first proc apply max stacks instead of 1.

### DIFF
--- a/engine/player/unique_gear_midnight.cpp
+++ b/engine/player/unique_gear_midnight.cpp
@@ -3529,17 +3529,30 @@ void sporelords_mycelium( special_effect_t& effect )
 // 1295057 Tidal Insight Buff
 void wavecallers_seastone( special_effect_t& effect )
 {
-  effect.custom_buff = create_buff<stat_buff_t>( effect.player, effect.player->find_spell( 1295057 ) )
-                  ->set_stat_from_effect_type( A_MOD_STAT, effect.driver()->effectN( 1 ).average( effect ) )
-                  ->set_stack_change_callback( [ effect ]( buff_t* b, int, int new_ ) {
-                      if ( new_ == b->max_stack() )
-                        make_event( *effect.player->sim, 0_ms, [ b ] { b->set_reverse( true ); } );
-                    } )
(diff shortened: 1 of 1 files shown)


engine/player/unique_gear_midnight.cpp
```diff
@@ -3535,8 +3535,10 @@ void wavecallers_seastone( special_effect_t& effect )
                       if ( new_ == b->max_stack() )
                         make_event( *effect.player->sim, 0_ms, [ b ] { b->set_reverse( true ); } );
                     } )
-                  ->set_expire_callback( [ effect ]( buff_t* b, int, timespan_t ) {
-                    make_event( *effect.player->sim, 0_ms, [ b ] { b->set_reverse( false ); } );
+                  ->set_expire_callback( []( buff_t* b, int, timespan_t ) {
+                    // must be immediate; combat_end() expires buffs after the event manager stops, so a
+                    // delayed reset would never run and reverse would leak into the next iteration
+                    b->set_reverse( false );
```
nyterage: rather than removing the event for buff modification (this should be present for iteration time buff changes to be sure things are sequenced correctly), we should be checking in the callback if `sim->event_mgr.cancelled`, if so, immediately set reverse to false. 

other options to solve this would be a player arise callback, which sets reverse to false whenever the player arises (basically the first thing after init at the start of an iteration). Or changing the buff to a custom struct, where you can set reverse to false in a `reset() override`

nyterage (changes requested):

engine/player/unique_gear_midnight.cpp
Lucasmingus: Went with the `reset() override` route. It's the most robust as it runs unconditionally for every iteration.

Thanks

PR #11881, opened 2026-09-18
Update MID2_Mage_Arcane_Sunfury.simc

lb1598:
Added condition


ArcanePorom:
Hello!
At face this appears to be a dps loss to the default profile in AOE.
This is also not the correct file to update if you want to contribute to the baseline APL, you would want to use the mage.cpp file where all 3 Mage APLs are stored.
Feel free to reach out to me on discord (Porom.) if you want to discuss further or have questions.

